PR #6, opened 2014-11-12
suffix JSON keys containing HTML with _html

philschatz:
**DO NOT MERGE** until Tutor and Exercises have been updated.

This came out of a discussion on which fields are just text (like a title) and which fields may contain markup (like math)

An exercise would become:

```json
{
  "stimulus_html":"The spaceship moves at <span data-math='\\dfrac{-b \\pm \\sqrt{b^2 - 4ac}}{2a}'></span> 1 m/s",
  "questions":[
    {
      "id": "123",
      "format": "short-answer",
      "stem_html":"What is the rest mass in kg?"
    },{
      "id": "234",
      "format": "multiple-choice",
      "stem_html":"What is the force if it slams into a wall?",
      "answers":[
        {"id":"id1","content_html":"10 N"},
        {"id":"id2","content_html":"1 N"}
      ]
    }
  ]
}
```

### DIFF
--- a/api/tasks/3
+++ b/api/tasks/3
@@ -3,19 +3,19 @@
   "steps": [{
     "type": "exercise",
     "content": {
-      "stimulus":"The spaceship moves at <span data-math='\\dfrac{-b \\pm \\sqrt{b^2 - 4ac}}{2a}'></span> 1 m/s",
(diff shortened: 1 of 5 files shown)


philschatz:
this update was included in another PR

PR #14, opened 2015-02-10
Reading step styles

jivewise:
screen shot 2015-02-11 at 2 19 39 am

Adding in initial styles for reading step for the following reading components:

- Grasp Checks
- Tips for Success
- Link to Physics

### DIFF
--- a/api/tasks/1
+++ b/api/tasks/1
@@ -3,6 +3,6 @@
   "steps": [{
     "type": "reading",
     "content_url": "http://archive.cnx.org/contents/3e1fc4c6-b090-47c1-8170-8578198cc3f0@8.html",
-    "content_html": "<strong>This is a strong tag</strong><em>Emphasize me</em>"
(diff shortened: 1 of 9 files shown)


kathi-fletcher:
FYI -- The Grasp Check is going to be an actual interactive problem, and thus will be handled differently from the book. It should be a referenceable assessment in 'Exercises'.

philschatz:
ok. when should it be visible in the book content?

1. Never in the reading but as a question after the reading
2. inside the book in Tutor
3. or inside the CNX book content

If 3, then there needs to be a bunch of markup changes (because a `<note>` cannot contain an `<exercise class="grasp-check">` or a `<note class="grasp-check">`).

kathi-fletcher:
In an interactive reading, the grasp check follows the item and is
interactive and looks like any other practice problem.

In CNX the grasp check can either be that simple show/hide thing that
exercises are, or it can just be the problem stem.

In the reference view of the book in tutor, we should do whatever we
decided to do in CNX.

Agreed this has impact on markup.

On Wed, Feb 11, 2015 at 11:33 AM, Philip Schatz <notifications@github.com>
wrote:

> ok. when should it be visible in the book content?
>
>    1. Never in the reading but as a question after the reading
>    2. inside the book in Tutor
>    3. or inside the CNX book content
>
> If 3, then there needs to be a bunch of markup changes (because a <note>
> cannot contain an <exercise class="grasp-check"> or a <note
> class="grasp-check">).
>
> --
> Reply to this email directly or view it on GitHub
> <https://github.com/openstax/tutor-js/pull/14#issuecomment-73925041>.
>



-- 
Katherine Fletcher, kathi.fletcher@gmail.com
<kathi.fletcher@gmail.com>
Twitter: kefletcher <http://www.twitter.com/kefletcher> Blog:
kefletcher.blogspot.com
<kathi.fletcher@gmail.com>

PR #15, opened 2015-02-12
WIP: Call backend API as user steps through the task

philschatz:
All the routes so far:

- `GET "/api/tasks/#{id}"` : JSON contains `steps: [...]`
- `PUT "/api/tasks/#{task.id}/steps/#{step.id}/completed"`
- `PATCH "/api/tasks/#{task.id}/steps/#{step.id}" payload: {free_response}`
- `PATCH "/api/tasks/#{task.id}/steps/#{step.id}" payload: {answer_id}`
- `DELETE "/accounts/logout"`
- `GET "/api/user/tasks"`

I assume the JSON for a Step contains:

- `id: "123"`
- `is_completed: true` (could switch to `completed_at`)
- `free_response: 'Some String'`
- `answer_id: 'id234'`
- `correct_answer_id: 'id345'`  (only when `is_completed: true`)
- `feedback_html: '<em>Good Job</em>'` (only when `is_completed: true`)

An example of the Reading Task: <https://github.com/openstax/tutor-js/blob/master/api/tasks/4>


jpslav:
@philschatz, if possible we prefer

* `GET "/login"`
* `DELETE "/logout"`

These routes are common among all of our apps (they come from our gem that connects us to OX Accounts).

philschatz:
Yep, that's fine (should it be in this sprint?). I was just using the logout someone else gave me.

jpslav:
It doesn't have to be a focus of the sprint, it is just the default that already exists. If /accounts/logout works I'm now sure how :-)

PR #16, opened 2015-02-12
Exercise question styling

jivewise:
screen shot 2015-02-13 at 4 40 00 am
screen shot 2015-02-13 at 4 40 13 am

Adding in styles for questions and answers.  Will need to check more question types later as we get access to them.

### DIFF
--- a/style/question.less
+++ b/style/question.less
@@ -1,63 +1,31 @@
+@correct-answer-color: #78b042;
+@correct-answer-background: #e4f4b9;
+@wrong-answer-color: #c22032;
+@wrong-answer-background: #ffe0e0; 
+@answer-hover-color: #f9f9f9;
+@answer-select-color: #f0f0f0; // darker than hover
+
 .question {
     counter-reset: answer 0;
 
-    &.has-correct-answer {
-      .answer-input-box { display: none; }
-
-      .answer-correct .answer-label { background-color: #0f0; }
-      .answer-checked:not(.answer-correct) .answer-label { background-color: #f00; }
-    }
-
-    .answers-table {
-      display: table;
-      border-collapse: collapse;
-
-      border-top: 1px dotted #ccc;
-      margin-top: 1rem;
-
-      .answers-answer {
-        counter-increment: answer 1;
-        display: table-row;
-
-        .answer-letter {
-          display: table-cell;
-          width: 2rem;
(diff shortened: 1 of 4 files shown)


style/question.less
```diff
@@ -86,40 +104,115 @@
...
+      margin: 0.5em 0;
+      padding-left: 0;
+      line-height: 1.5em;
+      display: inline-block;
       width: 100%;
+      float:left;
+      transition: background-color .2s ease-in;
+      transition: padding-left .2s ease-in;
+
+      //Materialize hack
```
philschatz: Are these hacks still used?

style/question.less
```diff
@@ -86,40 +104,115 @@
...
+    &.has-correct-answer {
+      .answer-input-box { display: none; }
+
+      .answers-table .answers-answer .answer-label {
+        padding-left: 10px;
+      }
 
+      .answers-table .answer-correct:before,
+      .answers-table .answer-checked:not(.answer-correct):before { 
+        content:"\f00c";
```
philschatz: Can you pull this from the `@fa-*` variables from the `font-awesome.less`?

style/question.less
```diff
@@ -1,63 +1,81 @@
...
+    line-height: 150%;
+  }
+
+  .free-response,
+  .free-response:before,
+  .question-feedback,
+  .question-feedback:before {
+    font-size: 0.75em;
+  }
+  .free-response:before {
```
philschatz: Could you put this inside the `.free-response {` selector?

```
.free-response {
  &::before {
    ...
  }
}
```

style/question.less
jivewise: Ah, that's a good idea, I'll check.

style/question.less
jivewise: I'm using the transitions, the Materialize hacks, I haven't tried taking them out, but I will to see if it makes a difference.

jivewise:
screen shot 2015-02-13 at 5 30 48 am
Here's a screenshot of what it looks like with long answers and feedback.

style/question.less
jivewise: Yep, just did that.  Thanks for the suggestion.

jivewise:
Uploading a screenshot of the answering state

screen shot 2015-02-13 at 6 01 43 am

PR #23, opened 2015-02-19
adding in ability to open external reading links in new window

jivewise:
screen shot 2015-02-19 at 2 29 09 pm

Adding in a target blank for external links in the reading html (anything that isn't a fragment).  Let me know if I'm missing some logic that I didn't think of.


src/components/html.cjsx
```diff
@@ -12,3 +12,10 @@ module.exports = React.createClass
       <div className={classes.join(' ')} dangerouslySetInnerHTML={__html:@props.html} />
     else
       <span className={classes.join(' ')} dangerouslySetInnerHTML={__html:@props.html} />
+
+  componentDidMount: ->
```
philschatz: Can you include $ = require 'jquery'?

I've had a preference for not including jQuery unless necessary (so far it was only used for `$.ajax`) especially since we are only targeting modern browsers. we can chat about it though ; )

jivewise: Sure, just added it.

PR #25, opened 2015-02-20
adding in moment.js and fomatting due at time

jivewise:
screen shot 2015-02-20 at 4 03 16 pm

Adding in moment.js and formatting due_at timestamp.  Is there a better way of doing this than in the view directly?


bower.json
```diff
@@ -24,6 +24,7 @@
     "jquery": "~2.1.1",
     "bootstrap": "~3.2.0",
     "katex": "~0.1.1",
-    "bootstrap-material-design": "~0.2.2"
+    "bootstrap-material-design": "~0.2.2",
+    "moment": "~2.9.0"
```
philschatz: Is the bower `moment.js` used anywhere?

src/components/task.cjsx
```diff
@@ -105,7 +105,7 @@ module.exports = React.createClass
           {breadcrumbs}
           <BS.Panel bsStyle="default" footer={footer}>
             <h1>{model.title}</h1>
-            <p>Due At: {model.due_at}</p>
+            <p>Due At: {moment(model.due_at).format('LLL')}</p>
```
philschatz: Should we have a `<Time>` component that generates a `<time>` HTML element? That way date/time formatting is done consistently throughout the app?

jivewise: Sure, I can do that.

bower.json
jivewise: Nope I'll remove it.

PR #27, opened 2015-02-23
adding in a simpler animation and style changes for breadcrumbs

jivewise:
animation-prototype

Adds a small animation (and colors) to breadcrumb steps.

@fredasaurus: Thoughts on if it should be included in the codebase?


src/components/breadcrumbs.cjsx
```diff
@@ -30,12 +30,16 @@ module.exports = React.createClass
...
         classes.push('active')
+        bsStyle = 'primary'
         # classes.push('disabled')
         title = "Current Step (#{step.type})"
 
       if step.is_completed
         classes.push('completed')
-        bsStyle = 'primary'
+        if step.correct_answer_id is step.answer_id or not step.correct_answer_id
+          classes.push('btn-success')
```
philschatz: Should this just be `bsStyle = 'success'`? Is there a need for both `btn-primary` and `btn-success` to be on the element?

jivewise: No, I just use that class in the css to get the correct color, for the current step.  Bad naming, I'll update to reflect it as we go further.

jivewise:
Looking for feedback on this animation.  It's far simpler, in terms of code and the animation.  Do we want something more complex?

philschatz:
Closing this for now. may reuse code once we start putting animations back in.

PR #28, opened 2015-02-25
ireading, question and answer styles, navbar styles

Fredasaurus:
Updates to ireading, and question answer styles. Also changes to navbar, logo-brand svg.

screen shot 2015-02-27 at 5 16 30 pm


index.html
```diff
@@ -2,6 +2,8 @@
 <html>
   <head>
     <link rel="stylesheet" href="/dist/tutor.css" />
+    <link href='http://fonts.googleapis.com/css?family=Lato:400,300,300italic,400italic,900,700' rel='stylesheet' type='text/css'>
+    <link href='http://fonts.googleapis.com/css?family=Architects+Daughter' rel='stylesheet' type='text/css'>
```
philschatz: Can you include these in the CSS file instead? This file is only used for mockup-testing.

style/buttons.less
```diff
@@ -0,0 +1,20 @@
+@btn-border-radius: 4px;
+@primary-background: #f47642;
+@primary-hover-background: #fa804d;
+
+// Override the default text color from bootstrap-material-design
+.btn-primary,
+.btn-success { color: @white !important; }
+
+.btn-primary {
+	font-size: 18px;
```
philschatz: Looks like dreamweaver uses tabs instead of spaces. Can you change these to 2-spaces?

index.html
Fredasaurus: Cool, where to put these was one of the questions I had for you.

style/buttons.less
Fredasaurus: erg, I set the preference in dw to tab= two spaces. Looks like that isn't recognized elsewhere, annoying.

style/buttons.less
```diff
@@ -0,0 +1,20 @@
+@btn-border-radius: 4px;
+@primary-background: #f47642;
+@primary-hover-background: #fa804d;
+
+// Override the default text color from bootstrap-material-design
+.btn-primary,
+.btn-success { color: @white !important; }
```
philschatz: Why is the `!important` necessary here? Either:

1. `bootstrap-material-design`'s selector is more specific *or*
2. `@import 'bootstrap-material-design';` occurs after this file

If 1. then this selector should probably be as specific as the original (unless it's really ugly). If 2. then the import should be moved up.

style/question.less
```diff
@@ -23,23 +24,28 @@
...
 
     &:before {
       content: "My Answer:";
       display: inline-block;
       float: left;
-      margin-left: -90px;
-      color: #999;
+      margin-left: -145px;
+			font-family: 'Architects Daughter', cursive;
+			font-size: .9em;
```
philschatz: Hmm, @jivewise : should we be using rem's instead of em's (to avoid cascading font size changes)?

style/buttons.less
Fredasaurus: Actually, I don't know. Just moved this code from tutor.less to buttons.less. I'll make the change.

style/question.less
```diff
@@ -23,23 +24,28 @@
...
+    background: #f5f5f5;
 
     &:before {
       content: "My Answer:";
       display: inline-block;
       float: left;
-      margin-left: -90px;
-      color: #999;
+      margin-left: -145px;
+			font-family: 'Architects Daughter', cursive;
```
philschatz: This should probably be in a mixin

style/question.less
Fredasaurus: I had the same question. I prefer rems.

index.html
philschatz: I think it should go in a separate less file which defines 2 mixins (for both of the fonts). When are they "generally" used?

Fredasaurus: OK, What I'm looking at is this:
Lato - headings and question typography
Some kind of handwriting font (currently trying Architects Daughter) - margin comments and label
Merriweather - book font

Fredasaurus: Should I combine with H1, h2, etc styles in a file called typography?
Or should typographic styles for things be rolled up into a whatever view they are included in (i.e. dashboard, ireading, etc.)?

style/question.less
jivewise: Yea, rems would be better.

PR #29, opened 2015-02-26
Ireading for teachers

jivewise:
screen shot 2015-02-26 at 11 46 09 pm
screen shot 2015-02-26 at 11 45 58 pm
screen shot 2015-02-26 at 11 45 47 pm

Adding in initial component for creating iReadings.

### DIFF
--- a/src/components/teacher/reading.cjsx
+++ b/src/components/teacher/reading.cjsx
@@ -62,12 +62,13 @@ Reading = React.createClass
     <BS.Panel bsStyle="default" className="create-reading" footer={footer}>
       <h1>{headerText}</h1>
       <div>
-        <label for="title">Name</label>
+        <label htmlFor="title">Name</label>
         <input id="title" type="text" value={@props.title}/>
       </div>
       <div>
-        <label for="due-date">Due Date</label>
+        <label htmlFor="due-date">Due Date</label>
         <Datepicker 
+          id="due-date"
           format="MMM dd, yyyy" 
           time={false} 
           calendar={true} 
--- a/src/components/teacher/reading.cjsx
+++ b/src/components/teacher/reading.cjsx
@@ -1,0 +1,84 @@
+React = require 'react'
+BS = require 'react-bootstrap'
+Router = require 'react-router'
+Widgets = require 'react-widgets'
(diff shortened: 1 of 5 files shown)


src/components/teacher/reading.cjsx
```diff
@@ -0,0 +1,84 @@
...
+    @setState({
+      isModalOpen: !@state.isModalOpen
+    })
+
+  render: ->
+    <BS.Button onClick={@handleToggle} bsStyle="primary">Edit Readings</BS.Button>
+
+  renderOverlay: ->
+    if !@state.isModalOpen
+      return <span/>
```
philschatz: Would an empty return still work?

src/components/teacher/reading.cjsx
```diff
@@ -0,0 +1,84 @@
...
+    dueDate = if @state.due_at then @state.due_at else @props.due_at
+
+    <BS.Panel bsStyle="default" className="create-reading" footer={footer}>
+      <h1>{headerText}</h1>
+      <div>
+        <label for="title">Name</label>
+        <input id="title" type="text" value={@props.title}/>
+      </div>
+      <div>
+        <label for="due-date">Due Date</label>
```
philschatz: Should this be `htmlFor="..."`?

src/router.coffee
```diff
@@ -10,6 +11,8 @@ routes = (
     <Route path='dashboard' name='dashboard' handler={Dashboard} />
     <Route path='tasks' name='tasks' handler={Tasks} />
     <Route path='tasks/:id' name='task' handler={SingleTask} />
+    <Route path='reading' name='createReading' handler={CreateReading} />
```
philschatz: This should probably be plural to match `/tasks/`

jivewise: I updated it to be /readings, but still like the non-plural form better :).

I also noticed something weird with routes; if I go to /readings/ or /tasks/, I get the invalid state.  Is there a way we can avoid having to add duplicate routes for those?

src/components/teacher/reading.cjsx
jivewise: No, looks like we need to return something there in order for this to work.  Just the way the react-bootstrap modal component works.

PR #34, opened 2015-03-05
Create ireading bugs

jivewise:
No UI changes

1. Fixed CrudConfig to be a function that has it's own scope
2. Fixed task store hasTopics/removeTopics to use settings.page_ids
3. Fixed error when user clears the date field

### DIFF
--- a/src/flux/task-plan.coffee
+++ b/src/flux/task-plan.coffee
@@ -1,5 +1,4 @@
 _ = require 'underscore'
-moment = require 'moment'
 {CrudConfig, makeSimpleStore, extendConfig} = require './helpers'
 
 CREATE_KEY = "CREATING"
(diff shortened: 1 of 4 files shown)


src/flux/task-plan.coffee
```diff
@@ -18,7 +19,7 @@ TaskPlanConfig =
     _.extend(plan, {title})
     @emitChange()
 
-  updateDueAt: (id, due_at) ->
+  updateDueAt: (id, due_at=new moment()) ->
```
philschatz: Why set this to `new moment()` instead of `new Date()`?

jivewise: Because I wasn't thinking :(

Updated to use Date.

PR #37, opened 2015-03-10
publish button greyed out unless add reading form complete

darrenmason:
tested with @state.id.  datetimepicker can be emptied after population but retains last value.

### DIFF
--- a/src/components/task-plan/reading.cjsx
+++ b/src/components/task-plan/reading.cjsx
@@ -116,10 +116,20 @@ ReadingPlan = React.createClass
 
   render: ->
     id = @getParams().id
-    footer = <BS.Button bsStyle="primary" onClick={@publish}>Publish</BS.Button>
+
+    if id
+      plan = TaskPlanStore.get(id)    
+    else
+      plan = TaskPlanStore.get(@state.id)
+     
+    if plan.title and plan.due_at and plan.settings.page_ids.length > 0
+    then btnDisable = '' else btnDisable = 'disabled'
+     
+    footer = <BS.Button bsStyle="primary" className={btnDisable} onClick={@publish}>Publish</BS.Button>
     headerText = if id then 'Edit Reading' else 'Add Reading'
     dueDate = if @state.due_at then @state.due_at else @props.due_at
     topics = TaskPlanStore.getTopics(@state.id)
+    
 
     selectedReadingList =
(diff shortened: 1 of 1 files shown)


darrenmason:
image

src/components/task-plan/reading.cjsx
```diff
@@ -116,10 +116,20 @@ ReadingPlan = React.createClass
 
   render: ->
     id = @getParams().id
-    footer = <BS.Button bsStyle="primary" onClick={@publish}>Publish</BS.Button>
+
+    if id
+      plan = TaskPlanStore.get(id)
```
philschatz: is `@state.id` ever not the same as `id`?

src/components/task-plan/reading.cjsx
```diff
@@ -116,10 +116,20 @@ ReadingPlan = React.createClass
...
     id = @getParams().id
-    footer = <BS.Button bsStyle="primary" onClick={@publish}>Publish</BS.Button>
+
+    if id
+      plan = TaskPlanStore.get(id)    
+    else
+      plan = TaskPlanStore.get(@state.id)
+     
+    if plan.title and plan.due_at and plan.settings.page_ids.length > 0
+    then btnDisable = '' else btnDisable = 'disabled'
```
philschatz: FYI: There is already an existing syntax in tutor-js for optionally adding classes: https://github.com/openstax/tutor-js/blob/master/src/components/question.cjsx#L36

darrenmason: ok I will update syntax.  Also I noticed this module may be helpful : https://github.com/JedWatson/classnames

It was originally here : http://facebook.github.io/react/docs/class-name-manipulation.html

philschatz: Yep, want to make a separate Pull request that changes all `className={classes.join(' ')}` to use `classes`?

src/components/task-plan/reading.cjsx
darrenmason: console.log on initial load in render:

getParams().id: undefined
@state.id: _CREATING_0
getParams().id: undefined
@state.id: _CREATING_0

it seems getParams() checks only the location.href?

philschatz: Yep, that's correct. `@getParams()` only pulls from the `window.location`.

PR #38, opened 2015-03-11
Update classes syntax

darrenmason:


philschatz:
Oh, I was suggesting switching them to use the `classset` (`cx`) package. but thanks for making them more consistent ; )

darrenmason:
ah :-)

PR #39, opened 2015-03-11
Render Math in iReading book content

philschatz:
image

Optionally run MathJax on any arbitrary HTML content if it is loaded.

need to update tutor-server to load MathJax as well


Dantemss:
Giving up on Katex?

philschatz:
hehe, not yet ; ) CNX content uses MathML which katex does not support at all but this is probably the first step in the death of `katex` ; (

Dantemss:
MathML? I was just planning on leaving the $$ in the content :D

index.html
```diff
@@ -2,6 +2,7 @@
 <html>
   <head>
     <link rel="stylesheet" href="/dist/tutor.css" />
+    <script src="http://cdn.mathjax.org/mathjax/2.3-latest/MathJax.js?config=MML_HTMLorMML"></script>
```
philschatz: This same line would need to be in exercises. Also, should probably switch to 2.5. I pulled this from cnx production but I think they're moving to 2.5

Dantemss: While discussing this, let me recommend http://stackoverflow.com/a/3832530 just in case MathJax's CDN is offline.

Dantemss: Also let me point out that Rails' philosophy (which you may or may not want to follow) is to compile all JS into one file (and same for the CSS), so that it is downloaded in one request (less overhead and stays cached after the first page load).

philschatz: Yep, that's the major reason why I tried to avoid MathJax. It has its own loader (it made a loader before loaders/combiners existed) which cannot be combined into one file plus it loads images/fonts/etc depending on the browser. It's so difficult to include that cnx (which also minifies into 1 file) tried to host a local copy of MathJax (instead of the CDN) and gave up.

Dantemss: :)

Dantemss: MathJax claims on their site that you can install it with Bower (and I assume that means you can serve it with gulp?) since version 2.3, so you could give that a try.

philschatz:
so, if there are no other objections, @Dantemss can you (or someone) make a PR to `tutor-server` adding the same line as the one in `index.html` to include MathJax?

Dantemss:
Do we need to change something to make it render the <span data-math'=blah'> or is this all we need?

philschatz:
`tutor-server` will need to add a `<script>` tag to load MathJax similar to the one in `index.html`

PR #42, opened 2015-03-12
WIP keep track of unsaved changes in the stores

philschatz:
no UI changes.

This is useful for `PATCH` to only send the unsaved changes and for “Cancel” buttons or confirm dialogs when moving away from a page.

Each CRUD Store now has `.isChanged(id)`, `.getChanged(id)` and a `.get(id)` contains a combination of the unchanged fields and the locally changed ones.

Each CRUD Actions set now has an internal `._change(id, obj)`.

### DIFF
--- a/Gulpfile.coffee
+++ b/Gulpfile.coffee
@@ -70,7 +70,7 @@ buildTests = (isWatching) ->
   buildBrowserify(srcPath, destDir, destFile, isWatching)
 
 
-gulp.task 'test', ['build'], (done) ->
+gulp.task 'test', ['buildJS'], (done) ->
   buildTests(false)
   .on 'end', ->
     config =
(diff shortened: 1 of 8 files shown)


Gulpfile.coffee
```diff
@@ -70,7 +70,7 @@ buildTests = (isWatching) ->
   buildBrowserify(srcPath, destDir, destFile, isWatching)
 
 
-gulp.task 'test', ['build'], (done) ->
+gulp.task 'test', ['buildJS'], (done) ->
```
lakshmivyas: Oops.   Sorry about missing this one.

jivewise:
Merging, though I have a feeling this will conflict with one of mine.

PR #45, opened 2015-03-12
Ireading loading deleting

jivewise:
screen shot 2015-03-12 at 10 36 45 am
deleting-ireading

1. Adding in loading of an iReading to edit data
2. Adding in delete action of an iReading
3. General refactoring to keep jsx clean


src/flux/helpers.coffee
```diff
@@ -82,6 +84,12 @@ CrudConfig = ->
       @_local[result.id] = result
       @emitChange()
 
+    delete: (id) ->
+      @_asyncStatus[id] = DELETING
+
+    deleted: (result, id) ->
+      @_asyncStatus[id] = DELETED
+      @_local[id] = null
```
philschatz: This should probably be `delete @_local[id]` and you should call `@emitChange()` so any views (like a listing of plans) gets updated. I don't think that view exists yet, but it will in this sprint.

src/components/task-plan/reading.cjsx
```diff
@@ -38,7 +40,21 @@ SelectTopics = React.createClass
     return
 
   render: ->
-    <BS.Button onClick={@handleToggle} bsStyle="primary">Edit Readings</BS.Button>
+    selectedReadingList =
+      <ul className="selected-reading-list">
+        <li><strong>Currently selected sections in this reading</strong></li>
+        {_.map(@props.selected, @renderTopics)}
+      </ul> if @props.selected.length && TocStore.isLoaded()
```
philschatz: `and` instead of `&&`? Plus, the whole predicate thing

src/flux/toc.coffee
```diff
@@ -25,7 +25,8 @@ TocConfig =
     get: ->
       @_toc or throw new Error('BUG: Invalid course')
     getSectionInfo: (sectionId) ->
-      @_sections[sectionId] or throw new Error('BUG: Invalid section')
+      if (@_toc && @_sections)
```
philschatz: `and` instead of `&&`?

philschatz:
also, can you include an example JSON in `/api/courses/1/plans/1` ?

api/courses/1/plans/1
```diff
@@ -0,0 +1 @@
+{"id":17,"type":"reading","title":"Jing","opens_at":"2015-03-04T16:40:23.796Z","due_at":"2015-03-13T04:00:00.000Z","settings":{"page_ids":[3,2,1]}}
```
philschatz: all one line? (minor)

src/components/task-plan/reading.cjsx
```diff
@@ -38,7 +40,22 @@ SelectTopics = React.createClass
     return
 
   render: ->
-    <BS.Button onClick={@handleToggle} bsStyle="primary">Edit Readings</BS.Button>
+    if (@props.selected.length and TocStore.isLoaded())
```
philschatz: There should probably be an `else` with `selectedReadingList = <div className="loading">Loading...</div>`

Also, (minor) what if `TocActions.load()` errors?

src/components/task-plan/reading.cjsx
```diff
@@ -82,91 +99,99 @@ SelectTopics = React.createClass
       </div>
     </BS.Modal>
 
+ReadingFooter = React.createClass
+  render: ->
+    classes = []
+    classes.push('disabled') unless @props.enabled
```
philschatz: Is `enabled` defined in `propTypes`? Maybe this concern should be a separate PR. Or, this logic should be inside the footer instead of as a Prop.

src/components/task-plan/reading.cjsx
```diff
@@ -82,91 +99,99 @@ SelectTopics = React.createClass
...
+  publishPlan: ->
+    id = @getPlanId()
     TaskPlanActions.publish(id)
     @transitionTo('editReading', {id})
 
-  renderTopics: (topicId) ->
-    topic = TocStore.getSectionInfo(topicId)
-    <li>{topic.number} - {topic.title}</li>
+  deletePlan: () ->
+    id = @getPlanId()
```
philschatz: Should this have a `confirm('Are you sure you want to delete this?')`?

src/components/task-plan/reading.cjsx
```diff
@@ -82,91 +99,99 @@ SelectTopics = React.createClass
...
+    @transitionTo('dashboard')
+    TaskPlanActions.delete(id)
+    {}
```
philschatz: What is the purpose of this line?

src/components/task-plan/reading.cjsx
```diff
@@ -82,91 +99,99 @@ SelectTopics = React.createClass
...
+    topics = TaskPlanStore.getTopics(id)
+    
+    footer= <ReadingFooter enabled={isEnabled} onPublish={@publishPlan} onDelete={@deletePlan}/>
 
-    else
-      <div className="loading">Loading...</div>
+    <BS.Panel bsStyle="default" className="create-reading" footer={footer}>
+      <h1>{headerText}</h1>
+      <div>
+        <label htmlFor="title">Name</label>
```
philschatz: This `id` should probably be more unique than just "title". Could it be `reading-title`?

src/components/task-plan/reading.cjsx
```diff
@@ -82,91 +99,99 @@ SelectTopics = React.createClass
...
+        <input ref="title" id="title" type="text" onChange={@setTitle} value={plan?.title}/>
+      </div>
+      <div>
+        <label htmlFor="due-date">Due Date</label>
+        <DateTimePicker
+          id="due-date"
```
philschatz: is the `id` necessary? If so, could it be prefixed (since it is not used elsewhere) with `reading-`?

api/courses/1/plans/1
jivewise: Copied straight from chrome :)

I'll format it.

src/components/task-plan/reading.cjsx
jivewise: Hm, if it errors, the button will likely not display...we may need an on error handler.  We likely need to think about error handling as a whole for that app.

I'll add an error container though.

src/components/task-plan/reading.cjsx
jivewise: Do we need a confirm for delete actions?  Is this really where the delete button for readings go?  If so, I can add it in.

src/components/task-plan/reading.cjsx
jivewise: I felt like they needed to be transitioned to another state after they deleted the reading plan.  We may want to listen to the store and see if it is deleted instead.

src/components/task-plan/reading.cjsx
jivewise: Do you mean the label for?  I just added that in there for clarity a while back, I don't really think it's necessary.  I'll take it out.

src/components/task-plan/reading.cjsx
jivewise: Yea, I'll rename it.

src/components/task-plan/reading.cjsx
jivewise: I'll add in the propTypes, but we should probably have a sprint task/pull request somewhere to clean those up too.

src/components/task-plan/reading.cjsx
philschatz: right, that's the reason for `@transitionTo` (which should probably be after the delete action). but doesn't explain the `{}`

src/components/task-plan/reading.cjsx
philschatz: the `htmlFor` is actually important for accessibility. It links the label (which gets focus) with the element (when you press `space` to check/uncheck/focus the element).

src/components/task-plan/reading.cjsx
philschatz: It's OK if the button does not display; I was just concerned about the async state between loading the ToC (or plan) and letting the user edit it.

src/components/task-plan/reading.cjsx
philschatz: Yep, agreed. and then moving forward `props` are always defined in `proptypes` first.

src/components/task-plan/reading.cjsx
jivewise: Oh, hm...not sure why that's there.  I'll remove that as well.

src/components/task-plan/reading.cjsx
jivewise: Okay, I'll change that as I update the id.

PR #50, opened 2015-03-16
Add Teacher Task Plans listing

nathanstitt:
Names and file locations very much subject to change, I'm not at all sure if I've implemented any of this correctly :)

screen shot 2015-03-16 at 10 14 03 am

### DIFF
--- a/src/api.coffee
+++ b/src/api.coffee
@@ -139,7 +139,7 @@ start = ->
       TaskActions.loadedUserTasks(results.items)
 
   apiHelper TeacherTaskPlanActions, TeacherTaskPlanActions.load, TeacherTaskPlanActions.loaded, 'GET', (course_id) ->
-    url: "/api/courses/#{course_id}/plans.json"
+    url: "/api/courses/#{course_id}/plans"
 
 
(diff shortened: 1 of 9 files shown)


src/api.coffee
```diff
@@ -127,6 +128,9 @@ start = ->
     .then (results) ->
       TaskActions.loadedUserTasks(results.items)
 
+  apiHelper TeacherTaskPlanActions, TeacherTaskPlanActions.load, TeacherTaskPlanActions.loaded, 'GET', (course_id) ->
+    url: "/api/courses/#{course_id}/plans.json"
```
philschatz: I noticed what you had to do here because you couldn't create a `/plans` file. #51 should fix that little oversight on my part :blush:

nathanstitt:
I should also note that I didn't spend any time on the UI, since I figured that it would be moved to the calendar display at some point.  If there's an interim look you'd prefer let me know and I can implement it.

philschatz:
- changed the dashboard to link to the reading list
- added the "Add a Reading" to the bottom of the list
- made the items in the list clickable

image


image

PR #54, opened 2015-03-17
Adding in ConfirmLeaveMixin

jivewise:
screen shot 2015-03-17 at 12 29 35 pm


src/components/confirm-leave-mixin.cjsx
```diff
@@ -0,0 +1,12 @@
+$ = require 'jquery'
+
+module.exports =
+  # getFlux: -> {store, actions}
+  # getId: -> id 
+  componentDidMount: ->
+    $(window).on("beforeunload", @confirmLeave)
```
philschatz: Should this be unmounted at any point?

philschatz: oh, and instead of jQuery you can use:

```coffee
window.addEventListener 'beforeunload', (e) ->
  msg = 'Are you sure?'
  (e or window.event).returnValue = msg     # Gecko and Trident
  msg                                        # Gecko and WebKit
```

jivewise: Should we avoid jQuery?  I told Fred the other day to go ahead and use it where it was convenient.

philschatz: I've been avoiding it because:

1. we're targeting modern browsers
2. if has a bunch of tasks that perform DOM manipulation which is mostly nullified because of Reactand would lead to questions like "when should I use `$el.append()` or `$el.addClass()`"

jivewise: I feel like it's not bad in cases when:
1. We're not manipulating Dom
2. jQuery is saving us from additional logic

I'm not sure what the best solution in this case is.  Is there a jquery-like library without dom manipulation?

philschatz:
I noticed the "Delete" footer button is available but should probably be hidden if `TaskPlanStore.isNew(id)` and the save as draft button is completely hidden until all the fields are filled in. It should probably be disabled instead.

image

jivewise:
We do hide the delete button on isNew, but since we create a reading immediately, we never get the chance to hide the button.  

I'll update the Save button behavior though, and add some tests soon.  I had some tests stashed away somewhere, but I think they're likely out of date now.

PR #57, opened 2015-03-18
Updating gulp to use watchify instead of gulp watch

jivewise:
I want to leave this as a WIP for now until I'm sure it doesn't case funniness while developing.  Feel free to cherry pick the change locally and use it there though


philschatz:
Does it restart when you edit a `src/` file as well as a `test/` file?

jivewise:
It does the tests I believe, but I think I need to test more...

jivewise:
One thing that should be noted is that I changed the gulp serve dependencies to be watchJS instead of buildJS.  This is under the assumption that we will never use gulp serve in production.  If we are, I'll have to make some changes...

lakshmivyas:
@jivewise - We won't be using `gulp serve` in production.

jivewise:
Great, thanks @lakshmivyas!

Gulpfile.coffee
```diff
@@ -82,7 +82,7 @@ gulp.task 'test', ['buildJS'], (done) ->
 
   return # Since this is async
 
-gulp.task 'tdd', ['build'],  (done) ->
+gulp.task 'tdd', (done) ->
```
philschatz: Why remove the build from tdd?

Gulpfile.coffee
```diff
@@ -92,6 +92,12 @@ gulp.task 'tdd', ['build'],  (done) ->
   return # Since this is async
 
 gulp.task 'buildJS', ['cleanJS'], -> build(false)
+gulp.task 'watchJS', ['cleanJS'], (done)->
+  build(true)
+  buildTests(true).on 'end', ->
```
philschatz: Does the build need to complete before karma starts up?

Gulpfile.coffee
```diff
@@ -92,6 +92,12 @@ gulp.task 'tdd', ['build'],  (done) ->
...
+    config =
+      configFile: __dirname + '/test/karma.config.coffee'
+    karma.server.start(config, done)
```
philschatz: Could you pull this bit out into a shared function since test, tdd, and watchJS probably all have similar code (but different args to build/buildTests)

Gulpfile.coffee
jivewise: I'm not entirely sure we need a tdd task.  It is basically equivalent to the watchJS task.  I think one of them should get taken out. TDD does sound better IMO.

Gulpfile.coffee
jivewise: Good point, I'll try to add that on the end event

Gulpfile.coffee
jivewise: This is now in a separate function.  The singleRun option in karma would stop if it was false.  Now that is being accounted for.

PR #58, opened 2015-03-18
reading stats - adjusted progress bars, added spaced practice

darrenmason:
image

This PR renders progress bars for course overall, current_pages and spaced_pages.  It is currently not rendering periods.  Percent change is calculated on spaced_pages based on previous_attempt.

current issues:

1) the way either bootstrap progressbar or react-bootstrap calculates percentage width is causing smaller segments to overflow:

image

2) the overall course stats needs stat: 'not started' and/or an accurate total_count to render the progress bar properly.

### DIFF
--- a/src/components/task-plan/reading-stats.cjsx
+++ b/src/components/task-plan/reading-stats.cjsx
@@ -4,27 +4,17 @@ BS = require 'react-bootstrap'
 Router = require 'react-router'
 
 {TaskPlanStore, TaskPlanActions} = require '../../flux/task-plan'
+LoadableMixin = require '../loadable-mixin'
 
 Stats = React.createClass
-  mixins: [Router.State, Router.Navigation]
+  mixins: [Router.State, Router.Navigation, LoadableMixin]
 
-  getInitialState: ->
-    id = @getParams().id
-    if (id)
-      TaskPlanActions.load(id)
-    else
-      id = TaskPlanStore.freshLocalId()
-      plan = TaskPlanActions.create(id, due_at: new Date())
+  getFlux: ->
+    store: TaskPlanStore
+    actions: TaskPlanActions
 
-    {id}
+  getId: -> @getParams().id
 
-  componentWillMount: -> TaskPlanStore.addChangeListener(@update)
-  componentWillUnmount: -> TaskPlanStore.removeChangeListener(@update)
-
-  getPlanId: () ->
-    @getParams().id or @state.id
-
-  update: -> @setState {}
 
   percent: (num,total) ->
     Math.round((num/total) * 100)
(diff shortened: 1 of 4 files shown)


src/components/task-plan/reading-stats.cjsx
```diff
@@ -0,0 +1,117 @@
...
+Router = require 'react-router'
+
+{TaskPlanStore, TaskPlanActions} = require '../../flux/task-plan'
+
+Stats = React.createClass
+  mixins: [Router.State, Router.Navigation]
+
+  getInitialState: ->
+    id = @getParams().id
+    if (id)
```
philschatz: I think the reading stats only apply to an existing reading so the following code is never executed

src/components/task-plan/reading-stats.cjsx
```diff
@@ -0,0 +1,117 @@
...
+          <BS.ProgressBar className="reading-progress-bar" bsStyle="success" label="%(now)s%" now={@percent(data.correct_count,data.correct_count+data.incorrect_count)} key={1} />
+        </BS.ProgressBar>
+        <div className="reading-progress-delta">{@percentDelta(data.correct_count,data.previous_attempt.correct_count)}% change</div>
+      </div>
+    </div>
+
+  render: ->
+    id = @getPlanId()
+
+    if TaskPlanStore.isLoaded(id)
```
philschatz: Did you try using the Loadable mixin to handle this case (and the `componentWill(Un)Mount`)?

darrenmason: yes, changed to loadable-mixin

src/components/task-plan/reading-stats.cjsx
darrenmason: removed

philschatz: Is it removed? This comment should disappear if the text was removed.

style/reading-stats.less
```diff
@@ -0,0 +1,42 @@
...
+
+
+// debug:
+
+.progress{
+  overflow:visible;
+}
+
+.progress-bar {
+    // overflow-x:visible;
```
philschatz: Are these for documentation? If not, they should probably be removed.

Fredasaurus:
Cool, didn't know we were working on this yet, thought it was a sprint 9 task.

Have a design for this here. http://fredasaurus.github.io/tutor-designs/calendar/calendar.html

screen shot 2015-03-19 at 1 56 09 pm

philschatz:
Thanks! @darrenmason : I copied your comment into the Pull Request description field.

Fredasaurus:
A couple comments:
1. We're not sure about how the first section will be visualized, working on that still. May be a pie chart.
2. Probably not going to use SPARFA with %delta for spaced practice, not accurate enough. MOst likely will be right / wrong.

darrenmason:
Eventually, will stats view need to be rendered both in its own page and in a modal?  If so will those two layouts be different, or could they pull from the same view?

Fredasaurus:
1. Yes Here, click on "Read Chapter 4" 
http://mockups.openstax.org/Tutor_MVP_v2/start.html#p=course_calendar_-_teacher
and here
http://mockups.openstax.org/Tutor_MVP_v2/start.html#p=reading_review_metrics_-_class
2. The view should be the same.

philschatz:
currently there is no modal (since there is no calendar) but for the sprint the BE was tasked with making the stats and we needed a place to show them so I opted for `/readings/1/stats`

Fredasaurus:
Also, I just used a bs modal for demo purposes.

PR #62, opened 2015-03-20
add unit tests for ReadingPlan

philschatz:
No UI changes.

This adds a couple unit tests for a ReadingPlan. Should be useful for making more unit tests

**Notes:**

- To test a Component I needed to refactor out part of it that used the `Router.State` mixin
- had to add classes to buttons to test if they were disabled

### DIFF
--- a/test/components/reading-plan.spec.coffee
+++ b/test/components/reading-plan.spec.coffee
@@ -1,0 +1,81 @@
+{expect} = require 'chai'
+
+React = require 'react'
+
+{TaskPlanActions, TaskPlanStore} = require '../../src/flux/task-plan'
+{TocActions} = require '../../src/flux/toc'
+{ReadingPlan} = require '../../src/components/task-plan/reading'
+
+TEST_TOC = [
+  {
+    id: 2
+    title: 'Chapter title'
+    type: 'part'
+    children: [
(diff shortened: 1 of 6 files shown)


jivewise:
Good start for testing. Do we also want to eventually add testing for interactions (like when a user adds/remove a topic)? It's hard for me to know where to draw the line in terms of testing.

test/components/reading-plan.spec.coffee
```diff
@@ -0,0 +1,81 @@
+{expect} = require 'chai'
+
+React = require 'react'
+
+{TaskPlanActions, TaskPlanStore} = require '../../src/flux/task-plan'
+{TocActions} = require '../../src/flux/toc'
+{ReadingPlan} = require '../../src/components/task-plan/reading'
```
jivewise: Should this be ReadingShell now that it's renamed elsewhere?

philschatz: Ah, the reason I split `ReadingPlan` into `ReadingShell` and `ReadingPlan` was so that I could test just the Plan part. Adding the code to mock the ReactRouter would have been non-trivial and since most of the testing was on the interactions rather than how the id was loaded I opted for splitting the file up

PR #63, opened 2015-03-20
fix malformed URL when completing a step

philschatz:
No UI Change (except that UI now works again ; )

fixes the `GET /steps/[Object object]` error @Fredasaurus encountered when backend changed what was returned from `PUT /steps/1/completed`


pandafulmanda:
lol I was looking for this too

PR #64, opened 2015-03-20
Add video step

philschatz:
The video shows up (bug) and has a link to the video below.

I think the video should not show up because the whole `<iframe>` should be removed but it works for now.

image

### DIFF
--- a/src/components/task-step/all-steps.cjsx
+++ b/src/components/task-step/all-steps.cjsx
@@ -43,7 +43,9 @@ Video = React.createClass
   renderBody: ->
     {id} = @props
     {content_html, content_url} = TaskStepStore.get(id)
-    <ArbitraryHtmlAndMath className="-video-content" html={content_html} />
-    <a target="_top" src={content_url} />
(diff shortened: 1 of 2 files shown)


Dantemss:
So what behavior would you ideally want from the backend?

- Just remove the iframe around the video
- Just return the video url
- Replace the video content with our own (based on the video url)

philschatz:
I think, initially just keep the iframe in because it seems (as of this morning) that some videos are embeddable but others are not. And, I'm not sure how best to convey that information.

So far, the 2 options I think are:

1. If video is embeddable, just keep the HTML with the iframe code. otherwise populate the `content_url`
  - that way BE does not have to extract the height/width info and store it in the TaskStep
  - BE does not need to have an additional `is_embeddable` field in the TaskStep
2. If the video is embeddable, extract the height/width/src and add it to the TaskStep as fields specific to a video.

Dantemss:
Hm. Can you link an example of a non-embeddable video that we have?

PR #68, opened 2015-03-23
enable CORS to load font icons

philschatz:
The icons now load when the JS is loaded from `tutor-server`'s port (`:3001`). Also, `gulp serve` no longer opens up a browser window.

image


pandafulmanda:
:+1:

PR #69, opened 2015-03-24
tests for crud store for extending the crudstore and using the _loaded function

pandafulmanda:
3 tests added in crud-store spec

- [x] Extended action function
- [x] Extended store function
- [x] _loaded will change what is stored

### DIFF
--- a/test/crud-store.spec.coffee
+++ b/test/crud-store.spec.coffee
@@ -2,9 +2,18 @@
 
 {CrudConfig, makeSimpleStore, extendConfig} = require '../src/flux/helpers'
 
-CrudConfig = CrudConfig()
+TestCrudConfig = CrudConfig()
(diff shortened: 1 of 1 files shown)


philschatz:
Thanks! Does this test when:

- `_loaded` does not exist
- `_loaded` returns `falsy`

pandafulmanda:
no, this does not, i can add those.  although, what is the desired behavior when loaded returns falsy?  i'd imagine we don't really want to assume that _local[id] should be set to falsy if that is the case?

philschatz:
Right, I think `@_local[id] = @_loaded?(obj, id) or obj` should cover that case (just use the original `obj`)

pandafulmanda:
OH right =).  okay I'll add these as well

PR #70, opened 2015-03-24
Try another button

jivewise:
screen shot 2015-03-25 at 12 05 34 am

Adding in try another button and integration with backend.


jivewise:
This is running with tutor-server#master, to test, you'll likely have to do a: rake db:drop && rake db:migrate sprint:008

The Try Another problem is the 6th problem in the 3rd task.

jivewise:
You can also run the following in the ruby console to mark the problems as incomplete and see the try another button again, after going through it once, thought I'd recommend running the rake task to test thoroughly.

    te = TaskedExercise.find(15)

    te.recovery_tasked_exercise = TaskedExercise.new(url: 'http://exercises.openstax.org/exercises/70@1', content: '{"uid":"70@1","number":70,"version":1,"published_at":"2015-03-10T19:49:18.935Z","editors":[],"authors":[{"user_id":1}],"copyright_holders":[{"user_id":2}],"derived_from":[],"attachments":[],"tags":["k12phys-ch04-s03-lo01","k12phys-ch04-ex070","chapter-review-short-answer","tutor-only","dok2","time-short","display-free-response","display-rec-sensitive"],"stimulus_html":"","questions":[{"id":148,"stimulus_html":"","stem_html":"The same force is applied in the same direction to two objects at rest. Which of these will undergo greater acceleration?","answers":[{"id":"532","content_html":"The one with large size."},{"id":"531","content_html":"The one with small size"},{"id":"530","content_html":"The one with more mass"},{"id":"529","content_html":"The one with less mass"}],"hints":[],"formats":["multiple-choice","free-response"],"combo_choices":[]}]}');

    te.save!(validate: false)

src/flux/task-step.coffee
```diff
@@ -5,13 +5,17 @@ flux = require 'flux-react'
 
 TaskStepConfig =
 
+  _loaded: (obj, id) ->
+    _.extend(@_local[id], obj)
```
philschatz: Why extend the existing object? If it's just to move the `task_id` then maybe make that explicit?

src/components/task-step/exercise-multiple-choice.cjsx
```diff
@@ -98,13 +100,32 @@ ExerciseReview = React.createClass
   onContinue: ->
     @props.onNextStep()
 
+  tryAnother: ->
+    task_id = TaskStepStore.getTaskId(@getId())
+    TaskStepActions.getRecovery(@getId())
+    TaskActions.load(task_id)
+
+  showTryAnother: ->
```
philschatz: `show` could be a verb. maybe `canShow....` ?

src/components/task-step/exercise-multiple-choice.cjsx
```diff
@@ -98,13 +100,32 @@ ExerciseReview = React.createClass
   onContinue: ->
     @props.onNextStep()
 
+  tryAnother: ->
+    task_id = TaskStepStore.getTaskId(@getId())
+    TaskStepActions.getRecovery(@getId())
```
philschatz: Should probably be named `loadRecovery` so as not to be confused with a getter/setter

philschatz:
After reviewing the incorrect answer and pressing "Try again" I get sent to this screen (new question). Instead of being asked for a `free_response` answer it is sent directly to the multiple choice:

image

src/flux/task-step.coffee
jivewise: The reason for this is that it may come from either @_local or obj.  I've updated it to explicitly set it for now, with some additional logic.  Hopefully we can have the backend send use the task_id directly eventually.

PR #75, opened 2015-03-25
Test/practicetasks

pandafulmanda:
Will list tests here as tasks in a bit, but these tests will attempt to cover:

> 1. conditional logic in the `render:` methods
> 2. verify calls to the store occur
> 3. URL transitions occur

-- from @philschatz in the skoops.

Tests -- 
- [x] Practice button exists on the tasks page
- [ ] Clicking on the Practice button will load a practice
- [x] Practice will load going to the practice page directly
- [x] Practice can be worked through
- [ ] Clicking the "Do another practice" will load a new practice
- [ ] In a different PR at another time, will pull out Task related tests and put together a set of tests for Tasks


pandafulmanda:
will probably revert shell stuff since that didn't fix the problem all the way because Task component that Practice uses also depends on location state.  Router.TestLocation seems like it will work for testing, although we may want to consider abstracting dependence on getParams from components as much as possible (like in ReadingPlan and ReadingShell) anyways for modularity/flexibility

package.json
```diff
@@ -67,5 +67,8 @@
         "jquery:$"
       ]
     }
+  },
+  "dependencies": {
+    "es6-promise": "^2.0.1"
```
philschatz: is this a dependency or a `devDependencies` (keep 'em alphabetized ; )?

test/components/practice.spec.coffee
```diff
@@ -0,0 +1,148 @@
+{expect} = require 'chai'
+_ = require 'underscore'
+React = require 'react/addons'
+Router = require 'react-router'
+Promise = require('es6-promise').Promise
```
philschatz: This can just be `{Promise} = require 'es6-promise'`

test/components/practice.spec.coffee
```diff
@@ -0,0 +1,148 @@
...
+    Router.run routes, history, (Handler, state)->
+      router = @
+      React.render(<Handler/>, div, ()->
+        component = @
+        result = {div, component, state, router, history}
+        resolve(result)
+      )
+
+tasksHelper = (courseId) ->
+  routerHelper('/courses/' + courseId + '/tasks/')
```
philschatz: You can use coffeescript's String concatenation syntax to make this a bit more concise: `"/courses/#{courseId}/tasks"` (is the trailing slash necessary?)

test/components/practice.spec.coffee
```diff
@@ -0,0 +1,148 @@
...
+  beforeEach ->
+    CourseActions.reset()
+  afterEach ->
+    React.unmountComponentAtNode(div)
+
+  it 'should load the practice button on the course tasks page', (done) ->
+    tests = ({div}) ->
+      buttons = Array.prototype.slice.call(div.querySelectorAll('button.btn-primary'))
+      practiceButton = _.last(buttons)
+      expect(practiceButton.innerText).to.equal('Practice')
```
philschatz: Can you test on class names (`-practice`) rather than the text? Gives Fred more opportunities to style the pages (he'll go through and unprefix the class if he plans on using it)

test/components/practice.spec.coffee
```diff
@@ -0,0 +1,148 @@
...
+      steps = TaskStore.getSteps(CourseStore.getPracticeId(courseId))
+
+      # not ideal.  wanted to find the step id off of the component's children
+      # but that is a pain in the butt right now.
+      stepId = steps[0].id
+      step = TaskStepStore.get(stepId)
+
+      # Will eventually test based on task type.  Assuming exercise with free
+      # response for now.
+      expect(step.freeResponse).to.be.undefined
```
philschatz: should this be `free_response` (sorry, just skimming through the file)

package.json
pandafulmanda: ooops.  yes a dev dependency

test/components/practice.spec.coffee
pandafulmanda: yes thanks!

test/components/practice.spec.coffee
pandafulmanda: i forget we're using coffeescript sometimes.

test/components/practice.spec.coffee
```diff
@@ -0,0 +1,170 @@
...
+
+  afterEach ->
+    React.unmountComponentAtNode(div)
+
+  it 'should load the practice button on the course tasks page', (done) ->
+    tests = ({div}) ->
+      expect(div.querySelector('.-practice')).to.not.be.null
+      done()
+
+    tasksHelper(1).then(tests).catch(done)
```
philschatz: is catching the error and calling done correct? I would have expected something like `.catch(fail)` from `{expect, fail} = require 'chai'` (not sure if that's the exact syntax) or just dropping the `.catch` and letting the test time out.

package.json
```diff
@@ -67,5 +68,6 @@
         "jquery:$"
       ]
     }
-  }
+  },
+  "dependencies": {}
```
philschatz: any reason this is left in?

test/components/practice.spec.coffee
```diff
@@ -0,0 +1,170 @@
...
+
+
+div = document.createElement('div')
+
+routerHelper = (route) ->
+  history = new Router.TestLocation([route])
+  promise = new Promise (resolve, reject) ->
+    Router.run routes, history, (Handler, state)->
+      router = @
+      React.render(<Handler/>, div, ()->
```
philschatz: Minor: You can just use `....., div, =>` instead of setting a router variable.

philschatz:
These tests are nice! I'm guessing the remaining TODO items will take a while to complete (since they require stubbing out the API).

I'd be happy merging what you have now and submitting a separate PR for the remaining items (if you update the TODO and remove the `WIP` from the title).

test/components/practice.spec.coffee
pandafulmanda: letting the test time out causes the timeout to mask the reason for the failed test.  .catch(done) is working/properly returns the reason for failure

package.json
pandafulmanda: nope sorry, i used it commandline to uninstall and reinstall as --save-dev and didn't realized I'd left that in.

test/components/practice.spec.coffee
pandafulmanda: wanted to pass in both the router and component ```this``` into the resolve, just in case

test/components/practice.spec.coffee
philschatz: ok, thanks!

PR #76, opened 2015-03-25
fix practice route

philschatz:
No UI changes.

makes the last slash in the practice URL optional like the rest of the routes


pandafulmanda:
:+1:

jivewise:
+1

PR #79, opened 2015-03-30
crude table render of guide data

darrenmason:
This renders sample learning guide json into a simple table.

image


style/guide.less
```diff
@@ -0,0 +1,7 @@
+.-course-guide-table-pageids span:after {
```
philschatz: The `.less` files should never refer to classes that begin with `-`. I would just remove this file (since it's a "Crude" version anyway)

philschatz:
Can you merge from master?

PR #80, opened 2015-03-30
Tests for task components

pandafulmanda:
Basis for #78 

- [x] write tests for
  - [x] render first incomplete step
  - [x] shows empty free response
  - [x] submits free response
  - [x] shows answered free response
  - [x] shows empty multiple choice
  - [x] submits multiple choice
  - [x] shows answered multiple choice
  - [x] show assignment feedback to student
  - [x] render next step when step completed
  - [x] render appropriate done page
  - [x] breadcrumbs can navigate to completed step


jivewise:
+1

src/components/task-step/exercise-multiple-choice.cjsx
```diff
@@ -110,13 +113,15 @@ ExerciseReview = React.createClass
...
   renderFooterButtons: ->
-    isDisabledClass = 'disabled' unless @isContinueEnabled()
-    continueButton = <BS.Button bsStyle="primary" className={isDisabledClass} onClick={@onContinue}>Continue</BS.Button>
+    # TODO switch to using React.addons.classSet for classname
+    buttonClasses = '-continue'
+    buttonClasses += 'disabled' unless @isContinueEnabled()
+    continueButton = <BS.Button bsStyle="primary" className={buttonClasses} onClick={@onContinue}>Continue</BS.Button>
     tryAnotherButton = <BS.Button bsStyle="primary" onClick={@tryAnother}>Try Another</BS.Button> if @canTryAnother()
-    <span>
+    <div>
```
philschatz: Why the switch to `<div>`? Since you're changing it, it should probably have a `className="-..."` too

src/components/task/breadcrumbs.cjsx
```diff
@@ -24,6 +24,8 @@ module.exports = React.createClass
 
     stepButtons = for step, i in steps
 
+      continue unless step
```
philschatz: Can you add a comment explaining when `step` can be falsy?

src/components/task/index.cjsx
```diff
@@ -54,24 +42,24 @@ module.exports = React.createClass
   render: ->
     {id} = @props
     model = TaskStore.get(id)
-    steps = TaskStore.getSteps(id)
-    stepConfig = steps[@state.currentStep]
+    stepConfig = @state.steps[@state.currentStep]
```
philschatz: why store all the steps in the `state`? (Does the component need to be updated when new props are provided?)

src/components/task/index.cjsx
```diff
@@ -54,24 +42,24 @@ module.exports = React.createClass
...
     {courseId} = @context.router.getCurrentParams()
 
-    allStepsCompleted = true
-    for step in steps
-      unless step.is_completed
-        allStepsCompleted = false
+    allStepsCompleted = TaskStore.isTaskCompleted(id)
 
-    if steps.length > 1
+    classes = classNames({
```
philschatz: I'm fine with this use for setting class names but to keep the code a little more consistent, could you either make this change for all places that require logic when setting the `className` or use methods already in the codebase?

src/components/task-step/exercise-multiple-choice.cjsx
pandafulmanda: I was getting an error during testing at some point about duplicate elements and one of the possible causes I found was doing funky stuff with HTML like putting a block element inside an inline element.  This ended up not being the problem so I can change this back?

src/components/task/index.cjsx
pandafulmanda: I wanted to avoid getting all steps multiple times, and was thinking of a potential situation where the steps for a task are updated outside of tasks step complete (i.e. if we get updated steps from adapting to the user). I will just move it back out of the state for now since that is not the case

src/flux/task.coffee
```diff
@@ -3,6 +3,29 @@ flux = require 'flux-react'
...
+getCurrentStepIndex = (steps) ->
+  currentStep = -1
+  for step, i in steps
+    unless step.is_completed
+      currentStep = i
+      break
+  currentStep
+
+getCurrentStep = (steps) ->
+  _.find(steps, (step)->
```
philschatz: Why use `(` here but not 12 lines above (`_.map`)?

pandafulmanda: oops okay :smile:   which should it be?

PR #81, opened 2015-03-31
Hide delete button logic

jivewise:
I think the backend is adding some fields in.  Creating a WIP PR for now.


jivewise:
This should be good to go.

screen shot 2015-04-03 at 12 33 48 am
screen shot 2015-04-03 at 12 33 27 am

PR #84, opened 2015-04-01
include source maps for less files

philschatz:
No UI Change.

- [x] test with `gulp prod` to ensure minified CSS does not include source maps


philschatz:
Verified source maps are not included in the minified CSS.

jivewise:
+1

PR #85, opened 2015-04-01
Fix gulp prod

philschatz:
No UI Changes.

fixes `gulp prod` by not trying to import external fonts while minifying the CSS.

@gregfitch


gregfitch:
font import no longer trips the provisioning up; uncovered an issue with the database reset

PR #87, opened 2015-04-02
Replace loadablemixin with loadable

philschatz:
No UI Changes.

Instead of relying on a `LoadableMixin` use a wrapper component that safely only tries to render once the content it loaded (using a guard fn). Instead of defining `getFlux: -> {store, actions}` and `getId: -> id` use the `LoadableItem` guard to ensure the item is loaded before rendering.

```coffee
render: ->
  <LoadableItem
    id={id}
    store={TaskStore}
    actions={TaskActions}
    renderItem={-> <LoadedComponentName ... />}
  />
```

Also defines a `<Loadable>` component which can be used for lists of things or if the content isn't exactly defined using the `CrudStore` like a Practice, Learning Guide, List of Tasks, etc. Basically, any time you end up writing `className="-loading"` or having to check `store.is*Loaded(id)` where `*` is non-empty.


jivewise:
+1, this is a great idea.

pandafulmanda:
:+1:

pandafulmanda:
:+1: :+1:

PR #90, opened 2015-04-03
exercise step with true and false with initial fake data and interface

pandafulmanda:
When used with openstax/tutor-server#140, 2nd Homework, exercise 4 is a true-false question

screen shot 2015-04-06 at 4 45 45 pm

screen shot 2015-04-06 at 5 35 22 pm

screen shot 2015-04-06 at 5 16 24 pm

- [x] stub fake true-false json
  - [x] Update format/formats on all stubbed exercises to array with either ```["multiple-choice"]``` and ```["free-response", "multiple-choice"]```
- [x] make exercise step handle true false
  - [x] Update task exercise component to handle formats as array
- [x] testsssss, when #80 is merged


package.json
```diff
@@ -54,6 +55,7 @@
     "react-router": "~0.13.2",
     "react-widgets": "^2.2.6",
     "underscore": "~1.8.2",
+    "underscore.string": "^3.0.3",
```
philschatz: Is this used?

src/flux/task-step.coffee
```diff
@@ -49,10 +50,37 @@ TaskStepConfig =
...
+    # TODO: write a test for this.
+    getPanels: (id) ->
+      step = @_get(id)
+      # Assumed panel order for now
+      panelOrders =
+        freeResponse: 0
+        multipleChoice: 1
+        review: 2
+
+      if step.type == 'exercise'
```
philschatz: Is the use of `==` instead of `is` (`===`) intentional?

pandafulmanda: no. ill fix this

package.json
pandafulmanda: oops, thought i uninstalled it

PR #94, opened 2015-04-08
force load steps from steps api when needed

pandafulmanda:
force load from steps api when initial task load from server api does not allow full access to step contents.  you could see this error particularly when you sign in as no_history on tutor_server and then attempt to complete a task.

No UI changes.  have tested through using both client stub and server data, and karma tests


pandafulmanda:
=( Travis, I'm failinggg youuuu

PR #96, opened 2015-04-08
linking up practice button on crude guide to post with page_ids

pandafulmanda:
No UI changes, just putting things in place for when guide server side is working and these practice buttons can render.

### DIFF
--- a/src/components/practice-button.cjsx
+++ b/src/components/practice-button.cjsx
@@ -1,6 +1,7 @@
 React = require 'react'
 BS = require 'react-bootstrap'
 Router = require 'react-router'
+_ = require 'underscore'
 
 {CourseActions, CourseStore} = require '../flux/course'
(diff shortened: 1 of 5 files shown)


src/components/practice-button.cjsx
```diff
@@ -20,11 +21,13 @@ PracticeButton = React.createClass
     <BS.Button bsStyle="primary" className="-practice" onClick={@onClick}>{actionText}</BS.Button>
 
   onClick: ->
+    params = _.omit(@props, 'courseId')
```
philschatz: This worries me a bit; could you make the params explicit?

pandafulmanda: hehe, okay.  wanted to make it as flexible as possible for whatever possible properties, but i guess that does open it up a little too much

PR #98, opened 2015-04-09
Details in top right hand corner

pandafulmanda:
With and without description HTML, based loosely on http://mockups.openstax.org/Tutor_MVP_v2/start.html#p=question_1a_1

- [x] styles polish
- [x] tests, after #95 is merged for convenience
- [x] how to adapt to different lengths of text?  how long are they on average anyways?

screen shot 2015-04-10 at 12 50 32 pm
screen shot 2015-04-08 at 7 03 49 pm


src/components/task/details.cjsx
```diff
@@ -0,0 +1,23 @@
...
+ArbitraryHtml = require '../html'
+
+module.exports = React.createClass
+  render: ->
+    {model} = @props
+
+    if model.description_html
+      detailPopover =
+        <BS.Popover title='Instructions' className='-task-details-popover task-details-popover'>
+          <ArbitraryHtml html={model.description_html}/>
```
philschatz: I don't think we're allowing arbitrary HTML in the descriptions; just text with newlines. I'm not sure how to handle text wrapping with newlines. Maybe there is some CSS to respect newlines (like `<pre>`) but still allow the text to wrap?

style/scaffold.less
```diff
@@ -45,6 +45,33 @@ body {
...
+    top: -2em;
+  }
+  button.task-details{
+    cursor: pointer;
+    top: -2.75em;
+  }
+}
+
+.task-details-popover{
+  font-size: 0.8em;
```
philschatz: Where did these styles come from?

src/components/task/details.cjsx
pandafulmanda: okay, ill look into that.  i saw that the property was called description_html so I assumed that they would have html

style/scaffold.less
pandafulmanda: @Fredasaurus what do you think? what should the styles be for this little guy? I was just guessing at things that look kinda okay.

pandafulmanda:
oh oops =(

pandafulmanda:
@Dantemss, should ```description_html``` be ```description``` instead for tasks?  we are looking at having the description be in markdown

pandafulmanda:
:+1:

PR #100, opened 2015-04-11
Jfl ireading 2015 03 20

Fredasaurus:
Checking this in so Dak can work on it. Latest includes styles for fig captions.

screen shot 2015-04-11 at 2 55 30 pm
screen shot 2015-04-11 at 2 56 03 pm


philschatz:
Can you remove `npm-debug.log` from this PR?

Fredasaurus:
Removed.

style/reading.less
```diff
@@ -10,52 +10,108 @@
...
+        border: 0;
+        margin: 0;
+      }
+
+      figcaption {
+        margin: 0 40px;
+      }
+
+      + p {
+      clear: both;
```
philschatz: should indent ; )

style/reading.less
```diff
@@ -10,52 +10,108 @@
...
+    .learning-objectives {
+      width: 100%;
+      min-height: 100px;
+      display: inline-flex;
+      margin-top: -20px;
+      color: #ffffff;
+      padding: 10px 40px;
+      background: @physics-background-secondary;
+
+      > div {display: none;}
```
philschatz: when do you put a single rule on 1 line and when do you put a single rule on multiple lines (I noticed above when styling `.reading-step figcaption`)

style/reading.less
```diff
@@ -10,52 +10,108 @@
...
+
+      p {
+        color: rgba(255,255,255,.75);
+        #fonts > .sans;
```
philschatz: Can you use `()` so it is clear this is calling a mixin?

style/reading.less
```diff
@@ -10,52 +10,108 @@
...
     color: @physics-color-primary;
-    margin: 0 4px 0 -4px;
+    margin: 0 4px 0 -6px;
   }
 
-  .reading-step section[data-depth="1"], .reading-step .work-in-physics, .reading-step .links-to-physics {
-    padding: 0 140px;
+  .reading-step {
+    
+    section[data-depth="1"]:not(.section-opener), section[data-depth="2"], .work-in-physics, .links-to-physics, .worked-example, .virtual-physics {padding: 0 140px;}
```
philschatz: can you put multiple selectors on a separate line? Helps with diffing later.

```css
section[data-depth="1"]:not(.section-opener),
section[data-depth="2"],
.work-in-physics,
.links-to-physics,
.worked-example,
.virtual-physics {padding: 0 140px;}
```

Also, maybe this should be in a `.x-*` mixin so all the styling for a `.work-in-physics` is in the same place. See [Napkin notes: Phil's LESS suggestions](https://github.com/openstax/napkin-notes/blob/master/phil/conventions.md#lesscss-conventions)

style/reading.less
Fredasaurus: Ah, vestiges, used to multiple rules. Will fix.

style/reading.less
Fredasaurus: Moved to separate .less files. Can you explain how to implement .x-*

style/reading.less
```diff
@@ -10,59 +10,111 @@
...
+      width: 60%;
+      float: left;
+      display: table;
+      text-align: left;
+      margin: 0 30px 0 -80px;
+      padding: 0;
+
+      img {
+        width: 100%;
+        border: 1px solid (@border-color);
```
philschatz: Why are there parentheses around some variables in these files?

style/reading/links-to-physics.less
```diff
@@ -19,15 +20,11 @@
     text-transform: uppercase;
     letter-spacing: 2px;
     text-align: center;
-    background: @physics-background-secondary;
+    background: (@physics-background-secondary);
 
-    &:before {
-      content: "\2014  ";
-    }
+    &:before { content: "\2014  ";}
```
philschatz: Can you include a comment identifying which character this is supposed to be? Do the 2 spaces matter? and this seems to be identical to code in `work-in-physics.less`; should they be combined?

style/reading.less
```diff
@@ -10,59 +10,111 @@
...
-      display: none
     }
 
-    p {
-      color: rgba(255,255,255,.75);
-      #fonts > .sans;
-      text-transform: uppercase;
-      text-align: right;
-      margin: auto 15px;
+    figure {
```
philschatz: For a future PR: would it make sense for this styling to be scoped inside `.arbitrary-html`? That way it can apply to other book content? like markup inside a video step, or for the reference book.

Or, you could do `.reading-step .arbitrary-html { .book-styles(); }`

style/reading.less
Fredasaurus: Ooooh, I did it bc I thought you told me to. You said mixin, not variable, let me change that.

style/reading/links-to-physics.less
Fredasaurus: 1. done
2. yes
3. Very close, probably can be combines, but was waiting for styles for each to be finished, then refactoring as needed. I supposed I could have labeled this PR as WIP-ish.

style/reading.less
Fredasaurus: That makes sense.

style/reading/links-to-physics.less
philschatz: RE 3: ok, just making sure it was noticed. Seems fine to do it in subsequent PR's

PR #105, opened 2015-04-14
Learning guide refactor

philschatz:
refactor learning guide into a component and a vanilla coffee class (which does the `d3` rendering)

@nathanstitt and @darrenmason thoughts?


nathanstitt:
Looks good to me.  I think it's a good idea to split it off from React since it's really all D3.

src/components/learning-guide-chart.coffee
```diff
@@ -71,38 +66,38 @@ LearningGuide = React.createClass
...
 
 
-  drawXAxis: (container, fields, points)->
+  drawXAxis: (container, fields, points) ->
     wrap = container.append('g')
       .attr('class', 'x-axis')
       .selectAll('line')
       .data(fields)
-    me=this
+    me = @ # Since displayUnit needs this and the 'this' scope still needs @parentElement
```
philschatz: Initially I was wondering why you did it for `displayUnit` but not for `navigateToPractice` and then I read a little more ; )

PR #107, opened 2015-04-15
practice link and button nav, display title in chart and footer

darrenmason:
image

### DIFF
--- a/src/components/learning-guide-chart.coffee
+++ b/src/components/learning-guide-chart.coffee
@@ -31,7 +31,7 @@ module.exports = class LearningGuideChart
 
     fields = guide.fields
 
-    space_between = WIDTH/fields.length+1
+    space_between = WIDTH/fields.length + 1
     points = _.map(fields, (f,i) ->
         {
           x: Math.max(space_between * i + (space_between/4), 5)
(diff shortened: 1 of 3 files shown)


src/components/learning-guide.cjsx
```diff
@@ -1,21 +1,34 @@
...
+
   propTypes:
     courseId: React.PropTypes.any.isRequired
 
+  getInitialState: ->
+    unit: false
+
   navigateToPractice: (unit) ->
-    alert "TODO: Navigate to practice unit ID #{unit.id} (#{unit.title})"
+    console.log("Navigate to practice unit ID #{unit.id} (#{unit.title})")
```
philschatz: can you just remove the console logs? Since they are no longer stubs?

darrenmason: ok removed

src/components/learning-guide-chart.coffee
```diff
@@ -48,10 +48,19 @@ module.exports = class LearningGuideChart
...
+    wrap = container.append('g')
+      .append("svg:text")
+      .attr("text-anchor","middle")
+      .attr("x", WIDTH/2)
+      .attr("y", 5)
+      .attr('class', 'main-title')
+      .text("Your Flight Path | "+guide.title+" | All Topics")
 
 
-  # Future improvement: Place clouds so they aren't
```
philschatz: I was hoping for animated clouds

nathanstitt: Hah! You know, I haven't mentioned it yet since I figured we had plenty else to work on, but another neat thing would be to animate the plane flying along the path between the circles.  It's quite doable: http://bl.ocks.org/KoGor/8162640

src/components/learning-guide.cjsx
```diff
@@ -1,21 +1,32 @@
...
   propTypes:
     courseId: React.PropTypes.any.isRequired
 
+  getInitialState: ->
+    unit: false
+
   navigateToPractice: (unit) ->
-    alert "TODO: Navigate to practice unit ID #{unit.id} (#{unit.title})"
+    {courseId} = @props
+    @context.router.transitionTo('viewPractice', {courseId})
```
philschatz: I think this may need to create a new practice in addition to `transitionTo` and send the tags (in the unit) to call it complete. But that can be a separate PR

PR #108, opened 2015-04-15
Student dashboard

nathanstitt:
screen shot 2015-04-28 at 11 23 41 am

TODO in future PRs: 
  * links to "recover credit" and "view feedback" are only `alert` calls.
  * links for reading to "view reference" is also just an `alert`.  It seems like this maybe could render the reading, but not sure of the best approach to take there.
  * The "Don't Forget" panel just selects the last 4 exercises competed.  When the BE is able to give an indication of what's been viewed, we can get smarter on that.


philschatz:
As a skeleton commit what's still WIP (can you add the TODO's to the PR description)? I think this would be great to pull in for @Fredasaurus / @dak to tweak.

philschatz:
Is this still "WIP"?

src/api.coffee
```diff
@@ -19,6 +19,7 @@ _ = require 'underscore'
 {TocActions} = require './flux/toc'
 {ExerciseActions} = require './flux/exercise'
 {TeacherTaskPlanActions, TeacherTaskPlanStore} = require './flux/teacher-task-plan'
+{StudentDashboardStore, StudentDashboardActions} = require './flux/student-dashboard'
```
philschatz: Is the store necessary?

src/components/student-dashboard/all-events-by-week.cjsx
```diff
@@ -0,0 +1,28 @@
...
+    courseId: React.PropTypes.any.isRequired
+
+  renderWeek:(events, week) ->
+    startAt = moment(week, "YYYYww")
+    <EventsPanel
+      key={week}
+      courseId={@props.courseId}
+      events={events}
+      startAt={startAt}
+      endAt={startAt.clone().add(1, 'week')}
```
philschatz: is `.clone()` necessary (since `.add` is called)?

src/components/student-dashboard/all-events-by-week.cjsx
```diff
@@ -0,0 +1,28 @@
...
+    />
+
+  render: ->
+    weeks = StudentDashboardStore.eventsByWeek(@props.courseId)
+    if _.any(weeks)
+      <div>{_.map(weeks, @renderWeek)}</div>
```
philschatz: can you add a `-`-prefixed class name to this element?

src/components/student-dashboard/dashboard.cjsx
```diff
@@ -0,0 +1,80 @@
...
+    router: React.PropTypes.func
+
+  getInitialState: ->
+    selectedTabIndex: 1
+
+  viewFlightPath: ->
+    @context.router.transitionTo('viewGuide', {courseId: @props.courseId})
+
+  selectTab: (index) ->
+    this.setState(selectedTabIndex:index)
```
philschatz: any reason for using `this` instead of `@`?

src/components/student-dashboard/dashboard.cjsx
```diff
@@ -0,0 +1,80 @@
...
+
+  render: ->
+    courseId = @props.courseId
+    info = StudentDashboardStore.get(courseId)
+    {longTitle, shortTitle} = StudentDashboardStore.getTitles(courseId)
+
+    <div className={shortTitle.toLowerCase() + " bg"}>
```
philschatz: if this is used in the LESS file can you document the possible class names here? Also, you can use string interpolation here: `className={"#{shortTitle.toLowerCase()} bg"}`

src/components/student-dashboard/dont-forget-panel.cjsx
```diff
@@ -0,0 +1,49 @@
...
+EmptyPanel  = require './empty-panel'
+_ = require 'underscore'
+module.exports = React.createClass
+
+  displayName: 'DontForgetPanel'
+  propTypes:
+    courseId: React.PropTypes.any.isRequired
+
+  viewFeedback: (taskId) ->
+    alert "View Feedback for task ID: #{taskId} in course ID: #{@props.courseId}"
```
philschatz: Can you add `TODO: ` to the alert so it is easier to grep on?

src/components/student-dashboard/empty-panel.cjsx
```diff
@@ -0,0 +1,12 @@
+React = require 'react'
+BS = require 'react-bootstrap'
+
+module.exports = React.createClass
+
+  displayName: 'EmptyPanel'
+
+  render: ->
+    content = @props.children or "No events this week"
+    <BS.Panel header={@props.title}>
```
philschatz: can you add `className="empty"` to the panel instead of the inner div? That way the whole panel could be hidden.

src/components/student-dashboard/event-row.cjsx
```diff
@@ -0,0 +1,21 @@
+React  = require 'react'
+BS     = require 'react-bootstrap'
+moment = require 'moment'
+
+module.exports = React.createClass
+  displayName: 'EventRow'
+  
+  propTypes:
+    cssClass: React.PropTypes.string.isRequired
```
philschatz: any reason this is `cssClass` instead of `className`?

src/components/student-dashboard/event-row.cjsx
```diff
@@ -0,0 +1,21 @@
...
+    event:    React.PropTypes.object.isRequired
+    feedback: React.PropTypes.string.isRequired
+
+  render: ->
+    <div className="task row #{@props.cssClass}">
+      <BS.Col xs={1}  sm={1} className="icon"></BS.Col>
+      <BS.Col xs={11} sm={7} className="title">{@props.children}</BS.Col>
+      <BS.Col xs={6}  sm={2} className="feedback">{@props.feedback}</BS.Col>
+      <BS.Col xs={5}  sm={2} className="due-at">
+        {moment(@props.event.due_at).format("ddd, MMMM Do")}
```
philschatz: are these format strings standardized somewhere else in the code?

src/components/student-dashboard/events-panel.cjsx
```diff
@@ -0,0 +1,37 @@
...
+    endAt:    React.PropTypes.object
+    limit:    React.PropTypes.number
+    title:    React.PropTypes.string
+    className: React.PropTypes.string
+
+  renderTitle: ->
+    if @props.title
+      <h3>{@props.title}</h3>
+    else
+      <span>{@props.startAt.format("MMMM Do")} - {@props.endAt.format("MMMM Do")}</span>
```
philschatz: are these format strings standardized somewhere else in the code? @Fredasaurus  should they be?

Plus, can this element have a `-`-prefixed class name?

src/components/student-dashboard/events-panel.cjsx
```diff
@@ -0,0 +1,37 @@
...
+
+  renderEvent: (event) ->
+    switch event.type
+      when 'reading'  then <ReadingRow  courseId=@props.courseId key={event.id} event={event}/>
+      when 'homework' then <HomeworkRow courseId=@props.courseId key={event.id} event={event}/>
+      else
+        <UnknownEventRow courseId=@props.courseId key={event.id} event={event}/>
```
philschatz: Should this throw an `new Error("Bug: unknown event #{event.type}")`? I'm hoping that all `throw new` code will eventually `POST` to the server so we can fix the code.

src/components/student-dashboard/homework-row.cjsx
```diff
@@ -0,0 +1,40 @@
...
+    alert "View recovery for task ID: #{@props.event.id} in course ID: #{@props.courseId}"
+
+  recoveryLinks: ->
+    if @props.event.is_complete
+      <span> | <a
+        onClick={@viewFeedback}>view feedback</a> | <a
+        onClick={@viewRecovery}>recover credit</a>
+      </span>
+    else
+      <span/>
```
philschatz: Can these elements have a `-`-prefixed class name?

src/components/student-dashboard/reading-row.cjsx
```diff
@@ -0,0 +1,26 @@
...
+
+module.exports = React.createClass
+  displayName: 'ReadingRow'
+
+  propTypes:
+    event: React.PropTypes.object.isRequired
+    courseId: React.PropTypes.any.isRequired
+
+  viewReference: ->
+    alert "View Reference for task ID: #{@props.event.id} in course ID: #{@props.courseId}"
```
philschatz: Can you include `TODO: ` in the alert?

src/components/student-dashboard/unknown-event-row.cjsx
```diff
@@ -0,0 +1,15 @@
...
+
+module.exports = React.createClass
+  displayName: 'UnknownEventRow'
+
+  propTypes:
+    event: React.PropTypes.object.isRequired
+    courseId: React.PropTypes.any.isRequired
+
+  render: ->
+    <EventRow feedback="" {...@props} cssClass="reading">
```
philschatz: Why use `cssClass` instead of `className`?

src/flux/student-dashboard.coffee
```diff
@@ -0,0 +1,56 @@
+{CrudConfig, makeSimpleStore, extendConfig} = require './helpers'
+{TimeStore} = require './time'
+_ = require 'underscore'
+moment = require 'moment'
+
+addEvents = (hash, events) ->
+  for event in events
+    week = hash[ moment(event.due_at).startOf('isoweek').format("YYYYww") ] ||= []
```
philschatz: Should this format be in a common area? @Fredasaurus

src/flux/student-dashboard.coffee
```diff
@@ -0,0 +1,56 @@
...
+      return @_weeks if @weeks
+      data = @_get(courseId)
+      @_weeks = {}
+      addEvents(@_weeks, data.tasks) if data.tasks
+      addEvents(@_weeks, data.exams) if data.exams
+      @_weeks
+
+    weeklyEventsForDay: (courseId, day) ->
+      events = this.exports.eventsByWeek.call(this, courseId)
+      events[moment(day).startOf('isoweek').format("YYYYww")] or []
```
philschatz: Should this format be in a common area? @Fredasaurus

src/flux/student-dashboard.coffee
```diff
@@ -0,0 +1,56 @@
...
+
+    getTitles: (courseId) ->
+      data = @_get(courseId)
+      shortTitle = data.course.name.split(' ')[0]
+      longTitle = "#{data.course.name} | #{arrayToSentence(data.course.teacher_names)}"
+      {longTitle, shortTitle}
+
+    pastEvents: (courseId, options = {}) ->
+      _.defaults(options, {
```
philschatz: This could be written without the `(` or the `{`.

src/flux/student-dashboard.coffee
```diff
@@ -0,0 +1,56 @@
...
+      events = this.exports.eventsByWeek.call(this, courseId)
+      events[moment(day).startOf('isoweek').format("YYYYww")] or []
+
+    getTitles: (courseId) ->
+      data = @_get(courseId)
+      shortTitle = data.course.name.split(' ')[0]
+      longTitle = "#{data.course.name} | #{arrayToSentence(data.course.teacher_names)}"
+      {longTitle, shortTitle}
+
+    pastEvents: (courseId, options = {}) ->
```
philschatz: When is `options` not defined? That might need to be documented

style/student-dashboard.less
```diff
@@ -0,0 +1,164 @@
...
+  .panel-body {
+    a { cursor: pointer; }
+    .empty{ text-align: center; }
+    .icon {
+      min-height: 3rem;
+      background-size: 3rem;
+      background-repeat: no-repeat;
+      background-position: 50%;
+    }
+    .reading  .icon { .x-icon-bg(reading); }
```
philschatz: Can this code be changed to use "Method 1"? Namely, the `cjsx` file adds a `<i className="..."/>` element rather than adding a generic `.icon` element?

src/components/student-dashboard/all-events-by-week.cjsx
nathanstitt: Yeah it is.  A surprising thing about the `moment` lib is that all methods simply mutate the object vs returning a new one.  So the `.add(1, 'week')` would actually modify the startAt var, causing the earlier set of the `startAt` prop to be identical to `endAt`.  The call to `clone()` causes a new object to be created so it's not shared.  There's several other spots in the code that have the same pattern.  http://momentjs.com/docs/#/parsing/moment-clone/

src/components/student-dashboard/dashboard.cjsx
nathanstitt: nope, other than I thought I was writing JS for a second.  I thought coffee-lint would catch those, but looks like it's not configured to do so, perhaps we could add that?

src/components/student-dashboard/event-row.cjsx
nathanstitt: Not that I'm aware of, I just matched what the mockup appeared to use.  Would be nice to standardize display of dates though.

src/components/student-dashboard/events-panel.cjsx
nathanstitt: It's my thoughts that eventually we'll have generic events such as "Presidents Day".  When those occur, the UnknownEventRow will be expanded to handle them gracefully.

src/components/student-dashboard/event-row.cjsx
Fredasaurus: I like, when possible, to spell out the month. It avoids the "is 4/5/15 April 5th or May 4th?" thing.
So I vote for Apr 5, 2015.

src/flux/student-dashboard.coffee
Fredasaurus: I would say so.

src/components/student-dashboard/events-panel.cjsx
nathanstitt: Good feedback though, it's not terribly well named.  I've changed it to be "GenericEventRow"

style/student-dashboard.less
nathanstitt: I've done so, but now remember why I'd used my custom css before.  The "large" size isn't very large.  Should we perhaps make the icon-lg a bit larger in all.less?  It's set to 2rem now, I think 3.5 would be more appropriate?

screen shot 2015-04-28 at 10 38 56 am

src/flux/student-dashboard.coffee
nathanstitt: That format is actually not used for display.  It's a combination of the year and week number that's used as a key to group events together by week.

PR #112, opened 2015-04-16
Lint cjsx files

philschatz:
No UI changes.

`coffeelint` now supports linting `cjsx` files.

I started off by enabling several options and seeing how much was caught. But I'm interested in feedback for which checks should be enabled/disabled. See <http://www.coffeelint.org/#options> for all the options.

Check the travis build to see what was caught.

Ref: clutchski/coffeelint 346


nathanstitt:
130 chars seems a bit long, but looks great other than that :+1:

pandafulmanda:
:+1:

PR #115, opened 2015-04-16
adjust scaling, add y-axis labels, add margin

darrenmason:
image


philschatz:
Thanks! Can you merge in changes from master?

philschatz:
I wonder if this should select the right-most unit first; that way students know they can click the numbers and have a way to click "Practice" instead of discovering the circle button or the unit button. @Fredasaurus ?

darrenmason:
I can add first selected unit by default;  also I can use flag label and cloud images from comp and convert to svg;  or if you have svg for them @Fredasaurus ?

nathanstitt:
SVG would be much preferred since it's vector and will scale with the chart without getting a bad case of the jaggies :) 

It may make more sense to draw the Y axis flag labels in the code rather than loading remotely in order to keep the network requests to a minimum and display it faster.  Not sure about that; I'll also bloat the code a bit.

Fredasaurus:
I’ll be working on new vector art early next week. Would delivery mid-Tuesday work.

BTW - I was hoping for animated clouds :)

Fred

darrenmason:
sure, I can animate them; was thinking prioritize the functional/layout bits first

PR #119, opened 2015-04-17
Try Another / Refresh my Memory / Move On question review buttons

nathanstitt:
screen shot 2015-04-17 at 3 15 56 pm

This adds a "Refresh My Memory" button to the question review screen.  If the button is selected it displays the last reading that was shown prior to the question so that the student can review the source material.

In order to hook up the button to navigate to the prior reading, I needed the "goToStep" callback from the main Task view, which is how the breadcrumbs navigate.  Unfortunately, this change rippled out through various panels before it could get to ExerciseReview.  It would seem to be more React like if we had a "navigation state" data structure that could be shared, and each would render appropriately based on it.

If there's agreement on that, I'll rework to use that method.

### DIFF
--- a/src/components/task-step/exercise-multiple-choice.cjsx
+++ b/src/components/task-step/exercise-multiple-choice.cjsx
@@ -7,7 +7,7 @@ ArbitraryHtmlAndMath = require '../html'
 StepMixin = require './step-mixin'
 Question = require '../question'
 BS = require 'react-bootstrap'
-classNames = require 'classnames'
+
 
 ExerciseFreeResponse = React.createClass
   displayName: 'ExerciseFreeResponse'
@@ -134,9 +134,11 @@ ExerciseReview = React.createClass
     return step.has_recovery and step.correct_answer_id isnt step.answer_id
 
   renderFooterButtons: ->
-    continueBtnClasses = classNames('-continue', {disabled: !@isContinueEnabled()})
+    # TODO: switch to using classnames library
+    buttonClasses = '-continue'
+    buttonClasses += 'disabled' unless @isContinueEnabled()
     continueButton =
-      <BS.Button bsStyle="primary" className={continueBtnClasses} onClick={@onContinue}>
+      <BS.Button bsStyle="primary" className={buttonClasses} onClick={@onContinue}>
         { if @canTryAnother() then "Move On" else "Continue" }
       </BS.Button>
     if @canTryAnother()
--- a/src/components/task-step/exercise-multiple-choice.cjsx
+++ b/src/components/task-step/exercise-multiple-choice.cjsx
@@ -7,7 +7,7 @@ ArbitraryHtmlAndMath = require '../html'
 StepMixin = require './step-mixin'
 Question = require '../question'
 BS = require 'react-bootstrap'
-
+classNames = require 'classnames'
 
 ExerciseFreeResponse = React.createClass
   displayName: 'ExerciseFreeResponse'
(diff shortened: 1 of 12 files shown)


src/components/task-step/exercise-multiple-choice.cjsx
```diff
@@ -112,6 +122,18 @@ ExerciseReview = React.createClass
...
+    currentStepIndex = _.findIndex(steps, (step) -> step.id == id )
+    return if -1 == currentStepIndex # should never happen
+    # Find the last reading that we saw by starting at the
+    # current step and moving backwards in the list
+    for i in [currentStepIndex..0] by -1
+      if "reading" == steps[i].type
+        # goToStep returns an function with the step index in closure scope.
+        @props.goToStep(i)()
+
   canTryAnother: ->
```
nathanstitt: This logic might be moved into a `TaskStore.getPreviousReading(fromStepId)` method?

src/components/task-step/exercise-multiple-choice.cjsx
```diff
@@ -112,6 +122,18 @@ ExerciseReview = React.createClass
     TaskStepActions.loadRecovery(id)
     TaskActions.load(task_id)
 
+  refreshMemory: ->
+    {id} = @props
+    steps = TaskStore.getSteps(TaskStepStore.getTaskId(id))
+    currentStepIndex = _.findIndex(steps, (step) -> step.id == id )
+    return if -1 == currentStepIndex # should never happen
```
philschatz: Cool! Can you add `throw new Error('BUG: Invalid step index')` instead of silently returning?

src/components/task-step/exercise-multiple-choice.cjsx
```diff
@@ -112,6 +122,18 @@ ExerciseReview = React.createClass
     TaskStepActions.loadRecovery(id)
     TaskActions.load(task_id)
 
+  refreshMemory: ->
+    {id} = @props
+    steps = TaskStore.getSteps(TaskStepStore.getTaskId(id))
+    currentStepIndex = _.findIndex(steps, (step) -> step.id == id )
```
philschatz: can you use `is` instead of `==`?

src/components/task-step/exercise-multiple-choice.cjsx
philschatz: Yep, that seems a bit cleaner!

src/components/task-step/exercise-multiple-choice.cjsx
```diff
@@ -121,16 +143,32 @@ ExerciseReview = React.createClass
     # TODO switch to using React.addons.classSet for classname
     buttonClasses = '-continue'
     buttonClasses += 'disabled' unless @isContinueEnabled()
-    continueButton = <BS.Button bsStyle="primary" className={buttonClasses} onClick={@onContinue}>Continue</BS.Button>
-    tryAnotherButton = <BS.Button bsStyle="primary" onClick={@tryAnother}>Try Another</BS.Button> if @canTryAnother()
-    <div className="-footer-buttons">
-      {tryAnotherButton}
+    continueButton =
+      <BS.Button bsStyle="primary" className={buttonClasses} onClick={@onContinue}>
+        { if @canTryAnother() then "Move On" else "Continue" }
```
philschatz: I didn't realize you could put arbitrary code in the `{}`! :+1:

style/task-step.less
```diff
@@ -5,6 +5,12 @@
 @answer-hover-color: #e6f9fc;
 @answer-select-color: #def4f7; // darker than hover
 
+.task-step {
```
philschatz: Is there a specific class on the panel that this could be wrapped inside?

test/components/helpers/task/checks.coffee
```diff
@@ -87,6 +87,11 @@ checks =
     expect(div.querySelector('.question-feedback').innerHTML).to.equal(feedback_html)
     {div, component, stepId, taskId, state, router, history, correct_answer, feedback_html}
 
+  _checkRecoveryRefreshChoice: ({div, component, stepId, taskId, state, router, history}) ->
+    expect(div.querySelector('.footer-buttons').children.length).to.equal(3)
+    titles = _.pluck(div.querySelector('.footer-buttons').children, 'textContent')
+    expect(titles).to.deep.equal(['Try Another', 'Refresh My Memory', 'Move On'])
```
philschatz: Can you test on a class on the buttons instead of text content? It could be a `-` prefixed class if there's no styling in the LESS files yet

style/task-step.less
nathanstitt: `task-step` is the wrapping class.  It encloses the entire panel, and is used by the existing styles below.  Would you like it re-named to something more specific?

philschatz: oh, nm ; )

PR #123, opened 2015-04-20
fix InvariantViolation when switching screens

philschatz:
No UI changes.

Thanks @jivewise and @pandafulmanda for providing reproducible steps!


pandafulmanda:
:+1: interesting!

nathanstitt:
Makes sense to stop listening while the component's being updated. 

I wonder if we should also do the same to the [katex mixin](https://github.com/openstax/tutor-js/blob/master/src/components/katex-mixin.coffee#L37-L38)?  That's where I think I was receiving the error.  I'm not currently receiving it though so I can't really test :(

PR #124, opened 2015-04-20
update gulp file for different types of users

philschatz:
No UI Changes.

- `gulp test` runs unit tests
- `gulp prod` builds minified (and zipped) files for production
- `gulp serve` builds files and starts up a static webserver
- `gulp dev` watches and rebuilds CSS and JS files and starts webserver
- `gulp tdd` does what `dev` does plus reruns unit tests

Also, gulp tasks not in the list above are prefixed with an `_` to show they are private.


Gulpfile.coffee
```diff
@@ -190,5 +212,34 @@ gulp.task 'serve', ['watch'], ->
...
-
   connect.server(config)
+
+gulp.task '_tdd', (done) ->
+  buildAndTest(true, done)
+  return # Since this is async
+
+
+# -----------------------------------------------------------------------
+#  Public Tasks
```
pandafulmanda: :+1:  I like the composition of the public tasks  :smile:  would it make sense to even pull the above private methods into a separate file, and require that in this main Gulpfile where the publics will sit?

philschatz: It was convenient to grep and replace all the names of private tasks. thoughts on naming this other file (`gulp-private-tasks.coffee`?

Or, a `gulp` directory in the root and files named `javascript-tasks.coffee`, `style-tasks.coffee`, `prod-tasks.coffee`? This might obscure the fact that `prod-tasks.coffee` depends on tasks defined in other files.

pandafulmanda: I like the second idea, but yeah, that just make prod-tasks a little more confusing...or maybe prod/_archive can live in gulpfile as well?

philschatz: Mind if I split the files in a separate PR? I'm updating #112 with commits in this PR

PR #125, opened 2015-04-20
add instructions to README for handling local updates

kevinburleigh75:
This PR needs to be reviewed by someone who knows more about `npm` and `gulp` than I do.


philschatz:
This is related to #124 in that `gulp serve` will differ based on which `gulp` tasks were run.

PR #111, opened 2015-04-16
Feature/plan details on calendar

pandafulmanda:
<a href="https://cloud.githubusercontent.com/assets/2483873/7256467/fdd71942-e814-11e4-9201-92a3483e4637.png" target="_blank">screen shot 2015-04-21 at 10 48 22 am</a><a href="https://cloud.githubusercontent.com/assets/2483873/7256468/0133c1ee-e815-11e4-8b16-43bbc2248b5b.png" target="_blank">screen shot 2015-04-21 at 10 48 28 am</a>

* Details on Homework should show an average at the top
* Details on Readings should not
* Chapter data may not show if it does not exist

Matches backend openstax/tutor-server#196

- [x] open modal on plan click
- [x] show stats on modal
- [x] tests
  - [x] matching modal opens

PR #126, opened 2015-04-21
tweak calendar styling and loading

philschatz:
Adds a "Loading" screen until the calendar has loaded and removed some LESS variable interpolation in the calendar.

This is a PR to #111


pandafulmanda:
:+1:

PR #127, opened 2015-04-21
Pin Exercise Summary on Scroll

dak:
Attached:
screen shot 2015-04-22 at 12 07 56 pm

Normal:
screen shot 2015-04-22 at 12 07 49 pm

### DIFF
--- a/src/components/task-plan/homework.cjsx
+++ b/src/components/task-plan/homework.cjsx
@@ -243,7 +243,7 @@ ExerciseSummary = React.createClass
 
   componentDidMount: () ->
     el = @getDOMNode()
-    @staticPosition = el.getBoundingClientRect().top - document.body.getBoundingClientRect().top
+    @staticPosition = @getPosition(el)
     @handleScroll() # Update scroll position immediately on mount
     @optimizedScrollFunc = throttle('scroll', 'optimizedScroll')
     window.addEventListener('optimizedScroll', @handleScroll)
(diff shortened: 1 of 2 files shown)


style/homework.less
```diff
@@ -1,3 +1,53 @@
...
+
+  .tutor-added-later {
+    line-height: 1.8rem;
+    font-size: 1.3rem;
+    text-align: left;
+  }
+}
+
+.navbar-fixed-top-lower {
+  top: 60px !important;
```
philschatz: Could we use a bootstrap mixin instead of setting several classes and then relying on `!important`?

src/components/task-plan/homework.cjsx
```diff
@@ -184,6 +198,26 @@ AddExercises = React.createClass
...
+  componentWillUnmount: () ->
+    window.removeEventListener('scroll', @optimizedScrollFunc)
+    window.removeEventListener('optimizedScroll', @handleScroll)
+
+  handleScroll: (e) ->
+    el = @getDOMNode()
+    if not el.textContent then return # HACK: Ignore empty span that is also an ExerciseSummary
+
+    if document.body.scrollTop > @staticPosition
+      el?.classList.add('navbar', 'navbar-fixed-top', 'navbar-fixed-top-lower')
```
philschatz: When is `el` null?

src/components/task-plan/homework.cjsx
```diff
@@ -184,6 +198,26 @@ AddExercises = React.createClass
 
 
 ExerciseSummary = React.createClass
+  componentDidMount: () ->
+    el = @getDOMNode()
+    @staticPosition = el.getBoundingClientRect().top - document.body.getBoundingClientRect().top
+    @handleScroll() # Update scroll position immediately on mount
+    @optimizedScrollFunc = throttle('scroll', 'optimizedScroll')
```
philschatz: why not just define this on the component directly? `optimizedScrollFunc: -> throttle('scroll', 'optimizedScroll')`

dak: This is attaching an event listener to the DOM element, which must exist (been mounted) for it to attach correctly.  Defining it on the component directly does not guarantee the element exists in the DOM (it likely does not).

src/components/task-plan/homework.cjsx
```diff
@@ -227,6 +241,31 @@ ExerciseSummary = React.createClass
...
+
+  getPosition: (el) -> el.getBoundingClientRect().top - document.body.getBoundingClientRect().top
+
+  handleScroll: (e) ->
+    el = @getDOMNode()
+    if not el.textContent then return # HACK: Ignore empty span that is also an ExerciseSummary
+
+    if document.body.scrollTop + 60 > @staticPosition
+      el.classList.add('navbar', 'navbar-fixed-top', 'navbar-fixed-top-lower')
+      document.body.style.marginTop = '120px'
```
philschatz: Is there a reason this isn't a class on `<body>`? If so, can you include a description in the LESS file with a greppable string like `FANCY_SCROLL_MARGIN_TOP` in both places? I'm worried about having styling in the code that will end up out-of-sync or be difficult to find later on.

dak: The margin is there to offset the page jumping when the exercise summary becomes a fixed element, so it's very closely tied with the height and padding of that element.

So I'm not sure making it a LESS style would be entirely appropriate, although it would certainly be more resilient to calculate how much to offset the body rather than the hardcoded 120px (it was hardcoded as a bit of a shortcut since the calculation is actually quite a few lines and requires only performing the calculation when the element is not fixed).

pandafulmanda:
cool

PR #129, opened 2015-04-21
WIP: nested mock data, show all link, chapter/overall view flag

darrenmason:
the graph is able to switch between parent/children data.  still need to bind to click and re-render chart.  changing 'showAll' flag will show all topics vs nested mock data on load.


src/components/learning-guide-chart.coffee
```diff
@@ -65,45 +70,21 @@ module.exports = class LearningGuideChart
...
-    circles = wrap.enter()
-      .append('g')
-      .attr('transform', (f,i) ->
-        "translate(#{points[i].x},#{points[i].y})"
-      )
-      .on('click', (field) =>
-        @navigateToPractice(field)
+      .text("Your Flight Path | #{guide.title} | All Topics | ")
+      .append('svg:a')
+      .attr('class', 'show-course noselect')
```
philschatz: @pandafulmanda is already using a noselect mixin for the calendar; maybe the mixin should be moved someplace more common? And not be a separate class on the element?

src/components/learning-guide-chart.coffee
```diff
@@ -65,45 +70,21 @@ module.exports = class LearningGuideChart
...
+      .text('Show All '+guide.title)
```
philschatz: Can you use string interpolation (`"Show All #{guide.title}"`) similar to 3 lines up?

src/components/learning-guide-chart.coffee
```diff
@@ -65,45 +70,21 @@ module.exports = class LearningGuideChart
...
+      .on('click', () ->
+        console.log('clicked title')
+        that.displayTopic(guide)
```
philschatz: You should be able to use fat arrows `=>` instead of `that`.

style/learning-guide.less
```diff
@@ -3,8 +3,13 @@
   svg {
     width: 100%;
     .main-title{
-      font-size: 3px; 
+      font-size: 2.5px;
```
philschatz: Why `px`? I think font sizes elsewhere are in `rem`

philschatz:
Thanks for the quick change to the learning guide format. @joemsak is there a related `tutor-server` PR for this?

style/learning-guide.less
nathanstitt: That was my bad.  I didn't notice that fonts should be rem.   (and you didn't catch it first during the earlier review  :stuck_out_tongue_winking_eye: )

src/components/learning-guide-chart.coffee
darrenmason: for some reason two string interpolations in a row didn't show data, but when I used +guide it worked

src/components/learning-guide-chart.coffee
darrenmason: will do

src/components/learning-guide-chart.coffee
darrenmason: yes, I was thinking of adding the noselect css to tutor.less or someplace common.  What do you mean by not being a separate class?  This way we can just add the class when needed?

style/learning-guide.less
darrenmason: will change to rem :smile:

src/components/learning-guide-chart.coffee
philschatz: Just an opinion (but open to discussion!) Since all elements already have a class on them, using a mixin ends up being more descriptive when just editing the styling; you don't need to check the JS to see if this element *sometimes* has `noselect` or *always* has `noselect` on it (maybe not the best example). Admittedly, when using a 3rd party lib (like bootstrap) there is no other option. @Fredasaurus thoughts? 

Whatever the decision, my vote goes for consistency!

src/components/learning-guide-chart.coffee
philschatz: String interpolation only occurs in strings with double quotes, not single quotes.

darrenmason:
can now re-render all svg with new data when necessary to change data, vs updating only the footer panel via setstate

philschatz:
closing in favor of #137

PR #130, opened 2015-04-22
Fix lint warnings from common components and flux stores

philschatz:
No UI Changes.

Fixes lint warnings from common components and flux stores.

### DIFF
--- a/src/flux/course.coffee
+++ b/src/flux/course.coffee
@@ -1,3 +1,4 @@
+# coffeelint: disable=no_empty_functions
 _ = require 'underscore'
 
 {TaskActions, TaskStore} = require './task'
@@ -12,7 +13,7 @@ CourseConfig =
     # TODO check with backend about task.type = 'practice' since task.type for homework = 'homework'
     obj.steps? # TODO: Find a more reliable way to determine if a practice is being loaded
 
-  createPractice: (courseId) ->
+  createPractice: (courseId) -> # Used by API
   createdPractice: (obj, courseId) ->
     @_loadedPractice(obj, courseId) # TODO: Maybe this should emit practice.created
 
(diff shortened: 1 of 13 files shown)


src/flux/course.coffee
```diff
@@ -46,7 +47,7 @@ CourseConfig =
     @emit('practice.loaded', obj.id)
 
   _reset: ->
-    CrudConfig.reset.call(@)
+    CrudConfig.reset.call(this)
```
pandafulmanda: is explicitly saying `this` preferred over using `@` "standalone"/in these types of changes?

src/flux/task.coffee
```diff
@@ -1,12 +1,12 @@
...
 _ = require 'underscore'
 flux = require 'flux-react'
 {TaskStepActions, TaskStepStore} = require './task-step'
 {CrudConfig, makeSimpleStore, extendConfig} = require './helpers'
 
 getSteps = (steps) ->
-  _.map(steps, ({id}) ->
+  _.map steps, ({id}) ->
     TaskStepStore.get(id)
-  )
```
pandafulmanda: cool, so should we stop using unnecessary parens in current stuff as well?

pandafulmanda:
just had some questions but :+1:

src/flux/course.coffee
philschatz: While it *does* find code like `that = @` I agree it's awkward. I'll probably just remove `no_stand_alone_at` from `coffeelint.json`

philschatz: Original reason for discouraging empty `@`: https://github.com/jashkenas/coffeescript Issue 1601 (not making it linkable so we are not added to the issue)

pandafulmanda: i see.  also good job on not accidentally linking, i totally did that on the react-calendar stuff lol.

src/flux/task.coffee
philschatz: I would say yes, but it's not enforceable... yet

PR #131, opened 2015-04-23
Allow navigation through homework questions by showing all breadcrumbs

pandafulmanda:
# Homework
screen shot 2015-04-23 at 2 50 34 pm
screen shot 2015-04-23 at 2 50 49 pm
screen shot 2015-04-23 at 6 53 50 pm

# Not Homework
screen shot 2015-04-23 at 2 50 26 pm
screen shot 2015-04-23 at 2 49 49 pm

Using star now but can easily switch for something else.
- [x] tests
  - [x] breadcrumbs show for all steps of homework
  - [x] breadcrumbs are clickable
  - [x] end breadcrumb also clickable and shown

- [x] breadcrumbs are numbered

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,3 @@
 /bower_components
 /.tmp
 /dist
-
-.DS_Store
-*.sublime-project
-*.sublime-workspace
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@
 /dist
 
 .DS_Store
+*.sublime-project
+*.sublime-workspace
--- a/style/task-step.less
+++ b/style/task-step.less
@@ -8,17 +8,19 @@
 // WHOA. This is sneaky. And hilarious. Am I allowed?!
 // http://caniuse.com/#search=counter
 .task-homework{
-  .steps {
-    counter-reset: step;
-  }
-  .step {
-    &:before {
-      content: counter(step);
-      counter-increment: step;
+  .step:not(.end) {
+    .step-problem-number {
+      display: block;
     }
     i {
       display: none;
     }
+  }
+}
+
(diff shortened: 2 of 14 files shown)


pandafulmanda:
@philschatz in mockup, the breadcrumbs include the one for the final turn panel as well...should I add that as well?

philschatz:
Ah, yes I think that `^` thingy needs to be in this

pandafulmanda:
oh okay, that changes things

pandafulmanda:
What about the question mark thing?

philschatz:
I think the question mark thing might be for another PR unless there's already info in the Task JSON. The info would be a step `type="pending"` I believe. @kevinburleigh75 is that in `tutor-server#master`?

.gitignore
```diff
@@ -3,4 +3,6 @@
 /.tmp
 /dist
 
-.DS_Store
\ No newline at end of file
+.DS_Store
+*.sublime-project
```
philschatz: Should this be in `~/.gitignore`? (probably same with `~/.DS_Store` now that I think about it)

pandafulmanda:
type pending should be very easy to add...maybe ill do that in a stub and style it before this gets merged.

style/task-step.less
```diff
@@ -5,6 +5,25 @@
 @answer-hover-color: #e6f9fc;
 @answer-select-color: #def4f7; // darker than hover
 
+// WHOA. This is sneaky. And hilarious. Am I allowed?!
+// http://caniuse.com/#search=counter
```
philschatz: I'd prefer you use CSS counters rather than generate them in the JS ; )

PR #133, opened 2015-04-23
load fonts securely

philschatz:
No UI changes.

changes `@import url('http://...')` to `@import url('//...')` so fonts load securely


gregfitch:
Tested and much happier...

PR #134, opened 2015-04-24
Select Reading Dialog and various refactorings

jivewise:
screen shot 2015-04-24 at 9 10 27 pm
- [x] Refactored select reading to use dialog
- [x] Updated reading to use new select reading dialog
- [x] Refactored exercise components out of homework
- [x] Started adding in some specs


src/components/task-plan/reading.cjsx
```diff
@@ -150,45 +113,67 @@ ReadingPlan = React.createClass
...
+
+    if (@state?.showSectionTopics)
+      formClasses.push('hide')
+      selectReadings = <ChooseReadings
+                        hide={@hideSectionTopics}
+                        courseId={courseId} 
+                        planId={id} 
+                        selected={topics}/>
+
+    <div>
```
philschatz: Any reason this guy is missing a class?

src/components/task-plan/select-topics.cjsx
```diff
@@ -0,0 +1,136 @@
...
+    active: React.PropTypes.bool
+
+  render: ->
+    classes = ['section']
+    classes.push('selected') if @props.active
+    isChecked = 'checked' if @props.active
+    classes = classes.join(' ')
+
+    <div key={@props.section.id} className={classes} onClick={@toggleSection}>
+      <span className="-section-checkbox">
```
philschatz: Is this span needed?

src/components/task-plan/select-topics.cjsx
```diff
@@ -0,0 +1,136 @@
...
+
+
+  renderChapterPanels: (chapter, i) ->
+    <ChapterAccordion {...@props} chapter={chapter}/>
+
+
+  render: ->
+
+    {courseId, planId, selected, hide, header, primary} = @props
+    unless TocStore.isLoaded()
```
philschatz: Can you use LoadableItem here?

jivewise: I was having trouble adding in loadable item here, since TocStore isn't extended from CrudConfig, I wanted to ask you if you think it'd be alright to extend from there?

src/components/task-plan/reading.cjsx
jivewise: I don't think it really needs one, as the panel below has a class already.  The div is just there to contain both the dialog and the panel.

src/components/task-plan/select-topics.cjsx
jivewise: Likely no, just added that in there in case Fred/Dak wanted to use it to style a bit.  Should I remove it?

src/components/task-plan/homework/exercises.cjsx
```diff
@@ -0,0 +1,213 @@
...
+
+  render: ->
+    load = @renderLoading()
+    if (load)
+      return load
+    
+    {courseId, pageIds, planId} = @props
+
+    if not TaskPlanStore.getTopics(planId).length
+      return <div className='bug'>Failed loading exercises</div>
```
philschatz: Could you add this class into `tutor.less` and make it big and [blinking](https://stackoverflow.com/questions/16344354/how-to-make-blinking-flashing-text-with-css3/16344548#16344548) (or marquee)?

PR #137, opened 2015-04-25
Learning guide city design svg

darrenmason:
updated svg for chart design;  still need to update detail footer, etc.

image


philschatz:
Great! Can you update the tests?

darrenmason:
this commit contains all the changes from #129

philschatz:
This PR is tied to openstax/tutor-server#153

darrenmason:
image

style/resources/cityscape.svg
```diff
@@ -0,0 +1,383 @@
+<svg
```
philschatz: Is this file still used?

src/components/learning-guide-chart.coffee
```diff
@@ -3,44 +3,57 @@ d3 = require 'd3'
 
 CLOUDS_PATH = '/style/resources/clouds.svg'
```
philschatz: Also, do we still need this file (and var)?

style/resources/cityscape.svg
darrenmason: for the moment, I will replace it next

src/components/learning-guide-chart.coffee
```diff
@@ -3,44 +3,57 @@ d3 = require 'd3'
...
 PLANE_PATH = '/style/resources/openstax-plane.svg'
+CITYSCAPE_PATH = '/style/resources/cityscape.svg'
+
+# should be svg bound to the label text, but temp for now.
+FLAG_BLUE = '/style/resources/flag-blue.png'
```
philschatz: Any reason these aren't drawn as SVG or included (with text) in the flight path background? Maybe this could be a separate PR?

darrenmason: the flags should be separate from the background so they can be positioned independently.  I don't have svg assets for the flags yet

src/components/learning-guide-chart.coffee
darrenmason: I will remove it.

style/resources/cityscape.svg
darrenmason: @Fredasaurus, regarding the new svg asset, I need 1 single separate cloud svg and 1 separate city svg.  The gradient and green bar are already in the chart, so do not need those.  Also if you want the colored flags to be svg I need those separate .svg files as well. (w/o the text)

PR #139, opened 2015-04-25
Student dashboard icons

philschatz:
This is a PR to #108 rather than master.

Adds color SVG icons with 2 methods for including the icons (feedback on which method to use?).

image

Maybe I should explore the best way to stack icons before deciding on one method over the other.


nathanstitt:
This looks awesome.  

My weak preference would be for the font-awesome pattern (option 1).  The other one looks to much like a css selector and I first attempted to interpret it that way before I realized it was using a less expression to extend the icon.  That could be just me though, I'm a bit more familiar with sass than less.

PR #143, opened 2015-04-27
Adding new flightpath skyline svg

Fredasaurus:
screen shot 2015-04-27 at 4 11 16 pm


philschatz:
Maybe this file should be named `guide-background.svg` so as not to be confused with `cityscape.svg` in #137

PR #144, opened 2015-04-28
Update/match plans backend

pandafulmanda:
update API and stub api data to match `api/courses/#{courseId}/events` and `api/plans/#{planId}/stats`

will conflict with 5f11c57 (#108)


nathanstitt:
Thanks for the heads-up.  I moved the #108's test json to be "dashboard.json" to match the endpoint.  It may no longer conflict.

PR #148, opened 2015-04-28
Link tasks to viewTask route so they can be worked

nathanstitt:
Also fixed a few things so the dashboard conforms to the notes that were transcribed during it's review, mainly that the Upcoming panel was just showing only next week,not all future tasks.

Also fixes a few lint warnings

### DIFF
--- a/src/flux/student-dashboard.coffee
+++ b/src/flux/student-dashboard.coffee
@@ -30,6 +30,11 @@ StudentDashboardConfig = {
       addEvents(@_weeks, data.exams) if data.exams
       @_weeks
 
+    pastEventsByWeek: (courseId) ->
+      weeks = this.exports.eventsByWeek.call(this, courseId)
+      thisWeek = moment(TimeStore.getNow()).startOf('isoweek').format("YYYYww")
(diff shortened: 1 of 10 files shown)


src/flux/student-dashboard.coffee
```diff
@@ -40,6 +45,18 @@ StudentDashboardConfig = {
...
 
+    canWorkTask: (event) ->
+      # only homework and readings can be worked on
+      _.contains(['homework', 'reading'], event.type) and
+        # readings can always be performed
+        (event.type is 'reading' or
+            # other types (homework) must be incomplete and not past due
+            (not event.complete and moment(event.due_at).isAfter(TimeStore.getNow())))
+
+    upcomingEvents: (courseId) ->
```
nathanstitt: This controls whether a task gets a onClick handler and navigates to it.  I'm not entirely sure of the logic behind it, but it seems close to what the task panel used.

nathanstitt:
Currently there is no visual identifier if a task is "workable" and can be clicked.  We should highlight them or add a star or something.  I do add a css class "workable" to them so we can target that way if needed.

PR #149, opened 2015-04-28
Remove ranges from the calendar

philschatz:
Ranges are meant to be for block scheduling.

This unfortunately hides some fancy (and tested!) code showing calendar ranges.

image

I commented the code in case there is a better way to turn it off for now


pandafulmanda:
i was thinking about making a range vs day display/mode but this is good for now i think

PR #151, opened 2015-04-29
added and positioned new assets, showAll toggle, cleanup

darrenmason:
image


src/components/learning-guide-chart.coffee
```diff
@@ -97,25 +100,25 @@ module.exports = class LearningGuideChart
       .text("Show All #{guide.title}")
       .on('click', =>
         @displayTopic()
+        detailPane = document.querySelector('.learning-guide-chart .footer')
```
philschatz: Changing the React DOM outside of react worries me a bit. This is usually the cause of InvariantViolation Errors when transitioning to another page. Any reason code that removes the active class isn't in `@displayTopic` since that was created in the React component?

src/components/learning-guide-chart.coffee
```diff
@@ -291,7 +301,7 @@ module.exports = class LearningGuideChart
         else if lv <= .49 and lv >= .35
           'yellow'
         else if lv < .35
-          'orange'
+          'grey'
```
philschatz: Should these be the semantic name `high/medium/low` or `ace/cruising/...` rather than the colors?

src/components/learning-guide-chart.coffee
```diff
@@ -1,19 +1,21 @@
 _ = require 'underscore'
 d3 = require 'd3'
 
-CLOUDS_PATH = '/style/resources/clouds.svg'
-PLANE_PATH = '/style/resources/openstax-plane.svg'
-CITYSCAPE_PATH = '/style/resources/cityscape.svg'
+CLOUD_PATH = '/style/resources/learning-guide/cloud.svg'
```
philschatz: Also, these assets will probably need to be included via CSS files to load properly (using `background-image: `) because these assets do not load when loaded from `:3001`. I had to change each path to `CLOUD_PATH = '//localhost:8000/dist/cloud.svg'` to get it to work.

src/components/learning-guide-chart.coffee
darrenmason: ok

src/components/learning-guide-chart.coffee
darrenmason: will do

src/components/learning-guide-chart.coffee
darrenmason: not getting invariant error, see what you mean though.  I have some reservations as well; specifically some dom manipulations and ui changes need to wait until the elements are done being painted to get their attributes and using just using ReactDOM isnt enough.  If we can't do a simple css manipulation on an 'outer leaf' then we may have still have to select the dom node after painting the 'actual dom', for attributes relative to the document such as scrollHeight.

PR #152, opened 2015-04-29
added awareness of 'not_started' status in the performance book

jpslav:
It's a biggie!


philschatz:
tisk tisk, no screenshots or unit tests? ; )

PR #157, opened 2015-04-29
new breadcrumbs

pandafulmanda:
## Homework, due in future
screen shot 2015-04-29 at 3 16 05 am

## Homework, due in past, feedback is viewable
screen shot 2015-04-29 at 3 16 23 am

## iReading
screen shot 2015-04-29 at 3 16 35 am


philschatz:
@Fredasaurus Any objections on including these in?

Fredasaurus:
Go for it, though we may need to make some changes later.

style/icons/all.less
```diff
@@ -56,7 +56,7 @@
 
 // Note: Use tokens instead of strings to be able to do LESS compile-time validation
 .icon-feedback { .x-icon-bg(feedback); }
-.icon-homework { .x-icon-bg(homework); }
+.icon-exercise, .icon-homework, .icon-end { .x-icon-bg(homework); }
```
philschatz: Can you place selectors on separate lines? Maybe this should be done in the JS that way it's easier to grep for when changing.

style/task-step.less
```diff
@@ -14,13 +31,49 @@
...
+    font-weight: 500;
+    cursor: pointer;
+    margin: 2px;
+    display: inline-block;
+    text-align: center;
+    background: @white;
+    color: @tutor-light-gray;
+    .transition(all .1s ease-in);
+
+    &:hover, &.active {
```
philschatz: Can you put selectors on separate lines?

style/icons/all.less
pandafulmanda: i think the icon representing icon-end is bound to change to be a different icon altogether, so i don't know if that's best at the moment

PR #161, opened 2015-04-29
crude guide lint

darrenmason:
this page is not active, but was showing lint errror anyway, so linted.


philschatz:
Is this page still being used? If not, it should probably be removed.

PR #167, opened 2015-04-30
remove url warnings from tests

philschatz:
No UI Changes.


pandafulmanda:
!!!! does this mean we may no longer have to manual .loaded into a store when doing testing?@!?@?!?!?!??@?#??@?!?!?!?

if so, i am SUPER excited

philschatz:
hmm, I liked (marginally) the extra documentation of calling `.loaded` but yeah writing `require '../../../../../../../../../../../../../../../../../../../../../../../../api/foo.json'` was probably quite annoying ; )

PR #170, opened 2015-04-30
Auto detect the host & port for loading resources

nathanstitt:
No UI changes

Created a new AppConfigStore that has a `getAssetsHost` and `urlForResource` methods and configures the learning-guide to load it's assets using it.

Feedback welcomed on how `getAssetsHost` auto-detects the host and if that's a good idea or not.  We could also set it explicitly as when the app's loaded in the index.html files as part of a "Bootstrap" type call.

### DIFF
--- a/src/components/learning-guide-chart.coffee
+++ b/src/components/learning-guide-chart.coffee
@@ -1,14 +1,14 @@
 _ = require 'underscore'
 d3 = require 'd3'
+{AppConfigStore} = require '../flux/app-config'
 
-CLOUD_PATH = '/style/resources/learning-guide/cloud.svg'
-PLANE_PATH = '/style/resources/learning-guide/openstax-plane.svg'
-CITYSCAPE_PATH = '/style/resources/learning-guide/houston-skyline.svg'
-
-FLAG_BLUE = '/style/resources/learning-guide/flag-blue.svg'
-FLAG_GREEN = '/style/resources/learning-guide/flag-green.svg'
-FLAG_YELLOW = '/style/resources/learning-guide/flag-yellow.svg'
-FLAG_GREY = '/style/resources/learning-guide/flag-grey.svg'
+CITYSCAPE_PATH = 'learning-guide/houston-skyline.svg'
+CLOUD_PATH     = 'learning-guide/cloud.svg'
+PLANE_PATH     = 'learning-guide/openstax-plane.svg'
+FLAG_BLUE      = 'learning-guide/flag-blue.svg'
+FLAG_GREEN     = 'learning-guide/flag-green.svg'
+FLAG_YELLOW    = 'learning-guide/flag-yellow.svg'
+FLAG_GREY      = 'learning-guide/flag-grey.svg'
 
 # SVG is vector so width/height don't really matter.  100 is just a convenient # to multiple by
 WIDTH = 160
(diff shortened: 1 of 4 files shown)


src/components/learning-guide-chart.coffee
```diff
@@ -1,14 +1,14 @@
...
 
-CLOUD_PATH = '/style/resources/learning-guide/cloud.svg'
-PLANE_PATH = '/style/resources/learning-guide/openstax-plane.svg'
-CITYSCAPE_PATH = '/style/resources/learning-guide/houston-skyline.svg'
-
-FLAG_BLUE = '/style/resources/learning-guide/flag-blue.svg'
-FLAG_GREEN = '/style/resources/learning-guide/flag-green.svg'
-FLAG_YELLOW = '/style/resources/learning-guide/flag-yellow.svg'
-FLAG_GREY = '/style/resources/learning-guide/flag-grey.svg'
+CITYSCAPE_PATH = 'learning-guide/houston-skyline.svg'
```
philschatz: so far all the SVG files in the `dist/archive.tar.gz` (`gulp prod`) end up in the root dir. We should probably change the gulpfile.

src/flux/app-config.coffee
```diff
@@ -0,0 +1,31 @@
...
+AppConfig =
+
+  setAssetsHost: (host) ->
+    @_assetsHost = host
+
+  exports:
+    getAssetsHost: ->
+      return @_assetsHost if @_assetsHost?
+      # http://caniuse.com/#feat=css-sel3 (substring selector is IE 8+)
+      tutorStyle = document.head.querySelector('link[href$="tutor.css"]')
```
philschatz: This file (in `gulp prod`) ends up looking like `tutor-293783.min.css`

src/components/learning-guide-chart.coffee
nathanstitt: Yeah, that's not good.  I'll take a look at it

src/flux/app-config.coffee
nathanstitt: Ah, good to know.  I'll have to use the substring matcher then instead of the tail one.

src/flux/app-config.coffee
```diff
@@ -0,0 +1,31 @@
...
+      href = if tutorStyle then tutorStyle.getAttribute('href') else "/"
+      # By using this method we can easily support http, https, as well as protocol agnostic schemes
+      # without using a huge nasty regex.
+      # It's a bit more expensive (don't use in a loop), but we're caching the results
+      a = document.createElement('a')
+      a.href = href
+      @_assetsHost = "#{a.protocol}//#{a.host}"
+
+    urlForResource: (path) ->
+      this.exports.getAssetsHost.call(this) + "/style/resources/#{path}"
```
philschatz: This path depends on:

1. the path in `archive.tar.gz`
2. production dumps `archive.tar.gz` into `/assets`

so the path would become: `#{HOST}/assets/#{PATH_IN_ARCHIVE}` for production and `/dist` for dev

src/flux/app-config.coffee
philschatz: Like `link[href^="tutor"][href$=".css"]`

nathanstitt: Yeah, or just `link[href*="tutor"]`  would seem safe enough.  That just looks for the string tutor anywhere in a css link.

src/flux/app-config.coffee
lakshmivyas: @philschatz , @nathanstitt - Instead of figuring out the url in code, one option is to do it as part of the build.   The source files can have a place holder, say, `{% assetHost %}` and `gulp <task>` will substitute it with the appopriate host based on the target environment.

PR #174, opened 2015-05-01
set a default policy for task steps by step types

pandafulmanda:
no ui changes.

just makes it so that we dont need to define a new policy for every new task type.


philschatz:
I like that we explicitly error when there is a new task type. Could you still log an error message when it occurs?

PR #175, opened 2015-05-04
Fix assets serving under production

nathanstitt:
No UI changes.

TODO:
  * [x] Fix gulp task
  * [x] Test if we need to prepend "/assets" onto url when running in production environment and do so if needed.

### DIFF
--- a/src/flux/app-config.coffee
+++ b/src/flux/app-config.coffee
@@ -4,10 +4,15 @@ flux = require 'flux-react'
 
 {makeSimpleStore} = require './helpers'
 
+MinimizedPattern = /tutor\.min-\d+/
+
 AppConfig =
 
   setAssetsHost: (host) ->
     @_assetsHost = host
+
+  setProductionEnvironment: (production) ->
+    @_isProductionEnvironment = production
 
   exports:
     getAssetsHost: ->
@@ -17,6 +22,8 @@ AppConfig =
       # if we didn't find the styling css, use "/" string.
       # That'll cause the anchor test below to be a relative link and use the document's scheme/host/port
       href = if tutorStyle then tutorStyle.getAttribute('href') else "/"
+      # detect isProduction unless it's been explicitly set
+      @_isProductionEnvironment = MinimizedPattern.test(href) unless @_isProductionEnvironment?
       # By using this method we can easily support http, https, as well as protocol agnostic schemes
       # without using a huge nasty regex.
       # It's a bit more expensive (don't use in a loop), but we're caching the results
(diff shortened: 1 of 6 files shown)


lakshmivyas:
@nathanstitt - I think assets is already prepended in production.
https://github.com/openstax/tutor-deployment/blob/master/tasks/deploy_assets.yml#L5

nathanstitt:
@lakshmivyas: thanks, but that's not the "assets" prefix I was referring to :grinning:  .  The assets are stored in an "assets" directory, but the url that the front-end generates to load them may not contain the "/assets" prefix.

We're currently unsure if the mechanism that I employed to load the assets in local and server modes (PR #170)  will also work on production or not, hence the need to load up production on vagrant to find out.

lakshmivyas:
@nathanstitt - makes sense :)  FWIW - all the assets packaged in the zip file are extracted out to the assets directory and the urls need to be `assets` prefixed.

I can see some issues here: https://tutor-dev1.openstax.org/dashboard.

src/flux/app-config.coffee
```diff
@@ -4,11 +4,16 @@ flux = require 'flux-react'
 
 {makeSimpleStore} = require './helpers'
 
+MinimizedPattern = /tutor\.min-\d+/
```
philschatz: Why not something like the following (and remove the `learning-guide/` prefix in `learning-guide-chart.coffee`?

```coffee
AppConfig =

  setAssetsPrefix: (prefix) ->
    @_assetsPrefix = prefix

  exports:
    getAssetsPrefix: ->
      return @_assetsPrefix if @_assetsPrefix?
      # http://caniuse.com/#feat=css-sel3 (substring selector is IE 9+)
      tutorStyle = document.head.querySelector('link[href*="tutor"]')
      # if we didn't find the styling css, use "/" string.
      # That'll cause the anchor test below to be a relative link and use the document's scheme/host/port
      href = if tutorStyle then tutorStyle.getAttribute('href') else "/"
      @_assetsPrefix = href[0.. href.lastIndexOf('/')]

    urlForResource: (path) ->
      prefix = this.exports.getAssetsPrefix.call(this)
      "#{prefix}/#{path}"
```

nathanstitt: Great idea!  I've tested it out and works great under production.  It doesn't work in development on either port 3001 or 8000.  

What happens in development is the css is loaded from the /dist folder, i.e. `http://localhost:8000/dist/tutor.css`, which causes the url for images to also get the /dist prefix prepended.  
As an example the skyline image is:
`http://localhost:8000/dist/learning-guide/houston-skyline.svg` 
But it should be: 
`http://localhost:8000/resources/images/learning-guide/houston-skyline.svg`

That's an easy fix though, I'm just adding a `.replace("dist", "resources/images")`

nathanstitt: I'm not sure why you say: "and remove the learning-guide/ prefix in learning-guide-chart.coffee"

That needs to stay since those assets are in the learning-guide directory - or am I misunderstanding your intent?

philschatz: They're all in `/dist/*.svg` as part of the build so no need for the paths and no need for `.replace("dist", "resources/images")`

PR #176, opened 2015-05-04
General exercises code clean up

jivewise:
- [x] Code clean up
- [x] Transition to dashboard when adding and saving a task plan

### DIFF
--- a/src/components/task-plan/footer.cjsx
+++ b/src/components/task-plan/footer.cjsx
@@ -12,8 +12,14 @@ PlanFooter = React.createClass
     courseId: React.PropTypes.any.isRequired
     clickedSelectProblem: React.PropTypes.func
 
+  saved: ->
+    courseId = @props.courseId
+    TaskPlanStore.removeChangeListener(@saved)
+    @context.router.transitionTo('taskplans', {courseId})
+
   onSave: ->
     {id} = @props
+    TaskPlanStore.addChangeListener(@saved)
     TaskPlanActions.save(id)
 
   onPublish: ->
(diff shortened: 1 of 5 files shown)


src/components/task-plan/footer.cjsx
```diff
@@ -12,8 +12,14 @@ PlanFooter = React.createClass
     courseId: React.PropTypes.any.isRequired
     clickedSelectProblem: React.PropTypes.func
 
+  saved: ->
+    courseId = @props.courseId
+    TaskPlanStore.removeChangeListener(@saved)
```
philschatz: What if the `save` errors instead of calling `Actions.saved`? I think you might need to test the Action being called. I remember suggesting something similar (like `Actions.saved.on(@saved)`) and then detaching `@saved` on `componentDidUnmount` to @pandafulmanda in another part of the code.

src/components/task-plan/index.cjsx
```diff
@@ -82,7 +71,6 @@ PlanShell = React.createClass
       id={id}
       store={TaskPlanStore}
       actions={TaskPlanActions}
-      saved={@saved}
```
philschatz: hmm, why not always `transitionTo` the calendar from here?

And if not possible, then is there any other code that uses `<LoadableItem>` `saved` prop?

jivewise: Hm, I think the saved prop is actually a misnomer, as it only triggers when saving a new item, it should probably be "created", and I was actually going to remove it after this PR, since I don't think it's used anywhere else.

PR #177, opened 2015-05-04
Update package dependencies

philschatz:
No UI Changes.


pandafulmanda:
Any reason we are using exact as opposed to `~`?  Are we moving away from `~`?

philschatz:
@dak thoughts on whether to pin versions or not? We've had some minor Travis hiccups by not pinning the versions

dak:
@philschatz As long as the dependency is being strict about semver then I think `~` is fine.  For example, I don't think I've ever had any issue at all with chai (even for major updates).  Gulp plugins are usually pretty decent too.  I could definitely see something like react-bootstrap causing issues and needing to be pinned though.

That said, it's really easy to keep up-to-date on dependencies using the david-dm badges (on GitHub), and something like [npm-check-updates](https://github.com/tjunnone/npm-check-updates) locally, so if you have the time to manually bump versions once a week or so, it can definitely make it easier to identify what caused an issue and when.

philschatz:
Good point. closing this PR for now. There's no pressing need to upgrade deps yet

PR #181, opened 2015-05-04
Enable views other than "Loading..." in loadable item 

pandafulmanda:
* components using loadItem can add `renderLoading`, `renderError`, `renderBug` prop to change what the loading/status screens look like

* example implementation with calendar

  ## Loading
screen shot 2015-05-05 at 7 46 12 pm

  ## Loaded
screen shot 2015-05-05 at 7 46 26 pm


src/components/loadable.cjsx
```diff
@@ -15,6 +15,13 @@ module.exports = React.createClass
     isLoading: React.PropTypes.func.isRequired
     isLoaded: React.PropTypes.func.isRequired
     isFailed: React.PropTypes.func.isRequired
+    renderStatus: React.PropTypes.func.isRequired
+
+  getDefaultProps: ->
+
+    # Enables a renderStatus prop function with a component other than a div
+    renderStatus: (status, message) ->
```
philschatz: Why not have props named `renderLoading`, `renderError`, or `renderBug` (and react can statically check their types to make sure they're a component) and have default props for each?

That way components can just provide a custom status when needed rather than implement `renderStatus` with a switch statement?

src/components/task-plan/teacher-task-plans-listing.cjsx
```diff
@@ -80,6 +80,7 @@ TeacherTaskPlanListing = React.createClass
         isLoaded={-> TeacherTaskPlanStore.isLoaded(courseId)}
         isFailed={-> TeacherTaskPlanStore.isFailed(courseId)}
         render={-> <CourseCalendar plansList={plansList} courseId={courseId}/>}
+        renderStatus={-> <CourseCalendar/>}
```
philschatz: If you just render a course calendar when loading it's unclear if the calendar is loaded but there is nothing to show or if it has not loaded yet. Why not just render `Loading Calendar...`?

philschatz:
thanks for making `<Loadable>` customizable!

src/components/loadable-item.cjsx
```diff
@@ -42,10 +47,13 @@ module.exports = React.createClass
...
 
+    renderModes = {renderLoading, renderError, renderBug}
+
     <Loadable
       store={store}
       isLoading={isLoadingOrLoad}
       isLoaded={-> isLoaded(id)}
       isFailed={-> store.isFailed(id)}
       render={renderItem}
+      {renderModes}
```
philschatz: Woah, I didn't know about this notation!

style/calendar.less
```diff
@@ -80,6 +80,24 @@ body.modal-open {
     .bootstrap-column-padding-override();
   }
 
+  .calendar-loading {
+    .rc-Month:after {
+      position: absolute;
+      width: 100%;
+      height: 100%;
+      content: 'Loading Plans...';
```
philschatz: I don't think teachers know what plans are. Maybe `Loading Calendar...`? Thoughts on the loading screen @Fredasaurus (see PR Screenshot)

PR #182, opened 2015-05-04
WIP: 1 course / 1 role redirect from dashboard

darrenmason:
currently having transitionTo in render causes errors, not sure if its the where I call transitionTo or maybe how.

### DIFF
--- a/src/components/index.cjsx
+++ b/src/components/index.cjsx
@@ -35,11 +35,11 @@ Dashboard = React.createClass
           singleCourseId = courses[0].id
           if roles[0].type is 'student'
             _.defer  =>
-              @context.router.transitionTo('viewStudentDashboard', {courseId: singleCourseId})
+              @context.router.replaceWith('viewStudentDashboard', {courseId: singleCourseId})
             null
           if roles[0].type is 'teacher'
             _.defer  =>
-              @context.router.transitionTo('taskplans', {courseId: singleCourseId})
+              @context.router.replaceWith('taskplans', {courseId: singleCourseId})
             null
       if courses.length
         courses = _.map courses, (course) ->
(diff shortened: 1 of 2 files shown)


src/components/index.cjsx
```diff
@@ -20,15 +20,26 @@ Dashboard = React.createClass
...
+
   render: ->
     if CurrentUserStore.isCoursesLoaded()
       courses = CurrentUserStore.getCourses()
+      if courses.length is 1
+        roles = courses[0].roles
+        if roles.length is 1
+          singleCourseId = courses[0].id
+          if roles[0].type is 'student'
+            @context.router.transitionTo('viewStudentDashboard', {courseId: singleCourseId})
```
philschatz: Have you tried `_.delay(fn, 10)` here to see if that will fix the `InvariantError`?

darrenmason: maybe this logic could be in the routehandler itself so we don't break history.  because even when this transition works it will always make stepping through /dashboard/ route unstable?

philschatz: That's true. I'm not sure how easy it is to change the handler itself. If it is not that simple then maybe we could replace all the `transitionTo('dashboard', ...)` code with a function that performs this logic of determining if there is only 1 course and then skipping the dashboard and going directly to the student list or teacher calendar.

nathanstitt: How important is it that the url always include "/dashboard" when what is displayed is the Dashboard?  

I'm thinking it might make sense to replace the task list with a shell which employ the above logic to decide whether to render either the task listing or the student dashboard.

That would remove the need for a redirect and the url would remain the same.  Only what is displayed would change depending on the number of courses/roles.

philschatz: Oh, that might be an option. But how would that impact the course name in the top nav bar (since the WIP PR relies on the `courseId` in the URL)?

nathanstitt: Well.  Hm.  I'd say that we'll have to figure out a different technique in #183 then since it seems like it'll be overly restrictive on other components.  Making additional comments there so we don't litter this up with 'em.

PR #184, opened 2015-05-05
Add global LESS color variables

dak:


pandafulmanda:
:+1: :fireworks: YAYYYY

PR #188, opened 2015-05-06
Student dashboard styling

nathanstitt:
screen shot 2015-05-08 at 11 40 27 am

Hover state on rows.  Keep in mind this is a 256 color gif - color's are not very accurate, the hover background looks a bit darker than it appears here. It's really tutor-neutral-lighter (#e5e5e5)

one

Todo:
 * [x] Reorganize files to match new locations
 * [x] Break apart less files into components to match cjsx, 
 * [x] Use pre-defined fonts/ colors
 * [x] remove link texts. Just click anywhere to go to the HW/iReading
 * [x] Homeworks should always be clickable.  When due date is past, they'll display the review
 * [x] Don't render "Don't Forget" panel for now.
 * [x] Jump to start first step instead of last when clicking.  In PR #202 
 * [x] Remove course title and teacher names now that it's in the navbar.
 * [x] Style clickable sections with hover effects
  * [x] Grow icons like breadcrumbs do (maybe w/shadows if easy)
  * [x] make background a light grey.


philschatz:
I believe "Check for updates and reload using etags" is covered in #193 but I think @Fredasaurus may have some more tweaks for the screen like:

- removing the link text (just click anywhere to go to the HW/iReading)
- make completed HW items still clickable if the due date has not occurred yet

nathanstitt:
Cool - wasn't aware that #193 covered the same ground.
@Fredasaurus I've updated the main TODO to reflect items that I'm awaiting input on.  When you're able please take a look

nathanstitt:
@Fredasaurus I've done a few different variations on the hover effect for you to pick from.

Growing the icon and 241 light grey bg:
icon-grow
----------
Applying a shadow to make the row "pop out":
row-shadow
--------------
Both effects combined:

both

Fredasaurus:
1. thanks. Also, can we make the over state on the dashboard circle buttons slight less large? I assume that's the same over state as on the breadcrumb now.

philschatz:
No votes for keeping the icon size the same?

Fredasaurus:
Wasn't an option ;) We could try that.

nathanstitt:
heh, I guess there wasn't one.  here you go:
one

Fredasaurus:
Hey sorry if I wasn't clear. No need for a separate border style on hover. just a border-bottom: 1px dotted rgba(229,229,229,1) that is there all the time.

Fredasaurus:
I think we'll want the icons to get larger on hover since that is the convention being employed across the app.

PR #191, opened 2015-05-06
update api backend for calendar

pandafulmanda:
No UI updates.

Should probably wait for #190 before merging


api/courses/1/dashboard.json
```diff
@@ -83,6 +83,63 @@
...
+                    "exercises_count_dynamic": 2
+            }
+        },{
+            "id": "5",
+            "type": "homework",
+            "title": "Complete exercise 12A on pg 12",
+            "opens_at": "2015-03-31T11:40:23.796Z",
+            "due_at": "2015-04-01T04:00:00.000Z",
+            "settings": {
+                    "exercise_id": ["11"]
```
philschatz: Should this be `exercise_ids`? And is the `/api/.../events.json` still used?

pandafulmanda: good point, also, looks like settings isn't supposed to be in this endpoint anyway

PR #193, opened 2015-05-06
reload an item when mounted so updates are seen

philschatz:
Student:
- Doing part of an iReading and then going back to the course list shows updated data

Teacher:
- Editing the title of a HW and going back to the calendar should show the new title

### DIFF
--- a/src/components/loadable-item.cjsx
+++ b/src/components/loadable-item.cjsx
@@ -18,18 +18,25 @@ module.exports = React.createClass
     renderItem: React.PropTypes.func.isRequired
     saved: React.PropTypes.func
     load: React.PropTypes.func
+    renderLoading: React.PropTypes.func
+    update: React.PropTypes.func
 
-  componentDidMount: -> @reload()
-  componentDidUpdate: -> @reload()
+  componentDidMount: -> @reload({})
+  componentDidUpdate: (oldProps) -> @reload(oldProps)
 
-  reload: ->
+  reload: (oldProps) ->
     {id, store, load, actions} = @props
+
+    # Skip reloading if all the props are the same (the case in the Calendar for some reason)
+    if oldProps.id is id and oldProps.store is store and oldProps.actions is actions and oldProps.load is load
+      return
+
     load ?= actions.load
     unless store.isNew(id)
       load(id)
 
   render: ->
-    {id, store, actions, load, isLoaded, isLoading, renderItem, saved, renderLoading, renderError, renderBug} = @props
+    {id, store, actions, load, isLoaded, isLoading, renderItem, saved, renderLoading, renderError, renderBug, update} = @props
 
     load ?= actions.load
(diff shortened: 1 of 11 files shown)


src/components/loadable-item.cjsx
```diff
@@ -18,9 +18,25 @@ module.exports = React.createClass
     renderItem: React.PropTypes.func.isRequired
     saved: React.PropTypes.func
     load: React.PropTypes.func
+    renderLoading: React.PropTypes.func
+    update: React.PropTypes.func
+
+  componentDidMount: -> @reload({})
+  componentDidUpdate: (oldProps) -> @reload(oldProps)
```
pandafulmanda: can this be
```
  componentWillUnmount: -> @reload(@props)
```

instead?  I quickly tested through this and currently, this is causing problems for homework and navigating between problems using breadcrumbs.  really, it would seem we only need to reload on the component unmount and this worked for the use cases described in the PR for me

nathanstitt: Ohh sneaky.  That gave me another idea - why don't we just invalidate the store's data on unmount?  That way the next time it's visited (if it is), it'll reload naturally.

pandafulmanda: !  that's pretty cool. like [this](https://github.com/openstax/tutor-js/blob/master/src/flux/task-step.coffee#L17-L20)?   this could be part of the crud store config, and then
```
  componentWillUnmount: -> @props.store.forceReload()
```

philschatz: I was hoping to include https://github.com/openstax/tutor-js/compare/reload-when-mounted...reload-when-mounted-2 next. That way the user does not see the `Loading...` every time they visit the calendar and we can take advantage of React's core feature of only updating the DOM nodes that changed (so once the new data is loaded the screen just updates)

nathanstitt: @philschatz Ah, yeah that'd be the best user experience by far.  And is exactly what React promises - no flash of an entire screen rendering, just the affected text would update.

philschatz: @pandafulmanda : Why would you need to reload on unmount since the component is no longer going to be in the DOM?

pandafulmanda: That is the time at which the destination loadable needs to reload -- i.e. when the loadable for homework gets mounted and you are working from problem to problem, props will update, but the store for the step may not necessarily need to.  when we go from making progress on the homework back to the dashboard or anywhere else, the destination's loadable store needs to check the server for updates

philschatz: My understanding was that sometimes React just updates the `<Loadable>` with new props and sometimes (un)mounts it depending on the React DOM. I'll take a closer look

pandafulmanda: It unmounts when we change routes, but when were just stepping through tasks, etc, its updating

PR #194, opened 2015-05-06
fix calendar reloading when returning from screen

philschatz:
When returning from HW or iReading builder the calendar would be blank:

image


dak:
Verified :+1:

PR #195, opened 2015-05-07
Task breadcrumb updates

pandafulmanda:
screen shot 2015-05-07 at 7 52 45 pm
screen shot 2015-05-07 at 1 40 06 pm

- [x] tests
- [x] styles refactor by component


philschatz:
Cool! How much work would it be to remove the landing page (that just says `Practice` or `iReading #4 ...`)? I think Jason/Fred didn't want it anyway

pandafulmanda:
very easy.  what about the homework landing page?

philschatz:
Right, probably for all pages.

philschatz:
hmm, this seems to have tests; is it still WIP?

philschatz:
Oh, Jason mentioned that the practice breadcrumbs should match the HW ones (use numbers instead of icons)

resources/styles/components/question.less
```diff
@@ -0,0 +1,182 @@
...
+    width: 100%;
+    margin: 20px 0 5px 0;
+    background: #f5f5f5;
+
+    &:before {
+      content: "My Answer:";
+      display: inline-block;
+      float: left;
+      margin-left: -145px;
+      #fonts > .script;
```
philschatz: When calling mixins with no args can you use parentheses?

resources/styles/components/question.less
```diff
@@ -0,0 +1,182 @@
...
+      content: "Why?";
+      float: left;
+      margin: 3px 0 0 -56px;
+      #fonts > .script;
+      color: #888;
+    }
+  }
+}
+
+.question {
```
philschatz: Why not combine this with the previous scope?

resources/styles/components/question.less
```diff
@@ -0,0 +1,182 @@
...
+  &.has-correct-answer {
+    .answer-input-box { display: none; }
+
+    .answers-table .answers-answer .answer-label {
+      padding-left: 40px;
+    }
+
+    .answers-table .answer-correct:before,
+    .answers-table .answer-checked:not(.answer-correct):before {
+      content:@fa-var-check;
```
philschatz: Can you include a space between the rule name and the rule value?

resources/styles/components/task-step/ends.less
```diff
@@ -0,0 +1,5 @@
+.task-review-completed{
+  .panel-footer {
+    display: none;
```
philschatz: When you include a `display: none;` can you include a comment why? When revisiting the code it seems odd to have written code to generate the DOM and then hide it.

resources/styles/components/task/breadcrumbs.less
```diff
@@ -0,0 +1,59 @@
+.circle(@radius){
```
philschatz: Can you prefix the mixin with the name of the component so the names do not conflict with mixins elsewhere?

resources/styles/components/task/breadcrumbs.less
```diff
@@ -0,0 +1,59 @@
...
+  font-size: @radius;
+  border-radius: 2 * @radius;
+  line-height: 2 * @radius;
+  min-width: 2 * @radius;
+}
+
+
+// http://caniuse.com/#search=counter
+.task-homework, .task-practice, .task-chapter_practice{
```
philschatz: Can you place separate selectors on separate lines? (and a space before the curly brace)

resources/styles/components/task/details.less
```diff
@@ -0,0 +1,48 @@
+@details-popover-arrow-width: 17px;
+@details-popover-arrow-outer-width: 18px;
+
+.task-details-popover{
+  font-size: 0.8em;
+  max-width: 500px;
+  // THE GREATEST SIN IN CSS!  sorry, just really want the popover height to not matter =(
```
philschatz: Could you increase the specificity of this selector and avoid having to use `!important`?

src/components/task-step/exercise-multiple-choice.cjsx
```diff
@@ -37,7 +37,7 @@ ExerciseFreeResponse = React.createClass
     # TODO: Assumes 1 question.
     question = content.questions[0]
 
-    <div className='exercise'>
+    <div className='question'>
```
philschatz: Should this be `exercise-multiple-choice` instead (or rename the file)?

src/components/task/breadcrumbs.cjsx
```diff
@@ -1,72 +1,30 @@
...
 
 CrumbMixin = require './crumb-mixin'
+Breadcrumb = require './breadcrumb'
+BindStoreMixin = require '../bind-store-mixin'
 
 module.exports = React.createClass
   displayName: 'Breadcrumbs'
 
-  mixins: [CrumbMixin]
+  mixins: [CrumbMixin, BindStoreMixin]
```
philschatz: Is listening to the store here necessary? I'm worried about to many components all listening to store changes and recursively re-rendering rather than a shared parent component re-rendering once.

resources/styles/components/task/details.less
pandafulmanda: this is to override inline styles bootstraps' popover puts on them to position them

philschatz: oh, no way to override that; ok.

PR #197, opened 2015-05-07
Use woff format fonts, compat with modern browsers

nathanstitt:
This has subtle UI changes, and things look just a bit different now that fonts are loading. 

As an example - without fonts loading:
screen shot 2015-05-07 at 2 34 55 pm

and with fonts:
screen shot 2015-05-07 at 2 35 23 pm


resources/styles/global/typography.less
```diff
@@ -1,92 +1,27 @@
...
+.tutor-font-face(@name, @filename, @weight, @style:normal) {
+  @font-face {
+    font-family: @name;
+    src: local(@name),
+    url('fonts/@{filename}.woff2') format('woff2'),
+    url('fonts/@{filename}.woff') format('woff');
+    font-weight: @weight;
+    font-style: @style;
+  }
 }
```
nathanstitt: I elected not to move this to mixins.less, but could be easily swayed if anyone thing it has use outside of typography.

philschatz: I think @dak wrote these initially and noted that it was difficult to remove the copy/pasta.

1. Is there a reason for including the weight and style here?
2. can people that use this mixin use a token (no Quotes) instead of a string?
  - that way we **could** have LESS do some compile-time-checks

nathanstitt: 1) Oh, I see why they were there now.  @dak was using the same family name and setting the weight and style so the appropriate font would be selected when the weight and height matched.  I'll rework to use that method so they'll make sense.
http://stackoverflow.com/questions/28279989/multiple-font-weights-one-font-face-query

2) Sorry, I think you've reached the bounds of my less knowledge.  Do you mean the url() generation? If so I suppose we could, but then the caller would have to specify the complete filename for both files (woff and woff2).

I'm getting the font list from @darrenmason for inclusion.

PR #199, opened 2015-05-08
Bug/fix cal plan stacking

pandafulmanda:
This PR fixes this weird plan stacking thing and also the finnicky date hovering as a result of an extra div wrapping the plans for each week.

## From This

screen shot 2015-05-08 at 1 37 43 am

## To This

screen shot 2015-05-08 at 1 38 41 am

## Range mode is disabled
shown here for future reference when we need units as ranges.

screen shot 2015-05-08 at 4 42 45 am


src/components/course-calendar/plan.cjsx
```diff
@@ -74,10 +78,11 @@ CoursePlan = React.createClass
     planStyle =
       width: durationLength * 100 / 7 + '%'
       left: offset * 100 / 7 + '%'
+      top: (topOffset + 4 - order * 3) + 'rem'
```
philschatz: Can you include a comment on why/how this is calculated?

PR #200, opened 2015-05-08
Add "Admin Console" link in navbar dropdown if user is an admin

karenc:
Couldn't figure out how to add a navbar test for admins...

Depends on https://github.com/openstax/tutor-server/pull/257 "Return whether user is an admin in /api/user"

### DIFF
--- a/api/user.json
+++ b/api/user.json
@@ -1,3 +1,4 @@
 {
-  "name": "Johhny Tran"
+  "name": "Johhny Tran",
+  "is_admin": false
 }
--- a/src/components/navbar/index.cjsx
+++ b/src/components/navbar/index.cjsx
@@ -39,9 +39,13 @@ module.exports = React.createClass
   # changes, this component never update it's state with the course in that case.
   addBindListener: ->
     @context.router.getLocation().addChangeListener(@handleCourseChanges)
+    CurrentUserStore.addChangeListener(@updateAll)
 
   removeBindListener: ->
     @context.router.getLocation().removeChangeListener(@handleCourseChanges)
+    CurrentUserStore.removeChangeListener(@updateAll)
+
+  updateAll: -> @setState({})
 
   bindUpdate: ->
     @handleCourseChanges()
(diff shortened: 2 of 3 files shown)


api/user.json
```diff
@@ -1,3 +1,4 @@
 {
-  "name": "Johhny Tran"
+  "name": "Johhny Tran",
+  "is_admin?": false
```
philschatz: Why the question mark in the JSON key? I think the `is_` prefix is sufficient to know the user is an admin.

src/components/navbar/index.cjsx
```diff
@@ -93,6 +94,7 @@ module.exports = React.createClass
             title={<UserName/>}
             ref='navDropDown'>
             {menuItems}
+            <AdminConsole/>
```
philschatz: I don't think you need to create a whole `Component` just for the admin button. I think you can do something like the following at the top of `render`:

```coffee
if CurrentUserStore.isAdmin()
  adminLink = <BS.MenuItem
      href='admin'
      key='dropdown-item-admin'>Admin Console
    </BS.MenuItem>
```

and then replace this line with `{adminLink}`

api/user.json
karenc: That's the default in rails.  I can change it to `is_admin` instead?

src/components/navbar/index.cjsx
karenc: Oh I see, thanks!  I'll try it out.

karenc: `CurrentUserStore.isAdmin()` is undefined when `render` is called.  It only gets updated later.  So the admin link is never shown.

philschatz: oh I see. Could we just put the additional link in `src/components/navbar/username.cjsx` since it already listens to `CurrentUserStore`?

karenc: I can move what I have in `src/components/navbar/admin-console.cjsx` into `src/components/navbar/username.cjsx`.  I'm not sure how else I can do it...

PR #201, opened 2015-05-08
Pinned breadcrumbs and footer for tasks

pandafulmanda:
screen shot 2015-05-08 at 4 34 16 am

## Updated answer table as well

will adjust styles more as needed, this change just includes the change where unpicked answers are now hidden in review mode

screen shot 2015-05-11 at 2 07 40 pm

- [x] pull out fixed header footer panel type thing into it's own component or mixin
  * will probably do after #195
- [x] update tests
- [ ] adjust for review page -- new PR


package.json
```diff
@@ -30,6 +30,7 @@
     "react-bootstrap": "~0.18.0",
     "react-calendar": "^0.4.0",
     "react-router": "~0.13.2",
+    "react-scroll-components": "^0.2.2",
```
philschatz: I completely forgot about this package (even have it bookmarked). Could this be used for the exercise builder too @dak?

resources/styles/components/task/index.less
```diff
@@ -0,0 +1,35 @@
+.fixed-bar() {
```
philschatz: Can you prefix this with `task-` or include it in `mixins.less` (if it ends up being used elsewhere like the exercise builder)?

src/components/navbar/index.cjsx
```diff
@@ -53,6 +54,9 @@ module.exports = React.createClass
   transitionToMenuItem: (routeName, params) ->
     @context.router.transitionTo(routeName, params)
 
+  shouldHide: ->
+    @state.scrollTop > 60 and document.getElementsByClassName('task')
```
philschatz: Why not use `document.querySelector('.task')`?

philschatz: Also, @Fredasaurus Would it be OK to always hide the top nav bar when scrolling? (and have it "pop" back in when scrolling up?)

philschatz: This bit breaks the exercise builder since the header hides. I think the exercise builder should probably change its margins to be similar (to regain vertical screen space).

Not pinned:

image

Pinned:

image

resources/styles/components/task/index.less
```diff
@@ -0,0 +1,35 @@
...
+  left: 0;
+  position: fixed;
+  width: 100%;
+  background-color: fadeout(@tutor-white, 20%);
+}
+
+
+.task-view {
```
philschatz: Can you include `body.` to this to clarify that this is a class name on the root (worries me a little that this is outside of react)

resources/styles/components/task/index.less
```diff
@@ -0,0 +1,35 @@
...
+  padding-top: 2 * @tutor-navbar-height;
+  padding-bottom: @tutor-navbar-height;
+
+  .panel-header {
```
philschatz: Should this still be a `panel-header` since it is always pinned? I'm worried that we are overusing bootstrap terms and changing the semantics of what these elements are.

resources/styles/components/task/index.less
```diff
@@ -0,0 +1,35 @@
...
+    .fixed-bar();
+    padding: 1rem @tutor-navbar-padding-horizontal;
+    margin: 0;
+
+    button.task-details {
+      float: right;
+      position: static;
+    }
+  }
+  .panel-footer {
```
philschatz: Similarly, should this be a `.panel-footer` since:

1. it occurs inside `.panel-body`
2. is always pinned to the bottom

src/components/navbar/index.cjsx
pandafulmanda: ooopsss, big oops =).  fixing this.

resources/styles/components/task/index.less
pandafulmanda: will pull these things out into a different component or mixin =)

PR #202, opened 2015-05-08
Go to first step in task, update url on bc click

nathanstitt:
Adds a stepIndex param to the route and updates it when breadcrumb is clicked.  The new parameter is then used for jumping to the first step of a task on the dashboard.

The index is 1 based so that the URL parameter matches the text on the breadcrumb button.


pandafulmanda:
:+1: nifty

PR #203, opened 2015-05-08
Fix infinite loop on homework review page for past due homeworks

pandafulmanda:


src/components/task-step/ends.cjsx
```diff
@@ -60,6 +60,7 @@ HomeworkEnd = React.createClass
   onNextStep: ->
 
   renderReviewSteps: (steps, label, type) ->
+    TaskStepActions.HACK_DO_NOT_RELOAD(true)
```
philschatz: eep, I'm worried about including this in actual code; is there a better way to handle this?

pandafulmanda:
on it

PR #206, opened 2015-05-09
overall/chapter/back to dash navigation

darrenmason:
image


philschatz:
Thanks!

Should the chart title update when drilling down?

Also (added to pivotal task):

- `drawPlane` seems to fail when there is only 1 chapter (no `prevPoint`)
- The right-most chapter/section should be selected by default

darrenmason:
good idea ill update the title and add conditions for only 1 chapter

philschatz:
Thanks. can you fix the existing tests and add one for the case of only having 1 data point to show? (and maybe the ability to click and drill down)

darrenmason:
will do

philschatz:
I noticed the drilled-down version does not show the section number for some reason:

image

darrenmason:
i think this maybe related to chapter_section formatting, looking into it

darrenmason:
confirmed with joe that issue is related to BE returning "chapter_section": "1.1", versus FE expecting "chapter_section": ["1", "1"]

philschatz:
and https://github.com/openstax/tutor-server/issues/267

src/components/learning-guide-chart.coffee
```diff
@@ -10,6 +10,7 @@ FLAG_BLUE      = 'flag-blue.svg'
 FLAG_GREEN     = 'flag-green.svg'
 FLAG_YELLOW    = 'flag-yellow.svg'
 FLAG_GREY      = 'flag-grey.svg'
+ICON_PENCIL    = 'icon-pencil.svg'
```
philschatz: Since this file is not used like the other icons can you name it something different? That way it does not conflict with other icon files?

philschatz:
Can you update the PR description screenshot with the pencils? Also, does each pencil have the same background color as the circle in the current screenshot?

darrenmason:
image

darrenmason:
The pencil has a transparent bg so the color behind that changes depending on level shows through

PR #208, opened 2015-05-10
do not show "loading..." if local data is available

philschatz:
Pages will silently update if successful and error loudly otherwise.


pandafulmanda:
:+1:

PR #213, opened 2015-05-11
Only render katex nodes with Mathjax

nathanstitt:
I believe the MathJax Typeset call is what's causing the "Invariant Violation" exceptions in the task steps.  If it's disabled, the exceptions no longer occur. 

As part of that investigation, I discovered that the MathJax.Hub.Queue call's arguments were specified incorrectly.  The element should be given as part of the array.  Since it wasn't specified property, Mathjax is actually typsetting the entire document.  http://docs.mathjax.org/en/latest/queues.html#using-queues (*search for Typeset*)

Unfortunately, this does not correct the "Invariant Violation".  Even with this fix, MathJax still causes them.  When I set a breakpoint on the dom tree before the call, I see multiple mutations being performed, but they all appear to be inside the html of the Html component, so React shouldn't care about them.

The Tex component from Khan academy is an interesting implementation of MathJax with React: https://github.com/Khan/react-components/blob/master/js/tex.jsx

### DIFF
--- a/src/components/html.cjsx
+++ b/src/components/html.cjsx
@@ -31,3 +31,5 @@ module.exports = React.createClass
     links = root.querySelectorAll('a')
     _.each links, (link) ->
       link.setAttribute('target', '_blank') unless link.getAttribute('href')?[0] is '#'
+    # MathML should be rendered by MathJax (if available)
+    window.MathJax?.Hub.Queue(['Typeset', MathJax.Hub, root])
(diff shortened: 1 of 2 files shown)


philschatz:
Hmm, I think I moved it out of the katex mixin because the iReading book content contains HTML+MathML (which MathJax needs to render)

nathanstitt:
Ah!  knew I should have checked git history.  I'll move it back, but keep the args corrected

PR #215, opened 2015-05-11
Clean up stray elements created by MathJax

nathanstitt:
As pointed out by the awesome @pandafulmanda - MathJax creates multiple elements outside the container that it's working on.

If we remove them after MathJax does it's thing, the "Invariant Violation" exceptions no longer occur.

### DIFF
--- a/src/components/html.cjsx
+++ b/src/components/html.cjsx
@@ -25,11 +25,28 @@ module.exports = React.createClass
     if html
       __html: html
 
-  componentDidMount: ->
+  componentDidMount:  -> @updateDOMNode()
+  componentDidUpdate: -> @updateDOMNode()
+
+  # Perform manipulation on HTML contained inside the components node.
+  updateDOMNode: ->
     # External links should open in a new window
     root = @getDOMNode()
     links = root.querySelectorAll('a')
     _.each links, (link) ->
       link.setAttribute('target', '_blank') unless link.getAttribute('href')?[0] is '#'
+
     # MathML should be rendered by MathJax (if available)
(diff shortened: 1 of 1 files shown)


pandafulmanda:
yayyy :+1:

src/components/html.cjsx
```diff
@@ -25,11 +25,28 @@ module.exports = React.createClass
...
     _.each links, (link) ->
       link.setAttribute('target', '_blank') unless link.getAttribute('href')?[0] is '#'
+
     # MathML should be rendered by MathJax (if available)
     window.MathJax?.Hub.Queue(['Typeset', MathJax.Hub, root])
+    # Once MathML finishes processing, manually cleanup after it to prevent
+    # React "Invariant Violation" exceptions.
+    # MathJax calls Queued events in order, so this should always execute after typesetting
+    window.MathJax?.Hub.Queue([ ->
+      for nodeId in ['MathJax_Message', 'MathJax_Hidden', 'MathJax_Font_Test']
```
philschatz: I wonder if `document.querySelectorAll(['#MathJax_Message', '#MathJax_Hidden', '#MathJax_Font_Test'].join(','))` would be quicker

nathanstitt: Great question!  Looks like good ol' `getElementById` is 98% faster though

http://jsperf.com/test-multi-queryselectorall-vs-getelementbyid-2

nathanstitt: *On both Chrome and Firefox, didn't bother booting up IE

philschatz: OK, suggestion retracted ; )

PR #221, opened 2015-05-12
Quick calendar styles

pandafulmanda:
screen shot 2015-05-12 at 2 05 41 am

quick styles around current, past, and future dates, to set up logic needed for when styles are finalized

### DIFF
--- a/src/components/course-calendar/month.cjsx
+++ b/src/components/course-calendar/month.cjsx
@@ -97,24 +97,30 @@ CourseMonth = React.createClass
 
       if dayIter.isAfter(referenceDay, 'day')
 
-        day = <Day
-          date={dayIter}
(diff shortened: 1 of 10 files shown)


src/components/course-calendar/month.cjsx
```diff
@@ -29,7 +30,7 @@ CourseMonth = React.createClass
     router: React.PropTypes.func
 
   getInitialState: ->
-    date: @props.startDate or moment()
+    date: @props.startDate or moment(TimeStore.getNow())
```
philschatz: Are there other places where `moment()` should be converted to `moment(TimeStore.getNow())`?

pandafulmanda: ah, yes, good point

PR #222, opened 2015-05-12
Handle placeholder homework qs

pandafulmanda:
## Message says which questions need to be answered
screen shot 2015-05-12 at 3 32 44 am

## Message updates
screen shot 2015-05-12 at 3 48 17 am

## Breadcrumb updates when last core question has been completed
screen shot 2015-05-12 at 4 55 11 am

Uses [this server branch](https://github.com/openstax/tutor-server/tree/FE/amanda-personalized-testing) 
- [x] tests
- [ ] update with underscore.string (this PR or new PR?)


package.json
```diff
@@ -26,6 +26,7 @@
     "katex": "~0.2.0",
     "markdown-it": "^4.1.1",
     "moment": "^2.9.0",
+    "pluralize": "^1.1.2",
```
philschatz: @nathanstitt There was a similar library you suggested, what was it?

src/components/pluralize.cjsx
```diff
@@ -0,0 +1,18 @@
+React = require 'react'
+
+# TODO find a better pluralizer
```
philschatz: @nathanstitt There was a similar library you suggested, what was it?

nathanstitt: I think it was probably underscore.inflections.  It does pluralization and titleize, camelize and a few other useful ones.  It can also be customized for special words that need special pluralization rules

https://github.com/geetarista/underscore.inflections

nathanstitt: Sorry - I got the link wrong on that.  The one I've used is: https://github.com/jeremyruppel/underscore.inflection  (not inflections).

When I was closing the tab I noticed that the previous one hadn't been updated in two years - so don't use it.

pandafulmanda: That looks great! @philschatz i remember you were hesitant about adding something like this --[underscore string](https://github.com/epeli/underscore.string) -- because we didn't need it at the time.  If I switch to either of these, we can replace the camelcase library as well.  Should I switch out for underscore inflections or underscore string now?

PR #224, opened 2015-05-12
IReading visual design

nathanstitt:
Objectives now:
screen shot 2015-05-12 at 2 13 17 pm

Overlay on splash image:
screen shot 2015-05-12 at 5 14 07 pm

### DIFF
--- a/resources/styles/components/task-step/reading.less
+++ b/resources/styles/components/task-step/reading.less
@@ -17,7 +17,7 @@
   .reading-step {
     // pull in common stylings for book contents
     @import '../book-content';
-    
+  
     section[data-depth="1"]:not(.section-opener),
     section[data-depth="2"],
     .worked-example,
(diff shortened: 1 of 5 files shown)


philschatz:
Do you mind styling all the iReading book content in a separate `.less` file and then importing it via something like:

```less
.task-step {
   .any-html.reading-step { @import './book-content'; }
  ...
}
```

I'm hoping to eventually reuse the styling in the reference book view or in webview.

src/components/task-step/all-steps.cjsx
```diff
@@ -23,6 +22,17 @@ Reading = React.createClass
     {content_html} = TaskStepStore.get(id)
     <ArbitraryHtmlAndMath className='reading-step' html={content_html} />
 
+  componentDidMount:  -> @insertOverlays()
+  componentDidUpdate: -> @insertOverlays()
+  insertOverlays: ->
```
philschatz: Should this check if the overlay already exists?

resources/styles/components/task-step/reading.less
```diff
@@ -15,6 +15,9 @@
 .task-step {
 
   .reading-step {
+    // pull in common stylings for book contents
+    @import '../book-content';
+  
     section[data-depth="1"]:not(.section-opener),
```
philschatz: All of these should probably also move into `./book-content.less`

PR #226, opened 2015-05-13
Refactor book styling

philschatz:
No UI Changes.

Some minor edits to the book styling. @dak and @nathanstitt thoughts?

TODO:

- [x] test splash image (I was unable to get a splash image to show up)


resources/styles/components/task-step/reading.less
```diff
@@ -17,74 +17,6 @@
...
-        border-bottom: 1px solid;
-        border-color: @border-color;
-      }
-
-    }
-
-
   }
 
   .section-opener ~ section h1 + p::first-letter, div[data-type="document-title"] ~ p::first-letter {
```
philschatz: Should these parts move into `book-content` as well?

nathanstitt:
Looks good to me.  I wasn't sure if those styles were "universal" and would belong in book-content, or should remain task-step specific.

I was able to view the splash image by loading the setup:010 rake task, logging in as the first student and viewing: http://localhost:3001/courses/1/tasks/1/1

resources/styles/components/task-step/reading/book-content/index.less
```diff
@@ -0,0 +1,141 @@
...
+> section:not(.section-opener),
+> section > section,
+.worked-example,
+.virtual-physics { padding: 0 140px; }
+
+> section { padding: 0; }
+
+.splash {
+  width: 100%;
+  margin: -20px 0 0 0;
```
philschatz: Rather than setting negative margins to compensate for padding on `.card-body`, what if `.card-body` had no margins and `.card-body:first()` had a `margin-top`?  Or, instead of using `20px` here we could refer to a variable containing the `.card-body` padding. Thoughts @Fredasaurus?

Fredasaurus: variable sounds good.

PR #228, opened 2015-05-13
some small style changes

pandafulmanda:
## After
screen shot 2015-05-13 at 1 06 02 pm

* pull breadcrumb pinned par up next to the nav bar
* give more line-height to LOs
* decrease extra margins in pinned-footer

### DIFF
--- a/resources/styles/components/pinned-header-footer-card/index.less
+++ b/resources/styles/components/pinned-header-footer-card/index.less
@@ -1,4 +1,4 @@
-@pinned-top-heading-buffer: 51px;
+@pinned-top-heading-buffer: @navbar-height + 1;
 
 body.pinned-view {
   padding-top: 2.5 * @tutor-navbar-height;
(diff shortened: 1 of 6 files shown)


resources/styles/components/pinned-header-footer-card/index.less
```diff
@@ -1,3 +1,5 @@
+@pinned-top-heading-buffer: 51px;
```
philschatz: Is this calculated based on another number (like `@tutor-navbar-height + 1`)?

resources/styles/components/pinned-header-footer-card/sections.less
```diff
@@ -3,6 +3,11 @@
   .fixed-bar(fadeout(@tutor-white, 20%));
   padding: 10px @tutor-navbar-padding-horizontal;
   bottom: 0;
+
+  * {
```
philschatz: `*` worries me. could this be restricted to `.btn` ?

resources/styles/global/body.less
```diff
@@ -0,0 +1,3 @@
+html, body {
+  height: 100%;
```
philschatz: Can you comment on why this is necessary?

PR #229, opened 2015-05-13
remove katex

philschatz:
It's slow and in my opinion ugly (with all the flicker):

Sorry, had to pause so ignore the opaqueness of the screenshot
image

This option requires adding the following to `tutor-server/.../webview.html.erb`:

```html
  <script type="text/x-mathjax-config">
  MathJax.Hub.Config({
    showProcessingMessages: false,
    tex2jax: { 
      inlineMath: [['{{MATH}}','{{MATH}}']],
      displayMath: [['{{MATH_BLOCK}}', '{{MATH_BLOCK}}']]
    }
  });
  </script>
  <script src="//cdn.mathjax.org/mathjax/2.5-latest/MathJax.js?config=TeX-MML-AM_HTMLorMML-full" async></script>
```

Also, the MathJax z-index is set to be very high (may impact iReading as well):
image


nathanstitt:
I also turned off the processing messages in an attempt to fix the Invariant Violation exceptions.  However it still showed the "Loading" type messages, which was pretty disappointing. The docs for it: http://docs.mathjax.org/en/latest/options/hub.html do mention that mis-feature though.

gregfitch:
Did I miss a step?

screen shot 2015-05-13 at 3 31 02 pm

Dantemss:
whoa what is `{{MATH}}`?
Also you can use some CSS trickery to hide the content before it's rendered...

philschatz:
@gregfitch: Did you update `tutor-server`'s `webview.html.erb` and re-run `rails s`?

@Dante: I tried using `display: none` but MathJax just skips the element because it uses the height/width of the element to decide how to render the math (like splitting equations on multiple lines)

gregfitch:
Yup...forgot to add the change to the script. A bit slow, but it does work in most case.
One where it didn't:
screen shot 2015-05-13 at 3 58 43 pm

Dantemss:
Supposedly, `visibility: hidden` works: https://github.com/mathjax/MathJax-docs/wiki/Hiding-content-until-typesetting-is-complete

philschatz:
Thanks @Dantemss . Using `visibility: hidden` works. I'm going to try to use some funky unicode chars instead of `{{MATH}}` (because `visibility: hidden` still makes a rectangle) and need to add block math markers in addition to inline ones (which I think will address Greg's issue).

gregfitch:
I'm not sure what is going on with that one question, but it does work elsewhere. The initial render can take a bit but it works well enough at the moment.

gregfitch:
BE: master; FE: remove-katex; Data: demo-001; Student: 11 - Kasey Hettinger; Homework - 4.1-4.2

philschatz:
If we go with MathJax for rendering exercises I'm afraid this is about the best (speed-wise) that it'll get. I'll also create a PR for option 2 (use katex to render exercises and MathJax to render math in book content)

gregfitch:
Ok...homework Q10 looks like a case of too many quotes...must be in the exercise itself. I'll look into it.

gregfitch:
Yes/No?

    <script type="text/x-mathjax-config">
        MathJax.Hub.Config({
            showProcessingMessages: false,
            tex2jax: { inlineMath: [
                ['{{MATH}}','{{MATH}}'],
                ['{{MATH_BLOCK}}','{{MATH_BLOCK}}']
            ] }
        });
    </script>
    <script src="//cdn.mathjax.org/mathjax/2.5-latest/MathJax.js?config=TeX-MML-AM_HTMLorMML-full" async></script>

screen shot 2015-05-14 at 5 08 49 pm

philschatz:
@gregfitch What do you mean by "Yes/No?"

gregfitch:
Are we expecting to see the `{{MATH_BLOCK}}` in addition to the `{{MATH}}`?

philschatz:
yeah. `{{MATH_BLOCK}}` denotes a block-level math element (no html to the left or right) vs the more common `{{MATH}}` which is inline math

philschatz:
tied to https://github.com/openstax/tutor-server/pull/290 (should this get merged)

Dantemss:
We fixed the divs btw... at least in exercises-demo they should not be escaped.

nathanstitt:
Just added some changes that use invisible UTF characters (U+200C zero-width non-joiner, U+200B zero-width space) and sets all [data-math] to hidden, but the inner .MathJax to visible.  This has the effect of hiding the math altogether, while still reserving space for it.

Initially the screen looks like:
screen shot 2015-05-19 at 4 39 30 pm

Before it renders and then it snaps to:
screen shot 2015-05-19 at 4 39 06 pm


I also played with setting the initial content to display: none and then switching it back after render, but that didn't work as noted above by @philschatz and @Dantemss

nathanstitt:
I think we should also move the initialization code in  https://github.com/openstax/tutor-js/blob/master/index.coffee#L19-L23 to a `TutorClient.boot()` method and have it also initialize MathJax.  That way we wouldn't need to modify tutor-server's webclient.html whenever we need to change configuration.  That could be another PR once we figure out what exactly this should do.

index.html
```diff
@@ -3,7 +3,20 @@
   <head>
     <meta http-equiv="content-type" content="text/html; charset=UTF8">
     <link rel="stylesheet" href="/dist/tutor.css" />
-    <script src="//cdn.mathjax.org/mathjax/2.5-latest/MathJax.js?config=MML_HTMLorMML"></script>
+    <script type="text/x-mathjax-config">
+    MathJax.Hub.Config({
+      showProcessingMessages: false,
+      tex2jax: {
+        displayMath: [['‌‌‌','‌‌‌']], // 3 U+200C zero-width non-joiner
```
philschatz: Can you escape this character out so the file is still ASCII?

nathanstitt:
How often does content contain MathML?  And if it does contain it, is that content marked in anyway?  If we could figure out a method to identify content that needs MathJax ran on it even if they don't have any "data-math" attribute elements, it would speed up rendering of content considerably.

philschatz:
Book content is marked up with `<math>` elements but the only way to identify math in exercises is with the `data-math` attribute. We could also split the difference and try rendering in katex and fail back to MathJax if katex cannot parse it.

PR #231, opened 2015-05-13
Chapter titles for IReadings

nathanstitt:
UI:
screen shot 2015-05-15 at 9 52 45 am


Fredasaurus:
Can we font-size: 2.4rem?
Can icon be bigger (native is 30px high)?
Don't need a header for watch physics.

src/components/task-step/step-mixin.cjsx
```diff
@@ -23,8 +26,20 @@ module.exports =
...
+
+  setBannerClass: ->
+    has_banner = _.contains(@state.classNames, "with-ui-banner")
+    if not has_banner and this.getDOMNode().querySelector(":first-child[data-label]")
+      # The defer's needed to break out of the rendering loop so the store has time to
+      # mark itself as loaded
+      # TODO: investigate why
+      _.delay =>
+        @setState classNames: @state.classNames.concat ["with-ui-banner"] if @isMounted()
+      , 20
```
nathanstitt: I feel like this is "wrong" on many levels :grin:

The main thing that's complicating the code is that the "card-body" needs to have extra top margin if a banner is going to be displayed, otherwise it'll be rendered under the top navbar.

But the code has no way of knowing what the content looks like until it renders it.  A previous attempt of mine tried running a regex on it in the Reading component, but that was pretty brittle and felt even more wrong.

The other things that's weird is that their are two loaded events fired, one for the tasks endpoint, and then for the steps.  This causes the components to mount/unmount in rapid succession.  That's the root cause for needing the `.delay` and  `isMounted` checks.  I haven't investigated much further at this point - am hopeful that someone can point out a better way and I can remove this method altogether.

nathanstitt:
@Fredasaurus Sure, we can make the icon bigger - it's 40px right now though and the bg element is 50px.  I think what you're seeing might be that the icon's content is wider than it is tall.

I'll hide the title on "Watch Physics".  I showed that shot mainly so you could see what an "icon-less" one would look like.

Fredasaurus:
oh really? ok It should be 30px tall. Guess it was just the screenshot. thanks.

nathanstitt:
@Fredasaurus Icons are now 30px square and the container element is 40px tall with 5px padding.  I also refactored that into a variable so we can adjust at will

Fredasaurus:
sweetness.

resources/styles/components/task-step/reading/book-content/index.less
```diff
@@ -11,9 +11,42 @@
...
+
+.tutor-reading-step-banner-icon(@img){
+  background: @tutor-secondary url(@img);
+  background-repeat: no-repeat;
+  text-indent: @reading-ui-banner-height - 10px;
+  background-position: 5px;
+  background-size: @reading-ui-banner-height - 10px @reading-ui-banner-height - 10px;
+}
+
+// If the parent element that included this file has a .note[data-label] that is the first child
```
philschatz: eep! Maybe this should be a mixin so the inclusion behavior is more explicit.

src/components/task-step/step-mixin.cjsx
```diff
@@ -5,10 +5,13 @@ BS = require 'react-bootstrap'
 {TaskStepActions, TaskStepStore} = require '../../flux/task-step'
 LoadableItem = require '../loadable-item'
 {CardBody} = require '../pinned-header-footer-card/sections'
-
+_ = require 'underscore'
```
philschatz: Can you move global `require` calls above the local ones?

src/components/task-step/step-mixin.cjsx
```diff
@@ -22,8 +25,17 @@ module.exports =
...
     </CardBody>
+
+  componentDidMount:  -> @setBannerClass()
+  componentDidUpdate: -> @setBannerClass()
+
+  # If the rendered content has a matching element that will be used as the banner,
+  # add a "with-ui-banner" css class.  This way the component can be styled to match
+  setBannerClass: ->
+    has_banner = _.contains(@state.classNames, "with-ui-banner")
+    if not has_banner and this.getDOMNode().querySelector(":first-child[data-label]")
```
philschatz: I'm not sure if `data-label` is the correct attribute to check for, but probably OK for the demo.

Also, can you change this line to something like:

```coffee
unless hasBanner and @getDOMNode().querySelector(':first-child[data-label]')
```

src/components/task-step/step-mixin.cjsx
```diff
@@ -5,10 +5,13 @@ BS = require 'react-bootstrap'
 {TaskStepActions, TaskStepStore} = require '../../flux/task-step'
 LoadableItem = require '../loadable-item'
 {CardBody} = require '../pinned-header-footer-card/sections'
-
+_ = require 'underscore'
 
 module.exports =
 
+  getInitialState: ->
+    classNames: ['task-step']
```
philschatz: Why store the whole array of `classNames` in the state instead of just `hasBanner`?

nathanstitt: Good idea - I think I was going for extendability or something there, but I'm all for doing the simplest thing that would possibly work.  Thanks for pointing that out.

src/components/task-step/step-mixin.cjsx
nathanstitt: Yes, I'm not sure either, but don't see any really better way to identify the element.  I'm using the `data-has-label` attribute for the css selector though, so I'll update this to match it.

The unless won't work like that though.  We need to test if the state has not been set and the dom node is present.  If we flip it to be an unless that'd read as:   

`unless has_banner and not this.getDOMNode().querySelector(":first-child[data-label]")` 

 Which I think is even more confusing.  Unless and not don't really mix, so I've flipped the hasBanner to be bannerMissing and it reads better now

philschatz:
Is this PR still valid? If so, can you update the screenshot to have the styling inside the card?

nathanstitt:
I think @Fredasaurus is going to provide additional guidance on what the banner and title should look like.  I've made the change to bring it back inside and updated screenshot in the interim.

nathanstitt:
@Fredasaurus, something else you might consider is moving the banner to the right:

screen shot 2015-05-15 at 10 01 10 am

Fredasaurus:
Hmmm.... not sure about that one. I could say that it interrupts the F reading pattern or something -- mostly it just me feel uncomfortable.

nathanstitt:
Yeah, I wasn't sure either, which is why I edited the css in place and took the screen shot vs committing it.

PR #234, opened 2015-05-13
add steps to task URL

philschatz:
No UI Changes.

Thoughts @nathanstitt since I think you introduced the step index in the URL

### DIFF
--- a/src/router.cjsx
+++ b/src/router.cjsx
@@ -18,7 +18,8 @@ routes = (
     <Route path='dashboard/?' name='dashboard' handler={Dashboard} />
     <Route path='courses/:courseId/?'>
       <Route path='list/?' name='viewStudentDashboard' handler={StudentDashboardShell} />
-      <Route path='tasks/:id?/?:stepIndex?' name='viewTask' handler={SingleTask}/>
+      <Route path='tasks/:id/?' name='viewTask' handler={SingleTask}/>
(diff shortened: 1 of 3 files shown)


src/router.cjsx
```diff
@@ -18,7 +18,8 @@ routes = (
     <Route path='dashboard/?' name='dashboard' handler={Dashboard} />
     <Route path='courses/:courseId/?'>
       <Route path='list/?' name='viewStudentDashboard' handler={StudentDashboardShell} />
-      <Route path='tasks/:id?/?:stepIndex?' name='viewTask' handler={SingleTask}/>
+      <Route path='tasks/:id/?' name='viewTask' handler={SingleTask}/>
+      <Route path='tasks/:id/steps/:stepIndex/?' name='viewTaskStep' handler={SingleTask}/>
```
nathanstitt: Wow, so that works?

I tried that, `tasks/:id(/:stepIndex)?/?`,  and

    <Route path='tasks/:id/?' name='viewTask' handler={SingleTask}>
      <Route path='steps/:stepIndex/?' name='viewTaskStep' handler={SingleTask}/>
    </Route>

I think the nested one worked, but I had to also add a RouteHandler to the render and then the html wasn't correct?  Not sure on the details but I decided it wasn't feasible.

nathanstitt:
Go for it, glad you figured it out

PR #236, opened 2015-05-14
Feature/panel to split reading core and exercises

pandafulmanda:
screen shot 2015-05-13 at 6 47 39 pm

- [x] Update breaking tests
- [x] add tests


philschatz:
is there a corresponding BE PR for this or is it already in `tutor-server#master`?

pandafulmanda:
Yes, already on `tutor-server/master`.  It's purely FE

src/components/task/crumb-mixin.cjsx
```diff
@@ -57,7 +57,39 @@ module.exports =
...
+  # can possibly abstract this more and pull this out somewhere else if needed in the future
+  modifyCrumbs: (task, crumbs) ->
+    {currentStep} = @props
+
+    # insert spacer panel/crumb for reading task that have spaced practices or personalized problems
+    if task.type is 'reading'
+      notCore = _.find crumbs, (crumb) ->
+        (crumb.type is 'step') and not TaskStepStore.isCore(crumb.data.id)
+
+      if notCore?
```
philschatz: Could you change the logic so the spacer is only added if there are additional questions after the last core question?

src/flux/task-step.coffee
```diff
@@ -51,6 +51,14 @@ TaskStepConfig =
           isAnswered = false
       isAnswered
 
+    isCore: (id) ->
+      step = @_get(id)
+      coreGroups = [
+        'core'
+        'default'
```
philschatz: When is this `default`?

src/components/task/crumb-mixin.cjsx
pandafulmanda: ah, yes, good point

src/flux/task-step.coffee
pandafulmanda: I'm not sure when, but the BE sometimes returns it as 'default' and it is defined in the [napkin notes](https://github.com/openstax/napkin-notes/blob/master/jp/exercise_sources.md) as well

@kevinburleigh75 i think has insight into this

src/components/task/crumb-mixin.cjsx
pandafulmanda: oh wait, that line determines if there is at least one non-core question found, so if `notCore?` is true, that means there is at least one question after the last core question

philschatz: ok, thanks!

philschatz:
@kevinburleigh75 when is `core` is returned vs `default`?

PR #238, opened 2015-05-14
Fixes for reading overlay

nathanstitt:
I discovered that the overlay was leaving a blank area where it should have been due to the use of relative positioning adjustment.

Also put a check in to make sure the ui-overlay element isn't generated more than once.

Before:
screen shot 2015-05-14 at 9 50 58 am

After:
screen shot 2015-05-14 at 9 51 23 am


Fredasaurus:
Looking good. Is the chapter number not being supplied?

PR #239, opened 2015-05-14
Some objectives lack .ost-learning-objective-def

nathanstitt:
And are not styled.  This gives up identifiying them by class
and assumes that all li elements inside a .learning-objectives
are objectives.

Also add 2rem of indentation to MathJax so it looks neater.  The MathJax styling conflicts with the flex alignment on the list, causing it to be moved to a new line.  Since flex box is the only way to get the list to vertically space the elements equally I think that's the best we can do.

Before:
screen shot 2015-05-14 at 10 01 00 am

After:
screen shot 2015-05-14 at 10 03 50 am

### DIFF
--- a/resources/styles/components/task-step/reading/book-content/index.less
+++ b/resources/styles/components/task-step/reading/book-content/index.less
@@ -46,11 +46,9 @@ cnx-pi { display: none; }
     margin: 0 0;
     padding: 0 15px 0 30px;
     border-left: 1px solid rgba(255,255,255,.4);
-    ul {
-      display: flex;
-      flex-direction: column;
-    }
-    .ost-learning-objective-def {
+    display: flex;
+    flex-direction: column;
+    li {
       font-weight: 300;
       font-size: 1.5rem;
       line-height: 3rem;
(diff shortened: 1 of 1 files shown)


philschatz:
a missing `.ost-learning-objective-def` sounds like it should be reported to alina.

nathanstitt:
Ok, I'll re-add the `.ost-learning-objective-def` class shortly then.  I think we should still style MathJax a bit so the PR is still somewhat relevant.  

I've noticed a few others missing as well.  We should make a list and make sure they're fixed before demo since they stick out.

openstaxalina:
The content above seems to be old.  The new content is here: https://legacy-staging-tutor.cnx.org/content/col10004/latest

Please distribute this link to whomever needs it; I apologize that it did not reach everyone.

If you go into chapter 4, "Force" module (https://legacy-staging-tutor.cnx.org/content/m43217/latest/?collection=col10004/1.2) you will see that the third learning objective was removed from the final delivered manuscript (that the col10004 is based off).

philschatz:
I think the Backend bumped the version number so it might be fixed now.

PR #240, opened 2015-05-14
Video styles for iReading

nathanstitt:
@Fredasaurus: Not sure about what's intended for the "Visit Video" button.  It used to be a text link that just said "video" and appeared directly under the video.  Perhaps it should be hidden altogether?

screen shot 2015-05-14 at 2 04 33 pm


philschatz:
:+1: for removing the "Visit Video" link. I added it there because earlier we were unsure if BE would remove the iFrame and include the video URL in the JSON or keep the iframe.

PR #241, opened 2015-05-14
Custom bootstrap and materials import

pandafulmanda:
also, got rid of global/old-tutor

## Primary maps appropriately down to dropdown
screen shot 2015-05-14 at 2 06 13 pm

## Progress success and danger as well
screen shot 2015-05-14 at 2 09 25 pm


resources/styles/global/tutor-bootstrap.less
```diff
@@ -0,0 +1,50 @@
+// Core variables and mixins
+@import "@{bootstrap-path}variables";
```
philschatz: My only comment is it looks "weird" when there is no `/` between the `@{bootstrap-path}` and `variables`. Having 2 `/` is a safer fallback than having none (if the user forgot to add the trailing `/` to the variable definition)

nathanstitt:
:+1:  I think it looks good.  ( *although you wouldn't need to re-arrange the bootstrap includes if you used sass with it's support for default variables* :smirk: )

pandafulmanda:
actually, http://lesscss.org/features/#variables-feature-default-variables didn't have to move change the order of things, just loaded the variables afterward. I thought i had to, but this works.  we can go back to loading from node and bower, but this let's us omit some less files we aren't using

philschatz:
Looks good, but can you merge in changes from master?

PR #245, opened 2015-05-14
Work in Physics styles

nathanstitt:
screen shot 2015-05-14 at 4 38 06 pm

Todo: 
  * [ ] Possibly style "Here’s list of some of the years of satellite launches" sentence.

### DIFF
--- a/resources/styles/components/task-step/reading/book-content/work-in-physics.less
+++ b/resources/styles/components/task-step/reading/book-content/work-in-physics.less
@@ -28,4 +28,16 @@
 
   }
 
+  figure {
+    width: 100%;
+    margin: 0;
+    img {
+      width: 100%;
+    }
+  }
+
+  div[data-type="title"] + figure + p {
+    .tutor-reading-first-letter();
+  }
+  
 }
--- a/resources/styles/components/task-step/reading/book-content/work-in-physics.less
+++ b/resources/styles/components/task-step/reading/book-content/work-in-physics.less
@@ -35,8 +35,7 @@
       width: 100%;
     }
   }
-
-  div[data-type="title"] + figure + p {
+  > p:first-of-type {
     .tutor-reading-first-letter();
   }
   
--- a/resources/styles/components/task-step/reading/book-content/index.less
+++ b/resources/styles/components/task-step/reading/book-content/index.less
@@ -16,6 +16,16 @@
 @import './tips-for-success';
 
 
+.tutor-reading-first-letter {
+  &::first-letter{
+    float: left;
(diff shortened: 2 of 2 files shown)


nathanstitt:
@Fredasaurus On the mockup, you show the "Here’s list of some of the years of satellite launches" sentence as bold and a bit larger than the surrounding text. 

In the markup, that's just a `p` with no identifying information other than it's id.  Should I:

 * [ ] Style based on ID (would only apply to that p)
 * [ ] Style based on the fact that it's a para before a UL
 * [ ] Not style it at all.

Fredasaurus:
Heeeeeeeeere's list!

let's leave it as a <p> for now. We can revisit if need be, thanks.

Fredasaurus:
Ha ha ha ha ha. By blankspace I mean paragraph tag.

Fredasaurus:
So that would be option 3.

Fredasaurus:
Here is an extraneous comment.

nathanstitt:
:grinning:   I figured as much.  Awesome, then my work here is done (I think), since the title will be styled as part of whatever we figure out for #231.  Removing WIP prefix.

nathanstitt:
Here's your extraneous comment back, I don't need it but thank you anyways.

resources/styles/components/task-step/reading/book-content/work-in-physics.less
```diff
@@ -28,4 +28,16 @@
...
 
+  figure {
+    width: 100%;
+    margin: 0;
+    img {
+      width: 100%;
+    }
+  }
+
+  div[data-type="title"] + figure + p {
```
philschatz: Should this just be `p:first`?

Fredasaurus:
Think that is my doing, so yeah.

resources/styles/components/task-step/reading/book-content/work-in-physics.less
nathanstitt: `> p:first-child` doesn't work because it's not a direct descendant of the parent `.work-in-physics`.  `first-of-type` does though, switching to that

nathanstitt: I'm hesitant to modify the selector in `book-content/index.less` since that apply to all IReadings.

I took a stab at changing it to also use `first-of-type` but ran into problems with it matching `p` elements in the objectives, and not matching the page.   It would also match the first para on any reading, even ones that are a continuation of a previous page.  That may be desired behavior though.

Since it's been working fine thus far, I'd suggest we leave it until we can make a guided review of what elements it should and shouldn't match.

PR #246, opened 2015-05-15
Style/calendar continued

pandafulmanda:
screen shot 2015-05-14 at 11 42 53 pm

### Reading
screen shot 2015-05-15 at 2 52 35 pm

### Homework
screen shot 2015-05-15 at 2 52 42 pm


resources/styles/components/course-calendar/month.less
```diff
@@ -122,9 +122,20 @@
 
       &--upcoming {
         cursor: pointer;
-
+        &:after{
```
philschatz: technically it should be `&::after {` <http://www.w3.org/wiki/CSS/Selectors/pseudo-elements/:after> ; )

resources/styles/components/course-calendar/month.less
```diff
@@ -122,9 +122,20 @@
...
+          content: '+';
+          font-size: 2rem;
+          position: absolute;
+          top: 0.5rem;
+          left: 0.5rem;
+          color: @tutor-tertiary-light;
+          opacity: 0;
```
philschatz: Why make it transparent?

pandafulmanda: was gonna have them fade in, but i guess that might be a little much since the background already transitions.  ill put this in to the hover state

philschatz:
The `(20 students)` is still a bug in the iReading/HW details popup (not sure if it's BE or FE) but the rest looks great!

pandafulmanda:
yeah i have a question about that..what is it supposed to say?

philschatz:
I think it's supposed to show the number of students that have answered questions in this section. That way, students that partially complete an assignment still show up in the bar charts and the counts for the bar graphs would (roughly) be:

- 4.1 (15 Students)
- 4.2 (12 Students)
- 4.3 (3 Students)

PR #247, opened 2015-05-15
Feature/nav show on upscroll

pandafulmanda:
c-c-c-check it outtt

# Down scroll, past a certain point
### Pinned
screen shot 2015-05-15 at 10 56 13 am

# Slow up scroll
### Still Pinned
screen shot 2015-05-15 at 10 56 27 am

# Quick up scroll
### Un-pin as needed
screen shot 2015-05-15 at 10 55 58 am

### DIFF
--- a/src/components/pinned-header-footer-card/index.cjsx
+++ b/src/components/pinned-header-footer-card/index.cjsx
@@ -14,7 +14,7 @@ module.exports = React.createClass
 
   getDefaultProps: ->
     buffer: 60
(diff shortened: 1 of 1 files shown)


philschatz:
cool! Does this work for the HW builder (and the div that gets pinned)?

pandafulmanda:
The nav bar should hide there as well?

philschatz:
It seems (from the code) that we can enable/disable it for different screens. Thanks!

PR #249, opened 2015-05-15
Align text labels, move "Correct Answer" inside card

nathanstitt:
Before:
screen shot 2015-05-15 at 10 47 18 am
After:
screen shot 2015-05-15 at 12 10 46 pm

The main cause turned out to be that the :before element was attached to the incorrect element, which also caused the vertical alignment to be off.

### DIFF
--- a/resources/styles/components/question.less
+++ b/resources/styles/components/question.less
@@ -49,7 +49,7 @@
 
     &:before {
       content: "My Answer:";
-      margin-left: -145px;
+      margin-left: -130px;
       .answer-label();
     }
   }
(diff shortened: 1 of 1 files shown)


pandafulmanda:
uh oh, i did some stuff in #241 that will probably conflict with this

pandafulmanda:
actually, i doing some refactoring that will make it so that we dont need custom negative margins for each label right now

philschatz:
Thanks! but closing this in favor of #255

nathanstitt:
I agree 100%, #255 does much more and gets rid of the nasty negative left margin.

PR #255, opened 2015-05-15
styles refactoring for answer and label alignments, minor style changes

pandafulmanda:
removes the need to individually adjust each label margin-left

screen shot 2015-05-15 at 6 20 40 pm

screen shot 2015-05-15 at 5 37 12 pm

### DIFF
--- a/src/components/question.cjsx
+++ b/src/components/question.cjsx
@@ -35,7 +35,7 @@ module.exports = React.createClass
     if @props.feedback_html
       feedback = <ArbitraryHtml
         className='question-feedback has-html'
-        html={@props.feedback_html}
+        html="<div class='question-feedback-content'>#{@props.feedback_html}</div>"
         block={true}/>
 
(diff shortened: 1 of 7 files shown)


src/components/question.cjsx
```diff
@@ -33,7 +33,10 @@ module.exports = React.createClass
     hasCorrectAnswer = !! @props.correct_answer_id
 
     if @props.feedback_html
-      feedback = <ArbitraryHtml className='question-feedback has-html' html={@props.feedback_html} />
+      feedback = <ArbitraryHtml
+        className='question-feedback has-html'
+        html="<div class='question-feedback-content'>#{@props.feedback_html}</div>"
```
philschatz: Why is this appending HTML in the string instead of something like:

```coffee
feedback = <div className='question-feedback'>
  <ArbitraryHtml
    className='question-feedback-content has-html'
    html={@props.feedback_html}
    block={true}/>
</div>
```

PR #256, opened 2015-05-18
iReading styles integration

nathanstitt:
screen shot 2015-05-17 at 10 39 26 pm

screen shot 2015-05-17 at 10 40 08 pm

screen shot 2015-05-18 at 11 42 29 am


pandafulmanda:
:+1:

resources/styles/components/task-step/reading/book-content/index.less
```diff
@@ -25,20 +28,43 @@
...
     }
-    .title {
-      margin: @reading-ui-banner-height 20px 5px 20px;
-      #fonts > .sans(3rem, 3rem);
-      font-weight: 900;
-      letter-spacing: 2px;
+
+    > [data-type=exercise] {
+
+      content: "";
```
philschatz: What does this line do?

resources/styles/components/task-step/reading/book-content/index.less
```diff
@@ -25,20 +28,43 @@
...
+      > [data-type=title]:first-child {
+        
+        margin: @tutor-card-body-padding/2 @tutor-card-body-padding;
+        #fonts > .sans(3rem, 3rem);
+        font-weight: 900;
+        letter-spacing: 2px;
+      }
+      
+    }
+    .problem, .solution, .commentary {
```
philschatz: Can you include each selector on a separate line?

resources/styles/components/task-step/reading/book-content/index.less
```diff
@@ -25,20 +28,43 @@
...
+      margin-left: @tutor-card-body-padding;
+      margin-right: @tutor-card-body-padding
     }
   }
 }
 
+.note > .title, .commentary [data-type=title] {
+  color: #77af41;
```
philschatz: Should this color be defined at the top of the file?

resources/styles/components/task-step/reading/book-content/tips-for-success.less
```diff
@@ -1,19 +1,38 @@
...
-    #fonts > .sans(3.2rem, 3.2rem);
-    font-weight: 900;
-    color: @tutor-secondary;
-    white-space: nowrap;
-    background: #232e66;
-    padding: 10px 20px;
-    margin: 0 500px 15px 0;
+    height: 35px;
+    content: attr(data-label);
+    background: @tutor-quaternary url("icon-light-bulb.svg");
```
philschatz: Does splitting this into `background-color: ` and `.tutor-reading-step-banner-icon(...)` not work?

nathanstitt: Yeah, I started with the `tutor-reading-step-banner-icon` mixin but ended up changing every single thing on it for the tips-for-success, so I just made it be it's own thing.  It's a different size, color, and needs to have different text flow semantics.

resources/styles/components/task-step/reading/book-content/index.less
nathanstitt: That entire block is gone now.  The new mockups aren't using 0 margin on the left/right sides so there's now no need to have the margins set to 0 and then add it back on all the elements.

nathanstitt:
@Fredasaurus / @philschatz  I'd like your opinion of the video steps that have text before the video and no titles like this one:

screen shot 2015-05-18 at 12 01 10 pm

Because of the way the HTML is laid out, there's not an element that I can attach the frame to in order to style them like the other video steps such as this:

screen shot 2015-05-18 at 11 43 01 am

Is it important that the video have the frame at all times?  If it is, I'll have to add some code to manipulate the DOM manually after rendering to make sure that all the videos have a containing element.

nathanstitt:
Renamed title since this PR has grown to include styles other than worked examples, it now contains code to implement the latest mockup since when I changed one thing it'd ripple out and affect other step types as well.

nathanstitt:
The difference between the two video steps is that one has a wrapping div[data-type=media] container. 

The presence of that div allows us to position the wrapper as `position: relative`, and the iframe as `position: absolute` with it's right/left sides set at 3px so the drop shadow indent effect will appear. [full styles](https://github.com/openstax/tutor-js/blob/worked-example-polish/resources/styles/mixins.less#L26-L40)


The wrapped version:

```html
    <div data-type="note" data-label="Watch Physics">
      <div data-type="title" class="title">Newton’s Third Law of Motion</div>
      <div data-type="media">
        <iframe width="660" height="371.4" src="https://www.khanacademy.org/embed_video?v=By-ggTfeuJU"></iframe>
      </div>
    </div>
```

And the unwrapped:
```html
    <div data-type="note" data-label="Watch Physics">
      <div data-type="title" class="title">A Note called Watch Physics with YouTube Embedded</div>
      <p>Lorem ipsum dolo.</p>
      <iframe src="https://www.youtube.com/embed/40ETbLVkLKc"></iframe>
    </div>
```

Fredasaurus:
"drop shadow indent effect" What is this?

nathanstitt:
It's the subtle little frame that videos in the mockups have that makes them appear indented.  Maybe you didn't intend for that effect?  If not, we can just make the iframe 100% width and be done.

screen shot 2015-05-18 at 2 11 37 pm

Fredasaurus:
Ah, good eye. Not even sure where that came from. We can remove it, thanks.

nathanstitt:
Ok, disabled the drop shadow by default.  I left as a mixin in case we need to customize further later

PR #257, opened 2015-05-18
Style/steps cc and end panels

pandafulmanda:
quick panel paddings. need to check with @Fredasaurus about further styling

screen shot 2015-05-18 at 11 49 15 am
screen shot 2015-05-18 at 11 49 19 am


nathanstitt:
Ooh, I just added this as well on #256 since it was in the latest mockup from @Fredasaurus 

Mine looks like:
screen shot 2015-05-18 at 11 42 29 am


and I only used 20px margin for the concept coach step

pandafulmanda:
oh. oops

PR #260, opened 2015-05-18
Style iframe on sims to be 100% and height 800px

nathanstitt:
screen shot 2015-05-18 at 5 54 07 pm

TODO: (In a different PR) Figure out cross iframe communication so the iframe can resize to fit content.  Not sure if content will be same-origin but suspect not

### DIFF
--- a/resources/styles/components/task-step/reading.less
+++ b/resources/styles/components/task-step/reading.less
@@ -17,4 +17,5 @@
 
 .task-reading {
   @import './reading/completed';
+  @import './reading/interactive';
 }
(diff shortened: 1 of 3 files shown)


philschatz:
Is communication with the iFrame necessary? I don't think Patrick is planning on adding that code

nathanstitt:
It is to figure out the height we should make the iframe so it doesn't scroll.  800 seems large enough but there might be taller ones?

philschatz:
Ah, there are no taller ones due to restricting the sim to be designed to fit on the smallest screen size we are targeting.

PR #262, opened 2015-05-19
fix iReading publish

philschatz:
removes `exercise_ids` from iReadings

### DIFF
--- a/src/flux/task-plan.coffee
+++ b/src/flux/task-plan.coffee
@@ -32,6 +32,7 @@ TaskPlanConfig =
     # iReadings should not contain exercise_ids and will cause a silent 422 on publish
     if obj.type is PLAN_TYPES.READING
       delete obj.settings.exercise_ids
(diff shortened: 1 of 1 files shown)


src/flux/task-plan.coffee
```diff
@@ -21,12 +21,19 @@ TaskPlanConfig =
...
       @_local[planId].settings.exercises_count_dynamic ?= TUTOR_SELECTIONS.default
 
     #TODO take out once TaskPlan api is in place
     _.extend({}, @_local[planId], @_changed[planId])
+    obj = _.extend({}, @_local[planId], @_changed[planId])
+
+    # iReadings should not contain exercise_ids and will cause a silent 422 on publish
+    if obj.type is PLAN_TYPES.READING
+      delete obj.settings.exercise_ids
+
```
jivewise: We should delete the exercises count dynamic here as well, no?

``` delete obj.settings.exercises_count_dynamic ```

PR #276, opened 2015-05-19
Move book-content out of reading and use as mixin

nathanstitt:
No UI changes

Reorganizes book content styles into `.tutor-book-content` mixin and moves it's files into a top-level directory.


resources/styles/book-content/fun-in-physics.less
```diff
@@ -0,0 +1,9 @@
+.tutor-book-content {
```
philschatz: Is this defining a mixin named `tutor-book-content`? If so, can you include the parentheses to make it clear?

resources/styles/book-content/fun-in-physics.less
```diff
@@ -0,0 +1,9 @@
...
+  .fun-in-physics {
+    margin-top: 20px;
+    padding: 0 @tutor-card-body-padding-horizontal;
+    & > .title:first-child {
+      .tutor-book-content.tutor-reading-main-title();
```
philschatz: I'm not sure I understand what's going on here. Can you clarify?

nathanstitt: The `.tutor-book-content` prefix was needed so that the variables and mixins defined inside the `.tutor-book-content` could be accessed by a file that was re-opening the mixin vs being included from inside it.  I've now decided that we don't need to be that extreme and am moving the includes back inside `tutor-book-content`

PR #278, opened 2015-05-19
learning guide fixes

darrenmason:
safari fixes - fixed font size and parentNode bug in Safari8

### DIFF
--- a/resources/styles/components/learning-guide.less
+++ b/resources/styles/components/learning-guide.less
@@ -30,7 +30,7 @@
     margin: 0 auto;
 
     svg {
-      font-size:12px;
+      font-size: @html-font-size;
       width: @guide-width;
       height: @guide-height;
       min-width: @guide-min-width;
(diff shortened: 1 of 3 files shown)


resources/styles/components/learning-guide.less
```diff
@@ -30,6 +30,7 @@
     margin: 0 auto;
 
     svg {
+      font-size:12px;
```
philschatz: Can this use the same font size set on the `<html>` (maybe extract to a variable in `variable.less`)?

darrenmason:
adjusted if current level 0 the line is not below the graph

resources/styles/components/learning-guide.less
philschatz: That way Fred can globally change the rem size and make everything smaller/bigger

darrenmason: looks like we are getting root rem from the .sans font defined in body in typography.less. that overwrites html 10px.  the svg needs 12px equivalent, so it might make more sense to have a global var for svg root rem

PR #279, opened 2015-05-19
using the same animation for hover of dashboard icons and breadcrumbs

pandafulmanda:
screen shot 2015-05-19 at 1 44 49 pm

has drop-shadow and transition

### DIFF
--- a/resources/styles/components/student-dashboard/task.less
+++ b/resources/styles/components/student-dashboard/task.less
@@ -10,9 +10,8 @@
     }
     .icon {
       background-position: center;
-      height: @student-dashboard-row-height;
-      min-width: @student-dashboard-row-height;
-      width: 100%;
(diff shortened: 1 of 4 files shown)


nathanstitt:
Awesome!  I also tried this awhile ago and when I used the `scale` for the dashboard it also scaled the background making the icon appear to have a weird square border.  Wonder what I was doing wrong?

pandafulmanda:
ah! i snuck in a `border-radius: 50%;` [here](https://github.com/openstax/tutor-js/pull/279/files#diff-1aab5344055dab8ca418c01be9fe1412R59)

nathanstitt:
Aha, clever clever!  This is really cool, the two implementations were so close it really makes sense to combine them.

philschatz:
Thanks!

PR #280, opened 2015-05-19
Improvements on homework review page

pandafulmanda:
- [X] breadcrumbs update as you work homework problems on review page
  - try it, it's pretty kewl!
- [x] smooth out transition when problem completed
- [x] fix remaining questions needed for placeholder personalized question
- [x] disable continue for placeholder personalized question

- first problem leave may jump if you are not scrolled past a certain point :disappointed:

### Before
screen shot 2015-05-19 at 2 25 10 pm

### After
screen shot 2015-05-19 at 6 33 43 pm

### Before
Abruptly continues to next problem
screen shot 2015-05-19 at 6 53 20 pm
screen shot 2015-05-19 at 6 53 30 pm

### After
shows review screen for completed problem until user clicks continue
screen shot 2015-05-19 at 6 53 20 pm
screen shot 2015-05-19 at 6 53 26 pm
screen shot 2015-05-19 at 6 53 30 pm

PR #281, opened 2015-05-19
Practice pugs on done page

pandafulmanda:
## End panels

update with homework review, might wait till #280 is done for that

screen shot 2015-05-19 at 2 40 44 pm
screen shot 2015-05-19 at 2 41 11 pm

- [x] `do more practice` needs to be fixed
  - this is probably more complicated than I anticipated -- should this be a separate PR?
  - discussing with @kevinburleigh75 about a way to make this persist, but this is working for the user if they don't Reload during a practice right now.


philschatz:
can you merge from master?

PR #285, opened 2015-05-20
use form to sign out instead of AJAX

philschatz:
`GET` requests, while simpler to implement are not desirable because of URL prefetchers logging people out accidentally: https://stackoverflow.com/questions/3521290/logout-get-or-post/14587231#14587231

Styling the link still requires some work:
image

TODO:

- [ ] submit the form when `CurrentUserActions.logout` is called as well (or just redirect to an Error page)
- [x] fix backend or provide the authorization token


Dantemss:
Argh!

jpslav:
@Dantemss can we get the authenticity token to the FE so they can include it in the sign out form?  What's the lifetime of one of those tokens? I know in standard rails templates it is generated on every request, but I don't know how best to make that work in a JS client setting.

nathanstitt:
I was just thinking that the FE should really provide that on every request and was looking for where the BE was disabling them.   Looks like probably: https://github.com/openstax/openstax_api/blob/master/app/controllers/openstax/api/v1/api_controller.rb#L22

Shouldn't we require them?  Otherwise XSS could be employed by a malicious site to mark all answers on a HW to incorrect or something.

Dantemss:
Wait... that code doesn't disable the token when you're logged in through cookies... unless it's broken?

nathanstitt:
I'm ... not sure.  https://github.com/openstax/openstax_api/blob/master/app/controllers/openstax/api/v1/api_controller.rb#L47-L52

The last && clause checks `doorkeeper_token.nil?` which would indicate that the user is not using oauth correct?  Which all tutor accounts are, since they come from tutor-accounts.

I also see that the same check's used for setting CORS headers.  I popped open the network tab on the demo and don't see the `Access-Control-Allow-Origin` header.

Or maybe I'm just reading it totally wrong :)

nathanstitt:
Looks good.  I can style the logout link as part of #286 since that's also touching the menu

PR #286, opened 2015-05-20
Fix random orange color appearing on dashboard

nathanstitt:
On screenshots, pointer is hovering on Learning Guide.

Before:
screen shot 2015-05-20 at 2 16 17 pm
screen shot 2015-05-21 at 3 14 51 pm

I think this should also correct the orange background from showing on the testing site, but's hard to be sure since we can't replicate locally.

### DIFF
--- a/resources/styles/global/navbar.less
+++ b/resources/styles/global/navbar.less
@@ -9,20 +9,6 @@
   &.navbar.navbar{
     border-bottom: 1px solid @border-color;
     background-color: @tutor-white;
-    .dropdown-menu {
-      a:hover {
-        color: @tutor-neutral-darker;
-        background-color: @tutor-neutral-lighter;
-      }
-      .active a {
-        color: @tutor-neutral-darker;
-        background-color: @tutor-neutral-light;
-        &:hover {
-          background-color: @tutor-neutral-light;
-          cursor: default;
-        }
-      }
-    }
   }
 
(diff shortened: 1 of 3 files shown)


resources/styles/global/navbar.less
```diff
@@ -9,6 +9,20 @@
   &.navbar.navbar{
     border-bottom: 1px solid @border-color;
     background-color: @tutor-white;
+    .dropdown-menu {
+      a:hover {
+        color: @tutor-neutral-darker;
```
philschatz: Could we just change the active color var in `bootstrap-material-design`?

resources/styles/global/tutor-material.less
```diff
@@ -54,17 +54,17 @@ legend {
...
 .dropdown-menu {
-  .variations(~" > .active > a", background-color, @primary);
-  .variations(~" > .active > a:hover", background-color, @primary);
-  .variations(~" > .active > a:focus", background-color, @primary);
+  .variations(~" > .active > a", background-color, @tutor-neutral-light);
+  .variations(~" > .active > a:hover", background-color, @tutor-neutral-darker);
+  .variations(~" > .active > a:focus", background-color, @tutor-neutral-lighter);
 }
 
 // Alerts
```
nathanstitt: It sure seems like we should be able to do this, but it doesn't work, the drop-down menu still appears orange.

I'm continuing to hunt through the css to attempt to find the culprit.

PR #287, opened 2015-05-20
Read CSRF token from meta tag, send with requests

nathanstitt:
The simplest thing that could possibly work.  I'm also a fan of making a `Tutor.boot()` method that could initialize this.

screen shot 2015-05-20 at 3 24 40 pm

### DIFF
--- a/src/api.coffee
+++ b/src/api.coffee
@@ -7,7 +7,6 @@
 
 $ = require 'jquery'
 _ = require 'underscore'
-CSRF_Token = document.head.querySelector('meta[name=csrf-token]')?.getAttribute("content")
 {TimeActions} = require './flux/time'
 {CurrentUserActions, CurrentUserStore} = require './flux/current-user'
 {CourseActions} = require './flux/course'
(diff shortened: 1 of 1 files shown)


nathanstitt:
Corresponds with https://github.com/openstax/tutor-server/pull/300

philschatz:
And needs an entry in #285

src/api.coffee
```diff
@@ -7,7 +7,7 @@
 
 $ = require 'jquery'
 _ = require 'underscore'
-
+CSRF_Token = document.head.querySelector('meta[name=csrf-token]')?.getAttribute("content")
```
philschatz: Should an exception be thrown if this element is not found?

nathanstitt: Maybe, but if we do that then we'll also have to modify the dev server to set a dummy meta tag.  My thoughts were that the validation of the tag's presence and accuracy would be performed by the server when requests are performed.

I'd suggest that we split the difference and `console.warn` a "CRSF Token not detected" message.

PR #290, opened 2015-05-20
Fix for chapter sections on stats

pandafulmanda:
screen shot 2015-05-20 at 6 40 56 pm

@nathanstitt @darrenmason i changed something in learning guide and I checked that it's still working, but please confirm

uses openstax/tutor-server#301

### DIFF
--- a/api/plans/1/stats.json
+++ b/api/plans/1/stats.json
@@ -10,14 +10,14 @@
             "partially_complete_count": 0,
             "current_pages": [{
                 "id": "1",
-                "number": 1,
+                "chapter_section": [1, 2],
                 "title": "Introduction",
                 "student_count": 0,
                 "correct_count": 0,
                 "incorrect_count": 0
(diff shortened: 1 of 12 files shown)


philschatz:
:+1: 

Should non-React components (and mixins) live outside of `src/components`?

pandafulmanda:
oh hhhhmmm...i dunno, i dont think i have an opinion either way.  if so, separate PR?  sounds like a broader discussion...what about mixins that are currently group inside a folder in components?

philschatz:
The question was more for @darrenmason and @nathanstitt since the learning guide is the only non-Component so far; but yes, moving it would be a separate PR since the styling would need to move as well.

darrenmason:
I don't see it as an issue, however not sure how we are defining 'non-react'

darrenmason:
makes sense to pull out sectionFormat function

PR #293, opened 2015-05-21
makes it more obvious when student is transitioning from problems to review page on past due homeworks

pandafulmanda:
### Before
screen shot 2015-05-20 at 7 55 19 pm

### After
screen shot 2015-05-20 at 7 46 53 pm

styling is not the best, but i think the workflow is improved

### DIFF
--- a/src/components/task-step/ends.cjsx
+++ b/src/components/task-step/ends.cjsx
@@ -69,6 +69,15 @@ HomeworkEnd = React.createClass
   onNextStep: ->
     @setState({})
 
+  renderCompletionMessage: (completeSteps, totalSteps) ->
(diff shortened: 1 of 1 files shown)


src/components/task-step/ends.cjsx
```diff
@@ -69,6 +69,15 @@ HomeworkEnd = React.createClass
   onNextStep: ->
     @setState({})
 
+  renderCompletionMessage: (completeSteps, totalSteps) ->
+    congratsMessage = <h1>You are done! Great job!</h1> if completeSteps is totalSteps
+
+    [
```
philschatz: Any reason this is not a `<div>`? That way it is possible to style easier later.

PR #298, opened 2015-05-21
Guide ie10 and ie11 fixes

darrenmason:
IE11 JS errors specific to guide are resolved.  

WIP: IE11 is not respecting width:100% and height:100% on container, still looking for an adequate workaround.

### DIFF
--- a/resources/styles/components/learning-guide.less
+++ b/resources/styles/components/learning-guide.less
@@ -2,13 +2,22 @@
 @guide-height: 100%;
 @guide-min-width: 1000px;
 @guide-max-width: 1600px;
+@guide-min-height: 306px;
+@guide-max-height: 490px;
 // colors
 @guide-green: rgb(121, 177, 66);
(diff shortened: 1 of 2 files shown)


philschatz:
Could you use something like http://getbootstrap.com/components/#responsive-embed on the `<svg>` instead of an `<iframe>`?

darrenmason:
I used breakpoints that match the aspect ratio.  Resize now works in IE11.  This PR addresses IE10 as well.

resources/styles/components/learning-guide.less
```diff
@@ -245,3 +249,151 @@
...
   }
 }
+
+
+
+// IE10/11 resize hack
+// these breakpoints correspond approx. with what height:auto calculates in other browsers.
+// only targets IE10 and above.
+
+@media screen and (max-width: 1079px) {
```
philschatz: Can you combine these into a single mixin? That way the breakpoints are each one line

philschatz:
These changes were merged to master via #313

PR #299, opened 2015-05-22
Tutor Textarea auto expand

jivewise:
Textarea autoexpands.  Still flickers sometimes, but it seems fairly functional for now.

textarea-expand

### DIFF
--- a/src/components/tutor-input.cjsx
+++ b/src/components/tutor-input.cjsx
@@ -47,11 +47,11 @@ TutorTextArea = React.createClass
     onChange: React.PropTypes.func
     value: React.PropTypes.any
 
-  getInitialState: ->
-    {height: '2em'}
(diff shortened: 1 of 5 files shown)


src/components/tutor-input.cjsx
```diff
@@ -47,7 +47,11 @@ TutorTextArea = React.createClass
     onChange: React.PropTypes.func
     value: React.PropTypes.any
 
-  onChange: ->
+  resize: (event) ->
+    event.target.style.height = ''
```
philschatz: Why does the height need to be cleared first?

jivewise: That's so we can get the scroll height.  Because of the material css, we need to clear the height first.  This should probably be 0px to be safe.

PR #300, opened 2015-05-22
Calendar auto expand

jivewise:
- [x] Update date-picker to auto-expand when focused
- [x] Fix for safari browsers when coming from calendar
- [x] Updated plans to use a read-only date-picker instead of a label

### DIFF
--- a/src/components/task-plan/homework.cjsx
+++ b/src/components/task-plan/homework.cjsx
@@ -95,19 +95,15 @@ HomeworkPlan = React.createClass
     if @state?.showSectionTopics then formClasses.push('hide')
 
     if (TaskPlanStore.isPublished(id))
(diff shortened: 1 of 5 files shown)


gregfitch:
:thumbsup: Thanks for the quick turnaround, Jing.

philschatz:
Clicking a second time on the calendar (when it already has focus) does not reopen the calendar. Could the `<input>` be disabled so it never actually gets focus? Also, what happens when the user enters an invalid date?

jivewise:
Now clicking on the input should open up the date picker again.  I also added some initial handling for when an invalid date is entered in the input.

philschatz:
D'oh. It seems this causes a 422 when publishing (see #324) . Will try to merge after walking through the demo script

PR #302, opened 2015-05-22
Style teacher calendar like the student dashboard

nathanstitt:
With cursor hovered over `Add  iReading`:

screen shot 2015-05-22 at 3 58 50 pm

### DIFF
--- a/resources/styles/global/tutor-booksplash-background.less
+++ b/resources/styles/global/tutor-booksplash-background.less
@@ -1,0 +1,39 @@
+// Applies a fixed background effect to a container that includes a large fixed header
+//
+// It's intended to be customized for different looks depending on the type of content
+// the user is styling, i.e. Physics, Biology, etc
+.tutor-booksplash-background {
+
+  &:before {
+    content: attr(data-title);
+    position: fixed;
+    left: 0px;
+    right: 0px;
+    top:60px;
+    height:220px;
+    overflow: hidden;
+    background-color: @tutor-tertiary;
+    color: @tutor-secondary;
+    text-align: center;
+    padding-top: 85px;
+    font-size: 300px;
+    font-family: 'HelveticaNeue-Bold', 'Helvetica Neue Bold', 'Helvetica Neue';
+    font-weight: 700;
+  }
+  
+  // The background is applied as a :after block so the opacity can
+  // be applied to only the background image, not the content itself
+  &:after {
+    content: "";
+    position: fixed;
+    z-index: -1;
+    top: 10px;
+    bottom: 0;
+    left: 0;
+    right: 0;
+    opacity: 0.05;
+  }
+  
+  // custom backgrounds images for different types
+  &[data-title=Physics]:after{ .tutor-background-image("physics.jpg"); }
+}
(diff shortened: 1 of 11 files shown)


pandafulmanda:
:+1:

philschatz:
Can you remove the "Task Plans for Course 1" panel header too? that'll fix another pivotal ticket (which I think is assigned to you too)

nathanstitt:
Yeah the mockup doesn't have that so I've removed it.  New screenshot coming up shortly

nathanstitt:
@pandafulmanda Your `CourseAddMenuMixin` came through big time!  

It was ridiculous easy to add the "Add Assignment" button

resources/styles/global/tutor-booksplash-background.less
```diff
@@ -0,0 +1,44 @@
...
+    position: fixed;
+    left: 0px;
+    right: 0px;
+    overflow: hidden;
+    background-color: @tutor-tertiary;
+    color: @tutor-secondary;
+    text-align: center;
+    font-size: 300px;
+    line-height: 200px;
+    font-family: 'HelveticaNeue-Bold', 'Helvetica Neue Bold', 'Helvetica Neue';
```
philschatz: Should this be in `typography.less`?

resources/styles/global/tutor-booksplash-background.less
```diff
@@ -0,0 +1,44 @@
...
+    z-index: -2;
+    top: 200px;
+    bottom: 0;
+    left: 0;
+    right: 0;
+    opacity: 0.05;
+  }
+  
+  // custom backgrounds images for different types
+  &[data-title=Physics]:after{ .tutor-background-image("physics.jpg"); }
```
philschatz: Should this be based on the course type rather than the title?

resources/styles/global/tutor-booksplash-background.less
nathanstitt: I'm not sure.  This is the only place that I see Helvetica being used. `grep -r Helvetica` only returns this at least.  Is still worthwhile to have it available globally do you think?

resources/styles/global/tutor-booksplash-background.less
nathanstitt: Yeah, I think that'd be better. Adding a new attribute for that now

resources/styles/global/tutor-booksplash-background.less
philschatz: Since Helvetica is not included as a font I think it should either:

1. be included in `resources/fonts` (assuming we have rights to use it) and include it in `typography.less`
2. include a different font

nathanstitt: Yes, I believe it's a proprietary font that's included on OS X only.  I suppose we could license it but is probably better to find an alternative.

@Fredasaurus any thoughts on substituting a different font for HelveticaNeue here?  I'd used it because here because it was in the original mockup.  I'm not sure that was intended or perhaps it was just the default for your mockup tool.   I'd suggest using Lato Bold or we could bundle a different Google font of your choosing.

PR #314, opened 2015-05-27
Use html for interactives

nathanstitt:
screen shot 2015-05-27 at 12 38 10 pm


resources/styles/components/task-step/interactive.less
```diff
@@ -0,0 +1,13 @@
+@import '../../book-content/index';
+
+.task-step {
+
+  .interactive-step {
+    // pull in common stylings for book contents
+    .tutor-book-content();
+    iframe {
```
nathanstitt: Rather than mixin this twice without any changes, wouldn't it be better to just do a:
```css
.reading-step, .interactive-step {
.tutor-book-content();
}

```

nathanstitt:
I'm investigating why the scrollbar is appearing, it's a padding issue or something in the content, the height provided is acceptable.

nathanstitt:
The remaining scrollbar comes from a css rule in the iframe content.

There's a `<div class="app-view">` element that is nested under `body`.  It has a  `margin: 10px auto 0;` rule applied to it.  If the `10px` is removed the scrollbar no longer appears.

philschatz:
Is this one of Patrick's sims or one of PhET's?

nathanstitt:
I'm not sure?  And now I've lost that sim when I re-ran setup:001 to get fresh content, so can't verify sorry.

nathanstitt:
Per discussion, iframe height is now set to 560px and any scrolling will need to be addressed by content team.

Dantemss:
I believe the sim you guys were talking about was "The Moving Man" by Patrick.

PR #315, opened 2015-05-27
delay loading MathJax until configured

philschatz:
No UI Changes (except that MathJax should load and work).

Paired with https://github.com/openstax/tutor-server/pull/318

Source: http://docs.mathjax.org/en/latest/configuration.html?highlight=authorinit#configuring-mathjax-after-it-is-loaded

### DIFF
--- a/src/components/html.cjsx
+++ b/src/components/html.cjsx
@@ -48,7 +48,8 @@ module.exports = React.createClass
     hasMath = root.querySelector('math')
 
     # Return immediatly if no [data-math] or <math> elements are present
-    return unless window.MathJax and (_.any(nodes) or hasMath)
+    # TODO: If the MathJax Queue is not available then MathJax has not loaded yet. Add a load callback to enqueue.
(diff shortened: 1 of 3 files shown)


src/components/html.cjsx
```diff
@@ -48,7 +48,8 @@ module.exports = React.createClass
     hasMath = root.querySelector('math')
 
     # Return immediatly if no [data-math] or <math> elements are present
-    return unless window.MathJax and (_.any(nodes) or hasMath)
+    # TODO: If the MathJax Queue is not available then MathJax has not loaded yet. Add a load callback to enqueue.
```
philschatz: This may be solved by something like:

```coffee
unless window.MathJax.Hub?.Queue
  # TODO: Make sure this callback is not added more than once
  MathJax_AuthorInit = window.MathJax.AuthorInit
  window.MathJax.AuthorInit = => 
    MathJax_AuthorInit()
    if @isMounted()
      @setState({})
```

PR #319, opened 2015-05-27
striped incorrect for further differentiation for color-blindness

pandafulmanda:
screen shot 2015-05-27 at 3 32 59 pm

Need to check with @Fredasaurus

### DIFF
--- a/src/components/task-plan/reading-stats.cjsx
+++ b/src/components/task-plan/reading-stats.cjsx
@@ -36,20 +36,15 @@ Stats = React.createClass
     if total_count then @_percent(data[count], total_count) else 0
 
   renderPercentBar: (data, type, percent, correctOrIncorrect) ->
-    bsStyles =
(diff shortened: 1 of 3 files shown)


Fredasaurus:
Cool, can we make them less wide and run bottom left to top right?

pandafulmanda:
updated progress bar style as suggested

src/components/task-plan/reading-stats.cjsx
```diff
@@ -42,6 +42,7 @@ Stats = React.createClass
 
     classes = 'reading-progress-bar'
     classes += ' no-progress' unless percent
+    classes += ' progress-bar-striped' if correctOrIncorrect is 'incorrect'
```
philschatz: Could/should this be `progress-bar-incorrect` to be more semantic (since the class adds the red gradient)

PR #321, opened 2015-05-27
Bug/refresh location

pandafulmanda:
takes student the first step of a related reading when they click "Refresh my Memory" and gets a recovery step for the step as well
screen shot 2015-05-28 at 7 14 54 pm
screen shot 2015-05-28 at 7 15 01 pm

go to the recovery step when continuing from refresh
screen shot 2015-05-28 at 7 15 08 pm


philschatz:
@Fredasaurus thoughts?

pandafulmanda:
@philschatz I spoke with @Fredasaurus  and Jason in person today, and I think we settled that "refresh memory" will:
* first take student back to related reading
* on continue take student onto newly generated question

PR #322, opened 2015-05-28
guide - new header ui

darrenmason:
image


resources/styles/components/learning-guide.less
```diff
@@ -184,6 +188,51 @@
...
+        margin: 0 auto;
+        display:table;
+        .span-wrap {
+          display:table-cell;
+          vertical-align: middle;
+          color: @guide-green;
+          .path-label {
+            color: @tutor-neutral-darker;
+            &:after {
+              content: "\00a0\00a0";
```
philschatz: Can you include a comment explaining what these characters are?

resources/styles/components/learning-guide.less
```diff
@@ -256,144 +295,45 @@
...
-    height: 456px !important;
-  }
-}
-@media screen and (min-width: 1600px) {
-  _:-ms-lang(x), .learning-guide { 
-    height: 462px !important;
-  }
-}
+.ie-resize-fix(306px, 0px);
+.ie-resize-fix(306px, 1079px);
```
philschatz: @darrenmason Would a for-loop work here?

@Fredasaurus Do these breakpoints need to be so finegrained?

Fredasaurus:
Those breakpoints seem strange. I don't know a lot about svg scaling. Is there a reason we can't set width to 100% and height to auto?

darrenmason:
those breakpoints are for internet explorer 11 only.  It is a workaround to IE not scaling the container properly.  I tried responsive-embed but it makes the container absolute and breaks the layout.  Not sure what else to try, if you test it in IE11 you'll see it works pretty well.

philschatz:
When you tried the `responsive-embed` was it just on the chart or the entire guide (including title and footer)?

darrenmason:
the title/footer arent meant to be scaled.  The fix numbers aren't random, they reproduce approx height auto.  I can put it in a for loop?

philschatz:
a cursory test (by pasting the SVG wrapped in a <div class="responsive-item"><p>Foo</p>... into http://getbootstrap.com/components/#responsive-embed) seems to scale the svg and not scale the "Foo" text:

image

image

image

philschatz:
Also, @Fredasaurus did the "Back to dashboard" link go away?

Fredasaurus:
Yeah, let's put it back. Secondary button style top right.

darrenmason:
ok

PR #325, opened 2015-05-28
Datepicker Auto Expand

jivewise:
Another PR for date picker auto-expand.

### DIFF
--- a/src/components/task-plan/homework.cjsx
+++ b/src/components/task-plan/homework.cjsx
@@ -95,19 +95,15 @@ HomeworkPlan = React.createClass
     if @state?.showSectionTopics then formClasses.push('hide')
 
     if (TaskPlanStore.isPublished(id))
-      dueAtElem = <span>Due Date: {new moment(dueAt).format('MMM Do, YYYY')}</span>
-    else
-      dueAtElem = <TutorDateInput
-                    id='homework-due-date'
-                    format='MMM dd, yyyy'
-                    time={false}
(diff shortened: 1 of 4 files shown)


philschatz:
Clicking on "Today" in the calendar does not autofill in the date when creating an iReading or Homework. Other than that it seems to work!

jivewise:
I thought we weren't going to let teachers create homeworks that with a due date of today.  It seems like the minimum for that should be at least tomorrow.

philschatz:
@Fredasaurus thoughts on dates? and if today is not selectable, @pandafulmanda what would be needed to change the calendar?

Fredasaurus:
Same day is aok.

philschatz:
Maybe @jpslav has thoughts (since the assignments will be cron jobs)?

jpslav:
Assignments aren't going to be cron jobs per se, but they will be running in the background (immediately after someone hits "publish").  Same day assignments are potentially not good because the current default due time is 7am (chosen to avoid ambiguity about what "midnight" means).  So if someone hit the publish button at 8 am with a due date of today, the assignment would be past due immediately.  Even when we had a default due time of midnight, kef's assumption was that that meant the first second of a particular day (not 11:59:59pm), in which case there's no allowance for publishing something that is due today.

Kef/LAT have requested explicit due times post alpha, fwiw.

Sorry to be behind on this one.

/cc @kevinburleigh75

PR #327, opened 2015-05-29
student dashboard no longer caches tasks

philschatz:
Going from the dashboard to answering some questions and then back to the dashboard did not update the list of events. It does now


nathanstitt:
:+1:

PR #328, opened 2015-05-29
fix some invariant errors caused by MathJax

philschatz:
Seems to fix:

- dashboard -> learning guide
- initial load of the calendar

but @nathanstitt's code for cleaning up MathJax elements might be cleaner and good to include anyway

### DIFF
--- a/index.coffee
+++ b/index.coffee
@@ -42,4 +42,10 @@ api = require './src/api'
 api.start()
 router = require './src/router'
 
-router.start(document.body)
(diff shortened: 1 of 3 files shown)


nathanstitt:
Aha!  Yeah I bet that'd work great.  If we can, I think it'd be a good idea to remove the cleaning code.  That way Mathjax will have a chance of re-using it's elements on every pass and won't have to re-create them.

pandafulmanda:
:+1:

PR #329, opened 2015-05-29
Style the title for worked-example solutions blocks

nathanstitt:
WIP until @Fredasaurus lets us know what they should look like.  Right now I'm styling them the same as a sub-title element.

screen shot 2015-05-29 at 1 11 42 pm

### DIFF
--- a/resources/styles/book-content/worked-examples.less
+++ b/resources/styles/book-content/worked-examples.less
@@ -1,4 +1,6 @@
-.worked-example.note:first-child {
+// sometimes it's plural, sometimes not?
+.worked-example.note:first-child,
+.worked-examples.note:first-child {
 
   .book-content-step-size();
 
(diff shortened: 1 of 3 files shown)


resources/styles/book-content/worked-examples.less
```diff
@@ -9,5 +11,14 @@
...
     .tutor-reading-main-title();
   }
-  
+
+  .exercise {
+    .solution {
+      &:before {
+        content: "Solution";
+        display: block;
+       .book-content-subtitle();
```
philschatz: whitespace!

resources/styles/book-content/worked-examples.less
```diff
@@ -1,4 +1,6 @@
-.worked-example.note:first-child {
+// sometimes it's plural, sometimes not?
```
philschatz: Yep, that's correct. If there is > 1 example inside then it's plural.

nathanstitt: Ok, that makes sense.  I meant to comment this and point you toward it to ask about and forgot to, glad you noticed.

PR #334, opened 2015-05-29
Set IReading images to max-width 100%

nathanstitt:
Image without `tutor-ui-horizontal-img` class:
screen shot 2015-05-29 at 3 14 52 pm

Image with:
screen shot 2015-06-01 at 11 52 20 am

### DIFF
--- a/resources/styles/book-content/base.less
+++ b/resources/styles/book-content/base.less
@@ -117,26 +117,28 @@ div[data-type="document-title"] ~ p {padding: 0 @tutor-card-body-padding-horizon
 }
 
 figure {
-  width: 100%;
+  max-width: 100%;
   display: table;
   text-align: left;
   margin-right: 30px;
   padding: 0;
-
+  float: left;
   &.tutor-ui-vertical-img {
     @media (min-width: @screen-sm-min) {
       width: 50%;
       float: left;
     }
   }
-
   img {
-    width: 100%;
+    max-width: 100%;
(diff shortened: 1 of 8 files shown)


Fredasaurus:
the ideas was that we wanted to get the aspect ratio of the image. images with a vertical or even ratio, would align left, and text would flow around (fig 1 above). For images that have a horizontal ratio, we would scale to width:100% .

Dak was working on this but I'm not sure what the level of completeness is.

nathanstitt:
Oh, that makes sense.  I do see there's a `tutor-ui-horizontal-img` class on the span around some of the images like fig 2 above.  I'm guessing those should be set 100% then?

Fredasaurus:
Yeah, we may need to play with the threshold a bit. To make sure that images that maybe are just a little bit wider than they are tall don't get scaled 100% and become to tall for the view port.

Also, images that float left should have a max-width, so we don't end up with a skinny column of text on the right.

nathanstitt:
Updated so that text does not wrap around images that are marked `tutor-ui-horizontal-img`.  Their maximum size is set to the width of the panel, which is currently 720px.

nathanstitt:
@Fredasaurus You mentioned that images that are marked as horizontal should be shown at 100% width.  Do you mean that they should be scaled up to that if smaller, or that 100% should be the max-width?

Fredasaurus:
We can scale them up. If the quality looks bad, we can try something else

PR #336, opened 2015-05-29
Exercise loading indication

nathanstitt:
Disable continue button and show a spinner with "Loading ..." message while store is loading.

screen shot 2015-05-29 at 4 12 09 pm

### DIFF
--- a/resources/styles/global/forms.less
+++ b/resources/styles/global/forms.less
@@ -6,3 +6,5 @@ textarea {
     box-shadow: 0 0 10px fade(@tutor-tertiary-light, 25%);
   }
 }
+
+button i.fa { margin-right: 0.5rem; }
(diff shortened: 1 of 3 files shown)


philschatz:
ooh, cool. Will test this extensively @pandafulmanda I think this was your request, no?

nathanstitt:
yeah, it's hard to test locally since the requests complete so quickly.  A few console.log and breakpoints verify that it is shown very briefly.

I am a bit afraid of what it's error state might be if the request times out.  I attempted to test that locally by shutting down the server, but that just caused the "Load Error" message from loadable to display.  I think that's what will happen on a real timeout as well.

pandafulmanda:
YAY

philschatz:
Hmm, I was able to get the "Loading..." spinner to hang indefinitely by killing the server between the `PUT ./completed` and the `GET /steps/##` but I think it's no worse than it was before!

PR #341, opened 2015-06-01
Move .splash specifier below others for precedence

nathanstitt:
Fixes bug where .splash element wasn't resizing it's children to be 100% width

screen shot 2015-06-01 at 1 29 11 pm

screen shot 2015-06-01 at 1 29 28 pm

### DIFF
--- a/resources/styles/book-content/base.less
+++ b/resources/styles/book-content/base.less
@@ -145,7 +145,7 @@ figure {
     // This means child elements that normally have a fixed pixel width will not expand to fit
     // unless their widths are set to be 100%
     [data-type=media]{ width: 100%; }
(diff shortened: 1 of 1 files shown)


resources/styles/book-content/base.less
```diff
@@ -142,6 +138,12 @@ figure {
       width: 100%;
     }
   }
+  // splash images inside are always full width
```
philschatz: Can you add a comment on the rule that is overridden in this block (and why)?

nathanstitt: Does that make sense? I struggled with how much detail to go into there while still making the point somewhat concisely.

PR #344, opened 2015-06-01
Bug/fix task descriptions not showing

pandafulmanda:
screen shot 2015-06-01 at 3 40 20 pm


philschatz:
ah yes, the switch from `description_html` to `description`.

PR #347, opened 2015-06-01
Loading error msg

darrenmason:
image

This adds a refresh button on error and shows during Loading... only after 30 seconds.


resources/styles/global/loadable.less
```diff
@@ -0,0 +1,29 @@
+@-webkit-keyframes fadeIn { from { opacity: 0; } to { opacity: 1; } }
+@-moz-keyframes fadeIn { from { opacity: 0; } to { opacity: 1; } }
+@keyframes fadeIn { from { opacity: 0; } to { opacity: 1; } }
+
+.refresh-button-delayed {
```
philschatz: `<Loadable>` generates elements with classes (`-loading` and `-error`). Would it make sense to change these classes to `.loadable.loading`? Then, is there a need to explicitly have a `refresh-button-delayed` class?

src/components/loadable.cjsx
```diff
@@ -21,14 +23,17 @@ module.exports = React.createClass
...
 
+  getInitialState: ->
+    refreshButton: false
+
   getDefaultProps: ->
 
     # Enables a renderStatus prop function with a component other than a div
-    renderLoading: ->
-      <div className='-loading'>Loading...</div>
+    renderLoading: (refreshButton) ->
```
philschatz: How would this work with Components with a custom `renderLoading` method (I think the calendar may have this)?

It seems like to make sure a refresh button is always available either:

1. all `render*` method must always take a `refreshButton` argument (not a func)
2. render will always append a `refreshButton` for loading and error states

pandafulmanda:
:+1:

src/components/loadable.cjsx
pandafulmanda: I think if it always prepends, that would work fine and I can style the refresh button on the calendar loading screen as needed for now.

It would be nice if the button remained an argument that is a component for the custom render functions to use within the screens, but we can update that later.

pandafulmanda: From talking with @philschatz, when the loading times out (for example after the 30 seconds delay being used for the delayed button), should it just renderError()? If that's the case, then the renderLoading would not have to consider the button at all.

darrenmason: I found using _.delay to not work consistently within the render statuses, and using the css delay worked smoothly.  That way there is no timeout that could break render, it just shows a button.

resources/styles/global/loadable.less
darrenmason: I did that so one button could have either delay or no delay via css.  I could consolidate the classnames better.

src/components/loadable.cjsx
```diff
@@ -49,5 +50,27 @@ module.exports = React.createClass
...
+
+
+RefreshButton = React.createClass
+  displayName: 'RefreshButton'
+
+  reloadPage: ->
+    location.reload()
+
+  render: ->
+    if React?
```
philschatz: When would React not be available?

src/components/loadable.cjsx
```diff
@@ -49,5 +50,27 @@ module.exports = React.createClass
...
+      <div className="refresh-button">
+        <BS.Button onClick={@reloadPage}>
```
philschatz: I think this needs to be an `<a href={window.location}>` in the case of React no longer re-rendering (due to Invariant errors or something)

resources/styles/global/loadable.less
```diff
@@ -0,0 +1,29 @@
+@-webkit-keyframes fadeIn { from { opacity: 0; } to { opacity: 1; } }
```
philschatz: This name should probably be local to the file (in case something else declares `fadeIn` to be something different)

src/components/loadable.cjsx
darrenmason: I wondered that myself.  Was going off of pivotal checklist:

"check if reload() on click would work (and backup browser link in case React has crashed)"

PR #348, opened 2015-06-02
Fix the invariant exceptions when the dash redirects

nathanstitt:
For discussion in hopes that someone can find an better workaround.

This sucks because it doesn't display a "loading" message.  To work around this to some extent perhaps we should call `CourseListingStore.ensureLoaded()` as soon as possible in api.coffee?


src/components/course-listing.cjsx
```diff
@@ -0,0 +1,78 @@
...
+    # Called before the Router mounts and renders the component
+    # Uses the callback to delay rendering until the CourseListingStore is loaded
+    # and then calls DisplayOrRedirect above to perhaps redirect to a different component
+    willTransitionTo: (transition, params, query, callback) ->
+      unless CourseListingStore.isLoaded()
+        CourseListingActions.load() unless CourseListingStore.isLoading()
+        onLoadFn = ->
+          CourseListingStore.off('load', onLoadFn)
+          DisplayOrRedirect(transition, callback)
+        CourseListingStore.on('loaded', onLoadFn)
```
philschatz: Could you just use `CourseListingStore.once`?

nathanstitt: I'd forgotten about `.once`.  And should be `loaded`, so correct on both counts

src/components/course-listing.cjsx
```diff
@@ -0,0 +1,76 @@
...
+BS = require 'react-bootstrap'
+Router = require 'react-router'
+
+{CourseListingActions, CourseListingStore} = require '../flux/course-listing'
+
+# Called once the store is loaded
+# checks the course and roles and will redirect if there is only a single course and role
+DisplayOrRedirect = (transition, callback) ->
+  courses = CourseListingStore.allCourses() or []
+  if courses.length is 1 and courses[0].roles?.length is 1
```
philschatz: Minor code cleanup suggestion:

```coffee
[course] = courses
if courses.length is 1 and course.roles?.length is 1
    roleType = course.roles[0].type
    type = switch roleType
      when 'student' then 'viewStudentDashboard'
      when 'teacher' then 'taskplans'
      else
        throw new Error("BUG: Unrecognized role type #{roleType}")
    transition.redirect(type, {courseId: course.id})
  callback()
```

src/components/course-listing.cjsx
```diff
@@ -0,0 +1,76 @@
...
+    router: React.PropTypes.func
+
+  statics:
+    # Called before the Router mounts and renders the component
+    # Uses the callback to delay rendering until the CourseListingStore is loaded
+    # and then calls DisplayOrRedirect above to perhaps redirect to a different component
+    willTransitionTo: (transition, params, query, callback) ->
+      unless CourseListingStore.isLoaded()
+        CourseListingActions.load() unless CourseListingStore.isLoading()
+        CourseListingStore.once 'loaded', ->
```
philschatz: You could do `CourseListingStore.once 'failed', -> callback` and in `render: ->` display the error message `unless CourseListingStore.isLoaded()`. Agreed there's no loading message though

src/flux/course-listing.coffee
```diff
@@ -0,0 +1,59 @@
...
+{CourseActions, CourseStore} = require './course'
+
+LOADING = 'loading'
+LOADED  = 'loaded'
+FAILED  = 'failed'
+
+CourseListingActions = flux.createActions [
+  'load'
+  'loaded'
+  'reset'
```
philschatz: Missing `FAILED` action

src/components/course-listing.cjsx
nathanstitt: Hm, I don't think that'll work.  `courses` starts out as an array that may be empty.  Your `[courses]` statement converts it to a nested array but doesn't check if it's empty.

src/flux/course-listing.coffee
nathanstitt: Thanks!  Added, but the lines didn't match up enough to hide your comment.

src/components/course-listing.cjsx
philschatz: Are you sure? `[][0]` returns `undefined` and the 1st condition of the `if` statement is a guard that `course` is not null.

PR #351, opened 2015-06-02
Ireading style polish

Fredasaurus:
blue background for chapter title, drop cap
screen shot 2015-06-01 at 9 56 33 pm
subchapter title styles
screen shot 2015-06-01 at 10 21 38 pm
virtual physics title
screen shot 2015-06-01 at 10 21 47 pm
tips for success styling
screen shot 2015-06-01 at 10 21 59 pm
removing font-weight:600 from labels
screen shot 2015-06-01 at 10 35 26 pm
form-note
screen shot 2015-06-01 at 10 36 06 pm
aligning "Physics" with logo
screen shot 2015-06-01 at 10 35 15 pm


resources/styles/book-content/tips-for-success.less
```diff
@@ -16,25 +18,43 @@
...
-    padding-top: 10px;
-    border-top: 1px solid @tutor-neutral-light;
+  p {
+    #fonts > .sans(1.7rem, 1.5em);
+    color: @tutor-neutral-dark;
+    padding: 15px 20px;
+    border-left: 1px solid @tutor-neutral-lighter;
+    border-right: 1px solid @tutor-neutral-lighter;
+    background: @tutor-neutral-lightest;
+    &:first-child {
```
philschatz: Why place the borders on the children rather than the `p`?

resources/styles/book-content/tips-for-success.less
```diff
@@ -16,25 +18,43 @@
...
+    #fonts > .sans(1.7rem, 1.5em);
+    color: @tutor-neutral-dark;
+    padding: 10px;
+    border-left: 1px solid @tutor-neutral-lighter;
+    border-right: 1px solid @tutor-neutral-lighter;
     list-style: none;
-    
+    background: @tutor-neutral-lightest;
+    &:first-child {
+    border-top: 1px solid @tutor-neutral-light;
```
philschatz: Can you fix the indentation of this? Also, if this pattern is common would it warrant a mixin that takes the color as an arg?

resources/styles/components/task-step/all-steps.less
```diff
@@ -9,7 +9,7 @@
   .spacer-step {
     padding: @tutor-card-body-padding;
   }
-  .interactive-step, .reading-step {
+  .interactive-step > .interactive-content, .reading-step {
```
philschatz: Can you place each selector on separate lines? Otherwise, it's easy to miss the comma.

resources/styles/book-content/tips-for-success.less
Fredasaurus: That way the first <p> has a top border and the last one has a bottom border. I would prefer to place one border on the containing element, but that wasn't possible.

philschatz: Oh, I see. That way the yellow "Tips for success" piece does not span the entire width? If that is the case should the styling just be `:first-child` rather than `p:first-child` (or "What happens when the first child is not a `<p>`?

Fredasaurus: JUst realized that's not going to work anyway, so changed this up.

PR #353, opened 2015-06-02
force shy nav on task step through

pandafulmanda:
scroll is at top on first load
screen shot 2015-06-02 at 5 29 07 am

scroll is at buffer point, making nav shy, on changing step
screen shot 2015-06-02 at 5 29 12 am

### DIFF
--- a/resources/styles/components/pinned-header-footer-card/index.less
+++ b/resources/styles/components/pinned-header-footer-card/index.less
@@ -6,3 +6,8 @@ body.pinned-shy {
     .translate(0, -@pinned-top-heading-buffer);
   }
 }
+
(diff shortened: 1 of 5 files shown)


philschatz:
Actually, this makes the student dashboard always scrollable.

nathanstitt:
@philschatz I just pulled master and looks like the added css classes `task-view pinned-force-shy pinned-on` on the `<body>` element are not being cleared after the steps are navigated away from.   It's also affecting the learning guide for the same reason.  If I remove the classes from `<body>` in dev console, all is well.

We might be able to use `willTransitionFrom` on the breadcrumbs to remove the classes on navigation.  Investigating that now.

PR #354, opened 2015-06-02
reset the body's classlist on unmount

nathanstitt:
Fixes issue with Dashboard and learning guide always scrolling after visiting the a task

Identified during discussion on #353 

The original code was only remembering a single class to remove, but added several other classes.  This just remembers whatever the `body` had on mount and then restores it on unmount.


philschatz:
Thanks! Maybe we should take a look again at moving styling into the React root element

pandafulmanda:
ah!  sorry, that was silly of me

PR #356, opened 2015-06-02
Add temp support for calendar icon using FA

dak:
image


philschatz:
Was the calendar icon in the date picker fixed?

dak:
@philschatz Everything I can tell suggests that this is an issue with how the `RwWidgets` font is being served.  I've tried to duplicate the issue with various different nginx settings that match the same headers on tutor-demo but haven't been able to.

The font DOES show correctly locally.

Is it possible to get the nginx config being used in production?

PR #364, opened 2015-06-02
Set fixed height of navbar

nathanstitt:
Override's a media query that bootstrap material design uses to set the height to 50px on `(max-width: 1199px)`

Before (screen width 900px):
screen shot 2015-06-02 at 5 26 36 pm

After:
screen shot 2015-06-02 at 5 26 55 pm

### DIFF
--- a/resources/styles/global/navbar.less
+++ b/resources/styles/global/navbar.less
@@ -2,7 +2,13 @@
 @tutor-navbar-padding-horizontal: 4rem;
 
 .navbar-fixed-top {
-
+  
+  // override bootstrap material design media query for (max-width: 1199px)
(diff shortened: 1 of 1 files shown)


resources/styles/global/navbar.less
```diff
@@ -2,7 +2,13 @@
 @tutor-navbar-padding-horizontal: 4rem;
 
 .navbar-fixed-top {
-
+  
+  // override bootstrap material design media query for (max-width: 1199px)
+  .navbar-brand {
+    height: 60px;
```
philschatz: is there a `bootstrap` var we can set instead?

PR #368, opened 2015-06-03
Set and target a data-crumb-count attribute (AKA media queries gone wild)

nathanstitt:
Top-margin is determined by the interection of screen width and data-crumb-count values

Single row after answering last question which created new breadcrumb.  The content pushes down as it should: 
![screen shot 2015-06-03 at 3 50 46 pm]
(https://cloud.githubusercontent.com/assets/79566/7971747/b24199b6-0a0c-11e5-917b-41117b54d5b8.png)

Even supports 4 rows :grin: 

screen shot 2015-06-03 at 4 18 27 pm

### DIFF
--- a/resources/styles/components/task/breadcrumbs.less
+++ b/resources/styles/components/task/breadcrumbs.less
@@ -1,3 +1,35 @@
+// TODO: 2.9rem was determined by trial and error
+// Can't figure out why it doesn't correspond to the
+// 2rem radius (4rem diameter) specified below for .crumb-circle
+@crumb-size: 2.9rem;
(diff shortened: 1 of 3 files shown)


pandafulmanda:
hmm, just so happens i was working on something for this too, i make a PR with what have now

nathanstitt:
@pandafulmanda's approach may be more promising then this.  It'll potentially be able to handle any kind of content that increases the top navbar size, while this PR will only work with breadcrumbs.

philschatz:
Closing in favor of #369

PR #369, opened 2015-06-03
Margin on container should be based on headerHeight

pandafulmanda:
handles different header heights for pinned containers

working on handling header height changes properly

a different take on #368 

## Margin top automatically updates as header size changes
For example when breadcrumbs fill into a new row
screen shot 2015-06-03 at 5 26 30 pm

## Smoothly adjusts on window resize
screen shot 2015-06-03 at 6 36 27 pm

## Works with pinned headers with offsets as well
screen shot 2015-06-03 at 6 37 05 pm

screen shot 2015-06-03 at 6 58 32 pm

### DIFF
--- a/src/components/pinned-header-footer-card/index.cjsx
+++ b/src/components/pinned-header-footer-card/index.cjsx
@@ -25,6 +25,7 @@ module.exports = React.createClass
     pinned: false
     shouldBeShy: false
     headerHeight: 0
+    containerMarginTop: '0px'
 
   mixins: [ScrollListenerMixin]
 
@@ -53,7 +54,7 @@ module.exports = React.createClass
     @setState(offset: offset)
 
   shouldPinHeader: (prevScrollTop, currentScrollTop) ->
-    currentScrollTop >= @state.offset
+    currentScrollTop >= @state.offset - @props.buffer
 
   isScrollingSlowed: (prevScrollTop, currentScrollTop) ->
     Math.abs(prevScrollTop - currentScrollTop) <= @props.scrollSpeedBuffer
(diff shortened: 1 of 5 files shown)


src/components/pinned-header-footer-card/index.cjsx
```diff
@@ -111,17 +116,44 @@ module.exports = React.createClass
...
+    else
+      container.style.marginTop = @state.containerMarginTop
+
   componentDidMount: ->
     @setOffset()
     @updatePinState(0)
+    @setOriginalContainerMargin()
+    @setContainerMargin()
+
+    window.addEventListener('resize', @setContainerMargin)
```
philschatz: Should this listener be unmounted?

pandafulmanda: good point =)

nathanstitt: Also you might consider adding a _.debounce into the flow there.  Something like:

```coffeescript
@resizeListener = _.debounce( => @setContainerMargin , 250) # season to taste 
window.addEventListener('resize',  @resizeListener)`
```

And then on unmount:

` window.removeEventListener(@resizeListener) if @resizeListener`

Some browsers are **really** aggressive on sending resize events every time the window moves a few pixels.  Without a debounce the resizing can get jerky

nathanstitt:
BTW: I'd normally suggest using https://github.com/sdecima/javascript-detect-element-resize vs using a window resize listener, but I don't think that would play well with React since it inserts an invisible element inside the element you're monitoring, and then listens for scrollevents.   I've used it on other projects before though and works great.

philschatz:
The code looks portable-enough to rewrite as a React component thought it also uses a window listener (`requestAnimationFrame`)

src/components/pinned-header-footer-card/index.cjsx
philschatz: The exercise builder has a similar `throttle` method; maybe these can be consolidated?

nathanstitt:
Oops, I guess my Google skills failed me this morning.  The actual lib I've used is: https://github.com/marcj/css-element-queries, which does not use a window listener or `requestAnimationFrame`   It's also pure JS and could be be extracted into a React component.  I've just used the `ResizeSensor` class directly to bind to an element, and not bothered with the css based stuff that's in `ElementQueries`.

Since @pandafulmanda has gotten this working, I think I'll close #368 in favor of this one.

src/components/task-plan/homework.cjsx
```diff
@@ -68,6 +68,7 @@ ChooseExercises = React.createClass
         hide={hide} />
 
       <PinnedHeaderFooterCard
+        containerBuffer={50}
```
philschatz: Could/Should this be a var we refer to in other places as well? And make it required? That way the size is explicitly described?

pandafulmanda: This is describing the extra space desired between the bottom of the pinned header and the start of the main content, [line 135 of the component](https://github.com/openstax/tutor-js/pull/369/files#diff-0ca5c713d2c99710c0c8b04c3da4e07bR135).  The bulk of the margin height is still dependent on the height of the header itself.  Since the height of the header will change based on state changes/header content or window resizing, I don't know if we can make the size explicit =(

PR #372, opened 2015-06-05
NaN fix for average row

darrenmason:
This removes NaN error in perfbook when average number not available


src/components/performance.cjsx
```diff
@@ -18,7 +18,10 @@ Performance = React.createClass
     <th>{heading.title}</th>
 
   renderAverageCell: (heading) ->
-    classAverage = Math.round(heading.class_average)
+    if heading.class_average
+      classAverage = Math.round(heading.class_average)
+    else
+      classAverage = heading.class_average
```
philschatz: Does this else case ever do anything since `heading.class_average` is falsy?

PR #375, opened 2015-06-05
Specify -webkit prefixes for chapter headings

nathanstitt:
Before:
screen shot 2015-06-05 at 2 31 45 pm

After:
screen shot 2015-06-05 at 2 34 12 pm

Does anyone know of a good flex box less mixin?  I think this bit of styling would benefit from it since it now has both -ie and -webkit prefixes.

### DIFF
--- a/resources/styles/book-content/base.less
+++ b/resources/styles/book-content/base.less
@@ -44,16 +44,16 @@ cnx-pi
   > div[data-type=title] { display: none; }
 
   p {
-    #fonts > .sans(1.5rem, 1.2em);
+    #fonts > .sans(1.5rem, 1.8em);
     width: 220px; // keep line length consistent
     color: rgba(255,255,255,.75);
     text-transform: uppercase;
     text-align: right;
     margin: 0 20px 0 0;
-    justify-content: center;
     display: flex;
     flex-direction: column;
     display: -webkit-flex;
+    justify-content: center;
     -webkit-justify-content: center;
     -webkit-flex-direction: column;
   }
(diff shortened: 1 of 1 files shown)


Fredasaurus:
Can you shorten the left side to make more room for the right?

screen shot 2015-06-05 at 2 38 10 pm

nathanstitt:
@Fredasaurus sure.  Sorry about that, I now see that you recently changed that to have a fixed width on the left so the line would break that way.  I thought that was a bug and should be evenly spaced.  That'll teach me to always check "git blame" lines before modifying them.

screen shot 2015-06-05 at 2 47 11 pm

*All screenshots are in safari btw, but I've also tested in chrome and firefox.  IE is down for the moment but I'll test it tonight

Fredasaurus:
Cool. While I'm being picky I wanted to change the line-height on the'<p>'  '<li>s' so they match.

screen shot 2015-06-05 at 2 54 15 pm

Fredasaurus:
also, how do you escape chars in comments again?

nathanstitt:
@Fredasaurus this is about as best as I can get it.  It lines up as long as there is only two learning objectives, but if there's three the left hand side remains aligned the same vertically.

screen shot 2015-06-05 at 7 52 05 pm


screen shot 2015-06-05 at 7 55 20 pm


Also, I'm not sure how to escape chapters in the comments either :grin:

Fredasaurus:
That's actually perfect. You are the man.

Fredasaurus:
I mean the Dantch.

PR #380, opened 2015-06-06
use git sha suffix for deployment

philschatz:
Uses the git sha for the suffix of minified files used in production.

@lakshmivyas it does not generate the manifest file produced by `gulp-rev`. Is that a problem?


lakshmivyas:
@philschatz - yes it is right now.

https://github.com/openstax/tutor-deployment/blob/master/tasks/build_assets.yml#L23-L26

But, its super simple to fix -  (do git rev-parse --short and then suffix to get the file name).   Lets make this PR and the upcoming one for tutor-deployment go hand in hand.  \cc @cnuber

philschatz:
closing this for now but hope to revisit later :smile:

PR #381, opened 2015-06-07
Adding "This feature will be available soon." text to performance book page

Fredasaurus:
Per David and Kathi's request.

Also hiding contents of the page. I'm just hiding is css, so please find a better way. This is just to get the point across.

screen shot 2015-06-07 at 4 06 36 pm

### DIFF
--- a/resources/styles/components/performance.less
+++ b/resources/styles/components/performance.less
@@ -1,0 +1,3 @@
+.-course-performance-container {
+	display: none;
+}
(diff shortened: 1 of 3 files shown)


pandafulmanda:
is this a patch for alpha?

philschatz:
I believe @darrenmason already did this (not via CSS) for the `alpha` branch.

darrenmason:
yes this was done on alpha in the component.  css way works too :)

PR #384, opened 2015-06-10
Required Styles and functionality

jivewise:
screen shot 2015-06-10 at 9 05 58 pm


resources/styles/global/forms.less
```diff
@@ -33,6 +38,27 @@ textarea {
   }
 }
 
+.invalid-form {
```
philschatz: Hmm, what about `is-` prefixes for classes conveying boolean states? (like `is-valid`, `is-required`, and `is-empty`?)

src/components/task-plan/plan-mixin.coffee
```diff
@@ -39,8 +39,13 @@ module.exports =
...
     {id} = @props
-    TaskPlanStore.addChangeListener(@saved)
-    TaskPlanActions.save(id)
+    publishable = TaskPlanStore.isValid(id)
+
+    if (publishable)
+      TaskPlanStore.addChangeListener(@saved)
+      TaskPlanActions.save(id)
+    else
+      @setState({invalid: true})
```
philschatz: I think this is the point that implements the "Publish button is always enabled. Show the required fields when teacher tries to publish" piece. Can you add a comment here blaming fred for why this is done this way? ; )

resources/styles/global/forms.less
jivewise: Sure, I'll do that now.

src/components/task-plan/plan-mixin.coffee
jivewise: Haha, yep.

resources/styles/global/forms.less
```diff
@@ -1,3 +1,5 @@
+@form-error-color: #E35C3B;
```
philschatz: Should this be defined in `variables.less` or already in `bootstrap-material-design.less`?

jivewise: Hm, I believe that's the only time we use that color anywhere, but I can put it into variables.less.  I was under the impression that for colors that are only used in one file we would want to keep it locally defined.

philschatz:
Also, should `global/forms.less` be `components/tutor-input.less`?

jivewise:
Yes, I think we can put ```global/material-form-wrapper.less``` in there as well.

resources/styles/global/forms.less
philschatz: I think the goal was to *at least* move the colors up to the top of the file (to ease refactoring) but the eventual goal is to have all the colors in one place. Since this is a new file I think it should be up in variables.

jivewise: Gotcha.

PR #385, opened 2015-06-10
Display and style past dates on date picker

nathanstitt:
Just as I was starting to fork the component in order to add something like this, I discovered that they've just added a new `dayComponent` prop that allows you to render the dates yourself.

screen shot 2015-06-10 at 10 05 01 am

### DIFF
--- a/src/components/tutor-input.cjsx
+++ b/src/components/tutor-input.cjsx
@@ -1,5 +1,8 @@
 React = require 'react'
 BS = require 'react-bootstrap'
+moment = require 'moment'
+
+{TimeStore} = require '../flux/time'
 {DateTimePicker} = require 'react-widgets'
 
 TutorInput = React.createClass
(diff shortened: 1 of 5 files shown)


nathanstitt:
This requires version 2.7 of react-widgets. 
I've bumped the version requirement in package.json, but everyone will have to make sure that they `nom install` after pulling this change.

src/components/tutor-input.cjsx
```diff
@@ -28,6 +31,19 @@ TutorInput = React.createClass
...
+  propTypes:
+    date:  React.PropTypes.string.isRequired
+    label: React.PropTypes.string.isRequired
+
+
+  render: ->
+    className = if moment(@props.date).startOf('day') < TimeStore.getNow()
+      'invalid'
+    else
+      'valid'
```
philschatz: Can you prefix this with `is-` (or the `invalid` case) and remove the other class name? I think that would also require adding a component name on it for the less file

philschatz:
Howdy! Can you merge in changes from master?

philschatz:
I am not seeing that behavior in the date picker. Also, previous dates seem to be selectable now:

image

nathanstitt:
Hm.  It looks fine to me on my system,  I even manually deleted dist/* to force a complete style rebuild and cleared cache.  It's also strange that your picker isn't the same width as the input.  Did you `npm install` ?

philschatz:
ah, you're right. I failed to read the instructions

philschatz:
Hmm, so visually the dates are there but they are still clickable (and select Today). Is this a limitation of the `react-widget`

nathanstitt:
Hmm good point. The selectable bit used to be a  limitation.   Anytime a day is clicked it closes the drop down.    We would then validate the selection and set it to today for past due selections. 

Now that we draw the day ourselves I bet we can intercept the click and call stopPropogation on the event.  Playing with that now

PR #386, opened 2015-06-10
performance report sort by column

darrenmason:
added ascending/descending column sort to performance report component

image


resources/styles/components/performance.less
```diff
@@ -0,0 +1,18 @@
...
+    th {
+      .no-select();
+      &.sortable {
+        cursor: pointer;
+      }
+      &.asc:after {
+        content: '\00a0\f0de';
+        font-family: FontAwesome;
+      }
+      &.desc:after {
```
philschatz: Can you use `ascending` and `descending` or `is-ascending` and `is-descending`? `desc` could also be a `description`

resources/styles/components/performance.less
```diff
@@ -0,0 +1,18 @@
+.performance-report {
+  table {
+    th {
+      .no-select();
+      &.sortable {
+        cursor: pointer;
+      }
+      &.asc:after {
+        content: '\00a0\f0de';
```
philschatz: Can you use the `fa-*` mixins instead, to add this?

src/components/performance.cjsx
```diff
@@ -14,8 +14,27 @@ Performance = React.createClass
   propTypes:
     courseId: React.PropTypes.string.isRequired
 
+  getInitialState: ->
+    sortOrder: 'asc'
+    sortIndex: 0
+    isNameSort: true
+
+  sortClick: (event) ->
+    if not _.contains(event.target.classList, 'student-name')
```
philschatz: Can you include a comment saying this is a special-case?

src/components/performance.cjsx
```diff
@@ -14,8 +14,27 @@ Performance = React.createClass
...
+      @setState({isNameSort: false})
+    else
+      @setState({isNameSort: true})
+    headers = event.target.parentNode.querySelectorAll('th')
+    for header in headers
+      header.classList.remove('asc', 'desc')
```
philschatz: Any reason this isn't done in the heading `render: ->`?

darrenmason: I'm not sure how I would isolate the clicked element from render; it seems more straightforward to do the class manipulation this way.

resources/styles/components/performance.less
```diff
@@ -0,0 +1,18 @@
+.performance-report {
+  table {
+    th {
+      .no-select();
+      &.sortable {
+        cursor: pointer;
+      }
+      &.is-ascending:after {
+        content: '\00a0\f0de';
```
philschatz: I believe there is a `.fa-*()` mixin that adds this char as well as the line below this.

philschatz:
It seems:

- sort alternates between ascending and descending when clicking on different headings
- Should student names be sorted by first name or family name?

darrenmason:
regarding family name sort, I can use a regex for the expected 'firstName LastName' pattern, but for handling titles such as 'Jr.' 'PhD', 'Dr.' etc. it would probably be better to retrieve as separate fields so we could say sortBy student.lastname

philschatz:
I think if we need to sort by last name then BE needs to provide a first and last name. Sorting by First name is fine and we can fix after feedback from UX

src/components/performance.cjsx
```diff
@@ -14,8 +14,33 @@ Performance = React.createClass
...
     courseId: React.PropTypes.string.isRequired
 
+  getInitialState: ->
+    sortOrder: 'is-ascending'
+    sortIndex: 0
+    isNameSort: true
+
+  sortClick: (event) ->
+    isActiveSort =
+      _.contains(event.target.classList, 'is-ascending') or
```
philschatz: I believe you can just do `event.target.classList.contains(...)`

jason-holmes: Default sorting by student last names is nearly universal in LMS, so we should follow that pattern.

-------
Jason Holmes, Ph.D.
User Experience Manager
OpenStax | OpenStax.org
(713) 893-1326

philschatz:
RE @jason-holmes : Yep, but we have 2 days left in the sprint and this requires backend and frontend coordination so I'm not sure if it's feasible to get in this sprint.

philschatz:
Can you merge in changes from master?

darrenmason:
merged in master

PR #388, opened 2015-06-11
add async button to help make it clear when an error occurs

philschatz:
A reusable button that changes state based on if there was an async error (i.e. Loading or Saving)

Waiting (Loading or Saving):
image

Error State:
image

Thoughts @nathanstitt @pandafulmanda @jivewise ?

# TODO

- [ ] Decide: Should this component actually listen to the store for the event or should we use the `isWaiting/Failed` booleans?
- [x] change error button to be a `<a>` that reloads the page
- [x] use similar timeout as the `<Loadable>` so the waiting state does not spin forever


pandafulmanda:
Overall excited about this component.  I think listening to the store for the event would make it too tightly coupled to a store and prefer the booleans.

pandafulmanda:
closing in favor of #393

PR #392, opened 2015-06-12
Center and disable text wrapping on exercise images

nathanstitt:
Unfortunately, images on exercises are just a IMG tag without a class or wrapping element so more specific selection isn't possible.

The `margin: @auto` bit isn't strictly needed, but it makes some larger images look better since they have equal margin on both sides rather than just a slightly larger white space on the right.

Before:
screen shot 2015-06-12 at 11 54 39 am

After:
screen shot 2015-06-12 at 11 55 10 am

### DIFF
--- a/resources/styles/components/task-step/exercise/index.less
+++ b/resources/styles/components/task-step/exercise/index.less
@@ -1,6 +1,13 @@
 .exercise-free-response {
   .size-exercise-card();
 
+  // images in exercises are not wrapped and don't have a class.
+  // This will work fine until there's some images that shouldn't be targeted
+  img {
+    display: block;
+    margin: 0 auto;
+  }
+
   textarea {
     margin: 1em 0;
     padding: 5px;


philschatz:
Does this styling also apply to the multiple choice and review screens? I noticed it's scoped inside `.exercise-free-response`

Dantemss:
Since the exercises importer generates the img tags, I can make it add some class if needed.

philschatz:
I think in order to provide additional classes you'd need to test the height and width ratios of each image. How difficult would that be? The bits of information that would be useful are:

- very wide (ie `> 600px`)
- very tall (ie `> 400px`)
- very thin (ie `< 200px`)
- the aspect ratio (not sure how useful this is)

And maybe more @nathanstitt ?

Dantemss:
Hmm there are ways to calculate the image size, but they are not that pretty...
Basically requires me to reopen the file in ImageMagick

nathanstitt:
@Dantemss I'd recommend using https://github.com/sdsykes/fastimage  It's pretty efficient and only reads the as much of the image as it needs to in order to determine the image type and it's resolution.

I think all we need to know about the images is the width.  What we care about is whether text should be allowed to wrap around them at a given screen width.  To do that I'd suggest small < 400px, md < 750px and large > 750px

philschatz:
Can we implement this in JS 1st (as a safe fallback that causes flicker) and then maybe in the BE as an optimization?

PR #393, opened 2015-06-12
Feature/async button

pandafulmanda:
moving from #388 diffs off of alpha to async button off of master

## Waiting (Loading or Saving):
image
screen shot 2015-06-11 at 7 21 53 pm

Error or Timed-out state:
screen shot 2015-06-12 at 4 26 32 pm


src/api.coffee
```diff
@@ -71,7 +71,7 @@ apiHelper = (Actions, listenAction, successAction, httpMethod, pathMaker) ->
         statusCode = jqXhr.status
         if statusCode is 200
           # HACK For PUT returning nothing (actually, it returns HTML for some reason)
-          successAction(args...)
+          successAction('', args...)
```
philschatz: Does this additional arg break other success actions?

src/components/buttons/async-button.cjsx
```diff
@@ -0,0 +1,74 @@
...
+    {isTimedout} = @state
+
+    if isWaiting and not isTimedout
+      _.delay =>
+        @checkForTimeout()
+      , timeout
+
+  checkForTimeout: ->
+    {isWaiting} = @props
+    @setState(isTimedout: true) if isWaiting
```
philschatz: Wouldn't `isWaiting` always be true since it's a prop?

pandafulmanda: no, it's a prop that is being updated by the parent render

src/api.coffee
pandafulmanda: i checked, but i can check again.  this should only be happening when we are faking PUTs and PATCHes on our end, and the correct/expected behavior is that ```args...``` is the second argument to follow the pattern set throughout the api helper as in line 68.

src/flux/helpers.coffee
```diff
@@ -110,6 +110,9 @@ CrudConfig = ->
       _.extend(@_changed[id], obj)
       @emitChange()
 
+    _save: (id) ->
+      @_asyncStatus[id] = SAVING
+
```
philschatz: Do we need to emit a change?

pandafulmanda: no, the emitChange is being handled in functions that call this

PR #395, opened 2015-06-16
Period aware builder for homework and readings

nathanstitt:
Does not save tasks yet.  Needs to incorporate https://github.com/openstax/tutor-server/pull/380 in order to get the period names / ids to save them.

screen shot 2015-06-16 at 11 06 49 am

screen shot 2015-06-16 at 11 06 59 am


resources/styles/components/task-plan/builder.less
```diff
@@ -0,0 +1,20 @@
+.assignment {
+
+  label.all-periods {
+    .form-control-wrapper .floating-label;
```
philschatz: Are these mixins? If so, can you add parentheses to make it explicit?

resources/styles/components/task-plan/builder.less
```diff
@@ -0,0 +1,20 @@
...
+  .periods-listing {
+    .table;
+    .table-striped;
+    .table-bordered;
+    .form-control { background-image: inherit; }
+  }
+
+  .instructions {
+    color: @tutor-neutral;
+    #fonts .sans(1.1rem, 1.1rem);
```
philschatz: Can you include `>` to make the markup the same as other places where this is used?

PR #399, opened 2015-06-17
"All Periods" checkbox support

nathanstitt:
If the "All  Periods" checkbox is checked, the periods are hidden and any updates to the dates are copied to all the periods.

If dates are set while the checkbox is checked, all taskings are set to the date.

When the checkbox is unchecked and then checked, the tastings dates are compared and the "global" date inputs are updated if all taskings have the same opens_at/due_at
screen shot 2015-06-19 at 12 28 15 pm

### DIFF
--- a/src/components/task-plan/builder.cjsx
+++ b/src/components/task-plan/builder.cjsx
@@ -48,11 +48,11 @@ module.exports = React.createClass
 
   setOpensAt: (value, period) ->
     {planId} = @props
-    TaskPlanActions.updateOpensAt(planId, value, period?.target_id)
+    TaskPlanActions.updateOpensAt(planId, value, period?.id)
 
   setDueAt: (value, period) ->
     {planId} = @props
-    TaskPlanActions.updateDueAt(planId, value, period?.target_id)
+    TaskPlanActions.updateDueAt(planId, value, period?.id)
 
   togglePeriodsDisplay: (ev) ->
     @setState(showingPeriods: not ev.target.checked)
@@ -72,7 +72,7 @@ module.exports = React.createClass
     TaskPlanActions.updateDescription(id, desc)
 
   renderTaskPlanRow: (plan) ->
-    <tr key={plan.target_id}>
+    <tr key={plan.id}>
       <td>{plan.name}</td>
       <td>
         <TutorDateInput
@@ -176,5 +176,4 @@ module.exports = React.createClass
 
   render: ->
     plan = TaskPlanStore.get(@props.planId)
-
     if @state.showingPeriods then @renderShownPeriods(plan) else @renderHiddenPeriods(plan)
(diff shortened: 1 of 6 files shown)


philschatz:
@Fredasaurus Does that seem reasonable for checking and unchecking the Dates?

Also, when editing an existing HW/iReading is "All Periods" ever checked?

philschatz:
When creating a new HW from the calendar (by clicking the date in the calendar) the opens field is populated but then when unchecking the "All Periods" it is not filled in.

image

image

image

nathanstitt:
Fixed:
dates

philschatz:
hmm, the Due At (and OpenAt) dates should probably be restricted so Today <= OpensAt < DueAt. Maybe a separate PR?

philschatz:
Deleting the date adds in the Epoch instead of blanking (maybe we need checkboxes for all the periods)

image

nathanstitt:
Hm, what did we decide for allowing entering arbitrary dates by text?  If we can disallow that, perhaps we could just make the text field be readonly.

philschatz:
Also, typing in the assignment name is now very slow and publish doesn't work. 422 Tasking Plans is invalid:

```json
{"status":422,"errors":[{"code":"tasking_plans_is_invalid","message":"Tasking plans is invalid"},{"code":"tasking_plans_is_invalid","message":"Tasking plans is invalid"}]}
```

philschatz:
RE arbitrary dates by text: I'm hoping we replace this date picker so it's not related to this PR. The ranges might be though, ideas? I think the bigger bit is the 422 publish error. I'm guessing it's because we're sending the period name in the `tasking_plans` objects

nathanstitt:
I'm currently receiving a 500 error on server even though it seems to be submitting valid JSON:
screen shot 2015-06-18 at 1 42 15 pm

src/components/task-plan/builder.cjsx
```diff
@@ -100,7 +130,7 @@ module.exports = React.createClass
             </tr>
           </thead>
           <tbody>
-            { _.map periods, @renderPeriodRow }
+            { _.map CourseStore.get(@props.courseId)?.periods, @renderTaskPlanRow }
```
philschatz: Is the `?` needed or should this be covered by the `<Loadable>` guard (and an error otherwise)?

philschatz:
If I change 1 period's open date the other period's open date also changes so there is no way to set both periods to have different open or due dates ; )

image

nathanstitt:
@Fredasaurus: Note that this slightly differs from the mockup.  The mockup has a "set individual periods" link which reveals the individual periods when clicked.  

I did not implement that since it seems to duplicate the behavior of the "All Periods" checkbox, and I wasn't sure if we needed two controls that do the same thing.  Perhaps I've mis-understood one of both of the controls purpose though?

Fredasaurus:
I'm looking at this this in-detail for the first time today. Thoughts forthcoming.

philschatz:
Merging but @Fredasaurus can you send thoughts to Nathan (or the pivotal ticket?)

PR #400, opened 2015-06-17
merge fix

darrenmason:
fixes merge ovewrite

### DIFF
--- a/src/components/performance.cjsx
+++ b/src/components/performance.cjsx
@@ -14,6 +14,29 @@ Performance = React.createClass
   propTypes:
     courseId: React.PropTypes.string.isRequired
 
+  getInitialState: ->
+    sortOrder: 'is-ascending'
+    sortIndex: 0
+    isNameSort: true
+
+  sortClick: (event) ->
+    isActiveSort = event.target.classList.contains('is-ascending', 'is-descending')
+    # this is a special case for the name header data which is one level above nested data
+    if not _.contains(event.target.classList, 'student-name')
+      @setState({isNameSort: false})
+    else
+      @setState({isNameSort: true})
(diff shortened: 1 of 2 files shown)


src/components/performance.cjsx
```diff
@@ -14,6 +14,29 @@ Performance = React.createClass
   propTypes:
     courseId: React.PropTypes.string.isRequired
 
+  getInitialState: ->
+    sortOrder: 'is-ascending'
+    sortIndex: 0
+    isNameSort: true
+
+  sortClick: (event) ->
+    isActiveSort = event.target.classList.contains('is-ascending', 'is-descending')
```
philschatz: Can this be done using React? I'm worried this may introduce Invariant Errors

darrenmason: its just checking classNames before the state change; it does not deform/append/remove any dom elements so it should not cause any invariants

src/components/performance.cjsx
```diff
@@ -14,6 +14,29 @@ Performance = React.createClass
...
+    if not _.contains(event.target.classList, 'student-name')
+      @setState({isNameSort: false})
+    else
+      @setState({isNameSort: true})
+    headers = event.target.parentNode.querySelectorAll('th')
+    for header in headers
+      header.classList.remove('is-ascending', 'is-descending')
+    if not isActiveSort
+      @state.sortOrder = 'is-ascending'
+    event.target.classList.add(@state.sortOrder)
```
philschatz: Doesn't this change the DOM?

darrenmason: Oh I see what you mean.  I did not know that adding/removing classes on existing elements would cause invariant.  I can manipulate classlist via render

darrenmason:
classes are now manipulated in render

PR #401, opened 2015-06-17
Bug/smooth out review nav

pandafulmanda:
## Various details around review nav including:
- [x] Smoother scroll by removing unnecessary state updates from scroll points
- [x] No more flicker of URL state when clicking on breadcrumbs to navigate problems
- [x] Pruning irrelevant warnings
- [x] Loading state with stats showing if already loaded from calendar stats popup
- [x] Fix when period switching would mess up the states from scroll points when exercises are different between periods

## Loading state when stats are already loaded

screen shot 2015-06-17 at 4 19 21 pm

### DIFF
--- a/src/components/course-periods-nav.cjsx
+++ b/src/components/course-periods-nav.cjsx
@@ -35,7 +35,7 @@ CoursePeriodsNav = React.createClass
 
     period = periods?[key]
     unless period?
-      console.warn("#{key} period does not exist for course #{courseId}. There are only #{periods.length}.")
+      throw new Error("BUG: #{key} period does not exist for course #{courseId}. There are only #{periods.length}.")
       return
 
     handleSelect?(period)
--- a/src/components/course-periods-nav.cjsx
+++ b/src/components/course-periods-nav.cjsx
@@ -34,13 +34,15 @@ CoursePeriodsNav = React.createClass
     periods = CourseStore.getPeriods(courseId)
 
     period = periods?[key]
-    console.warn("#{key} period does not exist for course #{courseId}. There are only #{periods.length}.") unless period?
+    unless period?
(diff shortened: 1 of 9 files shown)


src/components/course-periods-nav.cjsx
```diff
@@ -34,13 +34,15 @@ CoursePeriodsNav = React.createClass
     periods = CourseStore.getPeriods(courseId)
 
     period = periods?[key]
-    console.warn("#{key} period does not exist for course #{courseId}. There are only #{periods.length}.") unless period?
+    unless period?
+      console.warn("#{key} period does not exist for course #{courseId}. There are only #{periods.length}.")
```
philschatz: Can you `throw new Error('BUG: ...`? I'm hoping to replace all throw's with making an AJAX call to BE for debugging

src/components/task-teacher-review/index.cjsx
```diff
@@ -158,14 +179,37 @@ TaskTeacherReview = React.createClass
...
 
-  onNextStep: ->
-    @goToStep(@state.currentStep + 1)()
-
 
 TaskTeacherReviewShell = React.createClass
   contextTypes:
     router: React.PropTypes.func
 
+  renderLoading: (refreshButton) ->
```
philschatz: Why not just use the default `renderLoading`? or use 2 nested `<Loadable>` items (one for the HW/iReading and one for all the problems)?

philschatz:
When clicking on the breadcrumb it seems the screen scrolls to the bottom of the section rather than the top.

pandafulmanda:
@philschatz this is ready for review again

philschatz:
I did once get `Uncaught TypeError: Cannot read property 'scrollPoint' of undefined` but it may have been because MathJax did not load yet.

PR #406, opened 2015-06-18
Async button footer

jivewise:
- [x] Get footer to use new async buttons

Moving publish/save logic to different PR.


philschatz:
I believe this is blocked on BE API changes, right? Adding "Blocked" label

jivewise:
Yea, but I can work off of that branch/locally in the mean time.

src/components/task-plan/plan-mixin.coffee
```diff
@@ -61,6 +61,7 @@ module.exports =
   saved: ->
     courseId = @props.courseId
     TaskPlanStore.removeChangeListener(@saved)
```
philschatz: Should this be `TaskPlanActions.saved.removeListener(@saved)`?

src/components/task-plan/plan-mixin.coffee
```diff
@@ -53,14 +53,15 @@ module.exports =
...
-      TaskPlanStore.addChangeListener(@saved)
+      TaskPlanActions.saved.addListener(@saved)
       TaskPlanActions.save(id)
     else
       @setState({invalid: true})
 
   saved: ->
     courseId = @props.courseId
-    TaskPlanStore.removeChangeListener(@saved)
+    TaskPlanStore.saved.removeListener(@saved)
```
philschatz: I think that if `save` ends up erroring it will leave a dangling listener (unless you listen to `FAILED` as well)

PR #407, opened 2015-06-18
Fixed updating plan descriptions

jivewise:
Updating plan descriptions was throwing an error before.  This should fix it.

No UI changes.


philschatz:
There were some other places I marked as `HACK` that did something like:

```coffee
{id} = @props
id ?= @props.planId
```

Is that additional check also needed? If not, can you remove them as well?

jivewise:
Yes, I'll do that.

philschatz:
can you merge from master?

PR #408, opened 2015-06-18
Cleanup Reference View Teacher's Edition notes

dak:


philschatz:
It seems like `section [data-type='note']` should never get any horizontal padding:

image

and TE content:
image

philschatz:
Any update on this?

PR #412, opened 2015-06-19
Fix "booksplash" banner so it displays in IE

nathanstitt:
The main issue was that the :before block was missing a height, which IE defaulted to 0.

The rest is tweaks so it displays identically in both IE and chrome/firefox/safari.

### DIFF
--- a/resources/styles/global/tutor-booksplash-background.less
+++ b/resources/styles/global/tutor-booksplash-background.less
@@ -7,19 +7,20 @@
   &:before {
     content: attr(data-title);
     position: fixed;
-    top: 48px;
+    top: 60px;
     left: 0px;
     right: 0px;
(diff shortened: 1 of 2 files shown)


nathanstitt:
@Fredasaurus One of the issues with getting the banner to display identically was specifying a font that was common between all browsers.  With text that large, small differences in the font makes it appear in drastically different locations.   `HelveticaNeue-Bold` wasn't being displayed in IE or Firefox.  I've changed it to be `Lato` since that's loaded as part of the build so it's always available.

Is there a different font you'd prefer me to try?

This changes the splash from:
screen shot 2015-06-19 at 4 08 19 pm

To be (as Lato)
screen shot 2015-06-19 at 4 07 56 pm

PR #413, opened 2015-06-20
Side stats on review overflow for when window is very short or when stats are long

pandafulmanda:
## Scrolling stats when content is too long or window is too short to accomodate
screen shot 2015-06-19 at 7 05 32 pm

## Along the way...
* Added resize listener mixin for use whenever we need to listen to resize
  * throttles resize for safety
  * has some handy state values such as window size and component size and which dimension is resizing
* Pulled out repeated getPosition function into a mixin
* Translate review stats up to get rid of extra spacing when nav is shy


pandafulmanda:
@philschatz ready for review

philschatz:
FWIW React is moving away from mixins (maybe deprecating)

src/components/resize-listener-mixin.cjsx
```diff
@@ -0,0 +1,67 @@
...
+      nextState.resizing.width = not (@state.$window.width is nextState.$window.width)
+
+  resizeEffect: (resizeEvent) ->
+    @setSizeState(resizeEvent)
+    @_resizeListener?(resizeEvent)
+
+  _getWindowSize: ->
+    width = window.innerWidth
+    height = window.innerHeight
+    $window = {width, height}
```
philschatz: Should this just be `{width, height}`

src/components/resize-listener-mixin.cjsx
```diff
@@ -0,0 +1,67 @@
...
+
+module.exports =
+  propTypes:
+    resizeThrottle: React.PropTypes.number
+
+  getDefaultProps: ->
+    resizeThrottle: 200
+
+  getInitialState: ->
+    $window: {}
```
philschatz: What is the reason for the distinction between `$` and `_` in the state? I'm guessing you have a prefix because this is a mixin?

pandafulmanda: yes, and $ as a prefix for DOM specific properties

pandafulmanda:
@philschatz i've updated the state names to using ```Initial``` and ```El``` as suffixes in replacement of ```_``` and ```$``` prefixes

philschatz:
Maybe I'm doing something wrong but clicking the tabs in the following screenshot does not update the URL:

image

PR #417, opened 2015-06-22
Course administration

nathanstitt:
screen shot 2015-07-07 at 3 10 31 pm

Issues outstanding with BE:
 * Reset password does not work.  `PATCH /api/students/123` completes without errors but no db updates are performed.  Can the teacher perform this action?


jpslav:
I believe the course settings stuff is not actually in this sprint? Not talking about the roster, but more the course name / time zone / radio button settings editing.   I think for this sprint we can just have the Course Information heading with an uneditable name under it.  @philschatz @nathanstitt is that good for you?

nathanstitt:
Sounds good to me.  I wasn't sure if that was intended or not and was just implementing what the mockup showed.  I'll rework the screen to only display an uneditable course name on the top with the Roster management below that.

Dantemss:
Timezone is in... at least on the BE side. Not sure about the others (or actually changing the timezone :D)

nathanstitt:
@Dantemss can you take a look at the BE issues mentioned above and comment on them?

src/components/course-administration/add-student.cjsx
```diff
@@ -0,0 +1,80 @@
...
+BS = require 'react-bootstrap'
+_ = require 'underscore'
+{RosterActions, RosterStore} = require '../../flux/roster'
+{TutorInput} = require '../tutor-input'
+BindStoreMixin = require '../bind-store-mixin'
+
+BLANK_STUDENT = first_name: '', last_name: '', email: '', password: ''
+
+Field = React.createClass
+  propTypes:
```
philschatz: Can you add a `displayName` field?

src/components/course-administration/add-student.cjsx
```diff
@@ -0,0 +1,80 @@
...
+    courseId: React.PropTypes.string.isRequired
+    period: React.PropTypes.object.isRequired
+
+  mixins: [BindStoreMixin]
+  bindStore: RosterStore
+  bindEvent: 'created'
+
+  getInitialState: ->
+    BLANK_STUDENT
```
philschatz: Does this need to be cloned? So as not to accidentally change the `BLANK_STUDENT`?

src/components/course-administration/add-student.cjsx
nathanstitt: I thought it wasn't needed if the class is assigned to a variable.  https://facebook.github.io/react/docs/jsx-in-depth.html#the-transform says: 
JSX will infer the class's displayName from the variable assignment when the displayName is undefined:

Still probably worth adding though, that way if it's later exported it'll be there.

src/components/course-administration/reset-password.cjsx
```diff
@@ -0,0 +1,43 @@
...
+PasswordResetField = React.createClass
+  mixins: [React.addons.LinkedStateMixin]
+
+  getInitialState: ->
+    password: ''
+
+  getValue: -> @state.password
+
+  componentDidMount: ->
+    React.findDOMNode(@refs.input).focus()
```
philschatz: Why is this focus done differently than in `add-student.cjsx`?

src/components/tutor-input.cjsx
```diff
@@ -17,18 +17,22 @@ TutorInput = React.createClass
   onChange: (event) ->
     @props.onChange(event.target?.value, event.target)
 
+  focus: ->
+    React.findDOMNode(@refs.input)?.focus()
```
philschatz: Why is this focus done differently than in `add-student.cjsx`?

src/router.cjsx
```diff
@@ -47,6 +47,7 @@ routes = (
           <Route path='homeworks/:id/?' name='editHomework' handler={HomeworkShell} />
           <Route path='readings/new/?' name='createReading' handler={ReadingShell} />
           <Route path='readings/:id/?' name='editReading' handler={ReadingShell} />
+          <Route path='admin' name='courseAdministration' handler={CourseAdministration} />
```
philschatz: Can this be `settings` to keep `admin` meaning site admin?

src/components/course-administration/add-student.cjsx
nathanstitt: No really because React will clone anything that's passed into state.  However, I don't think that's documented or anything so they might change the behavior at some point.  I'll add a `_.clone`

src/components/tutor-input.cjsx
nathanstitt: This adds a `focus` method to the `TutorInput` component, which is then called by `add-student-cjsx`  The `?` is used since it's possible that it might be called while the component is unmounted, in which case `findDOMNode` would return null.

The whole need for the explicit focus comes from them being rendered inside a popover component so they're invisible when first rendered.  That defeats the normal autofocus attribute on an input.

src/router.cjsx
nathanstitt: yeah, good call.  Making the change.

src/components/course-administration/reset-password.cjsx
```diff
@@ -0,0 +1,44 @@
...
+  displayName: 'ResetStudentPasswordField'
+  mixins: [React.addons.LinkedStateMixin]
+
+  getInitialState: ->
+    password: ''
+
+  getValue: -> @state.password
+
+  componentDidMount: ->
+    React.findDOMNode(@refs.input).focus()
```
philschatz: Why is this focus done differently than in `add-student.cjsx`?

nathanstitt: This focus is performed on a plain `<input />` component, while `add-student.cjsx` operates on a `TutorInput`  While it would be possible for `AddStudent` to reach down into `TutorInput`'s dom and muck around, it seemed better to ask the `TutorInput` to focus itself.

src/components/course-administration/add-student.cjsx
philschatz: O cool! I didn't know that! maybe we should always assign it to a variable before exporting and then get rid of the `displayName` everywhere

nathanstitt:
As discussed on https://github.com/openstax/tutor-js/pull/448#issuecomment-119301969, this changes the link color globally to #00c1de.

pandafulmanda:
:sunglasses:

PR #419, opened 2015-06-23
Feature/cal nav by route

pandafulmanda:
screen shot 2015-06-22 at 7 33 54 pm

a nice to have.

### DIFF
--- a/src/components/course-calendar/add-menu-mixin.cjsx
+++ b/src/components/course-calendar/add-menu-mixin.cjsx
@@ -15,7 +15,7 @@ CourseAddMenuMixin =
     addDate: null
 
   getDefaultProps: ->
-    dateFormat: 'MM-DD-YYYY'
+    dateFormat: 'YYYY-MM-DD'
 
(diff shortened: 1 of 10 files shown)


philschatz:
I like the ability to specify the month in the calendar (maybe @jivewise could send back to the correct month on save or just go to history `-1`).

How much work would it be to only include the year and month? I imaging each `transitionTo` call would have to handle it. Also, I wonder if `http://localhost:3001/courses/1/t/calendar/months/06-99-2015/` should do something other than:

image

pandafulmanda:
ah! did think about that edge case

pandafulmanda:
fixed the edge case, calendar should now redirect to today if an invalid date is put in the url

pandafulmanda:
need to take out some sneaky stuff that I did with abusing setting the value of nextState[some property] to values in willComponentUpdate

pandafulmanda:
@philschatz there should be no more tomfoolery

PR #421, opened 2015-06-23
Fix reference view toc position to left side of screen

dak:


philschatz:
Can you include a screenshot in the PR description?

PR #423, opened 2015-06-23
Teacher viewing tasks for individual student

pandafulmanda:
- [x] fix courseId as constant hack
- [x] make things like "Back to Dashboard" role appropriate -- i.e. "Back to Performance" instead
- [x] put student name somewhere

## Reading
screen shot 2015-06-26 at 2 14 51 pm

## Homework
screen shot 2015-06-25 at 5 18 52 pm

### DIFF
--- a/resources/styles/components/task/index.less
+++ b/resources/styles/components/task/index.less
@@ -5,6 +5,8 @@
 .task {
   max-width: @tutor-task-width;
   margin: 0 auto;
+
+  @import '../task-step/step-footer-mixin';
 }
 
 .task-view {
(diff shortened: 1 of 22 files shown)


pandafulmanda:
@philschatz readyyy =

resources/styles/components/task/index.less
```diff
@@ -5,6 +5,8 @@
 .task {
   max-width: @tutor-task-width;
   margin: 0 auto;
+
+  @import '../task-step/step-footer-mixin';
```
philschatz: Rather than nesting the imports can you make an explicit mixin in `task-footer-mixin.less` (maybe named `.task-footer-mixin()`)?

pandafulmanda: i've made a mixin in ```/task-step/step-footer-mixin``` so that file path is reflective of the components one.

PR #424, opened 2015-06-24
Publish api changes

jivewise:
No UI changes


philschatz:
Save as Draft does not seem to work for me.

PR #422, opened 2015-06-23
Browse book link for teacher task builder

nathanstitt:
Adds links to the topic selector to jump to the related chapter in the book.

screen shot 2015-06-23 at 4 55 42 pm

PR #426, opened 2015-06-24
Add "Reading covers <sections>" message to footer

nathanstitt:
screen shot 2015-06-30 at 12 23 51 pm

Also changes routes to support linking directly to a section:

screen shot 2015-06-30 at 12 21 46 pm


pandafulmanda:
:+1:

src/flux/task.coffee
```diff
@@ -169,6 +169,15 @@ TaskConfig =
...
 
+    getRelatedSections: (taskId) ->
+      _.chain(getSteps(@_steps[taskId]))
+        .pluck('chapter_section')
+        .compact()
+        .map( (cs) -> cs.join('.') )
+        .uniq()
+        .value()
+
+
```
pandafulmanda: is there a way to consolidate the chapter section formatting logic either here or to the mixin?

nathanstitt: Great idea.  I had the `.map` in there mainly so the later `uniq` would compare properly. But it's really not needed though since `.uniq` also accepts a comparator function.

nathanstitt:
WIP until https://github.com/openstax/tutor-js/pull/422 which in turn depends on https://github.com/openstax/tutor-server/pull/396 is merged.  #422 extends the reference routes to support jumping to chapters.

resources/styles/components/reference-book/toc.less
```diff
@@ -21,9 +14,33 @@
...
+      #fonts > .sans(1.3rem);
+      border-top: 1px dotted @tutor-neutral;
+      a {
+        color: @tutor-black;
+        padding: 5px;
       }
     }
-
+    ul { padding-left: 0rem; }
+    ul[data-depth='1'] a{
```
philschatz: Why not making these a file-local mixin?

src/components/buttons/browse-the-book.cjsx
```diff
@@ -2,27 +2,27 @@ React = require 'react'
...
-        {courseId: @props.courseId})
-    <BS.Button
-      bsStyle='primary'
-      target="_blank"
-      className='view-reference-guide'
-      href={link}
-    >
-      Browse the Book
-    </BS.Button>
+      <BS.Button bsStyle='primary', {...linkProps}>{text}</BS.Button>
```
philschatz: Is the comma necessary?

src/components/reference-book/navbar.cjsx
```diff
@@ -21,10 +21,15 @@ module.exports = React.createClass
     @refs.tocmenu.setDropdownState(true) unless cnxId
 
   renderSectionTitle: ->
-    page = ReferenceBookStore.getPageInfo(@getParams())
+    {cnxId, section} = @getParams()
+    if cnxId
+      page = ReferenceBookStore.getPageInfo(@getParams())
+      section = page?.chapter_section.join('.')
```
philschatz: Would the `<ChapterSection>` component work here?

src/components/task-step/step-footer-mixin.cjsx
```diff
@@ -20,16 +22,30 @@ module.exports =
 
     taskDetails
 
+  renderCoversSections: (sections) ->
+    sections = _.map sections, (parts) ->
+      combined = parts.join('.')
```
philschatz: Can this use the `<ChapterSection>` component?

src/components/task-step/step-footer-mixin.cjsx
```diff
@@ -20,16 +22,30 @@ module.exports =
...
+      <BrowseTheBook unstyled key=combined section={combined}>
```
philschatz: can you change to `key={combined}`?

src/flux/reference-book.coffee
```diff
@@ -25,10 +25,19 @@ ReferenceBookConfig = {
     getToc: (courseId) ->
       @_get(courseId)['0']
 
-    getChapterFirstPage: (courseId, chapterId) ->
+    # Takes a courseId and a chapter_section specifier
+    # which is a string joined with dots i.e. "1.2.3"
+    getChapterSectionPage: ({courseId, section}) ->
+      parts = _.map(section.split('.'), (part) -> parseInt(part, 10) )
```
philschatz: Can this use `<ChapterSection>`?

src/components/task-step/step-footer-mixin.cjsx
nathanstitt: Heh, I didn't know about the `<ChapterSection>` component.  Looking at it now, but I think it'll have to be reworked to use `join` since Bio has 3 section components. i.e. `1.1.2`

src/components/buttons/browse-the-book.cjsx
nathanstitt: The comma was an oversight, I've removed it.  

The react-bootstrap components translate the `bsStyle='primary'` prop into `btn-primary`, and the `btn` class is applied since it's a `Button` component.  If you think it'd be clearer I can use explicit classes though.

src/flux/reference-book.coffee
nathanstitt: Not really since it's a flux store and doesn't do any formatting.  What it does is attempt to find the first first page for a given section.

src/components/buttons/browse-the-book.cjsx
philschatz: I only included the `className="..."` since `<a>` is not a `BS` element. I guess I was surprised `<BS.Button>` is able to open links in a new tab.

src/flux/reference-book.coffee
philschatz: ah ok ; )

PR #427, opened 2015-06-25
refactor sneaky parts of smart overflow stats

pandafulmanda:
* no setting nextState in component will update
* no using delay to set a state change on the event queue


pandafulmanda:
@philschatz ready for review

PR #434, opened 2015-06-30
Fix due date/open date on load

jivewise:
screen shot 2015-07-03 at 9 17 02 am
- [x] Due Date and Open Date Loading
- [x] Moved labels inside date inputs
- [x] Made date coming from calendar due date instead of open date
- [x] All periods date inputs disabled and not required where appropriate
- [x] Labels for checkboxes will now select or deselect checkboxes

### DIFF
--- a/src/components/tutor-input.cjsx
+++ b/src/components/tutor-input.cjsx
@@ -83,7 +83,11 @@ TutorDateInput = React.createClass
     wrapperClasses = ["form-control-wrapper", "tutor-input"]
 
     now = TimeStore.getNow()
-    value = if @props.value then new moment(@props.value) else null
+    value = @props.value
+    value = if value and value.getTime and not isNaN(value.getTime())
+      new moment(value)
+    else
+      null
     min = if @props.min then new moment(@props.min) else new moment(now).subtract(10, 'years')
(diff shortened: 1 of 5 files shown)


philschatz:
Thanks for working on this last night! I noticed a couple issues:

- When only the open date is set and then "Individual Periods" is selected then `Invalid Date` is set. It should be blank
- Clicking on the radio label should also select the radio (I noticed you changed the `htmlFor` attrib but that does not seem to fix the issue)

Also, (not part of this PR but...) the queryString date should go to the `dueAt` field instead of the `opensAt` field.

src/components/tutor-input.cjsx
```diff
@@ -83,7 +83,11 @@ TutorDateInput = React.createClass
     wrapperClasses = ["form-control-wrapper", "tutor-input"]
 
     now = TimeStore.getNow()
-    value = if @props.value then new moment(@props.value) else null
+    value = @props.value
+    value = if value and value.getTime and not isNaN(value.getTime())
```
philschatz: When is `value` set to an invalid time?

jivewise: This code is more for prevention, I think setting the max and required props worked helped to eliminate the error state you experienced earlier.

PR #436, opened 2015-06-30
performance report: view/switch tabs by period

darrenmason:
image

### DIFF
--- a/src/components/performance.cjsx
+++ b/src/components/performance.cjsx
@@ -27,8 +27,7 @@ Performance = React.createClass
 
 
   getInitialState: ->
-    periodIndex = 0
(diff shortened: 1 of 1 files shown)


src/components/performance.cjsx
```diff
@@ -27,8 +27,7 @@ Performance = React.createClass
 
 
   getInitialState: ->
-    periodIndex = 0
-    performance = @getPerfByPeriod(periodIndex)
+    period: "1"
```
philschatz: Should this be `null`?

philschatz:
hmm, in `render: ->`: `performance = PerformanceStore.get(@props.courseId)` returns null for me but it looks like there is no code for handling this case.

PR #438, opened 2015-07-02
Learning Guide - new layout, individual chapters

darrenmason:
image


darrenmason:
progress bars represent current_level out of 1.  level is currently randomized from server on load.

philschatz:
Thanks! the randomized level is not new, right?

Also, are `d3` in `package.json` and loading images via `app-config` still needed?

darrenmason:
yes, randomization not new, just noting it :)

I will remove d3 package & app-config methods

PR #442, opened 2015-07-03
Move continue and other navigation buttons to right

nathanstitt:
screen shot 2015-07-03 at 5 04 54 pm

screen shot 2015-07-03 at 5 04 39 pm

screen shot 2015-07-03 at 5 04 29 pm

### DIFF
--- a/resources/styles/components/task-step/step-footer-mixin.less
+++ b/resources/styles/components/task-step/step-footer-mixin.less
@@ -29,11 +29,16 @@
       display: inline-block;
       float: right;
     }
+    .navigation {
+      float: right;
+    }
     .details {
+      float: left;
+      padding-top: 10px;
       // Has a "Covers sections ..." message
       &.has-sections {
+        padding-top: 0;
         position: relative;
-        float: right;
         height: 40px;
         .task-covers {
           position: absolute;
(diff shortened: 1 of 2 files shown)


philschatz:
is this based on mockups? @Fredasaurus thoughts?

nathanstitt:
No mockups (that I'm aware of).  I was assigned multiple pivotal cards for it, a search for "label:fe button right" pulls them up.  Of course perhaps this isn't what was intended :grin:

jason-holmes:
The Continue button for readings needs to remain on the left because it supports demonstrated reading behavior.  It should not go on the right.

http://www.nngroup.com/articles/f-shaped-pattern-reading-web-content/ <http://www.nngroup.com/articles/f-shaped-pattern-reading-web-content/>

I’m not sure whose idea it was to move it right or why it’s gained so much traction.  Lots of UX is subjective and negotiable but I feel strongly the the original location on the left is the best location for this button based on eye tracking research.

-------
Jason Holmes, Ph.D.
User Experience Manager
OpenStax | OpenStax.org
(713) 893-1326

kathi-fletcher:
I added the tickets because David Harris mentions that it has to move every time he speaks to me. But then Jason found this research. And I haven't done the work yet to get us all on the same page. Did not intend the FE tickets to get done before the UX work to decide ultimately, but totally understand how it got picked up.

philschatz:
Thanks for doing the work! but closing because **"research"** :smile:

nathanstitt:
Yes, sounds for the best.  It's git we can always resurrect it later if we need to :grinning:

kathi-fletcher:
Folks -- I have just confirmed with David that we can officially leave the Continue button on the left for readings for now.

PR #444, opened 2015-07-06
Fix mathjax exercise answer rendering height

nathanstitt:
This corrects the height of the mathjax spans.

Before:
screen shot 2015-07-06 at 11 51 19 am

After:
screen shot 2015-07-06 at 11 50 49 am

This also removes call to `cleanMathArtifacts` on startup.  I'd noticed that occasionally I'd see a "this.queue is undefined" exception when the page was reloaded in chrome but hadn't been able to find the source of the error.  Mathjax continued to render fine, so I didn't pay it too much attention.  When debugging the mathjax positioning I loaded in Firefox where it occurred more often and was able to find it originating from the cleanup call.

### DIFF
--- a/resources/styles/global/maths.less
+++ b/resources/styles/global/maths.less
@@ -14,17 +14,3 @@
 // Hide the HTML and use MathJax for rendering
 // TODO: Decide whether to remove KaTeX (involves additional code for MathJax)
 .has-html .katex > .katex-html { display: none; }
-
-// HACK: Override some odd properties being directly added on to 2 MathJax-related elements
-.MathJax .math {
-  > span {
-    display: inline !important;
-    position: static !important;
-    width: auto !important;
-    height: auto !important;
-
-    > span {
-      position: static !important;
-    }
-  }
-}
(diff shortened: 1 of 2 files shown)


nathanstitt:
@Dak the changes I made to `maths.less` undo the commits you made to get rendering working better on the reference book in https://github.com/openstax/tutor-js/commit/29c651098a55019bece7f65f44f7c25aa68fea55

I'm leaving this WIP until we can discuss a method that works for both areas.

nathanstitt:
@dak I think I've traced the reference book issue down to cleaning out some mathjax elements that apparently shouldn't be removed.

The obviously incorrect rendering I saw was:
screen shot 2015-07-06 at 2 12 23 pm


-------------------------
The formula then rendered correct after I changed the mathjax cleanup not to remove the `MathJax_Hidden` element:


screen shot 2015-07-06 at 2 11 28 pm

philschatz:
Looks good to me, and less `!important` in the CSS is always :smile:

Fredasaurus:
Where is this in the content? I'd like to look at this page in context.

nathanstitt:
@Fredasaurus The first issue occurred on any exercise that had math as part of the answer.

The second issue was on section 3.2 of the Physics reference book.  The bug would only occur when you advance to the page though, not if you only load/reload the page.  If you load the reference book and page forward two pages you should see it.

Fredasaurus:
Cool, thanks. I was curious about the other fonts on the page and how they were rendering.

PR #445, opened 2015-07-06
fix step footer when viewing as teacher

pandafulmanda:
screen shot 2015-07-06 at 12 13 04 pm

### DIFF
--- a/api/courses/1/performance.json
+++ b/api/courses/1/performance.json
@@ -1,4 +1,5 @@
-{
+[{
+  "period_id": "1",
   "data_headings": [
     {
       "title": "Reading 1"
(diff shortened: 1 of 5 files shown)


philschatz:
Hmm, might https://github.com/openstax/tutor-server/pull/409 fix part of this?

philschatz:
and @darrenmason does the JSON change to `/api/courses/1/performance.json` match the format you're expecting?

pandafulmanda:
no, ```/api/courses/1``` is the route that is missing the roles and causing problems.  ```/api/user/courses``` is the one with all courses and roles as well

darrenmason:
@philschatz it matches

PR #453, opened 2015-07-07
Correct tag usage to accommodate Biology content

nathanstitt:
With the new Biology content, the LO tag can be called either `lo` or `aplo`, and the tag's name field no longer contains the description.

The exercise builder has JS errors without this when it attempts to find a LO tag and cannot


Dantemss:
BE PR merged.

PR #454, opened 2015-07-08
Force Mathjax text elements to use lato font-family

nathanstitt:
This changes the text in Mathjax from using it's built-in STIXGeneral-Regular font:
screen shot 2015-07-08 at 10 16 14 am

To use lato:
screen shot 2015-07-08 at 10 17 11 am

Unfortunately, the difference isn't very apparent on the screenshots.  Viewing them at full-size is a bit better but it's still not a big change.

If we want to spend more time on this, Mathjax does provide a method to import the Lato font so it knows about it's glyph data and can render everything with it.  http://docs.mathjax.org/en/latest/font-support.html  has the details.

### DIFF
--- a/resources/styles/global/maths.less
+++ b/resources/styles/global/maths.less
@@ -13,3 +13,6 @@
 // Hide the HTML and use MathJax for rendering
 // TODO: Decide whether to remove KaTeX (involves additional code for MathJax)
 .has-html .katex > .katex-html { display: none; }
+
+// Override the inline style on mathjax text elements to match the tutor default of lato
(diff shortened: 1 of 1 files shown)


Fredasaurus:
Awesome! I didn't even know that was a possibility.

Fredasaurus:
One of the things I want to test is laying all of the Physics copy out in Lato, so see if that reads better with the notation.

philschatz:
so the content is one large <math> tag rather than multiple <math> tags ; (

PR #455, opened 2015-07-08
Performance Export

pandafulmanda:
## No previous exports
screen shot 2015-07-08 at 12 01 54 pm

## Exporting for the first time
screen shot 2015-07-08 at 12 01 27 pm

## Previously exported
screen shot 2015-07-08 at 11 45 19 am

## Exporting again
screen shot 2015-07-08 at 12 06 15 pm

essentially working, needs work on the details:
- [x] previous export awareness
- [x] "Last exported" label
- [x] if job failed
  - discuss with @joemsak, in the future if needed, the BE can return ```failed``` as state, and FE will respond accordingly
- [x] fix checkuntil


src/flux/performance-export.coffee
```diff
@@ -0,0 +1,73 @@
...
+EXPORTING = 'exporting'
+EXPORT_QUEUED = 'export_queued'
+EXPORTED = 'exported'
+
+PerformanceExportConfig = {
+
+  _job: {}
+
+  getJobIdFromJobUrl: (jobUrl) ->
+    jobUrlSegments = jobUrl.split('/api/jobs/')
```
philschatz: Is parsing the URL necessary or can we just use the `jobUrl` instead of the `jobId`?

src/components/buttons/export-button.cjsx
```diff
@@ -0,0 +1,52 @@
...
+
+    exportClass = 'primary'
+
+    if downloadUrl?
+      downloadLink =
+        <BS.Button bsStyle='primary' href={downloadUrl}>Download</BS.Button>
+
+      exportClass = 'default'
+
+    exportButton =
```
philschatz: Can you show either the `Export` button xor the `Download` button?

src/flux/performance-export.coffee
pandafulmanda: wanted to be able to use the JobStore and the api.coffee in the same consistent way

src/components/buttons/export-button.cjsx
pandafulmanda: i can, but at the moment, the BE does not invalidate the exports if the perf report has been updated since last export.

philschatz:
d'oh! can you merge in changes from master?

src/components/buttons/export-button.cjsx
philschatz: Hmm, I wonder if the BE should just press the button themselves when the data updates ; ) @jpslav @joemsak ? We can still show the spinner if the job hasn't completed

jpslav: Does providing time stamp data on the export help? (which you can compare to the time the perf report page was loaded to see if it later than the data being shown on the page)

pandafulmanda: i guess that would mean, on page load, we can always assume the export needs to be run and show only the export button on load.  once the export has be run, show only the download button

src/components/buttons/export-button.cjsx
```diff
@@ -0,0 +1,89 @@
...
+    downloadUrl: null
+    lastExported: null
+
+  isUpdateValid: (id) ->
+    {courseId} = @props
+    id is courseId
+
+  componentWillMount: ->
+    {courseId} = @props
+    PerformanceExportActions.load(courseId)
```
philschatz: Any reason for not using `<Loadable>` here instead of attaching the listeners manually?

src/flux/performance-export.coffee
```diff
@@ -0,0 +1,85 @@
...
+    (jobData) =>
+      exportData = _.clone(jobData)
+      exportData.exportFor = id
+
+      @_asyncStatus[id] = exportData.state
+      @emit("performanceExport.#{exportData.state}", exportData)
+      @emitChange()
+
+  getJobIdFromJobUrl: (jobUrl) ->
+    jobUrlSegments = jobUrl.split('/api/jobs/')
```
philschatz: @joemsak  could the `POST` which returns the Job URL return the job id instead? That way Amanda does not have to parse the `id` from the URL

jpslav: Would you guys want both?

```ruby
{
  job: {
    id: blah,
    url: blah
  }
}
```

or similar

jpslav: also FWIW there is a larger pending issue to improve the Job json output and its documentation: https://github.com/openstax/tutor-server/issues/406

PR #457, opened 2015-07-08
Perf report fix

philschatz:
Performance report was not loading for 2 reasons:

- an assumption was made on the initial id (set to `"1"` but rarely true)
- the CRUD store assumed stored things were objects, not Arrays


pandafulmanda:
:+1:

PR #462, opened 2015-07-08
learning guide - updated design/ui, links, expanding panels

darrenmason:
image

### DIFF
--- a/api/courses/1/guide.json
+++ b/api/courses/1/guide.json
@@ -1,124 +1,449 @@
 {
-"title": "Physics",
-"page_ids": [ "234", "345", "456" ],
-
-"children": [
-
-  {
-    "id": "123",
-    "title": "Kinematics",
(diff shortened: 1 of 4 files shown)


Fredasaurus:
Looking good, almost there....

philschatz:
It seems the buttons do not go to a new set of practice problems

philschatz:
hmm, looks like there are some floating issues with long chapter titles:

image

Fredasaurus:
Here is how I handle that. Note the two line title. If it is any longer we can truncate.
screen shot 2015-07-14 at 5 34 32 pm
http://mockups.openstax.org/Tutor_MVP_beta_release/start.html#p=learning_guide_v2

darrenmason:
ill adjust for 2 lines and the truncation width just needs a few more px width.

darrenmason:
image

PR #467, opened 2015-07-09
handle when not all periods are assigned a task plan

pandafulmanda:
Course nav bar will only show as many stats as available on plan stats

There may be some backend or some frontend issues during plan creation for the full bug to be fixed, but this will work once the stats json only returns for as many periods are assigned the plan

screen shot 2015-07-09 at 3 25 07 pm


philschatz:
it also works in the review class aggregate review page!

image

PR #468, opened 2015-07-09
Handle iReading and reference book links

pandafulmanda:
Within page

## Link text replacement, iReadings and Reference Book
screen shot 2015-07-10 at 6 56 08 pm

## Highlight Target Area
screen shot 2015-07-13 at 10 48 11 am
1.5 second yellow background, fades out to transparent

## External Link
screen shot 2015-07-13 at 1 07 46 pm

## Link from iReading to figure in reference book not in reading
screen shot 2015-07-13 at 1 21 43 pm
linked as expected

Merged after #472


Fredasaurus:
Looks good - two things:
Can we adjust the color to look more highlightery?
Can there be padding around the highlighted element (20px) without affecting the rest of the layout? (this may be difficult / not possible)

kathi-fletcher:
Is this highlighting the Figure when it is linked to make it clear which Figure is intended? If so that is really, really a nice idea since we don't have linked numbering.

philschatz:
Yep, that's the idea!

src/components/book-content-mixin.cjsx
```diff
@@ -29,6 +33,36 @@ module.exports =
...
+    # form media tag text based on tag name or data-type
+    tag = media.tagName
+    tag = media.dataset.type if media.dataset.type?
+
+    # capitalize
+    # TODO find one place for all such string formating helpers to go
+    tag = tag.charAt(0).toUpperCase() + tag.substring(1).toLowerCase()
+
+  isMediaLink: (link) ->
+    link.innerText is '[link]' and link.hash.length > 0 and link.hash.search('/') is -1
```
philschatz: Should this just check for `#` being the 1st char? That way links to other content (`[UUID]#id-123`) are not matched here

pandafulmanda: good point, will do

pandafulmanda: oh wait, for href, they all will have the UUID, we will have to do more detailed matching for the different cases, working on that now

philschatz: Right. I think the 4 cases are:

1. `#id123` and `id123` is on the page (link within current step)
1. `#id123` and `id123` is not on the page (link to ref book but use the `page_id` in new tab)
1. `[UUID][@VER]#id123` links to ref book in new tab
1. `https?://...` external link in new tab

nathanstitt: Would be nice for case # 3 if we didn't open a new tab if it's already being viewed in the reference book.  I think we could do so by checking the router's context for the existence of `pageId`

philschatz: or have an optional method on the mixin to decide whether to make the link `_blank` or not (yes for iReading and no for ref book)?

src/components/book-content-mixin.cjsx
```diff
@@ -29,6 +33,57 @@ module.exports =
...
+        mediaLink.innerText = tag if tag?
+      else
+        # The link is external to this page
+        mediaLink.target = '_blank'
+
+        # If not already a reference book page,
+        # make link to a reference book page.
+        # Assumes same/full page of current reading for now.
+        # TODO - when module attr is added, link to exact page if different
+        unless @constructor.displayName is 'ReferenceBookPage'
```
philschatz: eep! Could this be a method on the mixin? like `shouldOpenNewTab`?

src/components/book-content-mixin.cjsx
```diff
@@ -29,6 +33,78 @@ module.exports =
...
+  isMediaLink: (link) ->
+    link.innerText is '[link]' and link.hash.length > 0 and link.hash.search('/') is -1
+
+  hasCNXId: (link) ->
+    trueHref = link.getAttribute('href')
+    link.hash.length > 0 and trueHref.substr(0, 1) isnt '#'
+
+  shouldOpenNewTab: ->
+    # Make link external only if this is not already a ref book page.
+    @constructor.displayName isnt 'ReferenceBookPage'
```
philschatz: Why rely on the `displayName` instead of `ReferenceBookPage::shouldOpenNewTab: -> true`?

pandafulmanda: ah good point okay, ill make that update

pandafulmanda:
@philschatz updated and merged with master

resources/styles/book-content/target.less
```diff
@@ -0,0 +1,61 @@
...
+@-webkit-keyframes highlight {
+  0% {
+    background-color: @tutor-highlight;
+  }
+  100% {
+    background-color: transparent;
+  }
+}
+
+@keyframes highlight {
```
philschatz: Since keyframes are global can this be something like `targetHighlight` that way it does not conflict or move it? so it is used in the iReading as well as the book reader?

pandafulmanda: yes

src/components/book-content-mixin.cjsx
```diff
@@ -29,6 +33,75 @@ module.exports =
...
+          # that is left out of the iReading.
+          # Assume that the cnxId is this same page.
+          mediaCNXId = @getCNXId()
+
+    # Link media that are not found within DOM.
+    @linkMediaElsewhere(mediaCNXId, mediaLink) if mediaCNXId?
+
+  processLinks: ->
+    root = @getDOMNode()
+    linkSelector = 'a:not(.nav):not([data-type=footnote-number]):not([data-type=footnote-ref])'
```
philschatz: Why exclude footnotes?

pandafulmanda: these are also excluded in webview

src/components/reference-book/page.cjsx
```diff
@@ -33,10 +40,47 @@ module.exports = React.createClass
...
+
+  componentDidMount: ->
+    return unless @hasTargetHash()
+
+    targetEl = @getTargetEl()
+    if targetEl?
+      @scrollToTarget(targetEl)
+      $imagesToLoad = $('img')
+      imagesLoaded = 0
+      $imagesToLoad.load(=>
```
philschatz: Can you add the load listener manually to each image (and remove jQuery)?

Something like:

```coffee
images = @getDOMNode().querySelectorAll('img')
imagesToLoad = images.length
onLoad = ->
  imagesToLoad -= 1
  if imagesToLoad is 0
    ...
for img in images
  img.addEventListener('load', onLoad)
```

pandafulmanda: okay, i guess we're ignoring IE everywhere else anyway

philschatz: I believe that works in IE9+

pandafulmanda: yeah i prefer no jquery, im not sure why i opted for it in this situation

PR #472, opened 2015-07-13
Slide out menu for reference book

nathanstitt:
This changes the drop down menu on the reference book to slide in from the side.

The icon on the top navbar toggles depending on the menu's state.

The menu behaves in two different ways depending on the screen width.   1350pixels wide was chosen as the break point since the page width is 1,000px and the menu is 300.

If the page is narrower than 1350px, the menu slides over the page content and is dismissed once a link on it is clicked.

screen shot 2015-07-13 at 4 32 11 pm

If the screen is wider than 1350px appears on the left and pushes the page content and the "previous page" control to the right.  It remains visible when links on it are clicked.

screen shot 2015-07-13 at 4 28 22 pm

### DIFF
--- a/src/components/reference-book/reference-book.cjsx
+++ b/src/components/reference-book/reference-book.cjsx
@@ -5,27 +5,42 @@ _  = require 'underscore'
 
 BindStoreMixin = require '../bind-store-mixin'
 NavBar = require './navbar'
+Menu = require './slide-out-menu'
 
 {CourseListingStore} = require '../../flux/course-listing'
 {CourseStore} = require '../../flux/course'
 
 module.exports = React.createClass
   displayName: 'ReferenceBook'
+  contextTypes:
+    router: React.PropTypes.func
 
-  mixins: [Router.State, BindStoreMixin]
+  mixins: [BindStoreMixin]
   bindStore: CourseListingStore
   bindEvent: 'loaded'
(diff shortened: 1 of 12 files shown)


pandafulmanda:
:sunglasses:

nathanstitt:
Oh and this also styles the teacher dashboard since they're driven from the same less mixin:
screen shot 2015-07-14 at 10 34 05 am

src/components/reference-book/reference-book.cjsx
```diff
@@ -5,27 +5,42 @@ _  = require 'underscore'
...
+  getInitialState: ->
+    {cnxId} = @context.router.getCurrentParams()
+    # Pop open the menu unless the page was explicitly navigated to
+    {isMenuVisible: not cnxId}
+
   showTeacherEdition: ->
     @setState(showTeacherEdition: not @state.showTeacherEdition)
 
+  onMenuClick: ->
+    @toggleMenuState() unless window.innerWidth > 1350
```
philschatz: Can you make this a const on top of the file and add a searchable string to find the corresponding values in the LESS file?

nathanstitt: good idea, updating now.  I'll also add a note in the less file referencing the const

PR #473, opened 2015-07-14
Dashboard biology styles

nathanstitt:
Todo: 
 * [x] make browse the book be on it's own panel
 * [x] Style physics browse the book thumbnail

screen shot 2015-07-14 at 10 10 06 am
screen shot 2015-07-14 at 10 08 13 am


nathanstitt:
@Fredasaurus I took the background from the updated mockup.  I think that works well, but please confirm.   If you'd like the font to be "Helvetica Neue" we'll have to make it be an SVG with the text converted to a path since that font isn't included in all browsers/operating systems.  There's also some reports of good luck using Liberation Sans or Roboto, which we could include and use. http://graphicdesign.stackexchange.com/questions/15462/is-there-a-free-helvetica-neue-alternative/15464#15464

nathanstitt:
@Fredasaurus The panels didn't stand out from the background very much, so I've added a light grey border around it and the "How am I doing" panel (which is scheduled for reworking).

nathanstitt:
This also styles the teacher dashboard since they both use the same less mixin: 
screen shot 2015-07-14 at 10 34 05 am

Fredasaurus:
Definitely needs to be Helvetica Neue. I can provide an svg.
Also, the background doodles have been heavily compressed by Axure. I'll send a better file, or I can replace in git, your choice.

nathanstitt:
@Fredasaurus, either works. Doesn't matter to me at all.  Send it along with the SVG if that's easier.

Fredasaurus:
Oh and on the left panels -- can we give them a box shadow similar to the main panel? That should eliminate the need for the outline.

Fredasaurus:
by left I mean right.

PR #475, opened 2015-07-14
A few calendar bugs

pandafulmanda:
## Bug, overlap due to mis-sorting
screen shot 2015-07-14 at 2 23 28 pm

## Fixed
screen shot 2015-07-14 at 2 23 07 pm

## Bug, height of days wrong when switching from a different month, occasional bug
screen shot 2015-07-14 at 3 05 40 pm

## Fixed
screen shot 2015-07-14 at 3 04 57 pm

### DIFF
--- a/resources/styles/components/course-calendar/month.less
+++ b/resources/styles/components/course-calendar/month.less
@@ -108,7 +108,7 @@
       }
       &-days {
         width: 100%;
-        height: inherit;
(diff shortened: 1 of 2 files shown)


philschatz:
Hmm, I wonder if the overlapping bug is enough reason to not use green for external assignments (since the overlap looks green ; )

PR #476, opened 2015-07-14
Book content styles

nathanstitt:
Mainly biology related but fixes a few bugs also.  Namely it:

  * Adds uniform padding to the book content container element vs the individual elements setting their margins.
  * Fixes the intro image overlay from not appearing on chapter_sections.
  * Adds highlighting to the current page being displayed on the table of contents

screen shot 2015-07-14 at 8 50 38 pm

screen shot 2015-07-14 at 8 49 30 pm

This is an interesting one.  Should the teacher always have content, or should I not style the misconception if it immediately follows `ost-teacher` ?

screen shot 2015-07-14 at 8 54 31 pm

### DIFF
--- a/src/components/reference-book/page.cjsx
+++ b/src/components/reference-book/page.cjsx
@@ -23,8 +23,9 @@ module.exports = React.createClass
     @props.cnxId or @context.router.getCurrentParams().cnxId
 
   getSplashTitle: ->
-    {cnxId} = @context.router.getCurrentParams()
+    {cnxId} = @props.cnxId or @context.router.getCurrentParams().cnxId
     page = ReferenceBookStore.getPageInfo(@context.router.getCurrentParams())
+    page = ReferenceBookStore.getPageInfo({courseId: @props.courseId, cnxId})
     page?.title
 
   prevLink: (info) ->
--- a/src/components/reference-book/page.cjsx
+++ b/src/components/reference-book/page.cjsx
@@ -19,12 +19,11 @@ module.exports = React.createClass
     router: React.PropTypes.func
   mixins: [BookContentMixin, GetPositionMixin]
 
-  getCNXId: ->
+  getCnxId: ->
     @props.cnxId or @context.router.getCurrentParams().cnxId
 
   getSplashTitle: ->
-    {cnxId} = @props.cnxId or @context.router.getCurrentParams().cnxId
-    page = ReferenceBookStore.getPageInfo(@context.router.getCurrentParams())
+    cnxId = @getCnxId()
     page = ReferenceBookStore.getPageInfo({courseId: @props.courseId, cnxId})
     page?.title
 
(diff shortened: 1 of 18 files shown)


Fredasaurus:
can you do the same for the reference view? How long will that take?

nathanstitt:
@Fredasaurus since it uses the common book-content styles, it is applied to both iReading and reference view.

Fredasaurus:
Ah, it just hasn't been merged yet, never mind, thanks. @philschatz Can we get this one merged? Helene is waiting on this.

nathanstitt:
@philschatz, yes we can go ahead and merge this.  Items remaining can be addressed in a follow up PR, but this has a good base to make the Biology content much less "yucky".

Things remaining are:
  * prototype blue bar at top
  * Removing or reword "Intro" 
   * move arrows to be next to content on reference view

src/components/reference-book/page.cjsx
```diff
@@ -81,7 +81,7 @@ module.exports = React.createClass
   render: ->
     {courseId} = @context.router.getCurrentParams()
     # read the id from props, or failing that the url
-    cnxId = @getCNXId()
+    cnxId = @getCnxId()
```
philschatz: Aside: I think we should have a quick Discussion about this. Since the letters do not stand for anything I think this is the correct casing of the word `cnx`

Fredasaurus:
Looks good. Can we:

Adjust the margin of the splash caption to be 20px off the container
screen shot 2015-07-15 at 10 01 51 am
screen shot 2015-07-15 at 10 02 38 am

Get the book font same as in the interactive reading
screen shot 2015-07-15 at 10 05 08 am
screen shot 2015-07-15 at 10 05 12 am

Find out why text isn't wrapping around some images? 
screen shot 2015-07-15 at 10 07 44 am

thanks.

src/components/reference-book/page.cjsx
nathanstitt: Hm, I thought it looked to much like a constant with all the uppercase letters and the "Id" didn't have enough visual segregation from the CNX.  The method is only used internally though so it's naming isn't terribly important.  I can revert

philschatz: Oops, I meant that `Cnx` was correct since `CNX` does not stand for anything else.

nathanstitt:
@Fredasaurus I've started on implementing these in https://github.com/openstax/tutor-js/pull/479

I'm just a bit unclear on the "Adjust the margin of the splash caption to be 20px off the container".  I'm guessing that it refers to the amount of space that the blue bar extends around the text?  That's 30px now, and I'll slim it down to 20px.  If you're referring to something else, let me know.

Fredasaurus:
Ah, talking about the sides -- it's padding actually I think -- of tha caption below the image.

screen shot 2015-07-15 at 1 16 37 pm

versus
screen shot 2015-07-15 at 1 17 11 pm

Fredasaurus:
Blue bar is fine.

PR #479, opened 2015-07-15
Additional reading and reference book styles

nathanstitt:
Adds Rice University logo onto navbar:

screen shot 2015-07-15 at 5 02 30 pm

Stretches the caption to full width so it matches splash image:
screen shot 2015-07-15 at 1 29 53 pm

Wraps text around vertical images, even if they're inside a section:
screen shot 2015-07-15 at 1 32 24 pm

CSS arrows that grow and shrink with page margin size.  Note that this page should also have the "Connection for AP Courses" title, but it's hidden:

screen shot 2015-07-15 at 3 49 30 pm


nathanstitt:
@Fredasaurus I think this is the last of the style changes, but it's quite possible I've missed some items.

Something I haven't found an implementation for is how to "un-scrunching" the green box at top.  Do you have any ideas on how to improve that?  It's difficult because anything I do to compact it make the smaller versions of it look worse.  And since most of the bullets have only 2-3 items, it seems like we should give priority to making them look the best.

resources/styles/book-content/ap-connection.less
```diff
@@ -0,0 +1,9 @@
+.ap-connection {
+  > h1:first-child {
+    // The "Connection for AP Courses" message appears on every page and is redundant.
+    // although we don't want it to display, it still should take up space in order to provide a top margin
+    height: 0;
+    margin: 30px;
+    visibility: hidden;
```
philschatz: @Fredasaurus Is this what was decided in the discussion? My understanding was that it was suggested but then kept in

resources/styles/components/reference-book/page-navigation-controls.less
```diff
@@ -0,0 +1,84 @@
+.reference-book {
+  .page-wrapper > .nav {
+    position: fixed;
+    height: 100%;
+    width: calc(~"50% - " @reference-book-page-width / 2 );
+    top: 50px;
+
+    display: -webkit-flex;
```
philschatz: Hmm, I wonder if it might be time to look at http://lesshat.madebysource.com/ rather than recreate these all over ; ) (great link from @pandafulmanda )

resources/styles/book-content/ap-connection.less
nathanstitt: Yes, I know we discussed it quite a bit, but thought we'd ultimately decided to just hide the heading.  The note don't mention it at all :frowning:

resources/styles/components/reference-book/page-navigation-controls.less
nathanstitt: Lesshat has some neat stuff, but in my opinion the `calc`  mixin doesn't read any better than just using `calc` directly:  https://github.com/madebysource/lesshat/blob/master/README.md#calc

And since `calc` is supported by all of our target browsers (http://caniuse.com/#search=calc) I don't see the reason for the vendor prefixes.

I was hoping lesshat would have support for css triangles but looks like it doesn't

Fredasaurus:
@nathanstitt I'll take a look at the green box and see what we can do.

PR #480, opened 2015-07-15
style adjustments on homework review

pandafulmanda:
- [x] Should be same font for correct and incorrect options
  - [x] font weight still seem slightly different somehow, FIXED YAY.
- [x] less padding on question stem
- [x] top-align MC options
- [x] use button style for hide/show responses
- [x] make active concept same font size

screen shot 2015-07-16 at 11 57 31 am

## Expects Free Response
screen shot 2015-07-15 at 6 20 50 pm

## Multiple choice only
screen shot 2015-07-15 at 6 20 46 pm

### DIFF
--- a/resources/styles/components/task-teacher-review/question.less
+++ b/resources/styles/components/task-teacher-review/question.less
@@ -1,20 +1,22 @@
 .question {
 
-  font-size: 1.8rem;
-  line-height: 2.4rem;
+  #fonts > .sans(1.8rem, 2.4rem);
 
   .answers-answer {
+    #fonts > .sans(1.8rem, 2.4rem);
+
     &::before {
       .answer-label();
       .answer-icon-label();
       text-align: center;
       width: 80px;
-      font-size: 1.5em;
+      font-size: 2rem;
+      margin-top: 14px;
     }
 
     .selected-count {
-      display: inline;
-      #fonts > .sans(1.8rem, 1.8rem);
+      display: inline-block;
+      #fonts > .sans(1.8rem, 2.4rem);
 
       &::before {
         content: attr(data-count);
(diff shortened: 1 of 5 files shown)


pandafulmanda:
@Fredasaurus any recommendations on what "No student text responses" should look like?  we could also just completely leave it out

philschatz:
I don't have an eye for these things but I think it depends on if we can distinguish multiple choice questions (which have free responses) from True/False questions (which should not). If we can, then it's nice to see that there is data missing

pandafulmanda:
it seems like in that case, the problem just hasn't been answered at all

resources/styles/components/task-teacher-review/question.less
```diff
@@ -1,16 +1,22 @@
 .question {
 
+  #fonts > .sans(1.8rem, 2.4rem);
+
   .answers-answer {
+    #fonts > .sans(1.8rem, 2.4rem);
```
philschatz: Why aren't these cascading down? Something in bootstrap?

resources/styles/old/global/buttons.less
```diff
@@ -7,18 +7,44 @@
...
+  border-radius: @btn-border-radius;
+  transition: box-shadow 0.2s, background 0.2s;
+
+  &:hover {
+    box-shadow: 0 4px 10px rgba(0, 0, 0, .1), 0 6px 28px rgba(0, 0, 0, .2);
+  }
+}
+
+.btn-tutor-default() {
+  color: #424242 !important; //!important to overide local styles in plan-details.less. we can remove after integration
```
philschatz: Is the CSS in `plan-details` not scoped to be in a `plan`? And any reason this color isn't with the other colors in `variables.less`?

PR #484, opened 2015-07-17
set today for default opensAt

philschatz:
No UI Changes.

Allows the "Publish" button to work when the default "Opens At" (today) is set.

Thoughts @jivewise ?


philschatz:
closed in favor of #485

PR #485, opened 2015-07-17
Default Open Date

jivewise:
- [x] Making default open date for today
- [x] Modified datepicker formatting
- [x] Removed default due dates

### DIFF
--- a/src/components/tutor-input.cjsx
+++ b/src/components/tutor-input.cjsx
@@ -88,6 +88,8 @@ TutorDateInput = React.createClass
     if (not valid)
       value = @props.min or null
 
+    offset = (new Date()).getTimezoneOffset() / -60
+    value = value.add(offset, 'hours')
     date = new Date(value)
     @props.onChange(date)
     @setState({expandCalendar: false, valid: valid, value: date})
--- a/src/components/tutor-input.cjsx
+++ b/src/components/tutor-input.cjsx
@@ -6,6 +6,7 @@ _ = require 'underscore'
 {TimeStore} = require '../flux/time'
 DatePicker = require 'react-datepicker'
 TutorErrors = require './tutor-errors'
+TutorDateFormat = "MM/DD/YYYY"
 
 TutorInput = React.createClass
   propTypes:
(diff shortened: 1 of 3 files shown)


philschatz:
Now, when I click a date in the calendar and select a new iReading the dueAt is not being populated:

image

philschatz:
A few other issues I noticed:

1. RE "Made tutor datepicker have an offset to get rid of timezones." I think we should just drop the timezone when sending to the BE
2. the PR uses `new Date()` instead of `TimeStore.getNow()`
3. when I select 7/24 in the datepicker, the datepicker shows 7/23

src/flux/task-plan.coffee
```diff
@@ -118,7 +110,8 @@ TaskPlanConfig =
     {tasking_plans} = @_getPlan(id)
     # do all the tasking_plans have the same date?
     dates = _.compact _.uniq _.map(tasking_plans, (plan) ->
-      plan[attr]?.getTime()
+      date = new Date(plan[attr])
+      date?.getTime()
     )
```
nathanstitt: Date will always be defined here, so the `?` has no effect.  If a Date object is created with an invalid value it still creates the `Date` object, but it's value is set to `Invalid Date`.  One method to detect if the date is invalid is to use `isNaN`  To use that method you could write line 114 as: `if isNaN(d.getTime()) then 0 else d.getTime()`

nathanstitt:
Looks good to me @jivewise. I pulled down the branch and it seems to work well.

src/components/tutor-input.cjsx
```diff
@@ -141,7 +141,7 @@ TutorDateInput = React.createClass
           weekStart={0}
         />
     else if @props.disabled and value
-      displayValue = value.toString("YYYY/MM/DD")
+      displayValue = value.toString("MM/DD/YYYY")
```
pandafulmanda: Could "MM/DD/YYYY" be set as a default prop or something throughout to help with potential future date formatting changes?

nathanstitt: Also, I'm guessing the spec or mockup has the dates formatted that way?   I personally dislike that format because you can't readily  tell what 1/2/2016 means.

jivewise: I had a ticket in pivotal that said to change it to this format.  I personally prefer YYYY/MM/DD as well.  I will pull this out and put it as a constant somewhere though...

src/flux/task-plan.coffee
jivewise: Good point.

PR #491, opened 2015-07-18
Dashboard hw status icons

nathanstitt:
screen shot 2015-07-18 at 8 04 15 pm

### DIFF
--- a/resources/styles/components/student-dashboard/task.less
+++ b/resources/styles/components/student-dashboard/task.less
@@ -22,7 +22,7 @@
     }
     i.info {
       .fa-icon();
-      margin-left: 1rem;
+      margin-left: 0.5rem;
       &.incomplete::before {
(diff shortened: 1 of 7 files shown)


nathanstitt:
@Fredasaurus If you like these tooltips, we could also use them on the learning guide as well.

philschatz:
Travis is failing ; (

nathanstitt:
That did it.  I'd removed a require statement that it needed on the hw.  Weird that the error was an invariant one though

philschatz:
It's still failing ; (

nathanstitt:
Ok, Finally.  I think the extra space that the new icons where inserting into the DOM was causing the specs to be non-deterministic.   I also rebased from master since it conflicted with the other PR to not work unopened tasks.

Fredasaurus:
Looks good.

PR #492, opened 2015-07-18
Events that are un-opened should not be worked

nathanstitt:
Since the student dashboard shouldn't know about them at all, just set the store's methods to filter them out.

### DIFF
--- a/api/courses/1/dashboard.json
+++ b/api/courses/1/dashboard.json
@@ -23,7 +23,7 @@
         {
             "id": "7",
             "title": "iReading 3: Newton's Second Law of Motion:",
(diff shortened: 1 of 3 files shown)


nathanstitt:
Looks like I've mis-interpretted this.  https://github.com/openstax/tutor-server/issues/383#issuecomment-122428061  I'll rework to display them, but not allow clicking.

nathanstitt:
@Fredasaurus I think we'll need some visual distinction for the unopened tasks so the student will know that they can't click on them.  Perhaps a :lock: icon beside the title like we're doing with the late indicators?

philschatz:
Code looks fine to me. We can add styling as a separate PR

PR #494, opened 2015-07-19
Teacher views learning guide for a student

nathanstitt:
screen shot 2015-07-19 at 8 20 52 pm


src/components/loadable-item.cjsx
```diff
@@ -13,6 +13,7 @@ module.exports = React.createClass
   displayName: 'LoadableItem'
   propTypes:
     id: React.PropTypes.string.isRequired
+    options: React.PropTypes.object
     store: React.PropTypes.object.isRequired
```
nathanstitt: The `LoadableItem` component is extended with an `options` property.  This allows a user of the component to pass additional options to the store methods that are called by `LoadableItem`.  In the case of this PR, the store needs two id's for the endpoint, `courseId` and `roleId`.  The `roleId` is passed to the store in the options and from there to `api.coffee`

src/flux/learning-guide.coffee
```diff
@@ -0,0 +1,98 @@
...
+    getAllSections: (courseId, {roleId}) ->
+      findAllSections(@_local[courseId]?[roleId] or {})
+
+    reload: (id, {roleId}) ->
+      @_reload[id]?[roleId]
+
+}, new CrudConfig
+
+
+module.exports = {Student, Teacher, TeacherStudent}
```
nathanstitt: This ended up longer than I'd hoped, mainly because it had to override so many of the helper methods.  It's almost to the point where I considered either breaking out the `TeacherStudent `store into it's own file, or possibly creating a new `multi-option` base that it could extend from.

src/components/loadable-item.cjsx
philschatz: Ahh, yep. Thanks for adding a comment!

src/flux/learning-guide.coffee
philschatz: I wonder if there's any benefit from extending `CrudConfig`

PR #495, opened 2015-07-20
restore locale settings when done with react-datepicker.

pandafulmanda:
Otherwise, moment locale will be messed up and the dashboard cal suffers as a result, for example.

react-datepicker [sets global locale](https://github.com/Hacker0x01/react-datepicker/blob/master/src/calendar.jsx#L44-L49) and causes moment to be confused

# Minimum steps to replicate bug
1. load builder
screen shot 2015-07-20 at 3 18 59 pm
2. cancel
screen shot 2015-07-20 at 3 19 05 pm
3. the dashboard is :cry: 
screen shot 2015-07-20 at 3 19 12 pm

# What happens now
screen shot 2015-07-20 at 3 21 59 pm
screen shot 2015-07-20 at 3 22 03 pm
3. YAY!
screen shot 2015-07-20 at 3 22 09 pm

### DIFF
--- a/src/components/course-calendar/month.cjsx
+++ b/src/components/course-calendar/month.cjsx
@@ -122,7 +122,8 @@ CourseMonth = React.createClass
           otherProps.classes =
             active: true
 
-      day = <Day date={dayIter} modifiers={modifiers} {...otherProps}/>
+      key = "day-#{dayIter.format(@props.dateFormat)}"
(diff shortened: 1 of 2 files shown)


nathanstitt:
Wow, nice catch!

pandafulmanda:
thanks! :)

PR #496, opened 2015-07-20
course data can pass title and category onto elements

pandafulmanda:
For iReading, reference book, calendar, and dashboard as data-attributes -- ```data-title``` and ```data-category```, following pattern set up by @nathanstitt for dashboard and calendar.

So that theming for content can happen following the same pattern for book-content.  As per discussion with @helenemccarron, this should allow us to do our theming something like this:

```less
// A mixin to style book content from CNX
.tutor-book-content() {
  @reading-ui-banner-height: 40px;
  padding: 0 @tutor-card-body-padding-horizontal @tutor-card-body-padding-vertical @tutor-card-body-padding-horizontal;
  display: block;
  @import './mixins';
  @import './base';
  @import './splash-image';
  // type specific styling comes last so it can override the above styles
  @import './link-to-learning';
  @import './worked-examples';
  @import './note';
  @import './grasp-check';
  @import './target';
  @import './visual-connection';
}

.tutor-book-category-physics() {
  // maybe subject specific vars here

  @import './links-to-physics';
  @import './virtual-physics';
  @import './fun-in-physics';
  @import './work-in-physics';
  @import './watch-physics';
}

.tutor-book-category-physics() {
  // maybe subject specific vars here

  @import './links-to-biology';
  @import './virtual-biology';
  @import './fun-in-biology';
  @import './work-in-biology';
  @import './watch-biology';
}

.reference-book {
  .tutor-book-content();

  &[data-category=physics] {
    .tutor-book-category-physics();
  }
  &[data-category=biology] {
    .tutor-book-category-biology();
  }
}
```

### DIFF
--- a/src/components/course-data-mixin.cjsx
+++ b/src/components/course-data-mixin.cjsx
@@ -1,0 +1,10 @@
+{CourseStore} = require '../flux/course'
+
+module.exports =
+  getCourseDataProps: (courseId) ->
+    unless courseId?
+      {courseId} = @context.router.getCurrentParams()
+
+    dataProps =
+      'data-title': CourseStore.getShortName(courseId)
+      'data-category': CourseStore.getCategory(courseId)
--- a/src/components/html.cjsx
+++ b/src/components/html.cjsx
@@ -17,10 +17,18 @@ module.exports = React.createClass
     classes.push(@props.className) if @props.className
     classes = classes.join(' ')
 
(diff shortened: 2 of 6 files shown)


nathanstitt:
Nice, that makes total sense.  Will hopefully keep things manageable once additional books are added.

I think the ultimate plan is that the "short name" and category will come from the BE as metadata, but we haven't locked that down yet.

PR #497, opened 2015-07-21
A few student dashboard style tweaks

nathanstitt:
Removes the default color from the booksplash, which was set to the physics.  I think this is the cause of the "shows physics colors, then rapidly switches to biology on load" bug.

Also fixes spacing on a 1024 wide display:
Before/after

screen shot 2015-07-21 at 9 36 36 am

screen shot 2015-07-21 at 9 35 50 am

### DIFF
--- a/resources/styles/global/tutor-booksplash-background.less
+++ b/resources/styles/global/tutor-booksplash-background.less
@@ -15,7 +15,6 @@
     right: 0px;
     height: 24rem;
     overflow: hidden;
-    background-color: @tutor-tertiary;
     color: @tutor-secondary;
     text-align: center;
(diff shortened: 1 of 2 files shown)


resources/styles/global/tutor-booksplash-background.less
```diff
@@ -15,7 +15,6 @@
     right: 0px;
     height: 24rem;
     overflow: hidden;
-    background-color: @tutor-tertiary;
     color: @tutor-secondary;
```
pandafulmanda: :+1:   Also I bet a transition on the background color here would be nice =)

PR #502, opened 2015-07-22
fix the builder's confusion around the sticky header

pandafulmanda:
was causing homework builder to have errors and get stuck at problems selection.

## Bug
screen shot 2015-07-21 at 7 27 39 pm

## Fixed
screen shot 2015-07-21 at 7 27 02 pm


Fredasaurus:
Hey while we are at it.... Can we get rid of the last vertical diver and stretch out the columns? We only need five now.

pandafulmanda:
I think I'll do that in a different PR -- without this PR things are pretty broken so it's urgent for merging in

PR #498, opened 2015-07-21
step lateness in footer

pandafulmanda:
screen shot 2015-07-22 at 10 14 05 pm
screen shot 2015-07-22 at 10 14 39 pm

recommended for merging after #503

next would be to see about consolidating the late icon on the dashboard with the component version here

PR #503, opened 2015-07-22
Performance report updates. Cell types, display First/Last name

nathanstitt:
This:
 * Display late icon and message when task is late
 * Add first and last name columns
 * sets columns to expand to fit available space. 
 * Fixes the header for external assignments to show the 'X of X clicked' message.

The late icon and extern assignments header parts depend on BE PR: https://github.com/openstax/tutor-server/pull/502
screen shot 2015-07-22 at 3 49 15 pm

### DIFF
--- a/src/components/performance/cell-status-mixin.cjsx
+++ b/src/components/performance/cell-status-mixin.cjsx
@@ -1,0 +1,32 @@
+React = require 'react'
+Router = require 'react-router'
+LateIcon = require './late-icon'
+
+module.exports = {
+
+  propTypes:
+    courseId: React.PropTypes.string.isRequired
+
+    task: React.PropTypes.shape(
+      status:          React.PropTypes.string
+      due_at:          React.PropTypes.string
+      last_worked_at:  React.PropTypes.string
+      type:            React.PropTypes.string
+    ).isRequired
+
+    student: React.PropTypes.shape(
+      name: React.PropTypes.string
+      role: React.PropTypes.number
+    ).isRequired
+
+  isLate: ->
+    new Date(@props.task.last_worked_at) > new Date(@props.task.due_at)
+
+  render: ->
+    <Router.Link className={"task-result #{@props.task.type}"} to='viewTaskStep'
+      params={courseId: @props.courseId, id: @props.task.id, stepIndex: 1}>
+      <span>{@statusMessage()}</span>
+      {<LateIcon {...@props}/> if @isLate()}
+    </Router.Link>
+
+}
--- a/src/components/performance/cell-status-mixin.cjsx
+++ b/src/components/performance/cell-status-mixin.cjsx
@@ -20,6 +20,9 @@ module.exports = {
     ).isRequired
 
   isLate: ->
+    # if the task was never worked then it's not late, it's incomplete
+    return false unless @props.task.last_worked_at
+
     new Date(@props.task.last_worked_at) > new Date(@props.task.due_at)
 
   renderLink: ({message}) ->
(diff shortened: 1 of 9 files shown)


nathanstitt:
Needs a bit of code cleanup and depends on BE PR https://github.com/openstax/tutor-server/pull/502

nathanstitt:
@pandafulmanda perhaps you can double check to see if the logic I'm using is correct?

pandafulmanda:
Looks good to me, I think. I'm liking the refactoring/pulling out the cells as their own components and that shape thing is cool.

src/components/performance/homework-cell.cjsx
```diff
@@ -0,0 +1,16 @@
...
+
+module.exports = React.createClass
+  displayName: 'HomeworkCell'
+  mixins: [CellStatusMixin] # handles rendering
+
+  render: ->
+    message = if @isLate()
+      'Incomplete'
+    else
+    message =  "#{@props.task.correct_exercise_count}/#{@props.task.exercise_count}"
```
pandafulmanda: does the else run if this is indented this way?

src/components/performance/cell-status-mixin.cjsx
```diff
@@ -0,0 +1,35 @@
...
+      name: React.PropTypes.string
+      role: React.PropTypes.number
+    ).isRequired
+
+  isLate: ->
+    # if the task was never worked then it's not late, it's incomplete
+    return false unless @props.task.last_worked_at
+
+    new Date(@props.task.last_worked_at) > new Date(@props.task.due_at)
+
```
pandafulmanda: Some late checking logic here as here (I want to take the homework out of it).  https://github.com/openstax/tutor-js/pull/498/files#diff-74c51a44b8ba99c878804b91182ab691R202

Would be nice to consolidate and either keep using this ```isLate``` or the TaskStore's ```getStepLateness``` but we can probably get to that in a different PR

src/components/performance/late-icon.cjsx
```diff
@@ -0,0 +1,22 @@
...
+    task: React.PropTypes.shape(
+      due_at:          React.PropTypes.string
+      last_worked_at:  React.PropTypes.string
+      type:            React.PropTypes.string
+    ).isRequired
+
+  render: ->
+    due = moment(@props.task.due_at)
+    msg = S.capitalize(@props.task.type) + ' was worked ' + due.from(@props.task.last_worked_at, true) + ' late'
+
```
pandafulmanda: same here as well, consolidate with ```getStepLateness``` ?

pandafulmanda:
hmm, something happened to the external task heading, need to fix

src/components/performance/late-icon.cjsx
nathanstitt: Hm.  How about an exported method(s) in a `helpers/task` file?  That way both could share it?
It would take an object with `type`, `last_worked_at` and `due_at` fields and return the same at the existing `getStepLateness` ?

philschatz:
I am unable to duplicate the "Late" icon. I tried using "Read Chapter 6. Metabolism" since some students said "In Progress" in the perf report but if I logged in as them and answered a question I did not see the late icon show up.

Is there a better way to trigger the late icon in the perf report?

philschatz:
looks great otherwise ; )

philschatz:
actually, the code shows up in #498 . I did notice that every task is late. Is that a coincidence?

image

nathanstitt:
Yeah, I just noticed that the latest does seem to mark too many as late.  Debugging now

nathanstitt:
The problem turns out to be on the demo importer.  When it works tasks it works them using the current date/time, regardless of when the assignment closed.  Therefore they're all marked past due.

Fix is in: https://github.com/openstax/tutor-server/pull/503

PR #505, opened 2015-07-22
Add a Task helper with a getLateness method

nathanstitt:
This way the logic can be shared between components and stores as needed


philschatz:
Is this code used or is it just for future use?

nathanstitt:
It's something that @pandafulmanda and I were discussing.  I've already pulled the commit into #503 but made a PR so she could possibly use it also.

pandafulmanda:
:+1:

PR #509, opened 2015-07-23
Spinners on one button at a time

jivewise:
screen shot 2015-07-23 at 11 40 34 pm

Now the publish button won't show a spinner when saving and the save as draft button won't show a spinner when publishing.  
Though maybe we should disable actions when either are occurring? @philschatz

### DIFF
--- a/src/components/task-plan/footer.cjsx
+++ b/src/components/task-plan/footer.cjsx
@@ -23,6 +23,14 @@ PlanFooter = React.createClass
       TaskPlanActions.delete(id)
       @context.router.transitionTo('taskplans', {courseId})
 
+  onSave: ->
+    @setState({saving: true, publishing: false})
+    @props.onSave()
+
+  onPublish: ->
+    @setState({publishing: true, saving: false})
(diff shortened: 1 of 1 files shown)


pandafulmanda:
oh oops, i was just working on this as lwell

src/components/task-plan/footer.cjsx
```diff
@@ -47,8 +55,8 @@ PlanFooter = React.createClass
         <AsyncButton
           bsStyle='primary'
           className='-publish'
-          onClick={onPublish}
-          isWaiting={isWaiting}
+          onClick={@onPublish}
+          isWaiting={isWaiting and @state.publishing}
```
philschatz: can you use `isSaving` instead of `saving` on `@state`? That way you can do:

`{isSaving, isPublishing} = @state` and then use it below

philschatz:
I am unable to reproduce this because I'm unable to publish an iReading.

philschatz:
Ah, merging master into this branch allows me to publish/save an iReading!

PR #511, opened 2015-07-23
Style/ref book exercises

pandafulmanda:
screen shot 2015-07-23 at 11 25 07 am

Initial taking out blue background of answers on ref book exercises

@Fredasaurus any other quick stylings?  this might be all we can do now without digging into where the exercises show up in the books, and how to style notes, etc


pandafulmanda:
I guess this is ready to merge for now, and we can update as needed later.  Having this change in would look better than to not.

PR #514, opened 2015-07-24
Perf report due date

nathanstitt:
* Adds due dates to the column headers
 * Formats the name columns a little narrower
 * Fixes sorting by names

screen shot 2015-07-22 at 3 49 15 pm

### DIFF
--- a/resources/styles/components/performance/index.less
+++ b/resources/styles/components/performance/index.less
@@ -49,9 +49,6 @@
     .tutor-late-icon();
     padding-left:8px;
   }
-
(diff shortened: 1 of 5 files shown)


philschatz:
For some reason I can't look at Period 2. I get a "Uncaught TypeError: Cannot read property 'type' of undefined" in `switch d.data[index].type` in `performance/index.cjsx`

nathanstitt:
Yeah, that occurred because it was sorted by a column index that didn't exist when the period was changed.  I've updated the code so it now resets the sort whenever the period is changed.

philschatz:
can you merge from master?

PR #515, opened 2015-07-24
Add time to due at date

nathanstitt:
screen shot 2015-07-23 at 7 52 05 pm

### DIFF
--- a/resources/styles/components/student-dashboard/task.less
+++ b/resources/styles/components/student-dashboard/task.less
@@ -1,4 +1,5 @@
 .student-dashboard {
+  .due-at, .due-at-label { padding-left: 0; }
   .task {
     background: @tutor-white;
     border-top: thin dotted @tutor-neutral-light;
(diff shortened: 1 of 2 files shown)


pandafulmanda:
I think @Fredasaurus is going to not like that hehehehe

nathanstitt:
Yeah, the card I was assigned explicitly said to put the time beside the date.  Otherwise I think we could put something like "All assignments are due by 7pm CST" at the bottom of the panel.

philschatz:
I think this is good for now; @Fredasaurus can suggest otherwise.

Fredasaurus:
If all time at 7:00 it does not need to be repeated. And yeah it look awful.

jason-holmes:
Something more like this would be better.  http://mockups.openstax.org/Tutor_MVP_beta_release/start.html#p=johnny_tran_dashboard_-_visual_design_-_1 <http://mockups.openstax.org/Tutor_MVP_beta_release/start.html#p=johnny_tran_dashboard_-_visual_design_-_1>



-------
Jason Holmes, Ph.D.
User Experience Manager
OpenStax | OpenStax.org
(713) 893-1326

PR #516, opened 2015-07-24
Fix "value.format is not a function" bug

### LINKED ISSUE #513: Adding exercises to HW Builder
When selecting today for the open date I get the following error:

image

Thoughts @jivewise or @pandafulmanda ?

nathanstitt:
Fixes #513

The value passed to isValid is a momentjs instance, which
can't be compared using > < operators.  This switches to
the isBefore and isAfer methods with the granularity set to day.

Also if false was returned from the date picker, the min property
would be used.  However if the min property is a plain date object
it would lack the .format method which was later called on it.

### DIFF
--- a/src/components/tutor-input.cjsx
+++ b/src/components/tutor-input.cjsx
@@ -100,15 +100,15 @@ TutorDateInput = React.createClass
 
   isValid: (value) ->
     valid = true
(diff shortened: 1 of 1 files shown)


pandafulmanda:
:+1:

PR #530, opened 2015-07-26
Fix practice buttons that lead to empty practice sets.

nathanstitt:
screen shot 2015-07-27 at 10 27 11 am

When BL doesn't return any pages the button needs to be aware of that and not attempt the practice.

**Note:** This adds a new component rendering helper, and also imports the sinon mocking library.  suggestions are welcome on whether this is a good idea or how it could be improved.  My goals with the helper are to allow testing a component with as little effort as possible.

### DIFF
--- a/src/components/learning-guide/practice-by-type-button.cjsx
+++ b/src/components/learning-guide/practice-by-type-button.cjsx
@@ -5,6 +5,8 @@ _ = require 'underscore'
 LearningGuide = require '../../flux/learning-guide'
 
 module.exports = React.createClass
+  displayName: 'PracticeByTypeButton'
+
   propTypes:
     courseId: React.PropTypes.string.isRequired
     practiceType: React.PropTypes.string.isRequired
--- a/src/components/learning-guide/practice-by-type-button.cjsx
+++ b/src/components/learning-guide/practice-by-type-button.cjsx
@@ -1,6 +1,5 @@
 React = require 'react'
 BS = require 'react-bootstrap'
-S = require '../../helpers/string'
 _ = require 'underscore'
 
 LearningGuide = require '../../flux/learning-guide'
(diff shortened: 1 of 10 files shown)


src/components/learning-guide/practice-by-type-button.cjsx
```diff
@@ -16,13 +15,25 @@ module.exports = React.createClass
 
   onClick: ->
     {courseId} = @props
-    all = LearningGuide.Student.store.getSortedSections(@props.courseId)
-    sections = if @props.practiceType is 'weaker' then _.first(all, 4) else _.last(all, 4)
-    page_ids = _.chain(sections).pluck('page_ids').flatten().uniq().value()
-    @context.router.transitionTo('viewPractice', {courseId}, {page_ids})
+    page_ids = LearningGuide.Student.store.getPracticePages(@props.courseId, @props.practiceType)
+    unless _.isEmpty(page_ids)
```
philschatz: If `page_ids` is empty should the practice button not exist rather than click and have nothing happen?

philschatz: nevermind. I need to read a couple of lines below. maybe this could `throw new Error('BUG: ....')`?

test/components/helpers/component-testing.coffee
```diff
@@ -0,0 +1,59 @@
+_ = require 'underscore'
+chai  = require 'chai'
+sinon = require 'sinon'
+sinonChai = require("sinon-chai")
```
philschatz: Is this boilerplate needed or can it be moved up to the `test/index` file?

philschatz:
Adding `sinon` is fine with me. Thoughts @darrenmason @jivewise @pandafulmanda ?

pandafulmanda:
ah, yes, i've been eyeing that

src/components/learning-guide/practice-by-type-button.cjsx
nathanstitt: I'm not sure that's a reasonable thing to do here.  It seems to me that the page_ids could be empty for non bug related reasons.  For instance, right now on tutor-demo, student01 has only once assignment that is past due and has not been started.  The performance report endpoint is returning an empty set, as it should because there's not any data to report yet.  

Your suggestion about the button's not rendering anything when empty is good, but I'd suggest that check move up into the parent component.  Otherwise we'll end up with a bold "**Practice**" heading and no buttons below it.  The button could still have the check as it does here, just in case the parent doesn't perform the check .

philschatz: While skimming I thought there already was a check in the `render: ->` method but I misread, it just does adds a tooltip.

PR #533, opened 2015-07-26
Remove Add, Delete and PW reset actions

nathanstitt:
WIP until it's decided if the "Tutor ID" column should be removed and if the link to it on the menu should be changed to read "Student Roster" or something else.

screen shot 2015-07-26 at 1 50 58 pm


nathanstitt:
We can merge and I'll adress the remainder in a followup

PR #536, opened 2015-07-26
Move due time to panel headings on student dashboard

nathanstitt:
screen shot 2015-07-26 at 4 42 13 pm

### DIFF
--- a/src/components/student-dashboard/events-panel.cjsx
+++ b/src/components/student-dashboard/events-panel.cjsx
@@ -46,7 +46,7 @@ module.exports = React.createClass
         <BS.Col xs={5} xsOffset={2} smOffset={0} sm={2} className='progress-label'>
           Progress
         </BS.Col>
-        <BS.Col xs={5} sm={2} className='due-at-label'>Due</BS.Col>
+        <BS.Col xs={5} sm={2} className='due-at-label'>Due (7:00am)</BS.Col>
       </div>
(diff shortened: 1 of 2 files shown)


src/components/student-dashboard/events-panel.cjsx
```diff
@@ -46,7 +46,7 @@ module.exports = React.createClass
         <BS.Col xs={5} xsOffset={2} smOffset={0} sm={2} className='progress-label'>
           Progress
         </BS.Col>
-        <BS.Col xs={5} sm={2} className='due-at-label'>Due</BS.Col>
+        <BS.Col xs={5} sm={2} className='due-at-label'>Due (7:00am)</BS.Col>
```
philschatz: Is/should this be available from course settings? Although, it's already hardcoded in the HW builder as well.

nathanstitt: It's not available on the corse settings as a property by itself, but the individual due dates do have time along with the date.  

I suppose we could pluck the time from `due_at` on all all the tasks and only display it on the heading if all the times are all the same.  If they weren't, we'd display them  beside the individual dates as we were previously doing.

Do you think this is worth it at this point, or should we wait until we allow the teacher to set the due time?

philschatz: I think it's fine to wait until the teacher can set the time; just was confirming BE does not yet send that time ; )

PR #537, opened 2015-07-26
Display assignment names with an ellipsis when they're too long for the column

nathanstitt:
Also updates fixed-data-table to 0.4.1, which includes a bug fix for the scrollbars appearing on top of the row headers.

screen shot 2015-07-26 at 6 09 24 pm

### DIFF
--- a/resources/styles/components/performance/index.less
+++ b/resources/styles/components/performance/index.less
@@ -25,6 +25,17 @@
     max-height: 100%;
   }
 
+  // removes the display as table, table-row, & table-cell
+  // from the wrappers so the header-cell below can set
+  // it's width to 100% with text-overflow: ellipsis;
+  .public_fixedDataTable_header {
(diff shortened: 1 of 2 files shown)


resources/styles/components/performance/index.less
```diff
@@ -25,6 +25,17 @@
     max-height: 100%;
   }
 
+  // removes the display as table, table-row, & table-cell
+  // from the wrappers so the header-cell below can set
+  // it's width to 100% with text-overflow: ellipsis;
```
philschatz: I believe there is an extra `'` somewhere here...

nathanstitt: Hrm?  I'm guessing that the ' in "it's" was throwing your syntax highlighting off?  I've removed it and reworded the comment to be a bit more concise.

PR #542, opened 2015-07-27
back button with fallback link option for when there is no browser history

pandafulmanda:
Also gets rid of "Back to Flight Path" bug/text :bug: 
Tests will be in a separate PR

# Practice
## Has history
screen shot 2015-07-28 at 2 27 24 pm
screen shot 2015-07-28 at 2 33 41 pm

## No browser history
screen shot 2015-07-27 at 5 12 55 pm

# Teacher Review
## Has history
screen shot 2015-07-28 at 2 13 19 pm
screen shot 2015-07-28 at 2 35 18 pm

## No browser history
screen shot 2015-07-27 at 5 12 19 pm

### DIFF
--- a/src/flux/transition.coffee
+++ b/src/flux/transition.coffee
@@ -23,9 +23,9 @@ TransitionStore = flux.createStore
   actions: _.values(TransitionActions)
   _local: []
 
-  load: (obj) ->
-    obj.type ?= 'push'
-    @_local.push(obj.path) if obj.type is 'push'
+  load: ({path, type}) ->
+    type ?= 'push'
+    @_local.push(path) if type is 'push'
 
   reset: ->
     @_local = []
@@ -36,10 +36,11 @@ TransitionStore = flux.createStore
   exports:
     getPrevious: (matchRoutes) ->
       history = @_get()
-      backPath = history[history.length - 2] if history.length > 1
-      matchedRoute = matchRoutes(backPath)
+      backPath = history[history.length - 2]
+      matchedRoute = matchRoutes(backPath) if backPath?
       deepestRouteName = _.last(matchedRoute.routes).name if matchedRoute?.routes?.length
 
+      # Both path and name will be undefined if history does not have a previous entry.
       path: backPath
       name: DestinationHelper.getDestination(deepestRouteName)
 
--- a/src/flux/transition.coffee
+++ b/src/flux/transition.coffee
@@ -1,28 +1,31 @@
 _ = require 'underscore'
 flux = require 'flux-react'
+DestinationHelper = require '../helpers/routes-and-destinations'
 
 TransitionActions = flux.createActions [
   'load'
   'reset'
   '_get'
-  '_getPathName'
 ]
 
-routesAndNames =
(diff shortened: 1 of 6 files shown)


philschatz:
I like the idea and wonder if more back buttons should be replaced with this one. One thing though: If there is history then the text of the button is only "Back", not "Back to ..." . Thoughts, @Fredasaurus ?

Fredasaurus:
yep.

pandafulmanda:
Yeah, im not sure how much we can get out of the History at that point, although..we can use other things to help us...

philschatz:
My concerns with relying on the browser history are:

- we need to make sure every screen that changes the URL correctly either just changes the URL or changes the URL + adds it to the history (so you don't accidentally click "Back" and just go to the previous step instead of going back to the Dashboard)

Calling a `TransitionStore` instead of `transitionTo` directly might work but it will probably need `@context.router` passed in as an arg.

pandafulmanda:
okay, will look into it

pandafulmanda:
teehehehe

philschatz:
Nice! It seems the `routesAndNames` is a bit brittle. I wonder if we could reuse the route name for matching so if the route is changed this matcher is also changed.

For example, with [rakt/react-router/modules/URLUtils.matchPattern](https://github.com/rackt/react-router/blob/master/modules/URLUtils.js#L88-L105) maybe something like:

```coffee
# router.cjsx

DASHBOARD = 'courses/:courseId/list/?'

routes = (
  <Route path={DASHBOARD} ... />
  ...
)

TransitionStore.register(DASHBOARD, 'Dashboard')
```

pandafulmanda:
ah, okay, so you'd be more comfortable if it matches by the route name.  I have something for that =).

philschatz:
ooh, I was unaware of that option. Would now be a good time to try and switch?

pandafulmanda:
hm..i don't think we should do that until after beta release, but perhaps we should discuss with the whole FE

src/flux/transition.coffee
```diff
@@ -0,0 +1,46 @@
...
+# TLDR: Use router.transitionTo or Router.Link for routes you want
+# in History and back button.  Only use router.replaceWith when you
+# want that route to be ignored in History and back.
+
+TransitionStore = flux.createStore
+  actions: _.values(TransitionActions)
+  _local: []
+
+  load: (obj) ->
+    obj.type ?= 'push'
```
philschatz: **Minor:** rather than change the object that was passed in, what about something like:

```coffee
load: ({path, type}) ->
  type ?= 'push'
  @_local.push(path) if type is 'push'
```

src/flux/transition.coffee
```diff
@@ -0,0 +1,46 @@
...
+  reset: ->
+    @_local = []
+
+  _get: ->
+    @_local
+
+  exports:
+    getPrevious: (matchRoutes) ->
+      history = @_get()
+      backPath = history[history.length - 2] if history.length > 1
```
philschatz: What if `history.length <= 1`?

PR #544, opened 2015-07-28
Disabling actions on plan footer

jivewise:
Disables publish, save, cancel and delete in the plan footer when either saving or publishing.


philschatz:
It seems "Delete" is not disabled for me if I open a draft and then click publish

philschatz:
any updates to this?

PR #545, opened 2015-07-28
Dashboard Style Cleanup

Fredasaurus:
Makes headers consistent, fixes a button bug is FF.

Before
screen shot 2015-07-28 at 10 58 42 am

After
screen shot 2015-07-28 at 10 59 09 am

Things left to do:

- [ ] Change "Recent Topics" to "Learning Forecast"
- [ ] Just have  "My Weakest Topics" button - move into Learning Forecast panel above "All Topics"
- [ ] Change "Current level of understanding" to "Recent topics" italics, grey, not bold
- [ ] Remove Stronger 
- [ ] Change "X worked" to "X problems worked"

### DIFF
--- a/resources/styles/components/student-dashboard/progress-guide.less
+++ b/resources/styles/components/student-dashboard/progress-guide.less
@@ -37,6 +37,7 @@
         width: 50%;
         border-top-left-radius: 0 !important;
         border-bottom-left-radius: 0 !important;
+        span { float: left; }
 
         & i:after {
           .fa-icon();
--- a/resources/styles/components/student-dashboard/progress-guide.less
+++ b/resources/styles/components/student-dashboard/progress-guide.less
@@ -2,74 +2,78 @@
 // dashboard specific styles only inside .dashboard-container below
 @import "../learning-guide/chapter-panel-mixin";
 
-.student-dashboard {
-  .progress-guide {
-    .chapter-panel { .learning-guide-chapter-panel(); }
-    h1,h2,h3 { margin: 0; text-align: left; }
-    .panel-title {
-      #fonts > .sans(2.5rem, 5rem);
-      font-weight: 900;
-      border-bottom: thin dotted @tutor-neutral-light;
-    }
-    .current {
-      #fonts > .sans(1.6rem, 5rem);
-      font-weight: 600;
-    }
-    .btn-group {
-      width: 100%;
-      // override material design
-      .btn:not(.btn-link):not(.btn-flat):not(.btn-fab) { margin-right: 0; }
-      .stronger {
-        width: 50%;
-        span { float: right; }
+#react-root-container {
 
-          i {
-          float: left;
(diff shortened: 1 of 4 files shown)


Fredasaurus:
I had to use !important in  a couple places (!I know), so if someone can figure that one out that would be great.

philschatz:
@nathanstitt any ideas on how to remove the `!important`?

resources/styles/components/student-dashboard/progress-guide.less
```diff
@@ -8,26 +8,42 @@
...
-      .weaker i:after {
-        .fa-icon();
-        padding-left:8px;
-        content: @fa-var-arrow-down;
-        color: @tutor-danger;
+      .weaker {
+        width: 50%;
+        border-top-left-radius: 0 !important;
+        border-bottom-left-radius: 0 !important;
+
```
nathanstitt: Are you sure the buttons have a border-radius set?  I don't see any rules specifying that and they appear to be in the "flat" style to me.

If they do have a border-radius, the `!important` clause could be dropped if you change the rule to be as specific as whatever applies it.  For instance, here's a case where I had to get pretty specific in order to override some material-design styles: https://github.com/openstax/tutor-js/blob/master/resources/styles/components/student-dashboard/progress-guide.less#L16

Fredasaurus:
The border radius is definitely there. What is strange is that the class "btn-group" should remove it.

Failing other ideas I can up the specificity, thanks.

Fredasaurus:
I also noticed that the box-shadow on buttons isn't a mixin. Should it be?

nathanstitt:
My general rule is that it should be extracted if that set of style are going to be used in more than one place.  I'm not aware of any other use of that so I think it's fine.

philschatz:
@Fredasaurus were you going to remove the remaining `!important` or just leave them in?

Fredasaurus:
Been in meetings all afternoon. Let me try and if I get stuck I'll see if someone can come in in relief.

Fredasaurus:
Looks lie the "btn-group" class is not being applied correctly. Can't figure why.

Here is markup pasted straight from here. http://getbootstrap.com/components/#btn-groups

And how it looks.

screen shot 2015-07-28 at 6 12 35 pm

nathanstitt:
I've noticed that the material design rules change `.btn-group` quite a bit using `not()`  rules.  That's my guess at least

Fredasaurus:
OK added #react-root-container to top of file but.....

I'm assuming that will duplicate this selector when everything compiles,
It seems more block headed that !important.

Let me know, thanks.

philschatz:
Just an idea: what if all CSS rules were prefixed with `#react-root-container`?

philschatz:
This is no longer mergeable. Can you pull from master and resolve conflicts? I think it might be because of Nathan's Learning Guide fixes

PR #546, opened 2015-07-28
Bug/confusing export

pandafulmanda:
# on first page on load
screen shot 2015-07-28 at 3 00 28 pm

# when export has been clicked
screen shot 2015-07-28 at 3 00 53 pm

Once download is completed, **download is forced** without any extra action needed from user

# After export and download
screen shot 2015-07-29 at 1 16 35 am

# If download URL doesn't validate for whatever reason
screen shot 2015-07-29 at 12 59 26 am

Export button dims to discourage multiple exports per page load.  Linked export time with last export download link as well.

Once the user leaves the perf report and comes back at another time, export button will be primary again, as in first screen.

Thoughts?  @Fredasaurus, @philschatz, @jpslav 

If your local exports does not work, you may need to use this: openstax/tutor-server#527

tests with be in separate PR after behavior is agreed on

### DIFF
--- a/src/components/performance/export.cjsx
+++ b/src/components/performance/export.cjsx
@@ -75,25 +75,28 @@ PerformanceExport = React.createClass
         if contentType is mime.contentType('.xlsx')
           @triggerDownload({downloadUrl, lastExported})
         else
-          @cancelDownload()
+          @cancelDownload({downloadUrl, lastExported})
 
     downloadUrlChecker.send()
 
-  cancelDownload: ->
-    @setState(
+  cancelDownload: ({downloadUrl}) ->
+    invalidDownloadState =
       tryToDownload: false
       exportedSinceLoad: true
       downloadHasError: true
-    )
+
+    invalidDownloadState.downloadUrl = null if @state.downloadUrl is downloadUrl
+
+    @setState(invalidDownloadState)
 
   triggerDownload: ({downloadUrl, lastExported}) ->
-    downloadedState =
+    downloadState =
       tryToDownload: false
       exportedSinceLoad: true
       downloadUrl: downloadUrl
       lastExported: lastExported
 
-    @setState(downloadedState)
+    @setState(downloadState)
 
   downloadCurrentExport: (linkClickEvent) ->
     linkClickEvent.preventDefault()
--- a/src/components/performance/export.cjsx
+++ b/src/components/performance/export.cjsx
@@ -1,8 +1,11 @@
 BS = require 'react-bootstrap'
 React = require 'react'
 moment = require 'moment'
+mime = require 'mime-types'
+
 BindStoreMixin = require '../bind-store-mixin'
 AsyncButton = require '../buttons/async-button'
+TimeDifference = require '../time-difference'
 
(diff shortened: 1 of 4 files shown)


src/components/performance/export.cjsx
```diff
@@ -41,8 +45,13 @@ PerformanceExport = React.createClass
         downloadUrl: lastExport.url
         lastExported: lastExport.created_at
 
+      @triggerDownload(lastExport.url) if @state.triggerDownload
       @setState(exportState)
 
+  triggerDownload: (downloadUrl) ->
+    window.location = downloadUrl
```
philschatz: Should the download replace the current browser tab (and hope that the file is downloaded rather than something like opened in the browser)?

src/components/performance/export.cjsx
```diff
@@ -3,6 +3,7 @@ React = require 'react'
 BindStoreMixin = require '../bind-store-mixin'
 Time = require '../time'
 AsyncButton = require '../buttons/async-button'
+$ = require 'jquery'
```
philschatz: Is `jQuery` still needed?

pandafulmanda: oops, nope

Fredasaurus:
Can we change the verbiage to read "Last export on July 28th at 12:33pm"?

Trying to make it a little more human readable.

I'm 50/50 on year. I'd like to have it out (we know what year it is) but I guess there is a use case where a teacher would want to re-export something from multiple years ago and wants to know the year of the last export (maybe?).

src/components/performance/export.cjsx
pandafulmanda: it doesnt actually open, the way the BE set up that route, it forces a download

philschatz:
RE Year: We use relative times for the lateness (`2 minutes late`, `3 days late`); Would that work here as well?

pandafulmanda:
how about, ```2 minutes ago```, ```5 days ago`` etc...

Fredasaurus:
Yeah I think so.

pandafulmanda:
@Fredasaurus and @philschatz like this?

src/components/performance/export.cjsx
```diff
@@ -39,10 +42,15 @@ PerformanceExport = React.createClass
 
       exportState =
         downloadUrl: lastExport.url
-        lastExported: lastExport.created_at
+        lastExported: moment(lastExport.created_at).fromNow()
```
philschatz: Should this use `TimeStore.getNow()`?

src/components/performance/export.cjsx
philschatz: ok. I wonder what the UX would be if that goes to a 404 page...

src/components/performance/export.cjsx
pandafulmanda: Initially that's what I attempted.  However, it gets confused because the initial export call to the server sets time at the request time from server, and the created_at is always a few seconds ** after ** that as opposed to before.

src/components/performance/export.cjsx
pandafulmanda: oh hmm...good point..I'll add a check 404 before doing that.

PR #547, opened 2015-07-28
Don't show progress bar when missing current_level

nathanstitt:
When there isn't enough data available to plot the student's progress, current_level isn't given.  In that case we should just display a "Practice Section" message

screen shot 2015-07-28 at 1 03 01 pm


Fredasaurus:
Cool. Jason / Kathi / I will work on messaging for this -- or feel free to chime in. What we need is something short that explains that they need to practice to get a score. I think I had "practice to get score". That can be improved upon.

PR #548, opened 2015-07-28
Notes Styles Refactor

helenemccarron:
The goal of this PR is to refactor the note specific styles to avoid code repetition. 

There are three different levels of notes so far which should be targeted without impacting notes contained within those. 

Every single note is supposed the look identical, notes within notes haven't been styled yet - this will be part of the next phase, as well as styling specific notes, which can be achieved in the relevant note include. 

I have removed any mention to a specific note styling both from the note.less and theming-mixin.less files. These should be used locally for each note, not in a global definition for notes.


resources/styles/book-content/hs-physics/virtual-physics.less
```diff
@@ -1,48 +1,2 @@
...
-  li {
-    padding: 5px 0;
-  }
-  .solution, .commentary {
-    [data-type=title]{
-      #fonts > .sans(1.8rem, 1.75em);
-      font-weight: bold;
-    }
-  }
+.virtual-physics {
```
philschatz: If this file is empty and not imported from anywhere can you remove it?

helenemccarron: This has the empty class for the note (similarly to all the notes). Currently, none of them are "doing" anything but we will need them individually when we style each note separately. I can remove all the unused notes, but we will need to recreate them on a per needed basis. I initially created all the bio notes files as a placeholder and complete list of all possible notes.

philschatz: If that is the case then can you uncomment the `@import` for each of these files? Otherwise, these files are just "dead code"

helenemccarron: done - removed all extra styles for now as well

philschatz:
I noticed this PR seems to remove styling from notes and places (ie `display: inline-block` was only removed, never added). Are there visible differences caused by this PR?

helenemccarron:
I have removed the code duplicated in individual notes, the display inline-block is applied at the note level .note,
> section > .note:not(.learning-objectives),
> section > section > .note 

and should apply to all notes listed in this declaration (which targets the three levels of nesting for notes I have identified)

PR #550, opened 2015-07-28
learning guide - message if no assignments worked

darrenmason:
image


philschatz:
If there is no weaker/stronger should the entire panel show up? @Fredasaurus ?

Fredasaurus:
The Ux-y folks are taking a long look at this tomorrow and Thursday. We'll try to make some decisions ASAP so we can get direction for implementation.

PR #551, opened 2015-07-28
Hides all inactive students

jpslav:
When students are removed from a course they are marked inactive, but up until this PR that has had no effect in the FE.   https://github.com/openstax/tutor-server/pull/530 adds an `active` property to the student JSON in the roster.  This PR filter out all inactive students from the roster -- later the FE can choose to show them differently so we can enable reactivation of students (undropping them).

### DIFF
--- a/src/components/course-settings/period-roster.cjsx
+++ b/src/components/course-settings/period-roster.cjsx
@@ -25,6 +25,7 @@ module.exports = React.createClass
 
   render: ->
     students = _.sortBy(RosterStore.getStudentsForPeriod(@props.courseId, @props.period.id), 'last_name')
+    students = _.filter(students, (student) -> student.active)
     <div className="period">
       <h3>Period: {@props.period.name}</h3>
       <BS.Table striped bordered condensed hover className="roster">
--- a/src/components/course-settings/period-roster.cjsx
+++ b/src/components/course-settings/period-roster.cjsx
@@ -25,7 +25,7 @@ module.exports = React.createClass
 
   render: ->
     students = _.sortBy(RosterStore.getStudentsForPeriod(@props.courseId, @props.period.id), 'last_name')
-    students = _.filter(students, (student) -> student.active)
+    students = _.where(students, is_active: true)
     <div className="period">
       <h3>Period: {@props.period.name}</h3>
       <BS.Table striped bordered condensed hover className="roster">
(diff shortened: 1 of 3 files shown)


philschatz:
Travis does **NOT** approve of this! @nathanstitt since you worked on the roster tests could you help?

jpslav:
Yeah I'm guessing some stub data doesn't have `active` and so everything getting filtered out.

philschatz:
Also, should it be `is_active`?

jpslav:
damn... good point... arg.

nathanstitt:
Nice try JP :grinning:   Looks like the spec that's failing is: [test/components/course-settings.spec.coffee#L43](https://github.com/openstax/tutor-js/blob/master/test/components/course-settings.spec.coffee#L43)

The 'Potter' student no longer renders because he's not active so it should be an empty array there.

I can take over the PR if you'd like ?

jpslav:
Hey I'm close, already fixed the bug :-)

Dantemss:
Oh I thought he was invisible because he was a wizard... carry on.

jpslav:
BE party on a FE PR!

src/components/course-settings/period-roster.cjsx
```diff
@@ -25,6 +25,7 @@ module.exports = React.createClass
 
   render: ->
     students = _.sortBy(RosterStore.getStudentsForPeriod(@props.courseId, @props.period.id), 'last_name')
+    students = _.filter(students, (student) -> student.active)
```
nathanstitt: Since you're just testing a property, you could use [where](http://underscorejs.org/#where)

`students = _.where(students, active: true)`

Fredasaurus:
Ain't no party like a BE party cause a BE party don't Break();

jpslav:
@philschatz the corresponding BE PR has been merged, so y'all can merge this whenever.

src/components/course-settings/period-roster.cjsx
jpslav: great tip @nathanstitt!

philschatz:
@jpslav Does this also need to be in other screens like the perf report, free responses when looking at the course aggregate review?

jpslav:
@jason-holmes might have to weigh in.  I could see teachers needing dropped students in the performance report, e.g. Jimmy drops part way through the semester but the teacher still needs to generate his grade for the first six weeks.  For course aggregate views maybe it is the other way, that teachers don't want to see responses from kids who are no longer in the class.

PR #552, opened 2015-07-28
Filter out tasks that are being deleted from calendar

nathanstitt:
When a large task is deleted, it takes the BE quite awhile to process the network request.  This causes problems due to how our flux stores are setup.  The sequence is:

* The delete action fires, which fires off the request to the BE and sets the store's asyncStatus = DELETING
* Teacher browses around, while the delete xhr request is still pending.  The data is still present in the store and shows on the calendar.
* xhr delete finally completes and than we finally delete the task from the store

Other options for this might be to delete the task from the store immediately when we fire the delete action to the BE.  This would stop the task from showing up, but if the delete failed we wouldn't be able to "resurrect" the task very easily.

### DIFF
--- a/src/flux/teacher-task-plan.coffee
+++ b/src/flux/teacher-task-plan.coffee
@@ -10,8 +10,11 @@ TeacherTaskPlanConfig =
     @_local[id] = plans
 
   exports:
-    getCoursePlans: (id) ->
-      @_local[id] or []
+    getActiveCoursePlans: (id) ->
+      plans = @_local[id] or []
+      # don't return plans that are in the process of being deleted
+      _.filter plans, (plan) =>
+        not @exports.isDeleting.call(@, plan.id)
 
 
 extendConfig(TeacherTaskPlanConfig, new CrudConfig())
--- a/src/components/task-plan/footer.cjsx
+++ b/src/components/task-plan/footer.cjsx
@@ -1,4 +1,3 @@
-
 React = require 'react'
 BS = require 'react-bootstrap'
 {TaskPlanStore, TaskPlanActions} = require '../../flux/task-plan'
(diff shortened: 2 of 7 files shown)


src/flux/teacher-task-plan.coffee
```diff
@@ -10,8 +10,11 @@ TeacherTaskPlanConfig =
     @_local[id] = plans
 
   exports:
-    getCoursePlans: (id) ->
-      @_local[id] or []
+    getActiveCoursePlans: (id) ->
+      plans = @_local[id] or []
+      # don't return plans that are in the process of being deleted
+      _.filter plans, (plan) =>
+        not @exports.isDeleting.call(@, plan.id)
```
nathanstitt: Is there a better way to call an exported method here?

philschatz:
@pandafulmanda Could something similar be done for publishing?

pandafulmanda:
yes, but we can use the jobs status to our advantage there

Dantemss:
I *think* delete blocks until done?

nathanstitt:
yes, it does.  Sorry, I only got my comment on the PR 1/3 written :grin:

philschatz:
What about using `<AsyncButton>` to show a spinner until the deletion completes (similar to publish/save/continue)?

nathanstitt:
Yes, it would be a nice UI to add the button, that way the teacher has some feedback that the task is progressing.  I'll do so shortly.

That alone won't solve the issue, because it wouldn't stop the teacher from clicking on any of the other navigation items or hitting their back button and returning to the calendar while the delete is still being performed.

philschatz:
ah, good point RE back button or other nav buttons

nathanstitt:
Actually, the asyncButton will never display as in-progress during "normal" operation.  The builder immediately redirects back to the calendar once a task deletion is initiated.  I think it still has some value though, because if the teacher then hits the browser back button they return to the builder and the task button will show the deleting spinner.

screen shot 2015-07-29 at 10 16 55 am

PR #553, opened 2015-07-29
exercises in reference book use absolute url returned from back end for the most part

pandafulmanda:
Matches with openstax/tutor-server#529

Take note of this [line](https://github.com/openstax/tutor-js/pull/553/files#diff-87c9b131b256593810e36390397c189fR89)


pandafulmanda:
closing in favor of #554 solution with openstax/tutor-server#529

PR #554, opened 2015-07-29
Update/exercises full link in ref book api

pandafulmanda:
Similar to #553, matches this openstax/tutor-server#533

### DIFF
--- a/src/components/reference-book/page.cjsx
+++ b/src/components/reference-book/page.cjsx
@@ -86,17 +86,13 @@ module.exports = React.createClass
   isExerciseLink: (link) ->
     link.pathname.search(EXERCISE_MATCHER) is 0
 
-  # absolute url returned does not include the path to the api
-  getExerciseAPIUrl: (link) ->
-    "#{link.protocol}//#{link.host}/api#{link.pathname}"
-
   renderOtherLinks: (otherLinks) ->
     ReferenceBookExerciseStore.setMaxListeners(otherLinks.length)
     _.each(otherLinks, @renderExercise)
 
   renderExercise: (link) ->
     if @isExerciseLink(link)
-      exerciseAPIUrl = @getExerciseAPIUrl(link)
+      exerciseAPIUrl = link.href
 
       if link.parentNode.parentNode?
         @_exerciseNodes.push(link.parentNode.parentNode)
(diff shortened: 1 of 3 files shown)


src/components/reference-book/page.cjsx
```diff
@@ -13,7 +13,7 @@ GetPositionMixin = require '../get-position-mixin'
 {ReferenceBookStore} = require '../../flux/reference-book'
 {ReferenceBookExerciseStore} = require '../../flux/reference-book-exercise'
 
-EXERCISE_MATCHER = new RegExp('#ost\/api\/ex\/(.*)')
+EXERCISE_MATCHER = new RegExp('\/exercises\/(.*)')
```
philschatz: ick, I wonder if we need/should match based on a CSS selector like `[data-type='exercise'] > [data-type='problem'] .os-embed` instead

philschatz:
1 minor issue about detecting what is an exercise link but otherwise fine ; )

src/components/reference-book/page.cjsx
pandafulmanda: Is that always going match all embedded exercises? I was thinking about that but I was unsure

philschatz: I think it should because BE is doing something similar for book import. There might be a better CSS selector to pull it out. @karenc what's the XPath used? (feel free to add Dantemss if he's not busy with book reimport)

karenc: I just check for links that starts with #ost. `//a[starts-with(@href, "#ost")]`

karenc: Not very useful for you I guess... The urls now look like `https://exercises-dev.openstax.org/api/exercises?q=tag%3Aapbio-ch02-ex026`.

philschatz: Ah, from https://github.com/openstax/tagging-legend#exercises `.os-exercise .os-embed` should do the trick!

philschatz: I just had to change this regexp in https://github.com/openstax/tutor-js/commit/880fe0ea6a032f45b1e5098704ebb4705a130df6 switching to use the CSS selector

PR #555, opened 2015-07-29
Show exercise uid on HW builder

nathanstitt:
TODO:
 * [x] Display on HW builder
 * [x] Student i-reading questions
 * [x] Student hw questions
 * [x] Student practice widget

Homework Builder:
screen shot 2015-07-28 at 9 57 43 pm

Multiple choice question.  The same component is used for both readings, homeworks, and practice exercise. The UID is displayed identically on them all.
screen shot 2015-07-29 at 10 29 09 am

### DIFF
--- a/resources/styles/components/question.less
+++ b/resources/styles/components/question.less
@@ -238,7 +238,15 @@
       }
     }
   }
-}
+
+  .exercise-uid {
+    position: absolute;
+    right: @tutor-card-body-padding-horizontal;
+    bottom: 10px;
+    .tutor-exercise-question-uid();
+  }
+
+} // end of .question
 
 .task-teacher-read-only {
   .question {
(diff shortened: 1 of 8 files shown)


philschatz:
Can you put the id in the exercise builder in the bottom-right corner (like in the student view)?

nathanstitt:
@philschatz Yeah, that'd be an easy change to make (and would make the code cleaner), but the mockup shows the id mixed in with the tags.  Should I go ahead?

philschatz:
@Fredasaurus thoughts on where/how these id's should be displayed?

Fredasaurus:
I think it looks fine where it is. Do we need to label it as "ID" in the second example?

philschatz:
I think light grey is sufficient but that's your call ; ) merging for now so Greg has the data sooner

Fredasaurus:
Cool.

PR #556, opened 2015-07-29
Feature/plan pending publish

pandafulmanda:
## Publishing state

will show as publishing if returning to calendar right after builder or even just on calendar load

screen shot 2015-07-29 at 2 43 55 am

## Clicking on the publishing plan
will bring up this modal
screen shot 2015-07-29 at 7 35 31 pm

## Publishing plan will automatically update
* with published state when job is completed
* with failed or killed state when job failed for whatevesies
screen shot 2015-07-29 at 7 36 03 pm

Depends on openstax/tutor-server#540, and running background jobs locally if testing locally

## To Do
- [ ] code clean up
- [ ] builder to calendar UX details
- [ ] styling
- [ ] tests


philschatz:
Can you merge from master?

philschatz:
Hmm, will the teacher need to refresh the calendar to see the "publishing..." text disappear?

pandafulmanda:
the teacher will not need to refresh.  I should add a case to handle errors.

Also, i realized there should be a different behavior when the plan is publishing -- clicking on it should not go to edit as it does now...what should it do?

philschatz:
I'm not sure what it should do. Thoughts @Fredasaurus ? Some options:

1. do nothing
2. popup that says "This item is currently being published. Please wait"
3. go straight to edit (like a draft)

I'm in favor of the last option because BE allows this (patching an existing TaskPlan that is in the process of being published) plus if the publish errors out, the teacher still has an option to edit/fix the assignment

Fredasaurus:
hmmmm.... as long as the info being shown on the edit screen is current and complete, then sure.

pandafulmanda:
Editing while publishing will be blocking for as long as the publish job is still running and will have unpredictable behavior.  I think will opt for option 2, with the ability to edit if the job errors out.

philschatz:
d'oh, something conflicts again. Can you re-merge?

Fredasaurus:
Did we decide on number 2 then?

philschatz:
I think so. Was there a better option?

PR #557, opened 2015-07-29
Replaced 'Tutor ID' role # in roster with deidentifier

jpslav:
The student's role number was displayed in the roster with "Tutor ID" as the heading -- this number is not useful to the teacher.  However, the "deidentifier" is -- it can be used to talk about students over insecure channels like email without divulging names.  This PR replaces the role with the deidentifier.

I also added a tooltip, which probably could use styling if we even want to keep it.  

I also removed the "Period: Period 1" heading that duplicates the name in the tab.

I'm totally fine if no one wants this PR :-)

image

### DIFF
--- a/src/components/course-settings/period-roster.cjsx
+++ b/src/components/course-settings/period-roster.cjsx
@@ -17,7 +17,7 @@ module.exports = React.createClass
     <tr key={student.id}>
       <td>{student.first_name}</td>
       <td>{student.last_name}</td>
-      <td>{student.id}</td>
+      <td>{student.deidentifier}</td>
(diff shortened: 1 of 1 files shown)


nathanstitt:
HE'S UNSTOPPABLE!

jpslav:
For my own edification, what's the most proper way in tutor-js to add space to the left of the question icon?  I resisted the urge to put in `style="padding-left:3px"` :-)

philschatz:
Sort of. You'd need to create a CSS class and add the padding there. `style=` is SOOO 1996 ; )

PR #561, opened 2015-07-29
Reference view menu

Fredasaurus:
adding the drop shadow, toggle button, cleaning up some styles, refactoring less.

screen shot 2015-07-29 at 6 10 10 pm


Fredasaurus:
A few things that still need work. Can someone look at these?

Indents don't look so great after two characters and a line wrap.

screen shot 2015-07-29 at 6 11 27 pm

Menu gets fugly on resize.

screen shot 2015-07-29 at 6 12 15 pm

nathanstitt:
@Fredasaurus I've fixed some of the above, want me to push them to your branch ?

Fredasaurus:
do it.

nathanstitt:
@Fredasaurus Give this a spin and see what you think.  The last commit changes the reference book to be 85% wide when the page is narrow, with 7.5% on each side used for the forward/back controls.  It seems to work pretty well I think, but I haven't yet explored all the content

nathanstitt:
Sigh. Somehow that commit messes up the padding on iReadings when the screen is narrow.  Attempting to debug that now

nathanstitt:
Ok, that should do it.   I'm ok to merge if @Fredasaurus is.

Fredasaurus:
The top menu disappears at resize 100px. Other than that, looks great.

Fredasaurus:
Ahhhh ICWYDT. That *might be ok. Let me check.

nathanstitt:
yeah, items from the nav were overflowing onto the page when it got too narrow.  We might also attempt render the title using a fixed with ellipsis when the screen gets narrow.  That way it would at least mostly appear.

resources/styles/components/reference-book/navbar.less
```diff
@@ -15,9 +24,26 @@
       }
     }
     .menu-toggle:before{ content: @fa-var-list; }
+    // don't collapse navbar when screen is narrow
+    @media (max-width: 979px) {
```
philschatz: Is there a var we can reference here? like `@book-content-collapse-breakpoint` ?

nathanstitt: Maybe  That's overriding the bootstrap default.   I'll investigate to see if bootstrap has a bar for it. If not I'll make a comment to that effect

nathanstitt:
Actually I was incorrect about the title.. I'm not currently hiding that in order to make more room, I'm just hiding the logo.

philschatz:
Looks good to me. Was there anything else that needed to be done?

nathanstitt:
I don't think there is, unless @Fredasaurus has found anything ?

philschatz:
Merging. Additional fixes should be a separate PR :smile:

Fredasaurus:
menu or mobile resize will need work eventually. It's OK for now. Thanks!

PR #563, opened 2015-07-30
Export button behavior matching export job status

pandafulmanda:
No UI change except the isWaiting maps properly now to statuses from the job as well


philschatz:
Is the code for the `HEAD` request to AWS still in the codebase?

pandafulmanda:
Yes, I'm leaving that in for this PR.  This is just meant for fixing the button's state.  Taking out the DL link checking code - if we are to - is dependent of  a decision between #565 and #564.

PR #564, opened 2015-07-30
Alt/export force dl iframe

pandafulmanda:
- [ ] tests

Teeeheheheh it works.  Just say yes =)  and do the CORS on s3...

Alternative to #565 

# on page load
screen shot 2015-07-30 at 2 21 57 pm

# exporting and validating dl link
screen shot 2015-07-30 at 2 22 01 pm

# auto download, button after export
screen shot 2015-07-30 at 2 22 06 pm
dims to discourage multiple exports on page load

# when theres a problem with any of the following
* queueing export
* job
* download url not validating as xlsx

screen shot 2015-07-30 at 4 55 08 pm

# Pros and cons
- User does not need know to wait and press DL button
- Tricky, but common method for auto DL's, I remember seeing this iframe method elsewhere


philschatz:
Closing in favor of #565. This requires CORS to be set up (deployment configs) and some 1-off IFrame trickery. If there was an external lib that accomplished this task that we could pull in it might be worth looking into but I think this method adds cruft that we'd need to test and support (although it *seems* to work in Chrome, FF, Safari)

pandafulmanda:
Reopened due to CORS being added to s3

philschatz:
closing for now because no complaints from users/UX

PR #565, opened 2015-07-30
Alt/export btn changes to dl button

pandafulmanda:
Another way, #564

# on page load
screen shot 2015-07-30 at 1 58 46 pm

# generating
screen shot 2015-07-30 at 1 59 06 pm

# after export generated and dl link verified
screen shot 2015-07-30 at 1 58 50 pm

# after first download
screen shot 2015-07-30 at 2 01 34 pm
 dims button to discourage repeated clicking

# when theres a problem with any of the following
* queueing export
* job
* download url not validating as xlsx

screen shot 2015-07-30 at 5 15 25 pm

# Pros and Cons
- User has to know to wait for export to finish and click the DL button
- Not able to export again per page load.
- Could potentially take out the HEAD checking code and avoid having to update CORS
  - I actually think it's a good idea to have this check in there regardless.  Prevents DL of other types of files accidentally.

### DIFF
--- a/resources/styles/components/performance/export.less
+++ b/resources/styles/components/performance/export.less
@@ -12,3 +12,9 @@
     .clearfix();
   }
 }
+
+.tutor-downloader {
+  width: 0;
+  height: 0;
+  float: right;
+}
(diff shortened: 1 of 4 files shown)


philschatz:
I think that for this one we decided to remove the "Last exported" text. @jason-holmes ?

jason-holmes:
I think we should leave the last exported text, but not make it link to the latest export file.

PR #569, opened 2015-07-31
Disallow actions after plan is visible

jivewise:
screen shot 2015-07-31 at 11 02 50 pm
screen shot 2015-07-31 at 11 03 12 pm

- [x] Readings cannot sort or remove sections after visible to students
- [x] HWs cannot sort or remove exercises after visible to students
- [ ] Add unit tests


philschatz:
I think the logic for this should be "Disallow actions after plan is visible and published". That way you can still edit/delete a draft that is open today (but not published)

PR #571, opened 2015-07-31
Adding iReading specific styles

helenemccarron:
Allowing for I-Reading specific styles. Added context-mixin for now.

TODO: full refactor, see napkin note https://github.com/openstax/napkin-notes/blob/master/phil/s016-book-styling-refactor.md

*Screenshots not available*
Removed grey background on all notes in the iReading view, removed extra padding within notes, excluded tips-for-success to display as regular notes in the iReading view. 

Reference book new styles:
bullets to match iReading + theming mixin for colors

Removed numbered list-styles for footnotes

Tightened up question styles

Tightened up learning objectives styles

Restyled grasp checks

### DIFF
--- a/resources/styles/book-content/context-mixins.less
+++ b/resources/styles/book-content/context-mixins.less
@@ -1,6 +1,6 @@
 .tutor-book-note(reading) {
   > .note,
-  > section > .note:not(.learning-objectives),
+  > section > .note:not(.learning-objectives, .tips-for-success),
   section > section > .note {
     background: none;
     padding: 20px 0;
@@ -8,7 +8,6 @@
       margin: -68px 0 0 0;
     }
   }
-
   section > h1 {
     clear: both;
   }
--- a/resources/styles/book-content/context-mixins.less
+++ b/resources/styles/book-content/context-mixins.less
@@ -8,8 +8,10 @@
       margin: -68px 0 0 0;
     }
   }
+}
+
+.tutor-book-titles(reading){
   section > h1 {
     clear: both;
   }
-
 }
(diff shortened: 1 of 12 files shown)


resources/styles/book-content/context-mixins.less
```diff
@@ -0,0 +1,17 @@
+.tutor-book-note(reading) {
+  > .note,
+  > section > .note:not(.learning-objectives):not(.tips-for-success),
+  section > section > .note {
+    background: none;
+    padding: 20px 0;
+    &:before {
+      margin: -68px 0 0 0;
```
philschatz: Where did this margin calc come from? Is there a var (or var + number) you could use in case we change margins?

resources/styles/book-content/theming-mixins.less
```diff
@@ -16,13 +16,19 @@
 .tutor-book-theme-note(@label-background: @tutor-book-primary; @label-text: @tutor-book-text-on-dark-labels) {
   > .note,
   > section > .note,
-  section > section > .note {
+  section > section > .note, .grasp-check {
```
philschatz: can you put each selector on a new line? That way the line does not, at a glance, look like `section > section > .note .grasp-check`

resources/styles/book-content/hs-physics/grasp-check.less
```diff
@@ -1,19 +1,18 @@
 // This file is part of the book-content mixin and should not be included directly
 .grasp-check {
-  background: @tutor-white;
-  display:  inline-block;
+  margin: 68px 0 32px 0;
```
philschatz: What is this `68px` from?

philschatz: Ah, it's just the bottom margin. ok

PR #574, opened 2015-08-02
Scroll reference ToC so the entry for the current page is visible

nathanstitt:
Corrects bug where a link to a chapter such as 7.2 was clicked but the table of contents menu wouldn't show it because it was below the scroll area.

### DIFF
--- a/src/components/reference-book/exercise.cjsx
+++ b/src/components/reference-book/exercise.cjsx
@@ -14,8 +14,9 @@ ReferenceBookExercise = React.createClass
   displayName: 'ReferenceBookExercise'
   render: ->
     {exerciseAPIUrl} = @props
-    {items} = ReferenceBookExerciseStore.get(exerciseAPIUrl)
-
+    exercise = ReferenceBookExerciseStore.get(exerciseAPIUrl)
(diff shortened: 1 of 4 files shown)


src/components/reference-book/exercise.cjsx
```diff
@@ -14,8 +14,9 @@ ReferenceBookExercise = React.createClass
   displayName: 'ReferenceBookExercise'
   render: ->
     {exerciseAPIUrl} = @props
-    {items} = ReferenceBookExerciseStore.get(exerciseAPIUrl)
-
+    exercise = ReferenceBookExerciseStore.get(exerciseAPIUrl)
+    return null unless exercise
```
philschatz: This removes the "Exercise Missing" text when an exercise is missing.

How about https://github.com/openstax/tutor-js/commit/7ce1035d1f629547a0ff95c64225521fd7af3567#diff-73b870b5a9922ddf9e623aa1807f684a instead?

nathanstitt: Yeah, that's more concise.  I'll do so ; )

PR #572, opened 2015-07-31
performance forecast

darrenmason:
Teacher:
screen shot 2015-08-03 at 8 27 30 am

Student:
screen shot 2015-08-03 at 7 55 46 am

Teacher viewing student:
screen shot 2015-08-03 at 7 57 59 am

- [X] updated teacher/student header
- [x] header info icon / hover
- [x] period tab buttons
- [x] single column weaker areas
- [x] single column chapter/section listing 
- [x] update all titles to 'performance forecast'

PR #575, opened 2015-08-03
Flexbox mixin with webkit and ie vendor prefixes

nathanstitt:
Thoughts @philschatz and rest of FE team?

If this looks good I'll switch out the remainder of the places we use flexbox with this.

### DIFF
--- a/resources/styles/book-content/learning-objectives.less
+++ b/resources/styles/book-content/learning-objectives.less
@@ -4,13 +4,10 @@
   width: 100%;
   min-height: 100px;
   .book-content-full-width();
+  .flex-display(inline-flex);
   margin-top: -@tutor-card-body-padding-vertical;
   padding: 10px 20px;
   border: none;
-  display: inline-flex;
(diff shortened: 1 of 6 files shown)


philschatz:
Is this commit already in #572?

nathanstitt:
It is, but was cherry-picked so the sha is identical and won't conflict.  Since there was agreement on using the mixin, I'm going to go through and switch out other use.

resources/styles/book-content/learning-objectives.less
```diff
@@ -21,34 +18,24 @@
...
+    .justify-content(center);
     .ost-learning-objective-def {
+      .justify-content(center);
       #fonts > .sans(1.5rem, 1.8em);
       margin-left: 20px;
       font-weight: 300;
-      justify-content: center;
-      -webkit-justify-content: center;
-      -webkit-flex-direction: column;
       &:before {
```
nathanstitt: Why were we setting `flex-direction: column` only on webkit?

PR #562, opened 2015-07-30
Learning Forecast née "Guide" tweaks

nathanstitt:
TODO:
LearningGuide
 * [x] Change name to "Learning Forecast"
 * [x] Default to open on chapters
 * [x] Remove "Stronger" section
 * [x] Change "Weaker" to "The weakest topics" - Student (Preferably "My Weakest topics")
 * [x] Change to "The weakest topics" - Teacher
 * [x] Change "X worked" to "X problems worked"

Dashboard
 * [x] Practice panel should just have "Practice" and "My Weakest Topics" button
 * [x] Change "Recent Topics" to "Learning Forecast"
 * [x] Change "Current level of understanding" to "Recent topics" italics, grey, not bold
 * [x] Remove Stronger
 * [x] Change "X worked" to "X problems worked"

Screenshots:
Learning Forecast:
screen shot 2015-07-30 at 1 54 22 pm

screen shot 2015-07-30 at 3 03 27 pm

View when no sections returned:
screen shot 2015-07-30 at 4 56 38 pm

PR #577, opened 2015-08-03
Restrict rendering to routes based on user role

nathanstitt:
In order to restrict access to a given route, the app needs to know what course roles the user has very early in it's lifecycle. 

The BE PR [#562](https://github.com/openstax/tutor-server/pull/562) provides it to the app via json data enclosed in a script tag.  This will allow the app to continue to function even if the tag isn't provided for some reason, such as running locally.

There's a few other advantages to bootstrapping the stores this way.
  * The FE needs to make two less XHR requests when it's loaded.
  * The "Flash of un-styled content" issue no longer occurs on the reference view. Currently if the view is loaded it will first display using the standard book colors and then switch to the appropriate physics or biology colors once the course store loads.
  * The navbar displays the username immediately rather than blinking it into existence a few seconds after load.

Screenshot taken as a student after pasting in the url and loading

screen shot 2015-08-06 at 9 44 51 pm

### DIFF
--- a/src/helpers/dom.coffee
+++ b/src/helpers/dom.coffee
@@ -10,4 +10,12 @@ module.exports = {
     if @matches(el, selector) then el else @closest(el.parentNode, selector)
 
 
+  readBootstrapData: (root = document) ->
+    el = root.querySelector('#tutor-boostrap-data')
+    if el
+      el.parentNode.removeChild(el)
+      JSON.parse(el.textContent)
+    else
+      {}
+
 }
--- a/index.coffee
+++ b/index.coffee
@@ -21,9 +21,12 @@ window._STORES =
   TIME: require './src/flux/time'
   TOC: require './src/flux/toc'
 
-api.start()
+api.start(
+  JSON.parse(document.getElementById('tutor-boostrap-data')?.textContent or '{}')
+)
 startMathJax()
 TransitionAssistant.startMonitoring()
+
 
 # This is added because MathJax puts in extra divs on initial load.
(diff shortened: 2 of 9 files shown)


src/helpers/dom.coffee
```diff
@@ -10,4 +10,12 @@ module.exports = {
     if @matches(el, selector) then el else @closest(el.parentNode, selector)
 
 
+  readBootstrapData: (root = document) ->
+    el = root.querySelector('#tutor-boostrap-data')
```
philschatz: Why not use a `<script>` tag and create a global var?

nathanstitt: Well it is a `<script>` tag, but it's holding json.    :grin: 

No real reason, other than a slight aversion to creating a global variable.  What I really wanted to accomplish was to create an optional variable so the app would continue to function fine if it was missing for some reason.

We could just as easily create a `window.TUTOR_BOOTSTRAP_DATA` and read that.

nathanstitt: Another slight advantage to the "json in script tag" is that it allows the BE to test that the values they're sending are correct.  Since the data can just be fed back into JSON.parse it's easy to validate.  If it was a javascript var, the BE would need to either boot up a javascript interpreter or fall back on regex to test the data.  [current spec](https://github.com/openstax/tutor-server/blob/cacd3f2f08d2cc22eb5ad59493592c6de2d7b59c/spec/controllers/webview_controller_spec.rb#L38-L48)

PR #578, opened 2015-08-03
Course Roster changes

nathanstitt:
* Renames Delete action to be Drop
 * Renames the link to be "Course Roster"
 * Removes the add student button
 * Fixes a dumb bug in the Icon component and adds a spec so it doesn't re-occur.

screen shot 2015-08-03 at 1 07 03 pm

### DIFF
--- a/src/components/icon.cjsx
+++ b/src/components/icon.cjsx
@@ -15,7 +15,8 @@ module.exports = React.createClass
     icon = <i className={classes.join(' ')} />
 
     if @props.tooltip
+      console.log @props
       tooltip = <BS.Tooltip>Useful for talking securely about students over email.</BS.Tooltip>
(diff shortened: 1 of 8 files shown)


src/components/icon.cjsx
```diff
@@ -15,7 +15,8 @@ module.exports = React.createClass
     icon = <i className={classes.join(' ')} />
 
     if @props.tooltip
+      console.log @props
```
philschatz: What's this?

kathi-fletcher:
How do I tell if this is released. There is a ton of stuff in sprint 17 that I am pretty sure is released. This is from Aug. 3

nathanstitt:
Yes, these changes were part of the beta release.   You can verify that they're present by checking that the teacher's menu has the "Course Roster" option.

PR #580, opened 2015-08-03
update state to status to make BE update for the jobs api

pandafulmanda:
Updates ```state``` to ```status``` when looking at jobs, matches BE PR openstax/tutor-server#535


philschatz:
Seems like the HTTP status would be enough but that probably requires changes to `api.coffee` ; )

PR #581, opened 2015-08-03
Link to accounts with confirmation dialog for unsaved data

nathanstitt:
Includes a `TutorDialog` component that can be called like so:

```coffee
    TutorDialog.show({
      title: "Are you sure you want to do this?",
      body: <h1>Delete everything!</h1>
    }).then( @callMassDeletion, @massDeletionCanceled )
```

Or you can also do:

      <TutorDialog title="Foo" onOk={@func} onCancel={@func}>
        <h1>Hello1</h1>
      </TutorDialog>

Links that use the `TransitionAssistant` will appear as:
screen shot 2015-08-04 at 8 17 30 pm

Others employ `window.onbeforeunload`
screen shot 2015-08-04 at 8 11 39 pm


nathanstitt:
@philschatz this is WIP until the BE completes the corresponding PR, but you can review away.  This one's got some *interesting* stuff with the global component registry by mixin.   feedback is welcome.

philschatz:
Should we just allow going to accounts and adding in the "Are you sure?" generically so it works on any page with unsaved changes?

nathanstitt:
Yeah, but how do we detect if a page has unsaved changes?  That requires the component to be able to generically be asked "Do you have any unsaved changes?  And what are they if you do?"

This method allows a given component to mixin this and then implement the `hasUnsavedState` function.   It can optionally have a `unsavedStateMessage` but we generate a generic one from from `displayName` if it's not present.

philschatz:
My gut says we'll need to have each component register somewhere that there are unsaved changes (maybe the Stores can do this via `@change: ->`?) because what should happen if:

1. The user tries to close the tab thinking they're done (answering a HW or editing an assignment)
2. we accidentally miss adding `target="_blank"` to an `<a>`

nathanstitt:
Apologies for the book. :wink: 

It seems to me that we're talking about two different approaches.

The method you've outlined would be for a component (or store) to register that there's unsaved changes and then remove the registration once the changes either canceled or are persisted to the server.

The other approach (that I've taken) is for a component to provide a hook where it can be queried if there's unsaved changes or not.

Both approaches have similar restraints.  They'd both have to employ `mount` and `unMount` methods to keep the registry updated.  The first method would checkin that the were changes and then unregister once the changes where compete.  The second would just register itself. Both would need to clear the "changes registry" on unmount.

As I see it, the first method's drawbacks is that the component (or store) has to remember to always keep the registration in sync whenever something wasn't saved and then remember to remove it once the data is saved.   It's creating additional state that needs to be juggled.

In my approach the component's "unsaved check" would most likely use the exact same logic as the first approach, but it would never be called unless the app was about to exit.  Much less overhead and no state to maintain.  The first method would have to call the "unsaved check" on every state change.

I also don't think the store has enough information *on it's own* to do a proper job of maintaining the registry.  The store also has no information about why they're being loaded and/or modified or if the component that manipulated them is even still around.  With that said, I think that in most cases the "unsaved check" mentioned above would reside in the store, I just don't think it should be required to be there.

As to your points about the user closing the tab or we forget to add target="_blank".  I think we could prevent that equally well with either approach by the "changes registry" setting a `window.onbeforeunload` that would only return a prompt if there's unsaved information.  It's how the registry determine the presence of  the unsaved data that's in question.

For that matter we could also put a check in [Router.run](https://github.com/openstax/tutor-js/blob/master/src/router.cjsx#L98) to prevent navigation.  Again, this would work equally well regardless of the approach we take.

test/components/helpers/component-testing.coffee
```diff
@@ -38,11 +43,12 @@ Testing = {
...
       props._wrapped_component = component
       wrapper = ReactTestUtils.renderIntoDocument React.createElement(Wrapper, props)
-      resolve({
+      RENDERED.push(wrapper)
+      resolve(
         wrapper,
         element: wrapper.refs.element,
         dom: React.findDOMNode(wrapper.refs.element)
-      })
+      )
```
nathanstitt: A puzzler.  Here i'm resolving a promise with the rendered element which works fine.  However I also need to unmount the component after all of the promise's `then` callbacks are complete.  I can't seem to figure out how to do so or even if it's possible.  Taking a `.then()` onto the `resolve({...})` here doesn't cause any errors but it's seemingly never called.

philschatz:
Should we hide the link if it is not in the JSON (or the JSON is not retrieved yet)?

nathanstitt:
Yes it's probably a good idea not to render the account link if the profile_url isn't present.  Of course it *should* always be loaded, but best not to assume that.

There homework builder is  currently only only component that is setup with the `UnsavedStateMixin` mixin,   I was also unsure what we should consider unsaved there or how best to test for unsaved data.  For right now all I'm doing is considering the state unsaved if the task plan is new and there has been a name entered.  

Accordingly the dialog pops up if you create a new assignment and type something in the name field.    Once you do that the Account link will pop up the JS dialog, and the logout button will use the `window.onbeforeunload`  method.  Future refinements would be to go through and figure out better logic for the builder and switch all external links like logout to use the dialog.

src/components/tutor-dialog.cjsx
```diff
@@ -0,0 +1,63 @@
...
+    buttons = @props.buttons or [
+        <BS.Button key='cancel' onClick={_.compose(@props.onCancel,  @hide)}>Cancel</BS.Button>
+        <BS.Button key='ok'     onClick={_.compose(@props.onProceed, @hide)} bsStyle='primary'>OK</BS.Button>
+    ]
+    classes = ['tutor-dialog']
+    classes.push @props.className if @props.className
+
+    <BS.Modal className={classes.join(' ')} onRequestHide={@hide} title={@props.title}>
+      <div className='modal-body'>
+        {@props.body}
```
philschatz: Would `@props.children` work here and (require less documentation)?

src/components/tutor-dialog.cjsx
```diff
@@ -0,0 +1,63 @@
...
+
+      props = _.extend(_.clone(props), {
+        onProceed, onCancel, show: true
+      })
+
+      if @dialog
+        @dialog.replaceProps(props)
+      else
+        div = document.body.appendChild( document.createElement('div') )
+        @dialog = React.render(React.createElement(TutorDialog, props), div)
```
philschatz: Why have a new `React` root, instead of just a `<BS.Modal>` in the existing DOM?

nathanstitt: It seemed to me that it'd be easiest if the Modal was callable from a component at any point in it's React lifecycle.  If we rendered the modal as a child of a component, it would have to track state and be more complex.  For instance the AccountLink would have to look something like:
```coffee
  redirectToAccount: ->
    @setState(askNavAllowed: true)

  render: ->
    if @state.askNavAllowed
       modal = <TutorDialog title="Can I nav?">
         {messages from somewhere}
       </TutorDialog>
    <BS.MenuItem onSelect={@redirectToAccount} >
      {modal}
      My Account
    </BS.MenuItem>
```

Whereas now it's just:
```coffee
  redirectToAccount: ->
    TransitionAssistant.checkTransitionStateTo("Account Profile Page").then ->
      window.location.href = CurrentUserStore.getProfileUrl()
```

src/components/tutor-dialog.cjsx
nathanstitt: Same reasoning as below, in order to pass children to the Modal, the component would have to be the parent, and render it as part of it's render lifecycle.  By de-coupling the two, `TutorDialog` will be callable from anywhere.

philschatz: ah, right. ok ;)

src/components/tutor-dialog.cjsx
philschatz: Thanks for explaining

src/components/tutor-dialog.cjsx
```diff
@@ -0,0 +1,97 @@
...
+    </BS.Modal>
+
+module.exports = TutorDialog = React.createClass
+  displayName: 'TutorDialog'
+
+  propTypes: _.omit(DialogProperties, 'body')
+
+  componentDidMount: ->
+    TutorDialog.show(_.extend({}, @props, body: @props.children)).then(
+      => @props.onOk?()
```
philschatz: Why not just `.then(@props.onOk, @props.onCancel)`?

src/components/tutor-dialog.cjsx
```diff
@@ -0,0 +1,97 @@
...
+  statics:
+    show: (props) ->
+      new Promise (onOk, onCancel) =>
+        props = _.extend(_.clone(props), {
+          onOk, onCancel, show: true
+        })
+        if @dialog
+          @dialog.replaceProps(props)
+        else
+          div = document.body.appendChild( document.createElement('div') )
```
philschatz: Is it possible to clean up this div when closed?

src/components/tutor-dialog.cjsx
nathanstitt: Because (while unlikely) the props could be updated. If they were then the dialog would call the old function.  We still should pass on the result through though.  fixing that now

test/components/helpers/component-testing.coffee
```diff
@@ -9,9 +9,11 @@ ReactTestUtils = React.addons.TestUtils
 {commonActions} = require './utilities'
 sandbox = null
 ROUTER = null
+RENDERED = []
```
philschatz: is this dead code?

src/components/tutor-dialog.cjsx
philschatz: oh, and add a `-dialog` class on it?

test/components/helpers/component-testing.coffee
nathanstitt: Yep.  That was left over from when I was trying to figure out how best to unmount components after the spec was done with them.  I see it's also being reset in the beforeEach below, but not used anywhere else.

src/components/tutor-dialog.cjsx
philschatz: ok, makes sense ; )

src/components/tutor-dialog.cjsx
nathanstitt: Doesn't appear to be very easy to do so.  I get errors when I remove it immediately because the dialog component attempts to fade out the backdrop div in a timeout.  I suppose we could set an even longer timeout to work around that but seems like that might be brittle if when we upgrade react-bootstrap.

I did add a `-tutor-dialog-parent` class to the div though

philschatz: ah, this only creates the div once. ok ; )

PR #582, opened 2015-08-03
subheading text, align main title

darrenmason:
image


Fredasaurus:
Sweet. Can we make "individual chapter" bigger and bolder?

darrenmason:
made it the 2nd largest after main title

PR #583, opened 2015-08-03
Bug/builder external edit restrictions

pandafulmanda:
## Allow delete of all draft plans, even if opened
screen shot 2015-08-03 at 4 59 20 pm

## Disable URL editing of external after plan is published and opened
screen shot 2015-08-03 at 4 58 33 pm

### DIFF
--- a/src/components/task-plan/external/index.cjsx
+++ b/src/components/task-plan/external/index.cjsx
@@ -45,7 +45,7 @@ ExternalPlan = React.createClass
     if @state?.invalid then formClasses.push('is-invalid-form')
 
     isURLLocked = TaskPlanStore.isOpened(id) and TaskPlanStore.isPublished(id)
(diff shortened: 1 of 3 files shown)


src/components/task-plan/external/index.cjsx
```diff
@@ -41,8 +41,12 @@ ExternalPlan = React.createClass
 
     footer = <PlanFooter id={id} courseId={courseId} onPublish={@publish} onSave={@save}/>
     header = [headerText, closeBtn]
+    label = 'Assignment URL'
     if @state?.invalid then formClasses.push('is-invalid-form')
 
+    isURLLocked = TaskPlanStore.isOpened(id) and TaskPlanStore.isPublished(id)
+    label = "#{label} (Cannot be changed once plan is opened and published)" if isURLLocked
```
philschatz: I don't think teachers know what a plan is but the wording can be fixed later

PR #588, opened 2015-08-04
Match title displayed to selected entry from ToC

nathanstitt:
This way the title will always match the selected entry

With chapter 7 selected, before:
screen shot 2015-08-03 at 9 26 55 pm

After:
screen shot 2015-08-03 at 9 27 36 pm

## linked by page will also show title and can show active TOC item
screen shot 2015-08-04 at 3 58 54 pm

## Subtle loading
screen shot 2015-08-05 at 10 04 54 am
screen shot 2015-08-05 at 10 04 58 am

### DIFF
--- a/src/components/chapter-section-mixin.cjsx
+++ b/src/components/chapter-section-mixin.cjsx
@@ -18,6 +18,8 @@ module.exports =
     # ignore 0 in chapter sections
     sectionArray.pop() if skipZeros and _.last(sectionArray) is 0
 
+    console.info('sectionArray')
+    console.info(sectionArray)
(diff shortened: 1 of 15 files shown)


Fredasaurus:
yay thanks!

src/components/chapter-section-mixin.cjsx
```diff
@@ -4,14 +4,21 @@ module.exports =
   getInitialState: ->
     sectionSeparator: '.'
     skipZeros: true
+    inputStringSeparator: '.'
```
philschatz: When would this change during the lifetime of the instantiated component?

PR #589, opened 2015-08-04
remove random id

philschatz:
Teacher should not see the researcher id. From Micaela: "Against policy". 

Maybe a BE update to the api @jpslav?

image


gregfitch:
Yes, this is correct.

PR #590, opened 2015-08-04
Selenium smoketests

philschatz:
Runs some selenium tests that:

- create/edit/publish/delete a Reading (saves screenshot on failure)
- steps through the reference book, finding missing exercises
- reviews all assignments from the performance report

Mostly just playing around with Selenium to get feedback from FE/QA folks.

# Instructions

1. `npm install -g mocha`
2. `npm install` (to get the new packages)
3. `mocha -R spec ./test-integration` 
4. To run just some tests: `mocha --reporter spec --bail --compilers coffee:coffee-script/register ./test-integration/task-plan.coffee`

TODO:

- [x] report JS errors
- [x] Abstract reading creation so it can be used for "Save Draft", "Publish" and partial completion

# Directory Structure Design

By keeping the integration tests in the same repo:

1. updates to class names or other DOM features can be easily renamed
2. developers are more likely to keep integration tests up-to-date
3. new tests written by QA benefit everyone

## Files match `src/components` dirs

Because integration tests can take a long time, the names of the coffee files match the path in `src/components`.
This way the paths of the diff'd files in a Pull Request define which Selenium tests need to run.
If a file changes and is not in this path then all tests will run.

By reducing the amount of tests to run coders are more likely to run the tests themselves.

## `helpers` dir

The names of these helpers provide the frequent actions on a screen.
Some examples of screens are: 'Performance Report', 'Calendar', 'Homework Builder'.
They loosely correspond to directories in `src/components` but should (with refactoring) match them better.

Frequently, integration tests require setting up some data which requires going through
several screens before the actual testing occurs.
The files in this directory make that process easier.
Also, by reusing these helpers, when a feature changes behavior it is easier than hunting
through each test and updating 1-by-1.

For example, to verify validation messages in the Homework Builder the test first needs to:

1. Log in
2. Choose a course
3. Click "Add new Homework" from the Calendar

So, the test would look something like:

```coffee
App.login(TEACHER_USERNAME)
CourseSelect.goTo('ANY')            # Choose any course
Calendar.createNew('HOMEWORK')
ReadingBuilder.edit(action: 'SAVE') # Try to save

(validation tests would go here)
```

Notes:

- tests that are marked `(readonly)` will never change data (and could be run in production)
- tests that are marked `(idempotent)` do change data (create drafts) but are cleaned up (could be run on a copy of production)
- otherwise, tests are unsafe and will cause irreversible changes (like publishing and opening today)
- if you see `Element is not clickable at point` then make sure to call `@scrollTop()` first


pandafulmanda:
:+1: very c:sunglasses:l!  Also, well organized, I think it makes sense.

PR #594, opened 2015-08-05
exercises werent rendering because oops

pandafulmanda:
I left a question mark in there from when we were switching jsons back from the absolute url to the queries url


src/components/reference-book/exercise.cjsx
```diff
@@ -14,7 +14,7 @@ ReferenceBookExercise = React.createClass
   displayName: 'ReferenceBookExercise'
   render: ->
     {exerciseAPIUrl} = @props
-    {items} = ReferenceBookExerciseStore.get(exerciseAPIUrl)?
+    {items} = ReferenceBookExerciseStore.get(exerciseAPIUrl)
```
philschatz: Wow. I would not have expected a single question mark would cause so much trouble.

PR #592, opened 2015-08-05
Correct bar lengths on performance forecast, collapse order on dashboard

nathanstitt:
Medium (tablet) size performance forecast:
screen shot 2015-08-05 at 4 11 56 pm

Student dashboard:
screen shot 2015-08-05 at 4 03 54 pm

PR #598, opened 2015-08-05
Dashboard styles

Fredasaurus:
Made some typographic and positioning changes. Also added new book covers.

screen shot 2015-08-05 at 5 20 17 pm

Can we make the bars span 100% here?

screen shot 2015-08-05 at 5 20 17 pm copy


nathanstitt:
Looks nice.  The fix for 100% width bars is in #592.  That also changes the color key to use flexbox so it aligns vertically.  It'll probably conflict with some of your styles but we can stitch it together without too much trouble.

Fredasaurus:
thanks. Ah cool. that's a lot better than the kludgey height: 30px I put in there.

nathanstitt:
@Fredasaurus #592 was just merged and now this PR has a merge conflict.  Want me to fix it up?

Fredasaurus:
yeah.

PR #600, opened 2015-08-05
Bug/breadcrumbs covering task

pandafulmanda:
screen shot 2015-08-05 at 5 36 01 pm

### DIFF
--- a/resources/styles/components/course-calendar/index.less
+++ b/resources/styles/components/course-calendar/index.less
@@ -14,9 +14,10 @@
 }
 
 .is-not-open() {
-  content: @fa-var-eye-slash;
-  .fa-icon();
-  color: @tutor-neutral;
+  font-weight: 400;
+  font-style: italic;
+  content: 'opens ' attr(data-opens-at);
+  color: @tutor-neutral-darker;
 }
 
 .is-draft() {
--- a/resources/styles/components/task/index.less
+++ b/resources/styles/components/task/index.less
@@ -22,9 +22,8 @@
 
   .pinned-header {
     .tutor-shadow(1);
-    height: 60px;
     background: fadeout(@tutor-white, 20%);
-    padding: 8px @tutor-navbar-padding-horizontal;
+    padding: 8px @tutor-navbar-padding-horizontal 0;
     margin: 0;
   }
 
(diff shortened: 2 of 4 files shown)


philschatz:
I remember this working before and then it got reverted. Do you remember why?

pandafulmanda:
I think the 60px height was an attempt to vertically center the breadcrumbs within the bar.  By making the bottom padding 0, the breadcrumbs are centered, even with the active breadcrumb pushing the bottom down.

Fredasaurus:
There is another issue as well. We can't have the icon bar wraps to another line or it will start to make some pieces of content unusable -- especially sims. We need to find another solution to this this issue. Jason suggests a window pane approach with arrows on either side to scroll to the next pane if needed.

nathanstitt:
I wonder if we could do an effect similar to how the Safari web browser handles a similar situation (too many tabs to display).

It squishes all the tabs that are furthest away from the selected one so they're barely visible.  Then if you click on one it expands the ones closest while squishing the others furthest away.  Something like:

screen shot 2015-08-06 at 9 59 26 am

This would make the breadcrumbs furthest away from the current one difficult to click directly onto.  But that's ok because 1) they're probably not the one's you care about anyway.  2) if you miss (which is likely), all the one's around the the clicked breadcrumb will expand so you can easily click the desired one.

philschatz:
like this?
image

Fredasaurus:
I think that could work. I think it would work better if they weren't circles, another reason to change them. Getting this to work on resize, especially on mobile sizes could be tricky. Maybe at a certain size, say 460, they get replaced with a mobile menu.

Fredasaurus:
@philschatz That's interesting, as long as we can make it clear which step you are on. We don't have the benefit of (inferred affordance?) that goes with the mouse pointer being over one of the selections. In this example I can't tell weather its mail or chrome being selected.

philschatz:
Don't we already increase the size of the currently-active breadcrumb?

Fredasaurus:
Yes. If we are scaling the size of all of them, the increment between the active step and the rest may need to be more obvious.

pandafulmanda:
how about stacking previous breadcrumbs? kinda like...coverflow-ish

PR #604, opened 2015-08-06
adding universal colors for teacher edition

helenemccarron:
Removed os-teacher from theming mixin, added style specific to os-teacher feature

### DIFF
--- a/resources/styles/book-content/theming-mixins.less
+++ b/resources/styles/book-content/theming-mixins.less
@@ -14,9 +14,9 @@
 }
 
 .tutor-book-theme-note(@label-background: @tutor-book-primary; @label-text: @tutor-book-text-on-dark-labels) {
-  > .note,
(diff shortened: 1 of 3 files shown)


resources/styles/book-content/theming-mixins.less
```diff
@@ -14,9 +14,9 @@
 }
 
 .tutor-book-theme-note(@label-background: @tutor-book-primary; @label-text: @tutor-book-text-on-dark-labels) {
-  > .note,
-  > section > .note,
-  section > section > .note,
+  > .note:not(.os-teacher),
+  > section > .note:not(.os-teacher),
+  section > section > .note:not(.os-teacher),
```
philschatz: You could do this instead:

```less
> .note,
> section > .note,
section > section > .note {

  &:not(.os-teacher) {
    ...
  }
}
```

PR #606, opened 2015-08-06
Fix date range on dashboard this week panel

nathanstitt:
Luckily the store grabs all events for the week so at least this was
showing all the events, but was confusing for the title to be incorrect.

Snapshot taken on Aug 6th:

screen shot 2015-08-06 at 1 56 36 pm

### DIFF
--- a/src/components/student-dashboard/this-week-panel.cjsx
+++ b/src/components/student-dashboard/this-week-panel.cjsx
@@ -12,7 +12,7 @@ module.exports = React.createClass
     courseId: React.PropTypes.string.isRequired
 
   render: ->
-    startAt = moment(TimeStore.getNow())
+    startAt = moment(TimeStore.getNow()).startOf('isoweek')
     events  = StudentDashboardStore.weeklyEventsForDay(@props.courseId, startAt)
     if events.length
       <Events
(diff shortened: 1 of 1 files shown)


philschatz:
I'm guessing these dates should either be Aug 2-8 (Sun-Sat) or 3-7 (Mon-Fri). Thoughts @Fredasaurus ?

nathanstitt:
We're using monday-monday in the store to split the tasks into weeks, which was decided quite awhile ago for what a week should consist of.  If we change the display, I think we should also update the store logic to split the tasks.  Otherwise we could potentially end up with tasks showing on the panel that is  outside the displayed ranges.

philschatz:
Ah, in that case it sounds like `Aug 3-9 (Mon-Sun)` is the best option and maybe there's an off-by-1 with the top part of the range? Otherwise, why would "this week" contain items from "next monday".

nathanstitt:
Ah, I see.  Yes monday-monday does sound like it's giving them 8 days to work on it, when really it's not.  I subtracted a second from the range so it now reads Monday - Sunday.

screen shot 2015-08-06 at 3 16 42 pm

PR #607, opened 2015-08-06
Handling  too many breadcrumbs

pandafulmanda:
## Breadcrumbs shrink when there are too many to fit full-sized on one row
screen shot 2015-08-07 at 12 03 59 pm

### Before step completed and new breadcrumb
screen shot 2015-08-11 at 2 02 24 pm

### After new breadcrumb
screen shot 2015-08-11 at 2 02 30 pm

### self-adjusts as window size changes as well
screen shot 2015-08-11 at 2 02 34 pm

discussed with @Fredasaurus  and this is our temporary solution

### DIFF
--- a/src/components/task/breadcrumbs.cjsx
+++ b/src/components/task/breadcrumbs.cjsx
@@ -22,7 +22,7 @@ module.exports = React.createClass
   getInitialState: ->
     updateOnNext: true
     hoverCrumb: @props.currentStep
-    shouldCoverflow: false
+    shouldCoverflow: null
     coverflowWidth: null
 
   componentWillMount: ->
@@ -45,11 +45,13 @@ module.exports = React.createClass
     crumbs = @getCrumableCrumbs()
     @setState(coverflowWidth: (crumbs.length * 47))
 
-
   componentWillUnmount: ->
     TaskStepStore.setMaxListeners(10)
     TaskStore.off('task.beforeRecovery', @stopUpdate)
     TaskStore.off('task.afterRecovery', @update)
+
+  componentDidUpdate: ->
+    @_resizeListener(@state) if @state.windowEl.width? and not @state.shouldCoverflow?
 
   _resizeListener: (sizes) ->
     shouldCoverflow = @shouldCoverflow(sizes)
--- a/src/components/task/breadcrumbs.cjsx
+++ b/src/components/task/breadcrumbs.cjsx
@@ -22,8 +22,8 @@ module.exports = React.createClass
   getInitialState: ->
     updateOnNext: true
     hoverCrumb: @props.currentStep
-    shouldCoverflow: null
-    coverflowWidth: null
+    shouldShrink: null
+    crumbsWidth: null
 
   componentWillMount: ->
     listeners = @getMaxListeners()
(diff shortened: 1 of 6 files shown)


philschatz:
thoughts @Fredasaurus ?

Fredasaurus:
Not sure that I'm loving this. Also not sure how much space this will save.

pandafulmanda:
It's quite a bit, but i don't like how crowded it starts to look.  i might try make the breadcrumbs smaller first

Fredasaurus:
It seems to squish horizontally without providing perspective, the net result is that the icons just look weird. Compare to http://www.blogcdn.com/www.tuaw.com/media/2008/04/ical_events_in_cover_flow.jpg .

pandafulmanda:
hmm, we can definitely finagle the perspective until it looks right if that's the concern

philschatz:
Thoughts @Fredasaurus on the updated version of this? I noticed there's still a screen size that does not work for me:

image

But if it is slightly smaller it will shrink:
image

src/components/task/breadcrumbs.cjsx
```diff
@@ -39,11 +42,24 @@ module.exports = React.createClass
     # until the recovery step has been loaded
     TaskStore.on('task.afterRecovery', @update)
 
+    crumbs = @getCrumableCrumbs()
+    @setState(coverflowWidth: (crumbs.length * 47))
```
philschatz: How was `47` determined?

Fredasaurus:
Looks like there are some readings that will wrap before they shrink, so it's not an all-encompassing solution. However, I think it's still an improvement. One downside is that it reduces hit-state area on tablets. This is not a concern for beta necessarily. Amanda and I have an idea for a fix for this to work on after first release.

src/components/task/breadcrumbs.cjsx
```diff
@@ -39,11 +42,46 @@ module.exports = React.createClass
...
+
+  componentDidMount: ->
+    @calculateCrumbsWidth()
+
+  calculateCrumbsWidth: (crumbDOM) ->
+    if @isMounted()
+      currentCrumbWidth = 0
+      crumbsWidth = _.reduce(@refs, (memo, ref) ->
+        refDOM = ref.getDOMNode()
+        computedStyle = window.getComputedStyle(refDOM)
```
nathanstitt: No biggy since we're using this all over the place, but `getDOMNode` is being deprecated in favor of `React.findDOMNode`  At some point we'll have to go through and replace all instances of it.

src/components/task/breadcrumbs.cjsx
```diff
@@ -53,17 +91,31 @@ module.exports = React.createClass
...
+        onMouseLeave={@updateHoverCrumb.bind(@, @props.currentStep)}
+        onMount={@crumbMounted}
+        style={crumbStyle}
         crumb={crumb}
         currentStep={currentStep}
         goToStep={goToStep}
-        key="breadcrumb-#{crumb.type}-#{crumb.key}"/>
+        key="breadcrumb-#{crumb.type}-#{crumb.key}"
+        ref="breadcrumb-#{crumb.type}-#{crumb.key}"/>
+
```
nathanstitt: I like the effect this uses.  Seems pretty smooth

PR #609, opened 2015-08-06
Feature/calendar plan date to builder due date

pandafulmanda:
due_at is set instead of opens_at from calendar clicked date.  tests for cal updated. works with plan builder

if due date is tomorrow, sets open date to today

@jivewise does this look okay?


jivewise:
Yep, looks good, I'll do some testing as well.

jivewise:
Just tested, looks good!  I'll test a bit more during doomsday as well.

PR #615, opened 2015-08-07
Style tweaks for IE 10

nathanstitt:
Fixes 

**refbook arrows**
screen shot 2015-08-07 at 3 48 45 pm

**Guide key alignment**
screen shot 2015-08-07 at 2 03 48 pm

### DIFF
--- a/resources/styles/components/learning-guide/guide-key-mixin.less
+++ b/resources/styles/components/learning-guide/guide-key-mixin.less
@@ -4,6 +4,7 @@
   .item {
     .flex-display();
     .align-items(center);
+    -ms-flex-align: center;
     .box {
       width: 10px;
       height: 10px;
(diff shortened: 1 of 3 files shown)


resources/styles/components/learning-guide/guide-key-mixin.less
```diff
@@ -4,6 +4,7 @@
   .item {
     .flex-display();
     .align-items(center);
+    -ms-flex-align: center;
```
philschatz: Should these be in the `.align-items()` mixin?

nathanstitt: Yes, that makes sense.  The styles I added are the old flexbox types and are only used by IE 10.  Would be nice to yank them out eventually.

PR #617, opened 2015-08-10
Adding numbering to questions in Reference View

helenemccarron:
I used .ost-chapter-review to target every chapter review type. Let me know if it would be best to target each class individually to avoid using .ost- classes

### DIFF
--- a/resources/styles/book-content/questions.less
+++ b/resources/styles/book-content/questions.less
@@ -16,4 +16,5 @@
 }
 .answers-answer {
   clear: both;
+  margin-left: 30px;
 }
--- a/resources/styles/book-content/questions.less
+++ b/resources/styles/book-content/questions.less
@@ -5,9 +5,15 @@
   &:before {
     counter-increment: question;
     content: counter(question) ". ";
-    display: inline;
+    display: block;
+    float: left;
+    width: 4%;
   }
   .question-stem {
-    display: inline;
+    float: left;
+    width: 96%;
   }
+}
+.answers-answer {
+  clear: both;
 }
--- a/resources/styles/book-content/questions.less
+++ b/resources/styles/book-content/questions.less
@@ -1,0 +1,13 @@
+.ost-chapter-review {
+  counter-reset: question;
+}
+.question {
+  &:before {
+    counter-increment: question;
+    content: counter(question) ". ";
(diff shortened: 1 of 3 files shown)


Fredasaurus:
Awesome. Can we get the number to sit to the left of the question like the "a)'?

philschatz:
I think using `.ost-` in tutor is using the class correctly. It might be nicer to use another class if there is one (to eventually use the styling outside of tutor like in webview but that's far away)

resources/styles/book-content/questions.less
```diff
@@ -0,0 +1,13 @@
+.ost-chapter-review {
+  counter-reset: question;
```
philschatz: Can/should the `question` counter just reset once (on the top of the page)?

helenemccarron: You mean add the reset in the base instead each section? The question is do we want the question numbering to be continuous or to reset for each section (ie critical thinking, review questions etc). In bio, chapter 3.1, for example, you have several types of questions, in my opinion it makes more sense to reset the questions for each section.

helenemccarron: @Fredasaurus yep, done :)

philschatz: I think that if we start with simple numbering the teacher can tell the students: "Go to ref book, 2.2, and do exercises 1,4,5,7,22" instead of "... and do exercises 1,4 in Concept Review and 2,3 in ...."

helenemccarron: I talked to Derek about the pdf strategy for numbering and right now it's comparing apples and oranges since the items don't collate here. The numbering is ad-hoc for the pdf and varies widely from one pdf to another. So I think right now, we don't have a way of designing this particular problem with the pdf in mind. If we want to do so, we should ask W&N and figure out what they want, but the answer will be different for Bio and for HS Physics, and will apply to end of chapter collated items.

resources/styles/book-content/questions.less
```diff
@@ -0,0 +1,17 @@
+.question {
```
helenemccarron: current commit reflects the questions being reset at the page level

philschatz: Thanks!

PR #618, opened 2015-08-10
Fix plan label not opening details when there's 2 that represent that plan

pandafulmanda:
## Clicking on first half
screen shot 2015-08-12 at 12 53 43 pm

## Clicking on second half
screen shot 2015-08-12 at 12 53 47 pm

Order does not matter

## Handles multiple week beyond 2 weeks properly
screen shot 2015-08-12 at 5 25 17 pm

Spacing tweaked and improved as well

Changes are based on the idea that the plan should be one component, even if it's being displayed by 2 or more elements due to splitting over weeks.  Displays are then their own component, concerned only with handling the display of the plan.

Before, there could be multiple plan components per actual plan, and this required syncing hover and clicked state across components with [janky direct element maniputation](https://github.com/openstax/tutor-js/blob/master/src/components/course-calendar/plan.cjsx#L149-L173)  as opposed to now with [a shared state update on the plan which passes down to the displays](https://github.com/openstax/tutor-js/blob/bug/calendar-plan-details-mistrigger/src/components/course-calendar/plan.cjsx#L122-L132). 

Also, before, clicking on one chunk of a plan sometimes meant triggering the click on another part of the plan to open the modal, which led to problems.  Now, all displays for a plan are triggers for the same plan modal.

If a plan is published, will show quicklook.  If a plan is publishing for the first time, will show publishing modal.  If a plan is in draft, will link to edit page.

### DIFF
--- a/src/components/course-calendar/duration.cjsx
+++ b/src/components/course-calendar/duration.cjsx
@@ -227,6 +227,8 @@ CourseDuration = React.createClass
             plan: simplePlans[plan.id]
             index: counter[plan.id]
 
+          planForRange.offsetFromPlanStart = planForRange.rangeDuration.start.diff(plan.duration.start, 'days')
+
           # Add plan to plans in range
           rangeData.plansInRange.push(planForRange)
           counter[plan.id] = counter[plan.id] + 1
--- a/src/components/course-calendar/duration.cjsx
+++ b/src/components/course-calendar/duration.cjsx
@@ -12,6 +12,7 @@ CourseDuration = React.createClass
 
   propTypes:
     durations: React.PropTypes.array.isRequired
+    courseId: React.PropTypes.string.isRequired
     viewingDuration: React.PropTypes.instanceOf(twix).isRequired
     groupingDurations: React.PropTypes.arrayOf(React.PropTypes.instanceOf(twix)).isRequired
     referenceDate: (props, propName, componentName) ->
--- a/src/components/course-calendar/duration.cjsx
+++ b/src/components/course-calendar/duration.cjsx
@@ -41,8 +41,11 @@ CourseDuration = React.createClass
         rangedPlan.plan.id
       )
       .map((groupedPlans) ->
+        # Grab the first plan as representative plan for displays grouped by plan
         {plan} = _.first(groupedPlans)
 
+        # omit the plan on the individual displays since they share one common plan
+        # TODO: clean up so that plan doesn't get duplicated in the first place.
         displays = _.map(groupedPlans, (groupedPlan) ->
           _.omit(groupedPlan, 'plan')
         )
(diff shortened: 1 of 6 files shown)


philschatz:
Will this handle the case of an item spanning 2 weeks? like:

image

src/components/course-calendar/duration.cjsx
```diff
@@ -36,6 +37,18 @@ CourseDuration = React.createClass
     durationsByStartDate = _.chain(groupedDurations)
       .pluck('plansInRange')
       .flatten()
+      .groupBy((rangedPlan) ->
+        rangedPlan.plan.id
+      )
+      .map((groupedPlans) ->
+        {plan} = _.first(groupedPlans)
+
+        displays = _.map(groupedPlans, (groupedPlan) ->
```
philschatz: Why is the plan removed here? Can you add a comment?

src/components/course-calendar/duration.cjsx
```diff
@@ -183,6 +196,14 @@ CourseDuration = React.createClass
     if calcedHeight > rangeData.dayHeight
       rangeData.dayHeight = calcedHeight
 
+  _getSimplePlan: (plan) ->
+    simplePlan = _.omit(plan, 'due_at', 'opens_at', 'duration', 'durationAsWeeks')
```
philschatz: Would specifically including, rather than omitting be shorter? It is not clear why these fields specifically are excluded and which ones are included.

src/components/course-calendar/plan.cjsx
```diff
@@ -36,12 +40,10 @@ CoursePlan = React.createClass
...
 
   _isPlanNotMatchingRouteOpen: ->
-    {planId} = @context.router.getCurrentParams()
-    not @_doesPlanMatchesRoute() and @state.isViewingStats
+    not (@_doesPlanMatchesRoute()) and @refs.details?
 
   _isPlanMatchRouteNotOpen: ->
-    {planId} = @context.router.getCurrentParams()
-    @_doesPlanMatchesRoute() and not @state.isViewingStats
+    (@_doesPlanMatchesRoute()) and not @refs.details?
```
nathanstitt: These method names hurt my brain.  Which probably says more about my brain than the names, but still...

What I think they're doing is checking if the current route matches the given task, and if the modal is open.  If that's the case why not rename to `isMatchingRouteOpen`?  Or just `isCurrentStateValid()`, since it looks like they're used to determine if the stats should be updated or the modal closed below.

src/components/course-calendar/plan.cjsx
```diff
@@ -103,190 +106,104 @@ CoursePlan = React.createClass
...
-    # top is calculated by using:
-    #   weekTopOffset -- the distance from the top of the calendar for plans in the same week
-    #   order -- the order the plan should be from the bottom, is an int more than 1 when a plan needs to
-    #       stack on top of other plans that overlap in duration.
-    planStyle =
-      width: durationLength * 100 / 7 + '%'
-      left: offset * 100 / 7 + '%'
-      top: (weekTopOffset + 4 - order * 3) + 'rem'
+  removeHover: ->
+    @setState(isHovered: false) if @state.isHovered
```
nathanstitt: You might consider just having a single `setHover` method that would accept a boolean and hand that method off the the components in `render`.  You could still use the optimization of not calling setState unless you need to.

src/components/course-calendar/plan.cjsx
```diff
@@ -103,190 +106,104 @@ CoursePlan = React.createClass
...
+
+      <displayComponent
+        {...displayComponentProps}
+        ref="display#{index}"
+        key="display#{index}"/>
+    )
+
+    <div>
+      {planDisplays}
+    </div>
```
nathanstitt: `render` is getting right at the edge of too long for me. I'd consider splitting it up.  A simple way to do so would be to make a `renderDisplay` method and call change line 174 to be `_.map(displays, => @renderDisplay(...))`.

nathanstitt:
@pandafulmanda done with review.  The few things I found are really just quibbles, and the method names thing wasn't even introduced in this PR.  We can merge if you're ok with it.

PR #619, opened 2015-08-11
New clue attribute with all elements of data

nathanstitt:
This updates the FE to handle the additional CLUE values that the BE is now sending.

The Performance Forecast will now only hide the bar if the `sample_size_interpretation` is `below`.  

It now has a secret debug option that will display all the values for debugging.  The option is an invisible pi symbol [ [background](https://www.google.com/search?client=safari&rls=en&q=the+net+pi+symbol&ie=UTF-8&oe=UTF-8) ] at the bottom left corner of the screen.  It's normally transparent, but turns grey when the mouse is hovered over it.  When clicked, display of all CLUE values is toggled.

@Fredasaurus any thoughts on the debug option?

screen shot 2015-08-11 at 1 24 51 pm

### DIFF
--- a/src/components/learning-guide/guide.cjsx
+++ b/src/components/learning-guide/guide.cjsx
@@ -27,7 +27,7 @@ module.exports = React.createClass
 
   renderWeaker: ->
     # sort sections by current level of understanding
-    sortedSections = _.sortBy(@props.allSections, 'current_level')
+    sortedSections = _.sortBy(@props.allSections, (s) -> s.clue.value )
     # if there are less than 4 sections, use 1/2 of the available ones
     weakStrongCount = Math.min(sortedSections.length / 2, 4)
 
--- a/src/components/learning-guide/progress-bar.cjsx
+++ b/src/components/learning-guide/progress-bar.cjsx
@@ -7,17 +7,21 @@ module.exports = React.createClass
   displayName: 'LearningGuideProgressBar'
 
   propTypes:
-    section: React.PropTypes.object.isRequired
+    section:  React.PropTypes.object.isRequired
     onPractice: React.PropTypes.func
 
   render: ->
     {section, onPractice} = @props
     {clue} = section
 
-    percent = Math.round((clue.value / 1) * 100)
-
-    # always show at least 5% of bar, otherwise it apears empty
-    bar = <BS.ProgressBar className={clue.value_interpretation} now={Math.max(percent, 5)} />
+    bar = if clue.sample_size_interpretation is 'below'
+      <span className="no-data">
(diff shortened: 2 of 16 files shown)


src/components/learning-guide/statistics.cjsx
```diff
@@ -0,0 +1,29 @@
...
+  propTypes:
+    section:  ChapterSectionType.isRequired
+    diplaying: React.PropTypes.string.isRequired
+
+  render: ->
+    <div className='statistics'>
+      <ul className='clue visible-when-debugging'>
+        { for key, value of @props.section.clue
+          value = value.join(' ') if _.isArray(value)
+          <li key={key}>{key}: {value}</li>}
```
pandafulmanda: how about a ```<strong>```  around the ```key```?

src/components/learning-guide/guide.cjsx
```diff
@@ -27,7 +27,7 @@ module.exports = React.createClass
 
   renderWeaker: ->
     # sort sections by current level of understanding
-    sortedSections = _.sortBy(@props.allSections, 'current_level')
+    sortedSections = _.sortBy(@props.allSections, (s) -> s.clue.value )
     # if there are less than 4 sections, use 1/2 of the available ones
     weakStrongCount = Math.min(sortedSections.length / 2, 4)
```
pandafulmanda: Minor: Could the `4` be a prop with default value of 4 or something instead since it's a configuration that may change?  or weakStrongCount could even be a prop that's a function, with the default set to something likek `Math.min(sortedSections.length / 2, 4)`

src/components/learning-guide/progress-bar.cjsx
```diff
@@ -11,16 +11,17 @@ module.exports = React.createClass
...
-      <BS.ProgressBar className={section.interpretation} now={Math.max(percent, 5)} />
-    else
+    bar = if clue.sample_size_interpretation is 'below'
       <span className="no-data">
         {if onPractice then 'Practice more to get forecast' else 'Not enough exercises completed'}
       </span>
+    else
+      percent = Math.round((clue.value / 1) * 100)
+      # always show at least 5% of bar, otherwise it just looks empty
+      <BS.ProgressBar className={clue.value_interpretation} now={Math.max(percent, 5)} />
```
pandafulmanda: same here with the `5` -- could this be a prop?

src/components/learning-guide/guide.cjsx
nathanstitt: This is all changing on next PR, so if we could merge this, I double promise to fix in a bit :)

pandafulmanda:
cool new stats component =), looks great overall.  i had to be super nit-picky to have anything to say

PR #626, opened 2015-08-11
fix mis-sizing of component from resize listener

pandafulmanda:
Component was confused about it's size, so it would sometimes shrink things initially when it didn't need to.  It would size appropriate on window resize or next crumb update.

## Proper initial sizing
screen shot 2015-08-11 at 5 58 39 pm

## Bug/before -- confused initial sizing
screen shot 2015-08-11 at 5 58 14 pm

### DIFF
--- a/src/components/resize-listener-mixin.cjsx
+++ b/src/components/resize-listener-mixin.cjsx
@@ -19,7 +19,7 @@ module.exports =
     @resizeListener = _.throttle(@resizeEffect, @state.resizeThrottle or @props.resizeThrottle)
 
   componentDidMount: ->
-    _.delay(@setInitialSize, 0)
+    _.defer(@setInitialSize)
     window.addEventListener('resize', @resizeListener)
 
   componentWillUnmount: ->
--- a/src/components/resize-listener-mixin.cjsx
+++ b/src/components/resize-listener-mixin.cjsx
@@ -19,7 +19,7 @@ module.exports =
     @resizeListener = _.throttle(@resizeEffect, @state.resizeThrottle or @props.resizeThrottle)
 
   componentDidMount: ->
-    @setInitialSize()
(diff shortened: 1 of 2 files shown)


src/components/resize-listener-mixin.cjsx
```diff
@@ -19,7 +19,7 @@ module.exports =
     @resizeListener = _.throttle(@resizeEffect, @state.resizeThrottle or @props.resizeThrottle)
 
   componentDidMount: ->
-    @setInitialSize()
+    _.delay(@setInitialSize, 0)
     window.addEventListener('resize', @resizeListener)
```
nathanstitt: You could just use `_.defer` here.  It's almost exactly the same (but uses 1ms).  http://underscorejs.org/docs/underscore.html#section-81

src/components/task/breadcrumbs.cjsx
```diff
@@ -75,12 +76,12 @@ module.exports = React.createClass
   crumbMounted: ->
     @calculateCrumbsWidth() if @state.crumbsWidth?
 
-  _resizeListener: (sizes) ->
-    shouldShrink = @shouldShrink(sizes)
-    @setState({shouldShrink})
+  didWidthsChange: (prevState, currentState) ->
+    currentState.crumbsWidth isnt prevState.crumbsWidth or currentState.componentEl.width isnt prevState.componentEl.width
```
nathanstitt: Shouldn't it just be `didWidthChange` (singular?)  An element has only one width.

nathanstitt:
Done w/review.  These are just tiny quibbles.  We can merge if you're ok with it.

src/components/task/breadcrumbs.cjsx
pandafulmanda: refers to either the width of of the crumbs having changed or the component width having changed

PR #628, opened 2015-08-12
Add sampleSizeThreshold property to progress bar

nathanstitt:
Allow the teacher and student components to set a value for the # of samples needed to render the progress bar.  If the value given is exceeded the bar will render regardless of what the clue's sample_size_interpretation value is.

### DIFF
--- a/src/components/learning-guide/progress-bar.cjsx
+++ b/src/components/learning-guide/progress-bar.cjsx
@@ -6,15 +6,21 @@ ChapterSectionType = require './chapter-section-type'
 module.exports = React.createClass
   displayName: 'LearningGuideProgressBar'
 
+  getDefaultProps: ->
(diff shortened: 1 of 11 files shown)


src/components/learning-guide/progress-bar.cjsx
```diff
@@ -6,15 +6,21 @@ ChapterSectionType = require './chapter-section-type'
 module.exports = React.createClass
   displayName: 'LearningGuideProgressBar'
 
+  getDefaultProps: ->
```
philschatz: Why provide a default threshold *and* make it required? I'll merge anyway

nathanstitt: Because I got paranoid that I missed updating a component and it wouldn't get passed and would break horribly :)

PR #629, opened 2015-08-12
Feature/plan delete separate store

pandafulmanda:
Should (and appears to) fix bug where plans that have been requested for deletion remain on the calendar.

@nathanstitt what do you think?


nathanstitt:
:+1:  Looks good to me :grin: Better than my overly clever #630 anyways

PR #630, opened 2015-08-13
Hide deleting task plans from calendar

nathanstitt:


nathanstitt:
Closing in favor of #629  That one's less code and doesn't require overloading methods `CrudConfig` methods.

PR #631, opened 2015-08-13
Prevent text wrapping on exercise images

nathanstitt:
Before:

screen shot 2015-08-12 at 7 43 45 pm

After:

screen shot 2015-08-12 at 7 43 16 pm

### DIFF
--- a/resources/styles/book-content/questions.less
+++ b/resources/styles/book-content/questions.less
@@ -9,6 +9,7 @@
   .question-stem {
     float: left;
     width: 96%;
+    img { display: block; }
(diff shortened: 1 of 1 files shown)


nathanstitt:
@helenemccarron should wrapping text on all exercise question images be enforced or do we need to somehow only target smaller ones?

Fredasaurus:
I think we should not wrap on any.

helenemccarron:
I agree with @Fredasaurus, in my opinion, questions are their own thing and should stand out as such.

PR #633, opened 2015-08-13
Fix incorrect logic when evaluating sample size limit

nathanstitt:
The "no data" message should be shown if the sample_size was below the
threshold, not above it. Lukily the interpetation had to also be 'below'
otherwise the no-data message would be shown all the time

### DIFF
--- a/src/components/learning-guide/progress-bar.cjsx
+++ b/src/components/learning-guide/progress-bar.cjsx
@@ -6,10 +6,6 @@ ChapterSectionType = require './chapter-section-type'
 module.exports = React.createClass
   displayName: 'LearningGuideProgressBar'
 
-  getDefaultProps: ->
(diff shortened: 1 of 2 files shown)


src/components/learning-guide/progress-bar.cjsx
```diff
@@ -6,10 +6,6 @@ ChapterSectionType = require './chapter-section-type'
 module.exports = React.createClass
   displayName: 'LearningGuideProgressBar'
 
-  getDefaultProps: ->
-    # specify a threshold just in case a parent component fails to pass it on
-    sampleSizeThreshold: 3
-
   propTypes:
```
nathanstitt: Phil suggested we remove this since the prop is also required

jivewise:
+1, looks good to me.

PR #634, opened 2015-08-13
Bug: No way to come back to dashboard after click on Account

darrenmason:
Regarding issue:  https://www.pivotaltracker.com/n/projects/1156756/stories/101144278

image


pandafulmanda:
:+1:

nathanstitt:
This'll certainly work fine.  But the AccountLink also checks for unsaved work.  Since we're now opening in a new tab, it seems to me that we shouldn't bother the user with the dialog?

PR #637, opened 2015-08-13
Updated accordion UI styles, added mixin for fadein on selected readings

Fredasaurus:
Also made a couple to term changes:
Currently selected readings in this reading > Currently selected,
Browse the Book > Browse this Chapter

screen shot 2015-08-13 at 6 02 38 pm


Fredasaurus:
A word on the mixin: It is a fair amount of code for a small interaction, and it *seems* like something that can be reused, but as far as I know this is the only place something like this is being used currently, so if you want me to roll it back into reading .less or do something else I can.

nathanstitt:
It's a nifty effect, and I'd say there's value in having it be available to other components. 

It is a bit large, but not overly so.  If it bothers you, you could move it into the `mixins` subdirectory and include it a the top the way the `flexbox` mixin's are.

PR #639, opened 2015-08-14
Fixes for performance report behavior

nathanstitt:
Updates the student dashboard to show all sections, not the weakest ones.  Only hides the weaker section when there is none or only 1 section.  Adds quite a few tests for the components and store to make sure it's working as intended.

For example,

## Without Fix
screen shot 2015-08-17 at 12 18 19 pm

## With Fix
screen shot 2015-08-17 at 12 19 33 pm


src/components/learning-guide/weaker-panel.cjsx
```diff
@@ -16,22 +16,8 @@ WeakerPanel = React.createClass
...
-    # Sort by value and pick 'displayCount' of the weakest
-    weakestSections = _.chain(validSections)
-      .sortBy((s) -> s.clue.value )
-      .first(displayCount)
-      .value()
-
-    for section, i in weakestSections
-      <Section key={i} section={section} {...@props} />
-
   render: ->
```
nathanstitt: This is all done by the WeakerSections component below

pandafulmanda:
screen shot 2015-08-17 at 1 43 31 pm
is this intended?

nathanstitt:
Great catch!  I think that's been in there code for awhile actually.  Fixed now

screen shot 2015-08-17 at 2 14 46 pm

pandafulmanda:
nice, much cleaner

pandafulmanda:
lingering thing where switching students in teacher viewing as student view lags from trying to load from BE and does not show as loading -- to be addressed later

nathanstitt:
I investigated that a little bit.  Since that store uses two id's (courseId & roleId), it can't easily use the standard loading stuff.  We could probably just use bindStore and show our own message.

PR #640, opened 2015-08-14
plan is only editable if no due date is before today

pandafulmanda:
## Before
screen shot 2015-08-14 at 2 01 55 pm
screen shot 2015-08-14 at 2 02 06 pm

### Don't show edit button for plans with due date before today
screen shot 2015-08-14 at 1 59 11 pm

### Show edit button if due in future
screen shot 2015-08-14 at 1 59 21 pm

@Fredasaurus  would this work?

### DIFF
--- a/src/components/course-calendar/duration.cjsx
+++ b/src/components/course-calendar/duration.cjsx
@@ -173,6 +173,7 @@ CourseDuration = React.createClass
     plan.isPublished = (plan.published_at? and plan.published_at)
     plan.isPublishing = @isPlanPublishing(plan)
     plan.isTrouble = plan.is_trouble
+    plan.isEditable = not plan.duration.start.isBefore(referenceDate, 'day')
 
   # TODO see how to pull out plan specific logic to show that this
   # can be reused for units, for example
--- a/src/components/course-calendar/plan-details.cjsx
+++ b/src/components/course-calendar/plan-details.cjsx
@@ -37,6 +37,9 @@ CoursePlanDetails = React.createClass
     editLinkName = camelCase("edit-#{type}")
 
     reviewButton = @renderReviewButton()
+    editButton = <Router.Link to={editLinkName} params={linkParams}>
+      <BS.Button>Edit Assignment</BS.Button>
+    </Router.Link> if plan.isEditable
 
     <BS.Modal
       {...@props}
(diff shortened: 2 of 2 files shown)


src/components/course-calendar/duration.cjsx
```diff
@@ -173,6 +173,7 @@ CourseDuration = React.createClass
     plan.isPublished = (plan.published_at? and plan.published_at)
     plan.isPublishing = @isPlanPublishing(plan)
     plan.isTrouble = plan.is_trouble
+    plan.isEditable = not plan.duration.start.isBefore(referenceDate, 'day')
 
   # TODO see how to pull out plan specific logic to show that this
```
nathanstitt: Why not: `plan.isEditable = plan.duration.start.isAfter(referenceDate, 'day')`.  That removes the `not` and *seems* to read better to me.

src/components/course-calendar/plan-details.cjsx
```diff
@@ -37,6 +37,9 @@ CoursePlanDetails = React.createClass
     editLinkName = camelCase("edit-#{type}")
 
     reviewButton = @renderReviewButton()
+    editButton = <Router.Link to={editLinkName} params={linkParams}>
+      <BS.Button>Edit Assignment</BS.Button>
+    </Router.Link> if plan.isEditable
```
nathanstitt: I've struggled with the cleanest way to write conditional jsx with coffeescript.  The best I've found thus far is:

```coffee
    if plan.isEditable then editButton =
      <Router.Link to={editLinkName} params={linkParams}>
        <BS.Button>Edit Assignment</BS.Button>
      </Router.Link>
```

You can also put a `\` after the `if` line to maybe make it more clear that it's a continuation.
i.e. ` if plan.isEditable then editButton = \`  I'm not sure if that adds much or not though.

src/components/course-calendar/duration.cjsx
pandafulmanda: That will make ```isEditable``` include things that are due today :smile:. I tried that initially.

src/components/course-calendar/plan-details.cjsx
pandafulmanda: Hmmm, this something I do semi-consistently throughout other code I've written.  The other way I've written it is:
```coffee
    if plan.isEditable
      editButton = <Router.Link to={editLinkName} params={linkParams}>
          <BS.Button>Edit Assignment</BS.Button>
        </Router.Link>
```

nathanstitt: Nope.  Just pointed that out in case you didn't know it :)

PR #641, opened 2015-08-14
My Account link open in new tab/window

darrenmason:
Regarding issue:  https://www.pivotaltracker.com/n/projects/1156756/stories/101144278

This time not check unsaved state since it's a new window/tab being opened.

image

### DIFF
--- a/src/components/navbar/account-link.cjsx
+++ b/src/components/navbar/account-link.cjsx
@@ -11,8 +11,10 @@ module.exports = React.createClass
     router: React.PropTypes.func
 
   redirectToAccount: ->
-    TransitionAssistant.checkTransitionStateTo("Account Profile Page").then ->
-      window.open(CurrentUserStore.getProfileUrl())
+    # workaround for 'No way to come back to dashboard after click on Account'
+    window.open(CurrentUserStore.getProfileUrl())
+    #TransitionAssistant.checkTransitionStateTo("Account Profile Page").then ->
+      #window.location.href = CurrentUserStore.getProfileUrl()
 
   render: ->
     return null unless CurrentUserStore.getProfileUrl()
--- a/src/components/navbar/account-link.cjsx
+++ b/src/components/navbar/account-link.cjsx
@@ -11,10 +11,7 @@ module.exports = React.createClass
     router: React.PropTypes.func
 
   redirectToAccount: ->
-    # workaround for 'No way to come back to dashboard after click on Account'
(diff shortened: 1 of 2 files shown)


nathanstitt:
You've angered Travis :)  `AccountLink` has a spec that was expecting unsaved state to be checked.

darrenmason:
disabled state check in account-link spec

src/components/navbar/account-link.cjsx
```diff
@@ -11,8 +11,10 @@ module.exports = React.createClass
     router: React.PropTypes.func
 
   redirectToAccount: ->
-    TransitionAssistant.checkTransitionStateTo("Account Profile Page").then ->
-      window.open(CurrentUserStore.getProfileUrl())
+    # workaround for 'No way to come back to dashboard after click on Account'
+    window.open(CurrentUserStore.getProfileUrl())
+    #TransitionAssistant.checkTransitionStateTo("Account Profile Page").then ->
```
nathanstitt: Should we probably give the new window a name that way if they click the account link twice it'll re-use the first window rather than opening a new tab?  Something like `window.open(CurrentUserStore.getProfileUrl(), 'account-profit')` should do that I think?  https://developer.mozilla.org/en-US/docs/Web/API/Window/open

Also, why not remove the commented code for `TransitionAssistant` and `window.location`?  We can always pull it out of git history if we need to review.

test/components/navbar/account-link.spec.coffee
```diff
@@ -19,4 +19,4 @@ describe 'Account Link', ->
     CurrentUserActions.loaded(USER)
     Testing.renderComponent( Link ).then ({dom}) ->
       Testing.actions.click(dom.querySelector('a'))
-      expect(TransitionAssistant.checkTransitionStateTo).to.have.been.called
+      #expect(TransitionAssistant.checkTransitionStateTo).to.have.been.called
```
nathanstitt: If we do indeed want to disable the spec, we should use `xit` for the spec.  But in this case we don't want to use the `TransitionAssistant` at all, so why not change the spec name to be "does not use TransitionAssistant when redirecting" and then change the condition to be `not.to.have.been.called` ?

src/components/navbar/account-link.cjsx
pandafulmanda: :+1: agreeing with @nathanstitt

darrenmason:
roger.

nathanstitt:
:+1: Looks good

PR #644, opened 2015-08-17
Fix confirm leave

jivewise:
screen shot 2015-09-23 at 6 30 04 am

-Fixing the confirm leave functionality on task plan builders
-Updating the alert to use the tutor dialog that @nathanstitt wrote
-Updated select topics and close buttons to use tutor dialog as well
-Updated wording on confirm dialog


jason-holmes:
Could we have the OK/Cancel pattern on these dialogs be this instead?




-------
Jason Holmes, Ph.D.
User Experience Manager
OpenStax | OpenStax.org
(713) 893-1326

nathanstitt:
@jason-holmes Yes, we could update the dialog to have that appearance.  For some reason github didn't display the image, but this is it:

dialog_pattern

I'm not 100% sure what the current UX best practice is for the old "OK displayed first, OK displayed last" debate.  It used to be that you should follow the platform conventions of the majority of your users.  Mac displays OK last, Windows first.  I'd guess that most of our audience (students) will probably be Mac users?  @Fredasaurus, do we have any data on the predicted platform usage?

jason-holmes:
I’m recommending Ok/Cancel rather than Cancel/OK position not based on assumed or guessed OS of user, but instead on web form design best practice we’ve tried to keep throughout Tutor.  Actions are left aligned with text. Primary action first, secondary action second.

-------
Jason Holmes, Ph.D.
User Experience Manager
OpenStax | OpenStax.org
(713) 893-1326

philschatz:
This PR has conflicts with master, can you merge the conflicts?

philschatz:
It seems this confirm dialog only shows up when pressing "Cancel", not when pressing the "X" in the top-right. How much work would it be to get the "X" to work the same way (it always shows up, and shows up in the "Select Readings" dialog as well)?

Dantemss:
The button text is ambiguous.

Are you sure you want to Cancel?

OK vs Cancel? What does OK do?

Fredasaurus:
I would make this "yes" and "no"

PR #649, opened 2015-08-18
Links in ref book and iReadings need some updates

pandafulmanda:
- [X] links that are linking to a figure found in itself should have target='_self'
- [X] check link that are not [link] text for media linking as well
- [X] update to use book_id

### links that are linking to a figure found in itself should have target='_self'
1. Clicking on link
  screen shot 2015-08-17 at 7 33 37 pm
1. After link is clicked, focuses on targeted figure, on same page/tab
  screen shot 2015-08-17 at 7 33 27 pm

### check links that are not "[link]" as link text for media linking as well
screen shot 2015-08-17 at 7 31 36 pm
### update to use book_id
  * see link in bottom right hand corner of previous SS.  shows properly formed reference book link

### links linking to a figure within page can open in new window
1. Clicking on link
  screen shot 2015-08-17 at 7 43 00 pm
1. After link is clicked, opens to figure in new tab
  screen shot 2015-08-17 at 7 43 03 pm

### ignores links that don't have "[link]" as link text
* not properly linked because link does not get processed
   screen shot 2015-08-17 at 7 42 18 pm

### Errors on links that aren't in step, but in reference book
screen shot 2015-08-17 at 7 48 17 pm

### DIFF
--- a/src/components/book-content-mixin.cjsx
+++ b/src/components/book-content-mixin.cjsx
@@ -1,6 +1,7 @@
 _ = require 'underscore'
 S = require '../helpers/string'
 dom = require '../helpers/dom'
+{CourseStore} = require '../flux/course'
 
 # According to the tagging legend exercises with a link should have `a.os-embed`
 # but in the content they are just a vanilla link.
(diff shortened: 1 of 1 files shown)


nathanstitt:
looks good to me!

PR #654, opened 2015-08-19
Fast MathJax rendering

Dantemss:
Gotta render fast!

### DIFF
--- a/src/helpers/mathjax.coffee
+++ b/src/helpers/mathjax.coffee
@@ -37,11 +37,8 @@ typesetMath = (root) ->
     # mark node as processed
     node.classList.add('math-rendered')
 
-  # convert NodeList to Array
-  node_array = Array.prototype.slice.call(nodes)
-
   # submit all nodes at once for optimal rendering performance
-  window.MathJax.Hub.Queue(['Typeset', MathJax.Hub, node_array])
+  window.MathJax.Hub.Queue(['Typeset', MathJax.Hub, _.toArray(nodes)])
 
   # render MathML with MathJax
   window.MathJax.Hub.Queue(['Typeset', MathJax.Hub, root]) if hasMath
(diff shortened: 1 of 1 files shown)


Dantemss:
I was bored :)

src/helpers/mathjax.coffee
```diff
@@ -34,10 +34,15 @@ typesetMath = (root) ->
...
-    window.MathJax.Hub.Queue(['Typeset', MathJax.Hub, node])
     # mark node as processed
     node.classList.add('math-rendered')
 
+  # convert NodeList to Array
+  node_array = Array.prototype.slice.call(nodes)
+
+  # submit all nodes at once for optimal rendering performance
+  window.MathJax.Hub.Queue(['Typeset', MathJax.Hub, node_array])
+
```
nathanstitt: Wow if this works, it should indeed be more efficient.  I wasn't aware that you could pass an array into `Hub.Queue`

You could just use underscore's _.toArray though:
`window.MathJax.Hub.Queue(['Typeset', MathJax.Hub, _.toArray(nodes)])`

That's assuming that `Hub.Queue` doesn't work with just a  `NodeList` ?

Fredasaurus: When Fred gets bored he reads about sports.

Dantemss: Feel free to patch it, Nathan, or I can do it tomorrow.

Dantemss: Passing a NodeList makes Mathjax blow up.

nathanstitt:
@Dantemss looks good and does seem to render faster.

Dantemss:
Woohoo!

Dantemss:
Keep in mind that this will probably still be slow if some element has `class="math"`, since that uses separate rendering logic for whatever reason.

PR #656, opened 2015-08-19
Fix JS exception when navigating reference book

nathanstitt:
Sometimes the when ref book router params do not have courseId, they just have bookId.  

No need to look it up in that case.

### DIFF
--- a/src/components/book-content-mixin.cjsx
+++ b/src/components/book-content-mixin.cjsx
@@ -1,3 +1,4 @@
+React = require 'react'
 _ = require 'underscore'
 S = require '../helpers/string'
 dom = require '../helpers/dom'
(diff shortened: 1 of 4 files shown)


src/components/book-content-mixin.cjsx
```diff
@@ -51,14 +55,12 @@ module.exports =
...
-    course = CourseStore.get(courseId)
-    referenceBookParams =
-      bookId: course.book_id
-      cnxId: cnxId or @getCnxId()
-    pageUrl = @context.router.makeHref('viewReferenceBookPage', referenceBookParams)
-
-    pageUrl
+    {courseId, bookId} = @context.router.getCurrentParams()
+    if courseId and not bookId
+      bookId = CourseStore.get(courseId)?.book_id
```
pandafulmanda: cool, yay!

PR #658, opened 2015-08-19
Render blank cell when no assignment data is present

nathanstitt:
When a student is dropped or transfered to a different period, future
assignements will not have any data.  The student should still
appear in the report but not display any information for the assignments
they didn't participate in.

BE PR: https://github.com/openstax/tutor-server/pull/608

Note the assignments "No Atticus" and "No Lily"  They were assigned after Atticus was dropped from the period, and Lily was transferred to it.
screen shot 2015-08-19 at 8 36 26 pm

### DIFF
--- a/src/components/performance/absent-cell.cjsx
+++ b/src/components/performance/absent-cell.cjsx
@@ -1,0 +1,7 @@
+React    = require 'react'
+
+module.exports = React.createClass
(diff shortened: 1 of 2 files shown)


nathanstitt:
@Dantemss is prepping a BE PR that will need this.

PR #660, opened 2015-08-19
don't using router go back.  manually build out calendar return route from plan info

pandafulmanda:
trying to use router.goBack would cause problems sometimes in edge cases where:

1. teacher deletes a plan
1. teacher then creates a new plan
1. teacher publishes or cancels
  * sometimes would cause a 404 as it attempts to go back to calendar, but to the deleted planId's route
  * for whatever reason, go back is tracked to the wrong route.

This PR aims to fix this issue by:

1. building calendar back route manually from due_at and plan id information
1. pulling leaving builder functions from publish, save, delete, cancel, etc. to use the central `@goBackToCalendar` function
1. using same cancel function for both X close button and Cancel button in builder footer
1. fixing the resets to be plan specific and use a copy of the plan initially loaded from the server since plan local store gets manipulated directly by the builder
  * cancel and delete used to complete reset the full task plan store, which had unintended effects.
  * to test behavior now
    1. edit a plan
    1. cancel edit
    1. click to edit it again
        * note that changes are reverted properly

### DIFF
--- a/src/components/task-plan/plan-mixin.coffee
+++ b/src/components/task-plan/plan-mixin.coffee
@@ -1,6 +1,7 @@
 React = require 'react'
 {TaskPlanStore, TaskPlanActions} = require '../../flux/task-plan'
 {TimeStore} = require '../../flux/time'
+moment = require 'moment'
 
 module.exports =
   contextTypes:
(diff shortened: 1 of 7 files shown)


src/components/task-plan/plan-mixin.coffee
```diff
@@ -57,9 +58,23 @@ module.exports =
...
-    @context.router.goBack()
+    if confirm('Are you sure you want to cancel?')
+      TaskPlanActions.resetPlan(id)
+      @goBackToCalendar()
+
+  goBackToCalendar: ->
+    {id, courseId} = @props
+    calendarRoute = 'calendarByDate'
+    dueAt = TaskPlanStore.getFirstDueDate(id) or @context?.router?.getCurrentQuery()?.due_at
+    date = moment(dueAt).format('YYYY-MM-DD')
```
nathanstitt: Mixins can also specify `contextTypes`.  If you set the mixin to have the router, then you could be assured that `@context.router. getCurrentQuery()` would always work.  You'd still need to check due_at's existence tho.

That's probably a double good idea since you're assuming `@context.router.transistionTo` below :)

pandafulmanda: if `@context.router.getCurrentQuery().due_at` doesn't exist, it's okay since moment will create a date for today...actually lemme push in a thing that'll use TimeStore.getNow() in that case to be better

PR #661, opened 2015-08-20
Change history to only remember selected routes

nathanstitt:
This changes the history to only remember routes that can be safely
returned to.

The Back button's helper methods are also modified so that it will only
load the last paths from history that was for a screen other than the
current one.

screen shot 2015-08-19 at 8 04 11 pm

screen shot 2015-08-19 at 8 07 29 pm

### DIFF
--- a/src/components/app.cjsx
+++ b/src/components/app.cjsx
@@ -22,8 +22,7 @@ module.exports = React.createClass
     @storeHistory(path: HistoryLocation.getCurrentPath())
 
   storeHistory: (locationChangeEvent) ->
-    if DestinationHelper.shouldRememberRoute(locationChangeEvent, @context.router)
-      TransitionActions.load(locationChangeEvent)
+    TransitionActions.load(locationChangeEvent, @context.router)
 
   getInitialState: ->
     displayDebug: false
--- a/src/components/app.cjsx
+++ b/src/components/app.cjsx
@@ -1,5 +1,6 @@
 React = require 'react'
 {HistoryLocation, History, RouteHandler} = require 'react-router'
+DestinationHelper = require '../helpers/routes-and-destinations'
 
 Navbar = require './navbar'
 
@@ -7,6 +8,8 @@ Navbar = require './navbar'
 
 module.exports = React.createClass
   displayName: 'App'
+  contextTypes:
+    router: React.PropTypes.func
 
   componentDidMount: ->
     @storeInitial()
@@ -16,12 +19,11 @@ module.exports = React.createClass
     HistoryLocation.removeChangeListener(@storeHistory)
 
   storeInitial: ->
-    if History.length is 1
-      path = HistoryLocation.getCurrentPath()
-      TransitionActions.load({path})
+    @storeHistory(path: HistoryLocation.getCurrentPath())
 
   storeHistory: (locationChangeEvent) ->
-    TransitionActions.load(locationChangeEvent)
+    if DestinationHelper.shouldRememberRoute(locationChangeEvent, @context.router)
+      TransitionActions.load(locationChangeEvent)
 
   getInitialState: ->
     displayDebug: false
--- a/src/components/app.cjsx
+++ b/src/components/app.cjsx
@@ -1,6 +1,5 @@
 React = require 'react'
 {HistoryLocation, History, RouteHandler} = require 'react-router'
-DestinationHelper = require '../helpers/routes-and-destinations'
 
 Navbar = require './navbar'
 
--- a/src/components/buttons/back-button.cjsx
+++ b/src/components/buttons/back-button.cjsx
@@ -9,15 +9,18 @@ BackButton = React.createClass
   displayName: 'BackButton'
 
   propTypes:
+    bsStyle: React.PropTypes.string
     fallbackLink: React.PropTypes.shape(
       to: React.PropTypes.string
       params: React.PropTypes.object
       text: React.PropTypes.string
     ).isRequired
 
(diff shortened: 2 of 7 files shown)


nathanstitt:
@pandafulmanda  I've removed several routes from the list in `routes-and-destinations`.  It's possible that some of those should be considered for back button behavior, but I wasn't sure.

I'm still testing manual to make sure it works ok, but it does for the practice sets at least.

src/components/app.cjsx
```diff
@@ -16,12 +19,11 @@ module.exports = React.createClass
...
   storeInitial: ->
-    if History.length is 1
-      path = HistoryLocation.getCurrentPath()
-      TransitionActions.load({path})
+    @storeHistory(path: HistoryLocation.getCurrentPath())
 
   storeHistory: (locationChangeEvent) ->
-    TransitionActions.load(locationChangeEvent)
+    if DestinationHelper.shouldRememberRoute(locationChangeEvent, @context.router)
+      TransitionActions.load(locationChangeEvent)
```
nathanstitt: Ideally TransitionActions would perform this check itself.  It's unable to though since you can't pass the router into the actions.  Flux clones incoming params which strips all the methods.

src/flux/transition.coffee
```diff
@@ -23,9 +25,9 @@ TransitionStore = flux.createStore
   actions: _.values(TransitionActions)
   _local: []
 
-  load: ({path, type}) ->
-    type ?= 'push'
-    @_local.push(path) if type is 'push'
+  load: (change, router) ->
+    change.type ?= 'push'
```
pandafulmanda: hehehehe, this is a similar comment that @philschatz  made when I made this original PR -- this will change the `change` object. just to be more safe, could we do, `{type, path}`and `type ?= 'push'`?

src/components/app.cjsx
```diff
@@ -1,5 +1,6 @@
 React = require 'react'
 {HistoryLocation, History, RouteHandler} = require 'react-router'
+DestinationHelper = require '../helpers/routes-and-destinations'
```
pandafulmanda: `require`d but not used?

src/components/buttons/back-button.cjsx
```diff
@@ -18,35 +18,25 @@ BackButton = React.createClass
...
+    historyInfo = TransitionStore.getPrevious(@context.router)
     {fallbackLink, className} = @props
     {text} = fallbackLink
 
-    if hasHistory
-      backText = 'Back'
-      backText = "Back to #{historyInfo.name}" if historyInfo.name
-      backButton = <BS.Button className={className} onClick={@context.router.goBack}>
-        {backText}
-      </BS.Button>
```
pandafulmanda: Oh weird, why did I use `goBack` when the whole point was to use `historyInfo`?

src/components/app.cjsx
```diff
@@ -16,12 +19,10 @@ module.exports = React.createClass
     HistoryLocation.removeChangeListener(@storeHistory)
 
   storeInitial: ->
-    if History.length is 1
-      path = HistoryLocation.getCurrentPath()
-      TransitionActions.load({path})
+    @storeHistory(path: HistoryLocation.getCurrentPath())
```
pandafulmanda: hmm, should we also store `params` and `queryParams`?

src/components/buttons/back-button.cjsx
```diff
@@ -18,35 +18,25 @@ BackButton = React.createClass
...
-      classes += " #{className}" if className?
-
-      backButton = <Link {...fallbackLink} className={classes}>
-        {text}
-      </Link>
+    backText = if historyInfo.name then "Back to #{historyInfo.name}"  else fallbackLink.text
 
-    backButton
+    <BS.Button className={className} {...@props}
+      onClick={@navigate}>
```
pandafulmanda: Do you think this could done as a `Router.Link` instead of using `onClick` for `transitionTo`?  I think have links increases usability a little bit, like for weird ppl like me that like to open everything in a new tab.

src/components/app.cjsx
nathanstitt: Nah, we don't need to. `router.transistionTo` can navigate with using complete path string that it previously had.

src/components/app.cjsx
nathanstitt: yeah :) forgot to remove

src/helpers/routes-and-destinations.coffee
```diff
@@ -39,7 +30,15 @@ routesAndDestinations =
...
 destinationHelpers =
-  getDestination: (routeName) ->
-    routesAndDestinations[routeName]
+  getDestinationName: (routeName) ->
+    REMEMBERED_ROUTES[routeName]
+
+  destinationFromPath: (path, matchRoutes) ->
+    matchedRoute = matchRoutes(path)
+    deepestRouteName = _.last(matchedRoute.routes).name if matchedRoute?.routes?.length
+    @getDestinationName(deepestRouteName)
```
pandafulmanda: c:sunglasses:l move

test/components/task-plan/footer.spec.coffee
```diff
@@ -57,7 +57,6 @@ VISIBLE_HW = extendBasePlan({
 
 helper = (model) ->
   {id} = model
-  console.info(model)
```
pandafulmanda: oops.

nathanstitt: that actually came from your work last night I think.  And I didn't notice when reviewing either...

PR #663, opened 2015-08-20
prevent trying to load plan that has been requested for deletion -- prevents the 404

pandafulmanda:
the older, calmer sister PR to #660 calmer?

## How to reproduce
1. publish a plan that opens in the future
1. view and edit
1. click delete
  * at this point, the previous PR neglected to take the user back to the calendar without plan id in route
    * It only accounted for drafts, and drafts that were being deleted, and avoided the plan id route for those specific cases, as intended
  * this really aint no thang since it doesn't cause any 404's though since loading the deleted plan and plan stats is not actually triggered
1. click browser back
  * **this is where things start to go wrong.** back to plan edit route will trigger an api hit for the deleted plan
  * another back will trigger an api hit for the deleted plan's stats
  * locally, it takes awhile for both to return 404 and show the modal, since the single thread is still working on actually deleting the plan

### Back once to plan editor
* Doesn't re-fetch plan that's been recently requested for deletion
* not my fave UI ever, but for something that a user shouldn't be seeing, should be okay for now?
  screen shot 2015-08-20 at 1 42 36 am

### Back twice to calendar that had plan modal open
* Prevents route for plan that's been deleted
  screen shot 2015-08-20 at 1 42 42 am

### DIFF
--- a/src/components/task-plan/external/index.cjsx
+++ b/src/components/task-plan/external/index.cjsx
@@ -36,6 +36,7 @@ ExternalPlan = React.createClass
       onPublish={@publish}
       onSave={@save}
       onCancel={@cancel}
+      getBackToCalendarParams={@getBackToCalendarParams}
       goBackToCalendar={@goBackToCalendar}/>
 
     header = @builderHeader('external')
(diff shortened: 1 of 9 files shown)


nathanstitt:
Awesome!  Works nicely for me

PR #664, opened 2015-08-20
Try one mo' gain if server throws an error

pandafulmanda:
so we take it and flip and reverse it.  just once though.  targeted at [69](https://docs.google.com/spreadsheets/d/1P0eq49e7u1uM3SUI9pULo7EQYZIJt1KLOOfBidHdFSY/edit#gid=1446311939) wherein `422` happens and the last attempted action needs to be repeated before student can continue task

also, captures info about the request.  SS shows spoofed url to force error

1. Before error trigger
  screen shot 2015-08-20 at 11 54 29 am
1. Spoofed error
  screen shot 2015-08-20 at 11 54 15 am
1. after closing modal, page reloads and app state is consistent with fresh server load
  * In this example, the multiple answer choice patch caused the server error, and so the answer was never saved.  The FE reflects that after the forced reload.
  screen shot 2015-08-20 at 11 54 09 am

Also, I have yet to achieve the same `422` during an exercise as during doomsday to try this with that.

### DIFF
--- a/src/components/task/index.cjsx
+++ b/src/components/task/index.cjsx
@@ -93,6 +93,8 @@ module.exports = React.createClass
   shouldComponentUpdate: (nextProps, nextState) ->
     {id} = @props
 
+    console.info(TaskStore.getAsyncStatus(id))
+
     # if a step needs to be recovered, load a recovery step for it
     if @_stepRecoveryQueued(nextState)
       TaskStepActions.loadRecovery(nextState.recoverForStepId)
@@ -116,6 +118,9 @@ module.exports = React.createClass
     #   redirect to the step after the step we triggered refresh from.
     if @_leavingRefreshingStep(nextState)
       @continueAfterRefreshStep()
+      return false
+
+    if TaskStore.isSaving(id)
       return false
 
     # if we reach this point, assume that we should go ahead and do a normal component update
--- a/src/components/task/index.cjsx
+++ b/src/components/task/index.cjsx
@@ -93,8 +93,6 @@ module.exports = React.createClass
   shouldComponentUpdate: (nextProps, nextState) ->
     {id} = @props
 
-    console.info(TaskStore.getAsyncStatus(id))
-
     # if a step needs to be recovered, load a recovery step for it
     if @_stepRecoveryQueued(nextState)
       TaskStepActions.loadRecovery(nextState.recoverForStepId)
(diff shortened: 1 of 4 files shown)


src/components/task/index.cjsx
```diff
@@ -93,6 +93,8 @@ module.exports = React.createClass
   shouldComponentUpdate: (nextProps, nextState) ->
     {id} = @props
 
+    console.info(TaskStore.getAsyncStatus(id))
```
philschatz: What is this? ; )

pandafulmanda: OH dang it.

nathanstitt:
The BE folks have discovered that the 422 errors that prompted this are probably real errors, but were timing dependent due to issues related to inter-server communication. For task steps only, we should attempt to retry a 422 error, but not other types.

Additionally, @Dantemss thinks that when a different error (500 or whatever) is encountered on practice steps we're still marking the step as completed internally and attempting to go forward.  We need to correct that behavior if it's occurring.

philschatz:
Why should BE still return a 422? That seems like odd semantics; if timing matters then shouldn't there be an async job generated and/or a 20x returned?

nathanstitt:
As I understand it, the BE returns a 422 in some cases when a step is in the process of being marked as completed, and then the FE submits a new request while the original is still being processed.  It's possible the new request can hit a different server and obtain stale state.  The retry employed here is proposed as a solution for beta until the underlying issues can be sorted through.

nathanstitt:
I don't see where this is being restricted to just task_steps though, @pandafulmanda I thought we needed to do that?

pandafulmanda:
im not restricting it for just task_steps.  Thinking it over, it doesn't seem like we should -- when we get a server error, we should opt to refresh because that's the way our FE will become consistent with the BE again

nathanstitt:
Yeah, reload is the best thing to do in order to get the state back in sync.  I'm just afraid the retry will cause other errors with the BE. @jpslav care to comment?

jpslav:
Looks good to me for short term patch

nathanstitt:
Actually, I was behind on the commits.  The code no longer performs a retry, just the modal that reloads the page.  Merging now.  Sorry for the confusion @pandafulmanda

Dantemss:
We have other options for handling errors in the future too:
- Manual retry button instead of automatic retry
- Clicking OK considers the request as unfulfilled or reloads everything

Dantemss:
Oh OK according to Amanda it reloads everything when you click close so that's probably good.

PR #665, opened 2015-08-20
Fix media links to another module

pandafulmanda:
Fixes [142](https://docs.google.com/spreadsheets/d/1P0eq49e7u1uM3SUI9pULo7EQYZIJt1KLOOfBidHdFSY/edit#gid=1446311939)
Important to merge because this will fix quite a few links.
* link matching wasnt matching on links that were media links but weren't `#ed`
* update task test to handle making links

## Bug
screen shot 2015-08-20 at 4 49 42 am

## With fix
screen shot 2015-08-20 at 4 46 32 am

### DIFF
--- a/src/components/book-content-mixin.cjsx
+++ b/src/components/book-content-mixin.cjsx
@@ -63,7 +63,7 @@ module.exports =
     )
 
   isMediaLink: (link) ->
-    link.hash.length > 0 and link.hash.search('/') is -1
+    (link.hash.length > 0 and link.hash.search('/') is -1) or link.href isnt link.getAttribute('href')
 
   hasCNXId: (link) ->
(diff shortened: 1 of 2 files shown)


src/components/book-content-mixin.cjsx
```diff
@@ -63,7 +63,7 @@ module.exports =
     )
 
   isMediaLink: (link) ->
-    link.hash.length > 0 and link.hash.search('/') is -1
+    (link.hash.length > 0 and link.hash.search('/') is -1) or link.href isnt link.getAttribute('href')
```
nathanstitt: :question: Huh?  I'm confused by how `link.href` would ever differ from `link.getAttribute('href')`  I'm sure it must though since you added the check deliberately.

PR #668, opened 2015-08-20
Fix/answers reversal

pandafulmanda:
## Bug
screen shot 2015-08-20 at 5 43 31 am

## With Fix
* sort by answer id
  screen shot 2015-08-20 at 5 21 53 am

Need to test more probably due to splitting Answer out as a separate component


nathanstitt:
I'm not sure how to test the bug locally.  I'm unable to replicate the bug locally. The new code works fine though!

PR #669, opened 2015-08-20
yay sort by average step number

pandafulmanda:
yay we get step number now!  FE sorts before displaying...and things seem to work!  BE openstax/tutor-server#597

## Bug
* question order in review does not match assign order.  it is sorted by page, and then assign order
screen shot 2015-08-20 at 12 41 52 pm

## Fixed

screen shot 2015-08-20 at 12 40 22 pm


nathanstitt:
Yay!  Looks great

PR #670, opened 2015-08-20
loading for teacher-student guide

pandafulmanda:
screen shot 2015-08-20 at 1 48 45 pm

initial stab.  this was really bothering me


nathanstitt:
:+1: Much better than it is now :grinning:

philschatz:
Thanks for scratching that itch! I just noticed this while updating the selenium tests and was wondering why students never had any data :smile:

PR #673, opened 2015-08-20
interactives and videos can now have links on them that need to be processed

pandafulmanda:
## With Fix
screen shot 2015-08-20 at 6 57 57 pm

### DIFF
--- a/src/components/task-step/all-steps.cjsx
+++ b/src/components/task-step/all-steps.cjsx
@@ -9,8 +9,9 @@ Exercise = require './exercise'
 Markdown = require '../markdown'
 StepMixin = require './step-mixin'
 StepFooterMixin = require './step-footer-mixin'
-BookContentMixin = require '../book-content-mixin'
+{BookContentMixin, LinkContentMixin} = require '../book-content-mixin'
 CourseDataMixin = require '../course-data-mixin'
+ChapterSectionMixin = require '../chapter-section-mixin'
 
 {StepPanel} = require '../../helpers/policies'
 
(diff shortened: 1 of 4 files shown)


test/components/book-content-mixin.spec.coffee
```diff
@@ -5,9 +5,10 @@ COURSE_ID = '1'
 COURSE    = require '../../api/user/courses/1.json'
 
 {CourseActions} = require '../../src/flux/course'
+{BookContentMixin} = require('../../src/components/book-content-mixin')
 
 TestComponent = React.createClass
-  mixins: [ require('../../src/components/book-content-mixin') ]
+  mixins: [ BookContentMixin ]
   getCnxId: sinon.spy -> CNX_ID
```
nathanstitt: BAHAH.  I see you're not a fan of my mixin require inline :)  Probably for the best

pandafulmanda: haaha yes i didn't like it, but also I needed to change it since I changed what book-content-mixin was exporting

src/components/task-step/all-steps.cjsx
```diff
@@ -19,6 +20,13 @@ err = (msgs...) ->
...
   throw new Error(JSON.stringify(msgs...))
 
+# TODO make this logic more failsafe
+getCnxId = (id) ->
+  {content_url} = TaskStepStore.get(id)
+  if content_url.search('contents/') > -1
+    cnxId = _.last(content_url.split('contents/'))
+  cnxId
+
 Reading = React.createClass
```
nathanstitt: how about:
```coffeescript
getCnxId = (id) ->
  {content_url} = TaskStepStore.get(id)
  parts = TaskStepStore.get(id).content_url.split('contents/')
  if parts.length > 1 then _.last(parts) else undefined
````

nathanstitt:
Looks good.  Up to you if you want to play with that method or anything.  Otherwise we can merge

nathanstitt:
Nice!

PR #677, opened 2015-08-21
Fix for adding in periods

jivewise:
No ui changes


nathanstitt:
@jivewise Awesome fix to find this so quickly!  Tested locally and fixes issue with changing periods on an existing plan

PR #679, opened 2015-08-21
perf report - reduced bottom margin height

darrenmason:
regarding pivotal task: https://www.pivotaltracker.com/story/show/101628600

reduces table bottom margin to match with top.

image


nathanstitt:
Looks good @darrenmason.  You might consider redoing the resize listener to use the Mixin at https://github.com/openstax/tutor-js/blob/master/src/components/resize-listener-mixin.cjsx  That abstracts out the adding/removing the event listener.

darrenmason:
ok, will update listener to use mixin

philschatz:
Thanks!

PR #680, opened 2015-08-21
fixes when we would accidentally submit a moment date obj to BE

pandafulmanda:
## How to replicate
1. start adding a brand **new** plan
1. select to make due dates different periods
1. change due date for at least one of the periods
1. switch back to "All periods"
  * this calls the `TaskPlanStore.getMinDueAt` which was the culprit
1. try to save as draft or publish
  * BE yells at you with a `422`
  * you'll see in the modal that FE is trying to submit a moment obj as due_at

should be fixed with this PR.

### DIFF
--- a/src/flux/task-plan.coffee
+++ b/src/flux/task-plan.coffee
@@ -169,12 +169,15 @@ TaskPlanConfig =
     throw new Error('id is required') unless id
     throw new Error("#{attr} is required") unless date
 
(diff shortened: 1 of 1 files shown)


nathanstitt:
:+1: Looks good. Thanks for the detailed replication instructions. I couldn't figure out how to trigger this thing.

PR #681, opened 2015-08-21
first name last name use instead of name from BE

pandafulmanda:
use `first_name` and `last_name` as opposed to `name`


nathanstitt:
Looks good for now.  Post beta would be nice to create a `PersonsName` component or some such that would first try `.name` and then concat `first` and  `last` if it's not present.

PR #684, opened 2015-08-22
Plan editability

pandafulmanda:
## To trigger bug
1. start building a new homework or reading
1. change start date to today
1. click publish
  * at this point, the "Add Reading" or "Select Problems" disappear
  * errors show up as well
  * start date picker becomes disabled
  * this gets you stuck

* There was another bug as well, hidden within.  If you click "Publish", and later finish building your assignment, and click "Save as Draft", your plan will actually publish instead.

## Fix
On step 3,
* you should be able to edit all aspects on the plan, despite open date being past

#### For plans that are already published and opened,
* will be unable to change plan details, except for
  * Name
  * Description
  * Due dates

#### For plans that are in draft mode, regardless of dates,
* everything editable

#### For plans that are published, but not yet open
* everything editable

#### Plans past due,
* should not let you edit anything
* should have a back to calendar button

### DIFF
--- a/src/components/task-plan/builder.cjsx
+++ b/src/components/task-plan/builder.cjsx
@@ -26,16 +26,8 @@ module.exports = React.createClass
   getInitialState: ->
     isNewPlan = TaskPlanStore.isNew(@props.id)
 
-    # Whether date/periods inputs are disabled should only depend on the whether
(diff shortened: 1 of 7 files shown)


nathanstitt:
Works good, I like how you moved more of the logic into the store.  After beta we should figure out if the Publish button should still be visible on the plan after it's opened.

PR #685, opened 2015-08-23
Fix performance report homework status

nathanstitt:
The homework cells in the performance report had a bug where they would display "Incomplete" if the task was worked late, regardless of the completion status.

Note the display for Alden, Isobel, and Bernhard in the HW column

Before:
screen shot 2015-08-22 at 10 52 59 pm

After:
screen shot 2015-08-22 at 11 12 24 pm


jivewise:
:+1:  Looks good, plus with specs!

nathanstitt:
Thanks @jivewise.  Yeah figured specs would benefit this since we didn't get it right the first time around.  Plus, I think I wrote the original code to boot :grin:

PR #686, opened 2015-08-24
dont try to set a min due date for all periods before a common opens_at is set for all periods

pandafulmanda:
## How to replicate
1. create a new assignment
1. choose to set individual period dates
1. change a open date
1. change a due date
1. switch to all periods again
1. fill out whatever else is needed, but don't touch any of the dates stuff
1. try to save or publish
  * you should get a 422

Steps 3 and 4's order does not matter

## Why?
* The FE attempts to create a minimum due date from the common open date, but there was no common open date at the time we try to create the minimum due date.


pandafulmanda:
Oops, i need to branch off master

PR #692, opened 2015-08-24
hwbuilder sort rendered answers by id

darrenmason:
hwbuilder sort rendered answers by id

### DIFF
--- a/src/components/task-plan/homework/exercises.cjsx
+++ b/src/components/task-plan/homework/exercises.cjsx
@@ -33,10 +33,10 @@ ExerciseCardMixin =
   renderExercise: ->
     content = @props.exercise.content
     question = content.questions[0]
-    question.answers = _(question.answers).chain()
+    renderedAnswers = _(question.answers).chain()
       .sortBy('id')
+      .map(@renderAnswers)
       .value()
-    renderedAnswers = _.map(question.answers, @renderAnswers)
     tags = _.clone @props.exercise.tags
     # Display the exercise uid as a tag
     tags.push(name: "ID: #{@props.exercise.content.uid}")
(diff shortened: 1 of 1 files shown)


src/components/task-plan/homework/exercises.cjsx
```diff
@@ -33,6 +33,9 @@ ExerciseCardMixin =
   renderExercise: ->
     content = @props.exercise.content
     question = content.questions[0]
+    question.answers = _(question.answers).chain()
+      .sortBy('id')
+      .value()
     renderedAnswers = _.map(question.answers, @renderAnswers)
```
nathanstitt: Nice, `sortBy` using a property, instead of a function.  I think that's much clearer that way.

The `_(` in underscore is a shorthand for the chain method though.  So you don't need to append the `.chain`  http://underscorejs.org/#chaining

It also might make sense to also include the `_.map` below into the chain though. i.e.
```coffee
  renderedAnswers = _(question.answers)
    .sortBy('id')
    .map(question.answers, @renderAnswers)
    .value()
```

darrenmason: Yes that's a good way to combine;  though map must be called w/o the list param:
```
renderedAnswers = _(question.answers).chain()
.sortBy('id')
.map(@renderAnswers)
.value()
```

nathanstitt:
:+1: Looks good,   Glad you were able to decipher my message.

src/components/task-plan/homework/exercises.cjsx
pandafulmanda: Unfortunately, or fortunately, the answer ids are strings.  `.sortBy('id')` will be different from `.sortBy( (answer) ->
  parseInt(answer.id)
)`

in cases where there are answer ids '1234' and '123' for one question for example

PR #693, opened 2015-08-24
strip timestamps for json payload to BE to deal with possible timezone problems

pandafulmanda:
works with BE, in which timestamp is ignored when patching

Going to work on specs with this and #687


pandafulmanda:
tested tutor-server#618, with timezone set to eastern, and it seems to fix the timezone bug, which i was able to reproduce with the beta branch of server without the fix

PR #694, opened 2015-08-25
Test/builder

pandafulmanda:
tests for the builder, companion to #687 and #693

Did not realize #697 would be pushed up today.  consider closing this in favor of #697


test/components/task-plan/builder.spec.coffee
```diff
@@ -0,0 +1,82 @@
...
+describe 'Task Plan Footer', ->
+  beforeEach ->
+    TaskPlanActions.reset()
+
+  it 'should be hiding all periods on default', ->
+    helper(NEW_READING).then ({dom}) ->
+      checkedPeriodsToggle = dom.querySelector('input[name=toggle-periods-radio]:checked')
+      expect(checkedPeriodsToggle.id).to.contain('hide')
+      expect(wrapper.refs.element.state.showingPeriods).to.be.false
+
```
nathanstitt: The `renderComponent` helper's promise yields quite a few things beside `dom`, one of which is the element reference.  If you added that to the variables extracted along with `dom`,  you could just say `element.state` here.  Actually where is `wrapper` coming from?  I don't see it being assigned :grin:

https://github.com/openstax/tutor-js/blob/master/test/components/helpers/component-testing.coffee#L50-L53

pandafulmanda: oops, i think i messed up the check-in -- it's supposed to be `.then ({wrapper, dom})`

pandafulmanda: AH i see, element is refer to that exact thing, I got confused and thought it was referring to the dom or something.  okay, i'll definitely use element instead of re-grabbing it off of wrapper

nathanstitt: Yeah, `element` could probably be renamed to something better.  `instance` maybe?

pandafulmanda: i was expecting this to be called `component` probably just because that's what i used in the other helper though

nathanstitt: Durp.  yeah `component` matches the React naming convention

PR #695, opened 2015-08-25
Do not display  "Try Another" while store is not loading or saving

nathanstitt:
If try another is clicked while store is saving answer a 422 error is encountered.

### DIFF
--- a/src/flux/task-step.coffee
+++ b/src/flux/task-step.coffee
@@ -5,9 +5,12 @@ flux = require 'flux-react'
 Task = require './task'
 Durations = require '../helpers/durations'
 
+RECOVERY = 'recovery'
+
 {CrudConfig, makeSimpleStore, extendConfig} = require './helpers'
 
 TaskStepConfig =
+  _asyncStatus: {}
 
   _loaded: (obj, id) ->
     if not obj.task_id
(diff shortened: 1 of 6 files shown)


src/flux/task-step.coffee
```diff
@@ -85,6 +85,14 @@ TaskStepConfig =
...
       step.content.questions?[0].formats?.indexOf('free-response') > -1
 
+    canTryAnother: (id) ->
+      step = @_get(id)
+      step? and
+        (step.has_recovery and step.correct_answer_id isnt step.answer_id) and
+        not Task.TaskStore.isTaskPastDue(step.task_id) and
+        not @exports.isLoading.call(this, id) and
+        not @exports.isSaving.call(this, id)
+
```
pandafulmanda: `this`/`@`

src/flux/task-step.coffee
```diff
@@ -85,6 +85,14 @@ TaskStepConfig =
 
       step.content.questions?[0].formats?.indexOf('free-response') > -1
 
+    canTryAnother: (id) ->
+      step = @_get(id)
+      step? and
+        (step.has_recovery and step.correct_answer_id isnt step.answer_id) and
+        not Task.TaskStore.isTaskPastDue(step.task_id) and
```
pandafulmanda: instead of [module loading order in a one off way](https://github.com/openstax/tutor-js/pull/695/files#diff-74c51a44b8ba99c878804b91182ab691R8), could we pull `isTaskPastDue` into a helper/stateless function?

nathanstitt: I'd like to but couldn't figure out a way to do so, since `isTaskPastDue` needs access to the task's `due_at`.  Regardless of whether the helper method was in a different location, the `task-step` would still have to access the task somehow to get that attribute.

I also thought about just changing the check in the component to be `not TaskStore.isTaskPastDue(..) and TaskStep. canTryAnother(...)`  but that felt like a lie since you can't try another if the task is past due.  The logic should be in the single method.

And then I thought about flipping the module loading hocus-pocus so the task store was the one that exported a blank object initially.  I rejected that because the TaskStep store calls only one method on Task, while Task calls two on TaskStep store.

So yeah... suggestions on how to break the circular dependency are most welcome :)

PR #698, opened 2015-08-25
Test/builder

pandafulmanda:
Supposed to be off of beta, duplicate of #694.

tests for the builder, companion to #687 and #693

Did not realize #697 would be pushed up today.  consider closing this in favor of #697.  Although if #697 will not be merged into beta, this might be a beta only PR, and we can port over the tests to that PR

Also, these tests can definitely be improved, and the repeated code could be reduced, but could suffice for regression testing of different new plan date options switching


pandafulmanda:
closing in favor of #697 being merged in first and porting these tests in after

PR #699, opened 2015-08-25
handle course names like '10th'

pandafulmanda:
## JP observed this with real data
screenshot 2015-08-24 23 16 24

my initial intentional assumption that there would only be periods up to 7 was silly and incorrect :(

I probs should write tests for this as welllllll


src/helpers/period.coffee
```diff
@@ -0,0 +1,7 @@
+_ = require 'underscore'
+
+PeriodHelper =
+  getOrder: (period) ->
+    parseInt(period.name)
```
philschatz: Will `parseInt` work for non-integer periods like 'Block B'? I assume it will return `NaN` and not sort at all

pandafulmanda: Oh, okay, is there a list of valid period names/patterns somewhere?  I guess this should just plan for all cases?

philschatz: or should we depend on the order returned by BE?

nathanstitt: I belive @philschatz is correct. parseInt("Period 10") would return `NaN`.  

On a somewhat related note, you should always use the radix argument for parseInt, otherwise Javascript will choose base 8 or 16 if the first argument starts with a zero.  Probably unlikely in this case, but it's good practice to always use 10 for the radix.

Here I think you'll have to use replace with a regex: 
`parseInt(period.name.replace(/\D/g,''), 10)`

pandafulmanda: Ah cool, okay, I'll try some stuffs

pandafulmanda:
oh.  should i make this a PR on master instead?

philschatz:
yeah, I wish you could change the target branch in PR's ; (

pandafulmanda:
forrealz.

PR #707, opened 2015-08-28
Plan publishing and delete details

pandafulmanda:
#702, but against master as opposed to beta

## Bug
1. publish a plan
  * you will be taken back to the calendar
  * you should see your plan as publishing
    * there may be a flash of draft as well
1. change your month
1. change back
  * plan shows as draft and not as publishing
  * also, if you click to edit, "Save as Draft" and "Delete" are still options

## With Fix
* publishing status will show even if you change your month view, and then change back
screen shot 2015-08-26 at 2 15 11 pm
screen shot 2015-08-26 at 2 15 14 pm
screen shot 2015-08-26 at 2 15 20 pm
* Even if you do get to the edit page for a publishing plan, you will be prevented from doing anything
screen shot 2015-08-26 at 3 19 02 pm
* After publishing is done, you get to do stuff again
screen shot 2015-08-26 at 3 47 19 pm

## To Do
- [x] fix calendar publishing state
- [x] disable builder footer when publishing
- [x] some code clean up
- [x] Specs for
  - [x] Job helper
  - [x] Plan publish store
  - [x] Plan component
  * job store, probably later


philschatz:
wow, that's quite a bit of refactoring and might take me a while to read through! thoughts @nathanstitt? Also, can you merge/rebase from master?

pandafulmanda:
most of the refactoring was around creating a base job listener config for the PlanPublish and PerformExport stores, that were sitting around with a lot of repeated code

src/components/scores/export.cjsx
```diff
@@ -134,6 +136,7 @@ ScoresExport = React.createClass
         isWaiting={ScoresExportStore.isExporting(courseId) or tryToDownload}
         isFailed={ScoresExportStore.isFailed(courseId) or downloadHasError}
         failedProps={failedProps}
+        timeout={3600000}
```
philschatz: Why is the timeout 1 hour? And could/should all of these `3600000` be pulled together to the top of the file?

pandafulmanda: The timeouts don't need to be an hour, they could probably be more like up to 5 minutes, because the jobs sometimes run longer than our default async button timeout

philschatz: I think 10min and just pulling the constant up into the top of the file (or a common place for the job timeout). up to you :smile:

PR #708, opened 2015-08-28
Wait for DOMContentLoaded before running app

nathanstitt:
This is the main bug that prevented IE 9 from running.  `document.body.appendChild` was being called before the document was ready, causing an exception.

While I know we don't want to support IE9, I'm wondering if we should wait for the document to be ready before we startup?  It seems like it'd be good practice and I'm mildly surprised it hasn't bitten us in other browsers as well.

### DIFF
--- a/index.coffee
+++ b/index.coffee
@@ -24,16 +24,16 @@ window._STORES =
   TOC: require './src/flux/toc'
 
 
-api.start(dom.readBootstrapData())
 
-startMathJax()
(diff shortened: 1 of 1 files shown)


pandafulmanda:
BE loads the tutor.js script in the head using the async tag, and this causes the 'DOMContentLoaded' event to not be called once tutor.js is loaded, causing the frontend to never be loaded

PR #709, opened 2015-08-31
use readystatechange for loading the app to be compatible with async js loading

pandafulmanda:
follow-up to #708

### DIFF
--- a/index.coffee
+++ b/index.coffee
@@ -24,8 +24,10 @@ window._STORES =
   TOC: require './src/flux/toc'
 
 
+loadApp = ->
+  unless document.readyState is 'interactive'
+    return false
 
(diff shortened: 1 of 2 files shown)


index.coffee
```diff
@@ -35,5 +37,7 @@ document.addEventListener("DOMContentLoaded", ->
   mainDiv = document.createElement('div')
   mainDiv.id = 'react-root-container'
   document.body.appendChild(mainDiv)
-  router.start(mainDiv)
-)
+  hello = router.start(mainDiv)
```
philschatz: is this variable used?

pandafulmanda: oops.

pandafulmanda: updates

PR #712, opened 2015-09-01
Update for using ecosystem id instead of course id

pandafulmanda:
to handle multiple book versions, matching FE PR to openstax/tutor-server#627

## Todo
- [x] fix reference book forrealz.
  * maybe it's forreal?
- [x] update specs
- [x] look for problems
  - [x] test around plans with actual book import
    * need to check why import doesn't seem to work for me
    * oh i probably just needed to start my jobs
    * things seem to be working as expected
      1. imported a new ecosystem
      1. assigned ecosystem to course
      1. viewing/editing old plan makes requests to the old ecosystem
      1. creating/viewing/editing new plans makes requests to the new ecosystem
  - [x] test and fix things with the reference book
    * pay particular care to links and stuff
  - [x] test and fix things with the reading
- [x] SPECS!
  - [x] ref book
  - [x] browse the book button
  - [x] plans stuffs
- [x] TocStore needs to be linked to the ecosystem id
  * There was a bug -- Toc would think it's loaded despite a fresh ecosystem id being requested.  This was less noticeable when Toc was specific to courses, and once the Toc was loaded once, `Toc.get` and `Toc.isLoaded` would remain true.  There were problems hidden, however.  Once the teacher switched courses and tried to make new plans/view plans for another course, there would be errors.

specs

## Viewing book, specific previous ecosystem
screen shot 2015-09-02 at 12 27 52 pm

## Viewing book, current/default ecosystem
screen shot 2015-09-02 at 12 28 15 pm

## Book TOC preserves ecosystem of book
screen shot 2015-09-02 at 3 43 05 pm

## Specific previous ecosystem for plans (reading and homework) created before ecosystem import
screen shot 2015-09-02 at 7 29 34 pm
screen shot 2015-09-02 at 12 35 51 pm


src/components/buttons/browse-the-book.cjsx
```diff
@@ -13,23 +14,48 @@ module.exports = React.createClass
...
+      # only link with ecosystem id if needed.
+      if ecosystemId isnt courseEcosystemId
+        queryParams = {ecosystemId}
+
     # the router is smart enough to figure out which props are present and return the best route
     linkType = if @props.page then 'viewReferenceBookPage' else
       if @props.section then 'viewReferenceBookSection' else 'viewReferenceBook'
-    link = @context.router.makeHref( linkType, {courseId, cnxId: @props.page, section:@props.section} )
-    linkProps = {target:'_blank', className:'view-reference-guide', href: link}
+    link = @context.router.makeHref( linkType, {courseId, cnxId: @props.page, section:@props.section}, queryParams )
```
philschatz: Why use `.makeHref` instead of `Router.Link`?

pandafulmanda: good question, @nathanstitt was there a reason why it had to be `a` with a `.makeHref`?  was it because of the `target='_blank'?

nathanstitt: I think it's because the `link` variable is used to build up the `linkProps` which is then used in either a plain `<a />` tag, or a `BS.Button` depending on whether the `unstyled` prop is set.   The plain anchor tag could be switched out for a `Router.Link`, and the BS.Button could be as well if we added the bootstrap styles to it.  

I also see that the specs check that `context.router.makeHref` is called with appropriate arguments, but I think they could also be swapped to check Router.Link.

pandafulmanda: darn, http://stackoverflow.com/questions/30202755/react-router-open-link-in-new-tab

PR #714, opened 2015-09-02
printer safe reference book

darrenmason:
regarding task:  https://www.pivotaltracker.com/story/show/102331932

adjustments to clean up printing of refbook

chrome print preview:

image

ie11 print preview:

image


philschatz:
When the Table of Contents is expanded the book is still shifted to the right (see the left padding in the screenshot):
 
image

darrenmason:
prints full width now, regardless of menu state;  tested in IE11, Saf8, Chrome 45, FF 40

philschatz:
Thanks! The left margin does not appear but when I look at 1.2 I noticed some of the learning objectives are above the top margin (in Chrome):

image

Any ideas why?

darrenmason:
should be aligned now:

image

PR #715, opened 2015-09-02
Fix bug where external assignment not counted when manually opened in a new tab 

nathanstitt:
When a link is right or middle mouse button clicked to open
in a new window, the current page does not receive any events.  This
means the link will be loaded, but the step not marked as completed 
since the onClick handler is never called.

To work around this, we setup a new destination url and have the link
point to that.  If the onClick hander is called, we open the external
link in a new window (tab) and mark the step as complete.  If the
handler isn't called and the destination page is loaded, we mark it
complete there and open the window at that time.


Dantemss:
Ugh? Should I just allow you to mark an exercise as complete as part of the update call?

nathanstitt:
Hm. We could do that.  If the BE gave a `completion_endpoint` to the FE, it could just send the user to there and then that endpoint could 1) mark task_step as complete,  2) redirect to "real" url.  That would free the FE from performing the hocus pocus above.

Dantemss:
Hm never mind, I guess this PR is fine.

pandafulmanda:
Hmmmm cool

nathanstitt:
Ok, @Dantemss :)  We can circle back to the BE doing this one later perhaps

src/components/task/external.cjsx
```diff
@@ -0,0 +1,49 @@
...
+
+# This component will only be mounted if the external link
+# was right-clicked on and opened in a new tab.
+#
+# Normal "left-click" usage of the link will be intercepted
+# and the external task opened immediately
+ExternalTask = React.createClass
+
+  getDefaultProps: ->
+    redirectToUrl: (url) -> window.location.href = url
```
philschatz: Can this be a link instead of relying on `window`?

src/components/task-step/all-steps.cjsx
```diff
@@ -109,29 +109,29 @@ ExternalUrl = React.createClass
   displayName: 'ExternalUrl'
   mixins: [StepMixin, StepFooterMixin]
   hideContinueButton: -> true
+  getDefaultProps: ->
+    redirectToUrl: (url) -> window.open( url, '_blank')
```
philschatz: why open this using `window.open` instead of just making a link with `target='_blank'`?

src/flux/task.coffee
```diff
@@ -221,6 +221,13 @@ TaskConfig =
 
       result
 
+    getExternalStep: (id) ->
+      step = _.findWhere(@_steps[id], type: 'external_url')
+      unless step?.external_url.match(/^https?:\/\//)
```
philschatz: Is this not already accounted for when building an external link? I thought @pandafulmanda added the JS bit that prepends a link with `http://`

src/components/task/external.cjsx
nathanstitt: No, I don't think so.  For this component the redirect is called in `componentWillMount`, not in response to a click.  It's done that way because this component will only be mounted if it's opened in a new tab.

The reason the function is set as a prop-with-default is so the implementation can be swapped out in the spec to ensure that it's called.  Karma gets angry if `location.href` is changed mid-spec run.

src/components/task-step/all-steps.cjsx
nathanstitt: Because it needs to be called as an onClick hander which first marks the task as complete, and then redirect.  The redirect fn is a prop so the spec can override the implementation to ensure it's called, but it does have to be triggered from the `onClick`.  

Although! Looking at the implementation below, I think onContinue should be called before the redirect.   I'll swap those line's order.

src/flux/task.coffee
nathanstitt: Yep, this is her implementation, I've just moved it to the store so it can be shared between the "normal" click and the "right-click" components

src/components/task-step/all-steps.cjsx
nathanstitt: Hm, didn't mark this as outdated since the change happened below, but i've made the switch

philschatz:
closing this in favor of #731

PR #717, opened 2015-09-03
add keyboard shortcuts

philschatz:
hello.when answering multiple choice questions use the keys “a”-“d” to select an answer.

TODO:

- [x] only add the key listeners when answering a multiple choice question, not when reviewing an already answered 
- [ ] disable for review when more than one incomplete exercise
- [ ] add enter key handling (maybe by focusing the continue button)
- [ ] add left/right navigation for the reference book
- [ ] add tests


jivewise:
I feel like 1, 2, 3, 4 would be easier to type...

philschatz:
ah, good point. thoughts @Fredasaurus or @pandafulmanda ?

pandafulmanda:
Hmm...that's a good point...I wonder if there would confusion around 1, 2, 3, 4 mapping to a, b, c, d though.  I feel like @Fredasaurus would have a stronger take on this but I could see it working either way.

nathanstitt:
Yeah, I'd say either one would work.  I also suppose we could support both mappings

Fredasaurus:
That's a tricky one. I think Nathan mat have the best answer. Do we ever have answers that are labeled 1, 2, 3, 4? If so, would we map a, b, c, d to that?

philschatz:
I think multiple choice options are always lettered.

pandafulmanda:
guess this will get ported to openstax/react-components ?

philschatz:
closed in favor of openstax/react-components#16

PR #720, opened 2015-09-07
Enlarge student dashboard progress column

nathanstitt:
When screen was between 1,000 & 800px wide, The expanded status's like "10/12 answered (I)" take up too much space,
while the description always seems to have space to spare

Before:
screen shot 2015-09-07 at 10 58 26 am

After:
screen shot 2015-09-07 at 10 57 55 am

PR #721, opened 2015-09-07
Customize student dashboard styles for narrow screens

nathanstitt:
Fixes a few styling issues I noticed when working on #720.

When the screen is narrow, the progress and due columns collapse under
the description.  In that case the line height needs to be adjusted so
it doesn't double the row height

Before:
screen shot 2015-09-07 at 12 13 51 pm

After:
screen shot 2015-09-07 at 12 09 17 pm

### DIFF
--- a/resources/styles/components/student-dashboard/panel.less
+++ b/resources/styles/components/student-dashboard/panel.less
@@ -18,13 +18,18 @@
       font-weight: 900;
       color: @tutor-neutral-darker;
     }
-    
+
     .date-range {
       #fonts .sans(1.4rem, @student-dashboard-row-height);
       color: @tutor-neutral-darker;
       text-transform: uppercase;
     }
-
+    // Adjust height for when the columns collapse and progress appears below task title
+    @media screen and ( max-width: @screen-sm-min ){
+      // don't take up as much vertical space when collapsed
+      .panel-title { height: @student-dashboard-row-height / 2; }
+      .progress-label { padding-left: 0; }
+    }
   }
 
   .panel > .panel-body {
--- a/resources/styles/components/student-dashboard/task.less
+++ b/resources/styles/components/student-dashboard/task.less
@@ -35,13 +35,17 @@
       }
     }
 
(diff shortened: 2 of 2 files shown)


pandafulmanda:
:+1: although it looks a little busy since all the fonts are the same weight and size...could it help to have the second row of status and date be lighter weight and/or smaller and/or italic?

nathanstitt:
Yeah that's a good idea.  I'll play with that and see how it looks

Fredasaurus:
top and right aligning the due date would help too.

pandafulmanda:
i initially was expecting the dates to be in the top right too, but then i thought about how that would start making things wonky with the longer assignment titles :disappointed:

Fredasaurus:
Probably the best approach to this would be to take a mobile design and work out. What would this look like @ 320 wide?

nathanstitt:
@Fredasaurus, well this is interesting.  As the screen is currently setup it doesn't actually collapse on mobile, it only does so if the browser is shrunk smaller than mobile sizes.  So that's a total fail.   Here's how it appears on the iPhone 5 simulator (4 & 6 are same),  I also verified by loading tutor-dev on an iPhone 6.  The progress running into the text is addressed in #720

screen shot 2015-09-08 at 10 28 04 am

Fredasaurus:
doh. yeah it was out of scope. I'm doing a quick and dirty mobile mockup.

nathanstitt:
Cool - I wasn't trying to introduce new stuffs, just noticed it looked kinda horrid and wanted to fix that up.  I'll await your mockup before taking it further :grinning:

philschatz:
@Fredasaurus so, objections to this getting merged now? and then a mobile mockup for later?

philschatz:
@Fredasaurus I'm assuming the silence means "No Objections" :smile: If there are additional issues, then they can come in later

Fredasaurus:
Yeah, it''l work for now. It's something I'd like to revisit in the near future.

PR #723, opened 2015-09-08
Refactor/browse the book new tab link cleanup

pandafulmanda:
No UI changes, `BrowseTheBook` links and buttons throughout, still look as expected

screen shot 2015-09-08 at 1 18 50 pm
screen shot 2015-09-08 at 1 19 19 pm
screen shot 2015-09-08 at 1 19 25 pm
screen shot 2015-09-08 at 1 19 42 pm
screen shot 2015-09-08 at 1 20 46 pm

### DIFF
--- a/src/components/buttons/browse-the-book.cjsx
+++ b/src/components/buttons/browse-the-book.cjsx
@@ -1,6 +1,6 @@
 React = require 'react'
-BS = require 'react-bootstrap'
 _ = require 'underscore'
+NewTabLink = require '../new-tab-link'
 {CourseStore} = require '../../flux/course'
 
 module.exports = React.createClass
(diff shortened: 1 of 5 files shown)


src/components/buttons/browse-the-book.cjsx
```diff
@@ -21,25 +21,23 @@ module.exports = React.createClass
...
 
-  getLinkProps: (link) ->
-    linkProps = {target:'_blank', className:'view-reference-guide', href: link}
+  getLinkProps: ->
+    linkProps =
+      className: 'view-reference-guide'
+
     linkProps.className += " #{@props.className}" if @props.className
-    text = @props.children or 'Browse the Book'
+    linkProps.className += " btn btn-#{@props.bsStyle}" unless @props.unstyled
```
nathanstitt: Hrm, can we maybe add classnames lib?  We're already including it in the build since react-bootstrap uses it internally.

+1 for renaming the component, it's not always a `Button`

pandafulmanda: could classnames lib be a separate pr? and is that still the standard/best react classname lib at this point?

nathanstitt: yeah I suppose it should be.  They're still [recommending it](https://facebook.github.io/react/docs/class-name-manipulation.html), and like I said it's something we're already including indirectly thanks to react-bootstrap.

philschatz: :+1: for adding `classnames` lib but as a separate PR. It should be easy to find all the places by just searching for `className` :smile:

PR #724, opened 2015-09-08
Scroll window down so targeted elements are not hidden beneath navbar

nathanstitt:
After chatting with @pandafulmanda and @Fredasaurus they were in favor of a smooth scroll effect to navigate to elements.  This may be completely deprecated by @pandafulmanda's #725 but there may also be times where we want to smoothly scroll to an element. 

The gif doesn't really have enough frames to capture the smoooothness: 
scroll

### DIFF
--- a/src/components/scroll-to.cjsx
+++ b/src/components/scroll-to.cjsx
@@ -7,6 +7,8 @@ _ = require 'underscore'
 Position = require './get-position-mixin'
 
 DEFAULT_DURATION   = 750 # milliseconds
+# This is calculated to be enough for the targeted element to fit under the top navbar
+# The navbar's height is controlled by the less variable @tutor-navbar-height from global/navbar.less
 DEFAULT_TOP_OFFSET = 80  # pixels
 
 # http://blog.greweb.fr/2012/02/bezier-curve-based-easing-functions-from-concept-to-implementation/
--- a/resources/styles/global/navbar.less
+++ b/resources/styles/global/navbar.less
@@ -1,3 +1,4 @@
+// If this is updated, the scroll-to component's DEFAULT_TOP_OFFSET should also be adjusted
 @tutor-navbar-height: 61px;
 @tutor-navbar-padding-horizontal: 4rem;
 


pandafulmanda:
:+1:

src/components/book-content-mixin.cjsx
```diff
@@ -16,6 +17,10 @@ MEDIA_LINK_EXCLUDES = [
   '[data-targeted=media]'
 ]
 
+# elements that are currently targeted by the URL's hash are
+# scrolled down this much in order to appear below the top navbar
+FOCUSED_TARGET_TOP_POSITION = 100
```
philschatz: Is this variable still used?

src/components/scroll-to.cjsx
```diff
@@ -0,0 +1,65 @@
+React = require 'react'
+_ = require 'underscore'
+
+# Note that the Position mixin methods are called directly rather than mixing it in
+# since we're a mixin ourselves our consumers also include GetPosition and it causes
+# duplicate method name errors to mix it in
+Position = require './get-position-mixin'
+
+DEFAULT_DURATION   = 750 # milliseconds
+DEFAULT_TOP_OFFSET = 80  # pixels
```
philschatz: Is/can this be somewhere in the CSS near the selector or the bootstrap var that causes the navbar to be `80px` high?

src/components/scroll-to.cjsx
```diff
@@ -0,0 +1,65 @@
...
+    win       = @props.windowImpl
+    startPos  = win.pageYOffset
+    endPos    = Position.getTopPosition(el) - _.result(@, 'getScrollTopOffset', DEFAULT_TOP_OFFSET)
+    startTime = Date.now()
+    duration  = _.result(@, 'getScrollDuration', DEFAULT_DURATION)
+    requestAnimationFrame = win.requestAnimationFrame or _.defer
+
+    step = ->
+      elapsed = Date.now() - startTime
+      console.log "SCROP"
```
philschatz: wha?

src/components/scroll-to.cjsx
nathanstitt: I'd like it to be, but not sure how to share var's between CSS and JS.  Is there a pattern for that?  The navbar is 69px in total, so 80 gives it just a bit more space at the top to look nicer.

src/components/book-content-mixin.cjsx
nathanstitt: Nope :(  Removed.

src/components/scroll-to.cjsx
nathanstitt: Oh man.  Total fail there.  I think it was supposed to be "step" but I didn't bother to fix the spelling because all I was really testing was that it was being called.  (and then forgot to remove of course).

src/components/scroll-to.cjsx
philschatz: I wasn't suggesting sharing vars, just having a common string to grep for if we change the CSS to be something other than `69px`

philschatz:
How difficult would it be to update the hash when clicking on a link?

PR #727, opened 2015-09-09
Rename link from Scores Report to Student Scores

nathanstitt:
@philschatz I think I remember you saying there was another location that the earlier rename missed?


philschatz:
Not off the top of my head... just the XLS file in https://github.com/openstax/tutor-server/pull/648

nathanstitt:
Cool, I must have mis-remembered then.  Thanks

PR #729, opened 2015-09-09
Refactor/improve exercise loading

pandafulmanda:
Ref Book Pages renders much faster and is more reliable/less likely to crash/be blocked by stuffs

Improvements were made by:
* making one api request for exercises per page
  * as opposed to one for each exercise, which added up to 20 + requests sometimes
* ensuring each exercise only renders once per page load
  * there were at least 2 renders per exercise for whatever reason
* making exercise rendering strictly static
  * after they are loaded, the exercises don't need to be reactive
* link processing is deferred till after the rest of the page is mounted/updated/rendered

### DIFF
--- a/src/components/bind-store-mixin.cjsx
+++ b/src/components/bind-store-mixin.cjsx
@@ -2,7 +2,7 @@ module.exports =
 
   # can modify which event you want to bind on as needed.
   _bindEvent: ->
-    @bindEvent or 'change'
+    @bindEvent or @props.bindEvent or 'change'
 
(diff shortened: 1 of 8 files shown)


philschatz:
Does this require a backend change to interpret the API call?

pandafulmanda:
nope

nathanstitt:
:+1:  Just tested this locally and worked awesome.  The refview now scrolled smoothly after loading where it would stutter for a few seconds before.

Something I did notice is that the URL for requesting exercises is pretty long.  Do we need to worry about hitting the 2K char limit for url length?  For physics sect 1.2 it was 394 chars which is safely well below that though so probably not.

PR #731, opened 2015-09-10
Disable right-click context menu on external assignments

nathanstitt:
If someone selects "open in new tab" from the menu, the browser doesn't
receive any events and the code can't mark the step as completed.

No UI changes other than the menu doesn't appear.  Tested in Chrome/Firefox/Safari.  Will test IE later.


pandafulmanda:
:+1:

PR #732, opened 2015-09-12
types scores report headings for different assignment types

pandafulmanda:
screen shot 2015-09-16 at 6 12 27 pm


Fredasaurus:
You left work early, then sent this out from a bar later on. You're not fooling anybody.

I actually have a mockup for this that didn't get published. Let's catch up on Monday.

Fredasaurus:
Also, not the worst.

pandafulmanda:
oh bt dubs, i won't be in monday.  could you post an image of the mockup?

nathanstitt:
Something I was gonna play around with at some point was replacing the “Review” with a magnifying glass icon and shrinking the size of the "Due" text.  That would allow the columns to be about 1/2 the current width, which would in turn allow quite a few more assignments to fit horizontally.

That would also mean that the course title would start show ellipsis fairly often.  I'm not sure if we can detect when that occurs, but if I could figure out a method to do so, I'd  add a tooltip that would show the complete title.

Fredasaurus:
I think we can get rid of "due" altogether. In this context it doesn't carrying much meaning and it's repetitive. I'll get a mockup shortly.

philschatz:
Some other ideas on collapsing: What about always having a tooltip for all titles, have them be 2 rows, and a shortened due data (`9/14`)? RE Ellipses: Might be useful to get a list of assignments from production to get an actual sample set to see if ellipses are useful.

nathanstitt:
Sorry, wasn't clear there.  The titles already show ellipsis when they overflow.  I was referring to detecting when the ellipsis are shown and only setting the tooltip in that case.  You're essentially correct though, we could (should?) just set the tooltip regardless.

philschatz:
@Fredasaurus any updates on the mockup?

Fredasaurus:
Met with Amanda this morning along with UX folks. Still working on a solution. Will have it this afternoon.

Fredasaurus:
OK like late afternoon.

student-scores

resources/styles/components/scores/index.less
```diff
@@ -30,6 +30,7 @@
     max-width: 100%;
     height: 100%;
     max-height: 100%;
+    margin-top: -20px;
```
philschatz: Is this negative margin used elsewhere? If so, could we pull it out as a var?

resources/styles/components/scores/index.less
```diff
@@ -64,23 +70,63 @@
...
     &.is-descending:after { content: @fa-var-sort-desc; }
+
+    &[data-assignment-type] {
+      height: 100%;
+    }
+
+    .tutor-plan-set(heading);
+  }
+
+  .scores-cell, .assignment-header-cell, .student-name {
```
philschatz: can you put each selector on a separate line? (makes scanning/diffing easier)

resources/styles/components/scores/index.less
```diff
@@ -64,23 +70,63 @@
...
+
+    > div {
+      display: block;
+      width: 100%;
+      height: 50%;
+      padding: 8px;
+    }
+
+    > div:nth-child(even) {
+      background-color: #f6f7f8;
```
philschatz: Could this color be defined in a var?

resources/styles/components/scores/index.less
```diff
@@ -64,23 +70,63 @@
...
+
+      > div:nth-child(even) {
+        mix-blend-mode: overlay;
       }
     }
+
     .review-plan {
-      float: right;
-      margin-right: 1rem;
+      &:not(.btn-link):not(.btn-flat):not(.btn-fab).btn {
```
philschatz: Could we add a semantic class name to the button(s) to style rather than exclude the rest?

resources/styles/mixins/tutor-plans.less
```diff
@@ -56,11 +78,31 @@
...
 }
 
-.tutor-plan-header(@plan-color) {
+.tutor-plan-cell(@plan-color) {
+  color: @tutor-neutral-darker;
+  background-color: fade(@plan-color, 2.5%);
+  mix-blend-mode: multiply;
+  .transition(background-color 0.1s ease-in);
+
+  &:not(div):hover {
```
philschatz: could the `div` be selected better?

resources/styles/components/scores/index.less
pandafulmanda: this i'd like to do in a separate PR, we have this wonky selector stuff with buttons elsewhere as well, and it needs to be fixed

philschatz: yep ok

PR #733, opened 2015-09-14
Anti-aliasing merriweather

helenemccarron:
Add -webkit-font-smoothing: antialiased; -moz-osx-font-smoothing: grayscale; to Merriweather

### DIFF
--- a/resources/styles/global/typography.less
+++ b/resources/styles/global/typography.less
@@ -41,6 +41,8 @@
     font-style: normal;
     font-size: @size;
     line-height: @line-height;
+    -webkit-font-smoothing: antialiased;
+    -moz-osx-font-smoothing: grayscale;
   }
   .script (@size: 1.5rem, @line-height: 3rem, @style: normal) {
     font-family: 'Architects Daughter', cursive;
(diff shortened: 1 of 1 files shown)


resources/styles/global/typography.less
```diff
@@ -41,6 +41,8 @@
     font-style: normal;
     font-size: @size;
     line-height: @line-height;
+    -webkit-font-smoothing: antialiased;
+    -moz-osx-font-smoothing: grayscale;
```
philschatz: Is there a non-prefixed rule for this we can include as well?

Fredasaurus:
There is, but it is non-standard and generally not supported.

http://caniuse.com/#feat=font-smooth

resources/styles/global/typography.less
helenemccarron: Do you still want it or should we skip it?

philschatz: Does it break if we add it, if not, then probably? It seems like the rest of the vendor-prefixes we use also contain the non-prefixed versions (so browsers *just work* once the rule is officially adopted)

helenemccarron: Right, but in this case, the default value was dropped. I added it, even if it doesn't do anything. It's been removed from the specification and is currently not on the standard track, but who knows?

PR #734, opened 2015-09-14
removed 'practice weakest' button from student dash

darrenmason:
This is regarding task:  https://www.pivotaltracker.com/story/show/102283464

'Practice Weakest in dashboard is misleading '

This removes the button.

image

### DIFF
--- a/src/components/student-dashboard/progress-guide.cjsx
+++ b/src/components/student-dashboard/progress-guide.cjsx
@@ -9,7 +9,6 @@ ChapterSection = require '../task-plan/chapter-section'
 ChapterSectionMixin = require '../chapter-section-mixin'
 LearningGuideSection = require '../learning-guide/section'
 LearningGuideColorKey = require '../learning-guide/color-key'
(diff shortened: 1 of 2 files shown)


src/components/student-dashboard/progress-guide.cjsx
```diff
@@ -94,9 +94,6 @@ ProgressGuidePanels = React.createClass
 
         <ProgressGuide sections={recent} courseId={@props.courseId} />
 
-        <PracticeButton ref='practiceBtn' title='Practice my weakest topics'
-            courseId={@props.courseId} sections={practiceSections} />
-
         <BS.Button
```
nathanstitt: Is the `practiceSections` variable used now that the button is no more?  I suspect that we might be able to remove it from line 88-90 above.  Can we remove the `require` for the `PracticeButton` component as well?

darrenmason: will do

philschatz:
The original ticket says this is moved to the next sprint. Closing for now (thanks for doing it though!)

PR #737, opened 2015-09-15
Fixing builder set date

jivewise:
Fixing the bug described here: https://www.pivotaltracker.com/n/projects/1156756/stories/103133164

The bug was due to calling setDueAt and setOpensAt with a date object instead of a moment.  It should be able to handle both now.

Also adding in some specs.

### DIFF
--- a/src/components/task-plan/builder.cjsx
+++ b/src/components/task-plan/builder.cjsx
@@ -70,7 +70,6 @@ module.exports = React.createClass
     opensAt
 
   getQueriedDueAt: ->
-    window.moment = moment
     {due_at} = @context?.router?.getCurrentQuery() # attempt to read the due date from query params
     isNewPlan = TaskPlanStore.isNew(@props.id)
     dueAt = if due_at and isNewPlan then moment(due_at).toDate()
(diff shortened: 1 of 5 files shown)


philschatz:
When I select a date clicking the Calendar (or the little popup calendar) the due date is set to 1 day before the date I selected.

jivewise:
That bug should be fixed, also added specs to make sure it doesn't happen again.

philschatz:
When I select tomorrow (9/17) from the Calendar the due date is still (9/16)  (maybe timezone related?):

image


The little popup calendar however does seem to work properly now ; )

jivewise:
That calendar error was due to us setting the default dates before we set the correct timezone in the componentWillMount function.  That should be good to go now.

src/components/task-plan/builder.cjsx
```diff
@@ -69,6 +70,7 @@ module.exports = React.createClass
     opensAt
 
   getQueriedDueAt: ->
+    window.moment = moment
```
philschatz: Is this line needed?

philschatz:
can you add a test to check that the dates are set correctly when mounted?

jivewise:
Yep, adding a test now.

PR #738, opened 2015-09-15
show distractors and style free response tables

pandafulmanda:
## Before
screen shot 2015-09-15 at 10 49 33 am
screen shot 2015-09-15 at 11 56 29 am
Table styles are missing for free response, they are in multiple choice portion

## After
screen shot 2015-09-15 at 12 03 19 pm
screen shot 2015-09-15 at 11 55 06 am

## Tables cascading as expected
after changing the mixin to be a base one with variations in a separate mixin
## only under `.has-html`
screen shot 2015-09-17 at 12 32 51 pm
## in `.book-content-mixin`, custom colors on top of `.has-html`
screen shot 2015-09-17 at 12 33 10 pm
## in a question, in reference view, and in `has-html`
screen shot 2015-09-17 at 12 32 49 pm

### DIFF
--- a/resources/styles/components/html.less
+++ b/resources/styles/components/html.less
@@ -1,0 +1,4 @@
+.has-html {
+  .tutor-tables();
+  .tutor-tables(@tutor-white, @tutor-white);
+}
--- a/resources/styles/components/question.less
+++ b/resources/styles/components/question.less
@@ -192,10 +192,11 @@
       }
 
       // For questions showing feedback,
-      // if the answers are not correct and are not chosen, hide the distracting answer.
-      // answers-answer still needs to render so that answer-letter increments as expected.
+      // if the answers are not correct and are not chosen, tone down the distracting answer.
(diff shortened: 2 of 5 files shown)


philschatz:
:+1: Should the `.tutor-tables(@tutor-white, @tutor-white);` code go in the `.arbitrary-html` and `.book-content` classes so they are styled everywhere? CCing @helenemccarron

pandafulmanda:
I think the `.tutor-tables` is already in `book-content`.  the issue was that tasks do not have those full classes/styles for book content applied to the exercise task steps, and only to the reading steps.

philschatz:
but tasks do have the `.has-html` right? Like exercises, assignment descriptions, and maybe the popovers as well (anything that uses the `.has-html`)

pandafulmanda:
i think it would be nice to put it in `.has-html` but i think the table style in `book-content` is ever so slightly different.  i guess we could pull those differences out and make that specific in `book-content` and have tables be styled as in `tutor-tables` elsewhere?

PR #740, opened 2015-09-15
Show new tab indicator on external tasks

nathanstitt:
Converts the less styles to a mixin and shares it between book-content
and the external tasks.

This also shrinks the indicator a bit to make it only 80% of the text size.
Before:
screen shot 2015-09-15 at 3 20 20 pm

screen shot 2015-09-15 at 3 19 56 pm

After:
screen shot 2015-09-15 at 4 21 09 pm

screen shot 2015-09-15 at 4 19 14 pm

### DIFF
--- a/resources/styles/book-content/target.less
+++ b/resources/styles/book-content/target.less
@@ -8,13 +8,7 @@
   padding-top: @tutor-navbar-height;
 }
 
-a[target=_blank] {
-  &::after {
-    .fa-icon();
-    content: @fa-var-external-link;
-    margin: 0 4px 0 6px;
-  }
-}
+a[target=_blank] { .tutor-external-link-indicator(); }
 
 figure {
   &:target, .target {
--- a/resources/styles/components/task-step/all-steps.less
+++ b/resources/styles/components/task-step/all-steps.less
@@ -25,6 +25,7 @@
   }
 }
 
+@import './external';
 @import './reading';
 @import './video';
 @import './ends';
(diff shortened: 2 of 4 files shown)


philschatz:
rather than duplicate the `a[href='_blank']` could/should that be part of the mixin? and then just include the mixin in more places after this sprint? (like `.has-html`)?

nathanstitt:
Yes, I considered that but then it'd lock us into only being able to use the mixin with elements that match.  

I ran a small test where I applied it globally and discovered that the "My Account" link doesn't have that attribute.  There might be others as well that use `window.open` for legit reasons and I wanted to keep the mixin flexible enough to also style them.

My thoughts were that after the sprint we'd setup a rule to include the mixin globally onto all the `a[href='_blank']` elements, and then go through and add target any link-like elements that also need it.

philschatz:
thoughts on the icon size @Fredasaurus ? I think once we start including it in other places the size of the link won't be an issue but it would be nice to have a consistent rule for what the icon size should be (ie always the same as the text around it, always x%, etc)

nathanstitt:
Thanks for the font-size tip @philschatz .  I'm setting it at 85% and I think that's about right, can tweak up or down in response to @Fredasaurus feedback

Fredasaurus:
Can we try one at uppercase letter height (a liiiiiiitle smaller than it is currently) and one at lower? Thanks.

nathanstitt:
Sure.  

Uppercase letter height is 100%:
screen shot 2015-09-15 at 3 30 05 pm

Looks like we can't go below %80 because then the underline gets wonky.  This is 78%:
screen shot 2015-09-15 at 3 30 57 pm

And here's %80:
screen shot 2015-09-15 at 3 31 24 pm

Fredasaurus:
80% looks good.

nathanstitt:
80% it is then

PR #742, opened 2015-09-16
Updating gulp tdd

jivewise:
Making 'gulp tdd' use the chrome browser when running tests.

To enable chrome debugging, run the following command:

```
export KARMA_BROWSER='Chrome' && gulp tdd
```


test/karma-dev.config.coffee
```diff
@@ -0,0 +1,12 @@
+{webpack} = require './karma.common'
+_ = require 'underscore'
+commonConfig = require './karma.common'
+
+module.exports = (karmaConfig) ->
+
+  config = _.extend(commonConfig, {
+    browsers: ['Chrome']
+    reporters: ['nyan']
```
philschatz: Hehe, fun surprise :smile:

philschatz:
I'm not quite sure how to read the results of failed tests using the nyan reporter. Here is my result from running `gulp tdd`, adding `throw new Error('nyan error')` and saving in the middle of `builder.cjsx`, and then removing the error I just introduced:

image

jivewise:
Hm, I noticed something similar, I don't think it's so much the reporter, because that was happening when I switched to using Chrome.  I'll see if we can get better errors though.

jivewise:
I was wrong, it was a config option with the nyan cat reporter.  Here's an example of the new reporting:
screen shot 2015-09-17 at 12 43 32 am

philschatz:
What is the status of this PR? @nathanstitt mentioned TDD testing only the files changed in the other webpack PR

jivewise:
I think this is ready for another round of reviews, I integrated Nathan's changes into this PR, and it'd be nice to get some feedback.

nathanstitt:
@jivewise I really like the method of killing of karma and restarting it.  That actually seems faster than it was before so that's a win for me.

I also tested using phantomjs and that seemed even faster yet.  I'm wondering if we could set that via an environmental variable?  if you set the browser config in `karma-dev.config.coffee` to `ENV['KARMA_BROWSER'] or 'PhantomJs'` then we could pick and choose.  I've also discovered that you can run the specs in a browser even if the `PhantomJS` browser is in use by loading `http://localhost:9876/`.  That was tricky before when the Karma would exit immediately, but is much better now that you've got the "run until killed" working.

I'm not 100% sure about the use of `nyan`  reporter.  It's cool right now, but i'm not sure how well it'll be tolerated after the 1,000 time running it.  We could always go back if it does get old though so I'm fine with it for now.

philschatz:
Nice speed improvement and seems to be more stable!

:+1: on using PhantomJS by default but adding an env option for Chrome. Every time I switch branches Chrome starts up right as I start testing the code :smile: 

But this combined with the `react-hot-reloader` is a really quick response time for making changes!

jivewise:
Okay, I've updated it to use an environment variable.  Maybe in the future we can have something like: ```gulp tdd debug=1`` to trigger chrome?

PR #743, opened 2015-09-16
Fix homework review jumpiness -- make it animate nicely

pandafulmanda:
### Step should now move to completed list as completed

out

### DIFF
--- a/src/components/task-step/ends.cjsx
+++ b/src/components/task-step/ends.cjsx
@@ -96,7 +96,7 @@ HomeworkEnd = React.createClass
           goToStep={@goToStep}
           onNextStep={@onNextStep}
           review={type}
-          ref={"task-review-#{type}"}
+          key={"task-review-#{type}"}
           focus={type is 'todo'}/>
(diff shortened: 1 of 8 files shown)


pandafulmanda:
upon further review, this is not a bug, and may actually be a change request

pandafulmanda:
Although, this should be merged for the fixes it provides, and the added attention to the experience, and we can make a separate PR for the specific behavior of the button on the review page

src/components/task-step/ends.cjsx
```diff
@@ -86,6 +96,7 @@ HomeworkEnd = React.createClass
           goToStep={@goToStep}
           onNextStep={@onNextStep}
           review={type}
+          ref={"task-review-#{type}"}
```
philschatz: Is this ref used anywhere?

pandafulmanda: oh...hmm...i may have been trying to make that a key actually because of warnings about keys...

PR #745, opened 2015-09-16
Animated reference view menu toggle control

nathanstitt:
The old implementation worked by adding the font icon and overlaying it with an invisible SVG, then fading in the SVG on hover.  This worked fine in chrome but Firefox and IE didn't line up the two layers properly and looked jagged.

This implementation introduces a new SVG based component that is animated via css transforms.  Unfortunately IE doesn't support setting transforms via CSS so they have to also be inline, which causes a few duplicated values :(

The gif doesn't really show it but the icon's bars shrinking and the triangle sliding in are also animated.

menu-toggle

It used to look like this in Firefox and IE
screen shot 2015-09-16 at 2 48 28 pm


pandafulmanda:
! c:sunglasses::sunglasses:l!

PR #747, opened 2015-09-17
QA review of exercises

nathanstitt:
screen shot 2015-09-30 at 8 59 35 pm

Todo

 * [x] setup separate web pack build of assets
 * [x] Minimal router config for browsing book
 * [x] Modify reference book component to rely on passed properties instead of router context
 * [x] Load exercises in place of normal book content

Depends on BE PR https://github.com/openstax/tutor-server/pull/666


philschatz:
Why have a separate router?

nathanstitt:
A separate router is created in order to keep the build as small as possible.  If both builds share a common router, then there's not going to be any difference between them since the router would pull in all the dependancies view cascading `require` statements.

I'm also currently sharing `api.coffee` between the builds which pulls in all the flux stores even though the exercises will only need a few of them.  I figured it wasn't worth splitting that apart though.

philschatz:
Could we do the router splitting as a separate PR because I think the *biggest bang* (to quote JP) is splitting up the student and teacher screens which *could* be done using webpack chunking while using the same router (if I understand the chunking properly)

nathanstitt:
@philschatz I've investigated this a bit and It's as I'd assumed.  Webpack doesn't actually know what files are used by a build, it only knows which files were required.

The router requires all the main components, which in turn require everything else in the app.  If the router is required, the build will include all the files that make up the app.

I ran a small test of building for production with and without the router. Right now an minified but un-gzip'ed JS build is 910Kb for exercises and 1,495Kb for tutor.   

When I required the main router instead of the slimmed down exercise router, the exercise build jumped to be 1,496Kb.  I'm guessing the extra 1Kb is because of the exercises specific files that aren't in use by the main app.

philschatz:
Ah, I forgot that the router does not have an async require option (it needs all of the actual components). 

It still seems like (with code organization) we could still asynchronously load the actual components so, for example, a teacher that does not use the Perf Forecast, wouldn't download it yet.

PR #748, opened 2015-09-17
Add classnames lib

nathanstitt:
It looked to me like the tutor-input's would benefit the most from it so that was chosen as the place to introduce it at.  

I'm sure there's other spots that could also use refactoring as well, but not sure if we need/want to target them all at once.

### DIFF
--- a/src/components/tutor-input.cjsx
+++ b/src/components/tutor-input.cjsx
@@ -148,9 +148,12 @@ TutorDateInput = React.createClass
     classes = classnames 'form-control',
       empty: (not @props.value and not @state.hasFocus)
 
+    isDatePickerDisabled = @props.disabled and value
+
     wrapperClasses = classnames 'form-control-wrapper', 'tutor-input', '-tutor-date-input', @props.className,
       'is-required': @props.required
       'has-error': @state.errors?.length
+      'disabled-datepicker':  isDatePickerDisabled
 
     now = TimeStore.getNow()
     value = @props.value
@@ -176,9 +179,8 @@ TutorDateInput = React.createClass
           selected={value}
           weekStart={@props.currentLocale.week.dow}
         />
-    else if @props.disabled and value
+    else if isDatePickerDisabled
       displayValue = value.format(TutorDateFormat)
-      wrapperClasses += ' disabled-datepicker'
 
     <div className={wrapperClasses}>
       <input type='text' disabled className={classes} value={displayValue}/>
--- a/src/components/tutor-input.cjsx
+++ b/src/components/tutor-input.cjsx
@@ -2,6 +2,7 @@ React = require 'react'
 BS = require 'react-bootstrap'
 moment = require 'moment'
 _ = require 'underscore'
+classnames = require 'classnames'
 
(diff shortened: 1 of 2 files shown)


philschatz:
Just curious if you had thoughts on the following notation:

```coffee
classes = classnames
  'assignment': true
  'is-foo`: foo
```
rather than mixing the different args and sometimes having them run on to separate lines

pandafulmanda:
I :+1: @philschatz's comment

nathanstitt:
I think we can have the best of both worlds.  I kinda like the flexibility of setting classes using plain strings at the start.  Plus that automatically excludes null & undefined values.  How about:
```coffee
classes = classnames 'foo', 'bar', @anUndefinedClass,
  'is-foo': @props.isFoo
  bar: @barName
```
I've just reworked PR to use the syntax as well, what do you both think?

src/components/tutor-input.cjsx
```diff
@@ -47,15 +48,12 @@ TutorInput = React.createClass
   forwardLabelClick: -> @focus()
 
   render: ->
-    classes = ['form-control']
-    wrapperClasses = ["form-control-wrapper", "tutor-input"]
-    wrapperClasses.push(@props.className) if @props.className
+    classes = classnames 'form-control', @props.class,
```
philschatz: Might be too much to change but should `TutorInput` take `className` as a prop instead of `class`?

/ cc @jivewise

src/components/tutor-input.cjsx
```diff
@@ -181,10 +178,10 @@ TutorDateInput = React.createClass
         />
     else if @props.disabled and value
       displayValue = value.format(TutorDateFormat)
-      wrapperClasses.push('disabled-datepicker')
+      wrapperClasses += ' disabled-datepicker'
```
philschatz: why concatenate here instead of setting it above?

philschatz:
Yep, I think that format is best-of-both :smile:

pandafulmanda:
I'm okay with that format, but the neat thing about:

```coffee
classes = classnames
  'assignment': true
  'is-foo`: foo
  'is-published': true
```

is that removing, adding, or commenting out classes becomes very easy.

src/components/tutor-input.cjsx
nathanstitt: No, but I suspect that it probably should been.  There's a  `classes.push(@props.class)` on the old line 58 below.  I opted to leave it as-is for now.

src/components/tutor-input.cjsx
nathanstitt: Yeah, I was on the fence about this.  I didn't add it to the main setting above since I didn't want to duplicate the logic condition in two places.  Your comment forced me to think about it a bit more and realize that we have these nifty things called variables that can hold the results of a condition and be checked in multiple places :)   I've reworked to do that.

src/components/tutor-input.cjsx
philschatz: ah, I was only looking at the diff above this line, not below ; )

PR #750, opened 2015-09-18
stash answer change when waiting for server, save when next ready

pandafulmanda:
Alternate solution to #749, fix for #751

Clicking on answers while save is still pending shows a change to the student as choice has been selected.  The answer/server save doesn't happen until the the free response is saved

## Fix
1. Answer a free response
1. Choose a multiple choice answer
  * You are able to click and select answer
  * Answer does not actually send a `PATCH` to server until previous response is returned

### DIFF
--- a/src/components/task-step/exercise/modes.cjsx
+++ b/src/components/task-step/exercise/modes.cjsx
@@ -87,6 +87,17 @@ ExerciseMultiChoice = React.createClass
     onStepCompleted: React.PropTypes.func.isRequired
     onNextStep: React.PropTypes.func
 
+  queueAnswer: (answer) ->
+    @setState(queuedAnswer: answer)
+
+  unqueueAnswer: ->
+    @queueAnswer(null)
+
+  componentDidUpdate: (prevProps, prevState) ->
+    {id} = @props
+    isReady = not TaskStepStore.isLoading(id) and not TaskStepStore.isSaving(id)
+    @onAnswerChanged(@state.queuedAnswer) if @state.queuedAnswer? and isReady
+
   renderBody: ->
     {id} = @props
     {content, free_response, answer_id, correct_answer_id, feedback_html} = TaskStepStore.get(id)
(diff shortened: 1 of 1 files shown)


philschatz:
I like that #749 exposes (to QA and students) when we are slow (the choices are disabled until the free response is saved) and if an error occurs it is clear that it occurred during the free response. Thoughts @kajalp? In #750 the student is immediately able to select a multiple choice answer even though the free response is not saved yet (or has not errored yet)

kajalp:
I like #749 over #750. #750 has the potential to make things go very wrong? ex Scenario - free response not saved and student selects MCQ answer and then click on continue.. at this time we are trying to save free response, mcq answer, and trying to load the response (if needed)?

pandafulmanda:
you will not be able to click on continue until the multiple choice actually saves

pandafulmanda:
I've added gifs of the experience with #749 and #750 to help show what this looks like in action @kajalp and @philschatz .  This realistically won't happen very often since the free response saving doesn't usually take longer than 2s.

pandafulmanda:
any thoughts after review of the experience?

philschatz:
Thanks for updating the animations! The "continue" spinner *is* pretty clear; I just wonder what maintaining unit tests would be like. Maybe time to ask @Fredasaurus (and customer support) if it's *OK* for students to get an error after they've selected a Multiple Choice problem but before they clicked "Continue" (the Error is from the Free Response not saving, not from selecting a Multiple Choice response)

Fredasaurus:
MY issue with 750 is the the "saving" occurs after the screen has passed, leading to a potential disconnect about what is actually being saved. I think I like 749 better for this reason.

Fredasaurus:
OK, now that Amanda has explained these to me, I don't like either one.

What is bad about the current solutions is that:
1. the user is presented with choices they cannot select, or
2. the multiple choice options are available with a distracting non-relevant loading icon in the corner.

I would like us to either:
1. hold the user on the free response screen until the response is saved,
2. silently handle the save in the background so the the user can proceed to the multiple choice without delay.

pandafulmanda:
okay, ill come up with some adjustments

pandafulmanda:
updated #749 to not display the multiple choices until the free response has been saved, as stated by @Fredasaurus  as option 1.  I am close this PR in favor of #749

PR #753, opened 2015-09-18
ref book nav hide and show

pandafulmanda:
### candidate for closing in favor of #757
#757 addresses #754 as well and is better for components overall

## Fixes
- [x] on prev and next click, TOC preserves open or close state
- [x] on TOC section click, TOC closes in small view  -- @nathanstitt maybe a separate PR or this one?
  * setState toggling the menu close doesn't get to complete before the shell re-renders from the route change

## Tests
- [x] prev-next `menu-open` class preserved
- [x] TOC item `menu-open` absent after click

## Prev and Next click, preserves TOC open/close state
fix

## TOC item click, auto-closes TOC
fix

PR #757, opened 2015-09-22
Remove courseId and router references from reference view

### LINKED ISSUE #754: Reference book next and prev re-opens TOC
## Prev, Next re-opens TOC
This is annoying, prev next re-opens TOC even though it was manually closed
bug

## TOC does not auto-close
Also, for a small window, TOC should close automatically when a TOC item is navigated to
bug

nathanstitt:
This is a step towards repurposing the reference view components for QA in #747 

It removes all the logic for getting the ecosystem from a courseId into the shell so the main component can be wrapped by different components.

As a side effect of the extraction, I think it also fixes the same bugs that #753, fixes #754

### DIFF
--- a/src/components/reference-book/index.cjsx
+++ b/src/components/reference-book/index.cjsx
@@ -4,60 +4,19 @@ BS = require 'react-bootstrap'
 _  = require 'underscore'
 
 {ReferenceBookActions, ReferenceBookStore} = require '../../flux/reference-book'
-{ReferenceBookPageActions, ReferenceBookPageStore} = require '../../flux/reference-book-page'
 {CourseActions, CourseStore} = require '../../flux/course'
-{Invalid} = require '../index'
-
+ReferenceBookPageShell = require './page-shell'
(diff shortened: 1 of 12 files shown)


pandafulmanda:
:+1:

src/components/reference-book/index.cjsx
```diff
@@ -69,29 +28,47 @@ ReferenceBookShell = React.createClass
...
+
+  renderBook: ->
+    classnames = []
+    classnames.push('is-teacher') if @state.isShowingTeacherContent
     {courseId, ecosystemId} = @state
 
+    <ReferenceBook
+        navbarControls={@renderNavbarControls()}
+        section={@state.section}
+        className={classnames.join(' ')}
```
philschatz: should this use the classnames lib now that it's merged?

nathanstitt: It should indeed - just refactored to do so.

PR #760, opened 2015-09-23
HW summary table

jivewise:
screen shot 2015-09-30 at 12 18 34 am

-Fixing details column in hw summary table
-Updated teks column to only use the tek name
-Added in some unit tests for flux ExerciseStore


jason-holmes:
I'm not sure if this is a WIP, but the formatting of the data in the details and TEKS columns needs attention.  Something more like what's shown in the mockup...http://mockups.openstax.org/Tutor_MVP_future_release/start.html#p=add_homework_3 .  At minimum, we need consistent formatting to distinguish the label (e.g.  Blooms) from the value (e.g., 2).  Also, remove the parentheses from the TEKS value.  
Also, the order of the details is not consistent (e.g., DOK is first in some, Blooms in first in another).  It would be good to have them always in the same order (see mockup above).

philschatz:
Is tutor-server returning just the id or also human readable text for the tags?

jivewise:
Sorry just saw this was unmerged.  I must have uploaded the wrong picture a while back, we were already using the human readable names.

I did add consistent tag ordering though, and I'm now stripping the parentheses off of the teks.  Please see updated screenshot.

src/components/task-plan/homework/exercises.cjsx
```diff
@@ -211,7 +211,7 @@ ExerciseTable = React.createClass
       unless teksString
         teksString = "-"
 
-      teks = <td>{teksString}</td>
+      teks = <td>{teksString.replace(/[()]/g, '')}</td>
```
philschatz: Can this be done in the store?

philschatz:
what do the unreadable names on TEKS tags look like?

Dantemss:
No name attribute I suppose. I think there are some that didn't have the required html tags.

philschatz:
oh, ok.

PR #761, opened 2015-09-23
guide - # students worked display text

darrenmason:
this is regarding https://www.pivotaltracker.com/story/show/102331776

image


Dantemss:
BE PR: https://github.com/openstax/tutor-server/pull/680
This PR just needs some slight renaming (I kept the same name as Biglearn to avoid confusion, since this value is actually coming from there).
The student count variable is `unique_learner_count` and it's inside the `clue` object.

Dantemss:
A question has been brought up:

X students worked Y problems

Currently, X is coming from Biglearn, Y from Tutor.

Should both come from one or the other?

philschatz:
If Biglearn is already providing the student count then maybe BigLearn? What would the problem be (except older data)?

Dantemss:
No problem to switch to Biglearn, I'm just asking because the values currently come from 2 different sources.

test/components/learning-guide/chapter.spec.coffee
```diff
@@ -15,8 +15,20 @@ describe 'Learning Guide Chapter Panel', ->
     }
 
   it 'reports how many problems were worked', ->
+    count = @props.chapter.clue.unique_learner_count
+    total = @props.chapter.questions_answered_count
+    if count > 1
```
philschatz: what about using [`pluralize`](https://www.npmjs.com/package/pluralize)?

philschatz:
The number of students worked should not appear on the student forecast

PR #764, opened 2015-09-24
trouble flag on period tabs

darrenmason:
This is regarding task: https://www.pivotaltracker.com/story/show/102331900

image


philschatz:
@Fredasaurus should the trouble flag be on the section as well as the period?

Fredasaurus:
Well, to me the class period with the issue should be obvious when looking at the progress indicators, so it wouldn't be needed -- YMMV.

Dantemss:
I think Phil meant the Page :)

resources/styles/mixins.less
```diff
@@ -90,6 +90,18 @@
         color: @tutor-neutral-darker !important;
       }
     }
+    
+    &.is-trouble {
+      a {
+      display: inline-block;
```
philschatz: *minor:* these rules should be indented. Should there be a class on `<a>`?

src/components/course-periods-nav.cjsx
```diff
@@ -57,7 +58,14 @@ CoursePeriodsNav = React.createClass
     @setState(active: key)
 
   renderPeriod: (period, key) ->
-    <BS.NavItem eventKey={key} key="period-nav-#{period.id}">{period.name}</BS.NavItem>
+    {plan} = @props
+    isTrouble = _.pluck(plan.stats, 'is_trouble')
```
philschatz: It looks like this puts a trouble flag on all periods if any contain `is_trouble`. I do not think this is the desired behavior.

src/components/course-periods-nav.cjsx
```diff
@@ -57,7 +58,14 @@ CoursePeriodsNav = React.createClass
...
+    <BS.NavItem 
+      className={'is-trouble' if _.contains(isTrouble, true)}
```
philschatz: should this be `className={'is-trouble' if period.is_trouble}`?

If so, then is the `plan` prop used anymore?

darrenmason:
switched back to period.is_trouble.  Testing with https://github.com/openstax/tutor-server/pull/679

PR #768, opened 2015-09-24
Analytics for page views

nathanstitt:
Sets up a page view tracking.  

Depends on BE PR: https://github.com/openstax/tutor-server/pull/677


src/helpers/analytics.coffee
```diff
@@ -0,0 +1,64 @@
...
+    # track all work done on a course
+    Analytics.sendEvent 'Course', 'Work', label: courseId
+
+  viewStudentDashboard: ({courseId}) ->
+    # compare activity between courses
+    Analytics.sendEvent 'Student', 'Dashboard', label: courseId
+
+
+# Translators convert a url like '/foo/bar/123/baz/1' into a simplified one like just '/foo/bar'
+Translators =
```
philschatz: Is there a way to get the route href used to build the router (I think `react-router` allows you to build a router from JSON instead of components) and either use it directly or strip out the `:****`? hmm, or use the route name (`viewGuide`) instead of the URL?

That way, when we add more routes we do not have to update this file too?

src/helpers/analytics.coffee
```diff
@@ -0,0 +1,64 @@
...
+
+GA = undefined
+
+Analytics =
+
+  setTracker: (tracker) -> GA = tracker
+
+  onNavigation: (change, router) ->
+    route  = DestinationHelper.routeFromPath(change.path, router.match)
+    return unless GA and route
```
philschatz: When would `route` be `null`? Also, what happens when the user clicks an internal link? Does/should that get sent to GA?

nathanstitt: It can be null if the page that's loaded initially doesn't match a route, when the router will display the `Invalid` page.  We probably should send a 'page-not-found' hit here.  I'll do so shortly.

Any links that hit the router are caught here and either cleaned up by the `Translators`  or sent as-is if no Translator is present.   

Links that navigate outside the app, like to the accounts page do not appear here.  For those we'll have to add a `Analytics.sendEvent` call if we want to track them.

src/helpers/analytics.coffee
nathanstitt: We can just send the href.  That's the default if a Translator method doesn't exist.  We really don't need any of these, but I thought you wanted analytics to track clean urls that would remain stable even if the app changes paths around.  if that's not the case, then we can just remove all these (or maybe just leave a few?)

philschatz: Sorry, my comment was a little muddled.. The `<Route>` component has 2 fields on it that are already "unique-ish", the  path and name. I was wondering if we could use one of those instead instead of creating another list here.

nathanstitt: Ah, yeah I see your point.  Yes we could use that but we probably should clean them up.  For instance the `dashboard` name isn't actually the dashboard, it's the course chooser.  

I'm thinking that the end-user of the analytics might not find it very user-friendly to see "calendarViewPlanStats".  They'd have to figure out that that corresponds to the quick metrics modal.

Google Analytics has a "drill down" view that allows viewing all hits for url segments.  with the pseudo url scheme I've setup they could view all hits for 'teacher' , page type, and then course.  We shouldn't be able to easily support that if we just logged the name.

philschatz:
Aside: devops is looking to try out piwik instead of GA. I'm guessing piwik has roughly similar features?

cnuber:
Piwik will need to be tested for several months before we can determine if we want to adopt it or not.  The higher-ups will need to decide if it's a suitable alternative to GA or not once it's ready to test.

PR #772, opened 2015-09-28
updated period tab sorting

darrenmason:
regarding task: https://www.pivotaltracker.com/story/show/104121340

sort will now handle 'period 1 period 2' as well as '1st 2nd' name formats


philschatz:
I don't see in the test that it correctly sorts "Period 10" and "Period 2". Can you add it?

PR #773, opened 2015-09-28
use the production version of React

philschatz:
No UI Changes.

When building the files for production, build without extra runtime checks and warnings

### DIFF
--- a/Gulpfile.coffee
+++ b/Gulpfile.coffee
@@ -32,10 +32,11 @@ gulp.task '_cleanDist', (done) ->
   del(['./dist/*'], done)
 
 gulp.task '_build', ['_cleanDist'], (done) ->
-  env(vars:{ NODE_ENV: 'production' })
   webpackConfig = require './webpack.config'
   config = _.extend({}, webpackConfig, {
     plugins: [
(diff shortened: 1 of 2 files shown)


philschatz:
Thoughts @nathanstitt or @pandafulmanda?

nathanstitt:
I was confused by the need to set `PROCESS_ENV` for webpack since it was already being set and checked by the `isProduction` variable.  It seems like the uglify plugin needs it to be set via the `process.env` plugin in order for the setting to take?   

I support the concept, we should use the production build without warnings on production. The stuff I've read says it should give us around a 30% speedup.

philschatz:
Yup, in order to know what to optimize out uglify needs to know that `NODE_ENV` is always `production` I did the same thing (thinking "oh, I just set the var when running webpack")

philschatz:
Out of curiosity, @nathanstitt why is part of the webpack config defined in `webpack.config.coffee` and some of it in `Gulpfile.coffee`?

nathanstitt:
That used to be how the webpack config was modified for building in production.  Now that we're using the `NODE_ENV is production` inside the webpack config, there's no longer any reason to have it there, it just hasn't been refactored out yet :)

PR #774, opened 2015-09-28
Fixing clearing due dates

jivewise:
When editing a TaskPlan with period-specific dates and switching to common dates the period-specific dates were still being saved.

Now they are cleared in the Store but retained in the component state in case the user clicks back to the period-specific radio option before saving.

Pivotal ticket here: https://www.pivotaltracker.com/n/projects/1156756/stories/104120296

- [x] Adding a fix where due dates are not being cleared correctly in the store
- [x] Add in some unit tests


jivewise:
Hm, the sets the default due date when based on query string is still failing only on travis for some reason.  I'm starting to suspect it's a timezone issue.  Skipping it for now, so this can get merged.

PR #775, opened 2015-09-28
HW Review Page Loading Indicator

jivewise:
screen shot 2015-09-28 at 11 46 31 pm

Updating how the loading indicator looks to fix the double loading indicators described here: https://www.pivotaltracker.com/n/projects/1156756/stories/104015042


philschatz:
How does this get rid of the double loading texts? 

If it is loading exercises could it say "Loading Exercises..." or "Loading Table of Contents..."?

PR #778, opened 2015-09-29
Selenium fixes

philschatz:
No UI changes.

This updates some of the selectors used in the Selenium tests and is smarter about extending the timeouts. 

tl;dr call `waitAnd` or `waitClick` frequently.

Each time `waitAnd` or `waitClick` is called, the timeout is extended by 60sec and then, the timeout is decreased based on how long the wait actually took. That way, if a long test errors at the end the timeout has not increased so much that selenium waits an hour before actually failing.

This addition ensures that the tests time out after 60sec of inactivity (by default). If you have a step that takes longer (like publishing) you can pass an additional argument to `waitAnd` or `waitClick` to override the 60sec.

/cc @gregfitch since he was looking at this last week :smile:


gregfitch:
looks fine -- doesn't really affect what I'm doing with SL because it has its own default timeouts

PR #780, opened 2015-09-30
Fix for when saving default periods

jivewise:
A bug was introduced here: https://github.com/openstax/tutor-js/pull/644 when I was trying to modify how a task plan loads it's periods.  In order to use the isChanged function of flux stores, I set the @_local variable directly instead of calling _changed.  This caused the default tasking plans to never get sent over to the backend, since TaskPlanStore.getChanged would never send over the default tasking plans.  This PR fixes that, and adds in a hasChanged function into the plan store.

- [x] Updating setPeriods to use _change instead of setting _local
- [x] Add a hasChanged method for task plan
- [x] Add a hasChanged method for select topics dialog

I'm adding some unit tests to a separate branch, but I think maybe this should get merged first.


src/flux/task-plan.coffee
```diff
@@ -459,6 +463,8 @@ TaskPlanConfig =
 
     isStatsFailed: (id) -> !! @_stats[id]
 
+    hasChanged: (id) -> not _.isEqual(@exports.getChanged.call(@, id), @_local[id].defaultPlan)
```
philschatz: Hmm, should we switch `isChanged`? I'm not sure what the difference is between `.hasChanged(id)` and `.isChanged(id)` is.

jivewise:
I'm not sure if isChanged is useful for task plans, due to the way we handle tasking plans and periods right now.  I'm not sure what other stores make use of isChanged.

src/flux/task-plan.coffee
philschatz: looks like `confirm-leave-mixin.cjsx`, `builder.cjsx`, and `plan-mixin.cjsx` all use `.isChanged`. If `.hasChanged` contains different logic than `.isChanged` then will the confirm dialog show up at different times?

jivewise: You're right, it should be fairly easy to change.  I already updated the ones in plan-mixin.cjsx to use hasChanged, and the one in builder.cjsx is old code, so can be removed.  

I think that just leaves confirm-leave-mixin.cjsx, which is only used by the task plans, so I think I can change it there as well.  It might be a good idea to move that file into the task-plan folder, since that's the only place where it is included.

jivewise:
Closed in favor of #781

PR #782, opened 2015-10-01
Complete exercise QA view

nathanstitt:
Finish up the qa exercise using the teacher exercise preview component.
Adds:
 * tags
 * type of exercise display
 * toggle to switch between normal book content and exercises

screen shot 2015-10-01 at 5 40 05 pm

### DIFF
--- a/resources/styles/variables.less
+++ b/resources/styles/variables.less
@@ -32,8 +32,8 @@
 
 @reading-color: @tutor-quaternary;
 @homework-color: @tutor-tertiary-light;
-@practice-color:  darken(@homework-color, 10%);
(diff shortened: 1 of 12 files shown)


resources/styles/variables.less
```diff
@@ -32,6 +32,8 @@
 
 @reading-color: @tutor-quaternary;
 @homework-color: @tutor-tertiary-light;
+@practice-color:  darken(@homework-color, 10%);
```
philschatz: could these be prefixed with "qa-"? then, if they get used elsewhere we can drop the prefix

nathanstitt: sure :+1:

PR #783, opened 2015-10-01
Display feedback for exercise question choices

nathanstitt:
Displays the question feedback to the teacher when they select exercises.  Will also be displayed on QA view since it shares the mixin as part of #782 

I'm undecided if the indent should match the question or be increased (as shown in screenshot)..  I felt like it might be confused as part of the question if it matched, even though it is displayed as grey.

screen shot 2015-10-01 at 6 22 19 pm

### DIFF
--- a/resources/styles/components/task-plan/homework/exercises.less
+++ b/resources/styles/components/task-plan/homework/exercises.less
@@ -168,7 +168,12 @@ label.exercises-section-label {
         }
         .answer {
           display: table;
+        }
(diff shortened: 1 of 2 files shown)


philschatz:
Might need a quick glance-over from @jason-holmes

Fredasaurus:
I think we can clean up the typography on a bit. We also worked on some interactions for toggling the view. @jason-holmes can elaborate.

PR #784, opened 2015-10-02
Speedup Mathjax processing on a pages with many HTML components

nathanstitt:
On the Teacher and QA views of exercises, mathjax was being invoked
several hundred times (twice per each answer on each question).  This modifies it so it will
only perform once pass on the entire document.  There was a concern about seeing flicker as it renders but I don't observe any.  

Before Mathjax was taking 4,627ms to switch from section 2.0 to 2.1 of physics in qa view.  Both timings were taken after a full page reload.
screen shot 2015-10-02 at 5 12 16 pm

With a single pass the time drops to 85ms

screen shot 2015-10-02 at 5 14 15 pm

### DIFF
--- a/src/helpers/mathjax.coffee
+++ b/src/helpers/mathjax.coffee
@@ -3,6 +3,8 @@ _ = require 'underscore'
 MATH_MARKER_BLOCK  = '\u200c\u200c\u200c' # zero-width non-joiner
 MATH_MARKER_INLINE = '\u200b\u200b\u200b' # zero-width space
 
+MATH_SELECTOR = '[data-math]:not(.math-rendered)'
+
 cleanMathArtifacts = ->
   # Once MathJax finishes processing, cleanup the MathJax message nodes to prevent
   # React "Invariant Violation" exceptions.
@@ -19,13 +21,12 @@ cleanMathArtifacts = ->
   ])
 
 
-typesetMath = (root) ->
-  nodes = root.querySelectorAll('[data-math]:not(.math-rendered)') or []
-  hasMath = root.querySelector('math')
 
-  # Return immediatly if no [data-math] or <math> elements are present
-  # TODO: If the MathJax Queue is not available then MathJax has not loaded yet. Add a load callback to enqueue.
-  return unless window.MathJax?.Hub?.Queue? and (_.any(nodes) or hasMath)
+# Typesets the entire document
+TypesetDocument = ->
+
+  nodes = document.querySelectorAll(MATH_SELECTOR)
+  return if _.isEmpty(nodes)
 
   for node in nodes
     formula = node.getAttribute('data-math')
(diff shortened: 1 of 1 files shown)


Dantemss:
I might not fully understand the code that handles these math tags, but couldn't we apply the same logic as the other nodes? Seems weird that we have to enqueue the root node when we have one of these tags?

src/helpers/mathjax.coffee
```diff
@@ -19,13 +21,12 @@ cleanMathArtifacts = ->
...
 
-typesetMath = (root) ->
-  nodes = root.querySelectorAll('[data-math]:not(.math-rendered)') or []
-  hasMath = root.querySelector('math')
 
-  # Return immediatly if no [data-math] or <math> elements are present
-  # TODO: If the MathJax Queue is not available then MathJax has not loaded yet. Add a load callback to enqueue.
-  return unless window.MathJax?.Hub?.Queue? and (_.any(nodes) or hasMath)
+# Typesets the entire document
+TypesetDocument = ->
```
philschatz: since this is a function (& has side-effects) could you make this `typesetDocument` ?

src/helpers/mathjax.coffee
```diff
@@ -40,10 +41,26 @@ typesetMath = (root) ->
...
+  cleanMathArtifacts()
+
+# Install a debounce around typesetting function so that it will only run once
+# every 10ms even if called multiple calls times in that period
+TypesetDocument = _.debounce( TypesetDocument, 10)
+
+
+typesetMath = (root) ->
+  nodes = root.querySelectorAll(MATH_SELECTOR) or []
+  hasMath = root.querySelector('math')
```
philschatz: Should this be `root.querySelector("math, #{MATH_SELECTOR}")` to handle math in the books (`math`) as well as math in the exercises (`MATH_SELECTOR`)?

nathanstitt:
8.1 and 8.2 of Physics have `<math>` elements for testing

nathanstitt:
@Dantemss - right you are.  I was unable to find any of the `<math>` nodes previously to test with so I thought it best to leave that bit of code as-is.  Finally found a few and was able to get it working

philschatz:
Hmm, it seems math in the ref book is still not rendering... 

image

nathanstitt:
Sigh.  So... apparently mathjax will not render `math` elements directly, it will only search for them in children.  This also explains why the earlier code was passing the root node to mathjax if any `<math>` elements were found.   I must have viewed some math elements that rendered fairly well in the browser before so I didn't notice that they weren't enhanced.  

To work around this, I'm now passing `<math>`'s  parentElement to mathjax.  Seems to work well and performance gains are still intact.

screen shot 2015-10-05 at 2 29 04 pm

PR #791, opened 2015-10-05
update learning guide component

pandafulmanda:
No UI changes

- [x] move learning-guide as index into learning-guide component folder
- [x] rename learning guide to performance forecast
  * there are still a lot of references to guide ...
  * it sounds like we are saying we don't need to change anything else?


philschatz:
So, if we wanted to change the name in the future could we just do a replace on `/performance.?forecast/i` ? and the only things that remain with "guide" are the URLs?

pandafulmanda:
there are still things like class names and [guide.cjsx](https://github.com/openstax/tutor-js/blob/1a779d5a85ed21d90352e0b78b68076b281c6954/src/components/performance-forecast/guide.cjsx#L40) in `performance-forecast` component

PR #796, opened 2015-10-07
Make "tasks" stand alone

pandafulmanda:
- [x] decrease mixins
- [x] eliminate store dependency
- [x] make actions and data update flexible
- [x] abstract out router dep. as well
- [x] make footer flexible
- [ ] keep performance in mind
- [ ] tests would be good
- [ ] prototype out CC display in tutor
  - [x] try in CNX

works locally, need to fix on travis, uses openstax/react-components@3fbca87
Also tested in cnx webview as in Connexions/webview@8971f75


src/components/task-step/exercise.cjsx
```diff
@@ -0,0 +1,46 @@
...
+
+    stepProps.getCurrentPanel = (id) ->
+      unless TaskStepStore.isSaving(id)
+        currentPanel = StepPanel.getPanel(id)
+
+    <Exercise
+      {...@props}
+      {...stepProps}
+      step={step}
+      Footer={StepFooter}
```
philschatz: why is this prop TitleCase?

src/components/task-step/exercise.cjsx
```diff
@@ -0,0 +1,46 @@
...
+  propTypes:
+    id: React.PropTypes.string.isRequired
+    taskId: React.PropTypes.string.isRequired
+
+  render: ->
+    {id, taskId} = @props
+    step = TaskStepStore.get(id)
+    task = TaskStore.get(taskId)
+
+    stepProps = {}
```
philschatz: Why pull the props out instead of explicitly putting them on the `<Exercise>`? It looks like the props are always set

src/components/task-step/index.cjsx
```diff
@@ -33,7 +33,6 @@ TaskStepLoaded = React.createClass
   propTypes:
     id: React.PropTypes.string.isRequired
     onNextStep: React.PropTypes.func.isRequired
-    goToStep: React.PropTypes.func.isRequired
```
philschatz: Why is this no longer needed?

src/components/task-step/step-footer.cjsx
```diff
@@ -0,0 +1,29 @@
...
+  getDefaultProps: ->
+    controlButtons: null
+
+  renderFooterButtons: ->
+    @props.controlButtons
+
+  render: ->
+    {pinned, courseId, id, taskId, review} = @props
+
+    <div>
```
philschatz: Can you put a `className='-foobar'` on this?

src/components/task-step/exercise.cjsx
pandafulmanda: It's a React class that gets passed. I guess normally we would be passing in a component, like this:

```cs
footer = {<StepFooter/>}
```

Which i tried and then using `React.addons.cloneWithProps` to pass in props in `Exercise` but that was giving me issues so I did this for now.  I can try to address in the other repo.

src/components/task-step/exercise.cjsx
pandafulmanda: I feel like this is more readable, but I can put them directly on `<Exercise>`

src/components/task-step/index.cjsx
pandafulmanda: I did some checking, seems like it was extra to begin with -- we don't end up using it inside the steps themselves, but we do need it on the Task

src/components/task-step/exercise.cjsx
philschatz: oh, that makes sense (maybe `FooterType` to clarify?). Why does the component need to be instantiated inside the exercise component instead of passed in?

src/components/task-step/index.cjsx
philschatz: ah, ok. **RESOLVED**

src/components/task-step/exercise.cjsx
pandafulmanda: Okay.  I uses props that the parent passes down to it

package.json
```diff
@@ -34,6 +34,7 @@
     "mime-types": "^2.1.3",
     "moment": "^2.9.0",
     "moment-timezone": "^0.4.0",
+    "openstax-react-components": "openstax/react-components#v0.0.0",
```
philschatz: Were you going to change this to a commit later?

src/components/task-step/exercise.cjsx
```diff
@@ -0,0 +1,47 @@
...
+      TaskStore.getReadingForTaskId(taskId, id)
+
+    getCurrentPanel = (id) ->
+      unless TaskStepStore.isSaving(id)
+        currentPanel = StepPanel.getPanel(id)
+
+    <Exercise
+      {...@props}
+      step={step}
+      Footer={StepFooter}
```
philschatz: Why does the exercise component need to instantiate this? Does it have more information than this component?

philschatz: Oh, I see your comment from earlier: "I can try to address in the other repo."

PR #800, opened 2015-10-08
Set asset path to "/assets/" for production for QA view

nathanstitt:
Hopefully this will fix the QA js loading


nathanstitt:
The spec failure is the calendar thing, should be unrelated to this PR.

PR #802, opened 2015-10-09
to be falsy should be not.be.ok

pandafulmanda:
there is no such thing as `be.falsy` in chai -- it's [`to.not.be.ok`](https://github.com/chaijs/chai/blob/6d0ae350974aa20d8e903c6b84bbb9ce4d05f435/test/expect.js#L40-L41)


Dantemss:
This is so weird... `to.not.be.ok`?

nathanstitt:
could also just be `.to.be.true`, but that would require it to strict equal to the boolean true.  Or `to.not.exist` would maybe be better.  That makes sure it's not `null` or `undefined`.  Chai's got tons of different ways to chain assertions: 'em: http://chaijs.com/api/bdd/

PR #809, opened 2015-10-12
changed worked text for teacher-student

darrenmason:
regarding task: https://www.pivotaltracker.com/n/projects/1156756/stories/105344550

image

Teacher viewing a single student says "N problems worked"

### DIFF
--- a/src/components/performance-forecast/statistics.cjsx
+++ b/src/components/performance-forecast/statistics.cjsx
@@ -17,20 +17,27 @@ Statistics = React.createClass
 
   propTypes:
     courseId: React.PropTypes.string.isRequired
+    roleId:   React.PropTypes.string.isRequired
     section:  ChapterSectionType.isRequired
     displaying: React.PropTypes.string.isRequired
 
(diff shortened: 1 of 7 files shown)


src/components/performance-forecast/statistics.cjsx
```diff
@@ -17,20 +17,27 @@ Statistics = React.createClass
...
+        "#{pluralize(' problems', total, true)} worked"
 
-    teacherWorkedText =
-      "#{pluralize(' students', count, true)}
-       #{pluralize(' has', count)} worked #{pluralize(' problems', total, true)}"
-
-    studentWorkedText =
-      "#{pluralize(' problems', total, true)} worked in this #{@props.displaying}"
+  render: ->
+    if @props.roleId?
```
philschatz: The `roleId` is marked as required above. Either it should not be required (above) or change the places in the code that use `<Statistics>` to pass in a `teacher-student` role (slight preference for this option since it's more documented and less *"magic"*).

PR #812, opened 2015-10-14
take out adding and removing event listeners on updates

pandafulmanda:
This was added a long time ago to try to fix invariant errors, which we later discovered needed to be fixed other ways.  It looks like it's not good for performance, and taking it out doesn't seem to break stuff, so I say we take it out.


philschatz:
thanks for deleting those files :smile:

PR #813, opened 2015-10-21
Change Course Names in Roster

helenemccarron:


src/components/course-settings/roster.cjsx
```diff
@@ -14,10 +14,27 @@ module.exports = React.createClass
...
+         <input type="submit" className="submit" value="rename" onSubmit={@updateCourseName}/>
+      </form>
+    </BS.Popover>
+
   render: ->
     course = CourseStore.get(@props.courseId)
     tabs = _.map course.periods, (period, index) =>
       <BS.TabPane key={period.id}, eventKey={index} tab={period.name}>
+        <BS.OverlayTrigger rootClose={true} trigger='click' placement='right'
+          overlay={@changeCourseName()}>
```
philschatz: I think this could just be a variable instead of calling a method. The `change` in the name of the method is a bit confusing: does it mean calling `changeCourseName` will change the course name? By convention in tutor-js, when a method generates DOM Elements the method is prefixed with `render...` but I think in the case of a popover you could just create a variable inside `render: ->` named `courseNamePopover`

src/components/course-settings/roster.cjsx
```diff
@@ -14,10 +14,27 @@ module.exports = React.createClass
   propTypes:
     courseId: React.PropTypes.string.isRequired
 
+  updateCourseName:(courseId)->
+    RosterActions.save(@props.courseId)
+    console.log('waka kinda working');
+
+  changeCourseName:->
+    course = CourseStore.get(@props.courseId)
+    <BS.Popover tilte={'Change Course Name'} {...@props}>
```
philschatz: is `...@props` needed? It seems odd that the props sent to the `<Roster>` would be passed into the popover

src/components/course-settings/roster.cjsx
```diff
@@ -14,10 +14,27 @@ module.exports = React.createClass
...
+      <form onSubmit={@updateCourseName} >
```
philschatz: Does this work? If so, it would be a nice way to handle the Enter key by default.

src/components/course-settings/roster.cjsx
```diff
@@ -14,10 +14,27 @@ module.exports = React.createClass
   propTypes:
     courseId: React.PropTypes.string.isRequired
 
+  updateCourseName:(courseId)->
+    RosterActions.save(@props.courseId)
+    console.log('waka kinda working');
```
philschatz: `updateCourseName` probably needs to pull the value out of the `<input>` box. to do that you need to change it to be `<input ref='something'>` and then use `{something} = @refs` in here (and then `something.value()` to get the actual text.

src/components/course-settings/roster.cjsx
helenemccarron: yes, it works. Well, at least with  
```
 updateCourseName:(courseId)->
    console.log('working'); 
```
it does :)

philschatz:
Where'd it go?

PR #815, opened 2015-10-21
Stop using course title for styling

nathanstitt:
This uses the new `catalog_offering_identifier` from https://github.com/openstax/tutor-server/pull/721 to style course.

This also adds some margin between the courses, To me it looked strange when there were two courses with the same colors right beside one another.

TODO:
* [x] display course name in chooser for courses that aren't titled "Physics" or "Biology"

 
screen shot 2015-10-22 at 3 35 28 pm

Displays ellipsis when text is to long to fit

screen shot 2015-10-22 at 3 35 37 pm

### DIFF
--- a/resources/styles/components/course-listings.less
+++ b/resources/styles/components/course-listings.less
@@ -49,18 +49,14 @@
     }
 
     &[data-category=physics] {
-      .tutor-course-item {
-        .tutor-physics-cover-background();
-      }
+      &[data-title=Physics]{ .tutor-course-item { .tutor-physics-cover-background(); } }
       .tutor-course-alt-link {
         color: @tutor-book-hs-physics-primary;
         background-color: @tutor-book-hs-physics-secondary;
       }
     }
     &[data-category=biology] {
-      .tutor-course-item {
-        .tutor-biology-cover-background();
-      }
+      &[data-title=Biology]{ .tutor-course-item { .tutor-biology-cover-background(); }
       .tutor-course-alt-link {
         color: @tutor-book-ap-bio-primary;
         background-color: @tutor-book-ap-bio-secondary;
--- a/resources/styles/components/course-listings.less
+++ b/resources/styles/components/course-listings.less
@@ -5,9 +5,15 @@
 .course-listing {
   margin-top: @course-listing-buffer;
 
+  .row {
+    margin-top: @course-listing-buffer;
+    &:first-child{
+      margin-top: 0;
+    }
+  }
+
   .tutor-booksplash-course-item {
     position: relative;
-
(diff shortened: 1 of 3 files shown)


resources/styles/components/course-listings.less
```diff
@@ -49,18 +49,14 @@
     }
 
     &[data-category=physics] {
-      .tutor-course-item {
-        .tutor-physics-cover-background();
-      }
+      &[data-title=Physics]{ .tutor-course-item { .tutor-physics-cover-background(); } }
```
philschatz: Is the `catalog_offering_identifier` TitleCase?

nathanstitt: No that's targeting only courses that are actually named "Physics".  They're the only ones that we should use the SVG background for.  The others can have the colors if the `catalog_offering_identifier` is known, but we don't want to use an SVG that has an incorrect title.

philschatz: Is there an enumeration of all the values for `catalog_offering_identifier`? Seems odd that we'd continue to special-case based on the title of a course

nathanstitt: There isn't one currently. The `catalog_offering_identifier` can be set to arbitrary things on the BE to reflect what the course type is, and multiple courses will share the same identifier.  My thoughts on it are that the identifier should be used to style the courses, but it shouldn't be depended on for displaying the course name.

This presents a problem for the chooser though.  It uses SVG's for displaying the course title, and we only have ones that read "Biology" or "Physics".

In this case I think it makes sense to target the title because that's what we're replacing with an image.

nathanstitt:
_Pasting comment here so it isn't hidden in the diff:_

There isn't a listing of all possible values of `catalog_offering_identifier`.  It cand can be set to arbitrary things on the BE to reflect what the course type is, and multiple courses will share the same identifier. My thoughts on it are that the identifier should be used to style the courses, but it shouldn't be depended on for displaying the course name.

This presents a problem for the chooser though. It uses SVG's for displaying the course title, and we only have ones that read "Biology" or "Physics".

In this case I think it makes sense to target the title because that's what we're replacing with an image.

jason-holmes:
I’m not sure if this is relevant to this particular discussion, but here’s a mockup of the picker page we’re planning when Concept Coach launched.  http://3lwsnp.axshare.com/#p=course_picker_page__if_more_than_one_course_


-------
Jason Holmes, Ph.D.
User Experience Manager
OpenStax | OpenStax.org
(713) 893-1326

philschatz:
Yep, there's a separate ticket for updating the look-and-feel of the course selection page :smile:

PR #824, opened 2015-10-29
No assignment yet on Scores page

helenemccarron:
Display no assignment yet when empty


philschatz:
Can you exclude the `npm-debug.log` file? Maybe add it to `.gitignore`?

philschatz:
Looks like this broke Travis. can you fix the tests?

PR #826, opened 2015-11-03
Move control for displaying feedback below the card

nathanstitt:
Moves the control below the card so it doesn't select the exercise when its clicked.

I experimented with `stopPropagation` but it would only work if the checkbox itself was clicked, not the label. Clicks on the label would still trigger exercise selection callback event though they were above it on the dom tree.  Puzzlingly, clicks on the label would have the panel title as it's target, which makes no sense since that's not an ancestor of the label.  After awhile I gave up and figured we should just implement the mockup ;)

screen shot 2015-11-03 at 1 28 39 pm

### DIFF
--- a/src/components/task-plan/homework/exercises.cjsx
+++ b/src/components/task-plan/homework/exercises.cjsx
@@ -5,29 +5,31 @@ BS = require 'react-bootstrap'
 ArbitraryHtmlAndMath = require '../../html'
 ExerciseCard = require '../../exercise-card'
 ChapterSection = require '../chapter-section'
+Icon = require '../../icon'
 {ExerciseStore, ExerciseActions} = require '../../../flux/exercise'
 {TaskPlanStore, TaskPlanActions} = require '../../../flux/task-plan'
 {TocStore} = require '../../../flux/toc'
 
 ExerciseCardMixin =
 
-  toggleFeedbackDisplay: ->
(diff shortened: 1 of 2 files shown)


src/components/task-plan/homework/exercises.cjsx
```diff
@@ -5,29 +5,30 @@ BS = require 'react-bootstrap'
...
-      displayFeedback={@state?.displayFeedback}
-      panelStyle={@getPanelStyle()}>
-
-      <BS.Input type="checkbox" label="Display Feedback" standalone
-        wrapperClassName={"feedback-toggle"}
-        checked={@state?.displayFeedback}
-        onChange={@toggleFeedbackDisplay} />
-
-    </ExerciseCard>
+    <div>
```
philschatz: For selenium testing could you add a class to this (maybe `-exercise-wrapper`)?

Fredasaurus:
"After awhile I gave up and figured we should just implement the mockup"

Actual lol.

nathanstitt:
Hehe, figured you might find that admission *interesting*.   Was more of a "I know this worked when I originally implemented it, why won't it now..."  And also to short-circuit Phil's suggestion of "did you try to use stopPropagation" :)

philschatz:
it worked! I saw the PR title and was thinking "Oh, maybe that part of the discussion wasn't clear" :smile:

PR #828, opened 2015-11-06
Course Settings - teacher remove - fixed delete arg, update to courselist on delete current teacher

darrenmason:
image

### DIFF
--- a/src/components/course-settings/remove-teacher.cjsx
+++ b/src/components/course-settings/remove-teacher.cjsx
@@ -16,6 +16,8 @@ module.exports = React.createClass
     teacher: React.PropTypes.object.isRequired
     courseRoles: React.PropTypes.array.isRequired
     courseId: React.PropTypes.string.isRequired
+  contextTypes:
+    router: React.PropTypes.func
 
   isRemovalCurrentTeacher: ->
     role = _.chain(@props.courseRoles)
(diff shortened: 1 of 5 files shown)


src/components/course-settings/remove-teacher.cjsx
```diff
@@ -25,10 +24,11 @@ module.exports = React.createClass
       .value()
 
   goToDashboard: ->
-    @context.router.transitionTo('dashboard', {courseId: @props.courseId})
+    TeacherRosterStore.once 'deleted', ->
+      window.location.href = '/dashboard/'
```
philschatz: Does `@context.router.transitionTo('dashboard', {courseId: @props.courseId})` no longer work in `TeacherRosterStore.once 'deleted', =>` ?

darrenmason: When using transitionTo I encountered a SecurityTransgression error b/c the logged in course role had been deleted.  The combination of using once 'deleted' and window.location bypasses any missing course role errors and the courselist is always updated when reaching /dashboard/

philschatz: hmm, does the following not work (combining `transitionTo` without the course id but wrapped in a `.once 'deleted'`)?

```coffee
TeacherRosterStore.once 'deleted', =>
  @context.router.transitionTo('dashboard')
```

philschatz: Oh also, should the delete button in the modal be an `AsyncButton` so the user sees if the delete failed?

PR #829, opened 2015-11-06
Change Course Names in Roster

helenemccarron:
Update the course name in the roster, validate new name, update  course name in the nav


philschatz:
replaces #813

src/components/course-settings/rename-course.cjsx
```diff
@@ -0,0 +1,94 @@
...
+        label="Course Name"
+        name="course-name"
+        default={@props.course.name}
+        onChange={(val) => @setState(course_name: val)}
+        validate={@validate}
+        autofocus />
+      </div>
+
+      <div className='modal-footer'>
+        <BS.Button
```
philschatz: Could this be an `<AsyncButton>` so the teacher sees a spinner while we save?

helenemccarron: Yup!

src/components/course-settings/rename-course.cjsx
```diff
@@ -0,0 +1,94 @@
...
+    <TutorInput
+      ref="input"
+      label={@props.label}
+      default={@props.default}
+      required={true}
+      onChange={@onChange}
+      validate={@props.validate}
+      />
+
+module.exports = React.createClass
```
philschatz: JSX automatically gives the react component the displayName of the left-hand-side of the assignment. so you can just do `RenameCourse = React.createClass` and then later `module.exports = RenameCourse`

src/components/course-settings/rename-course.cjsx
```diff
@@ -0,0 +1,94 @@
...
+  getInitialState: ->
+    course_name: @props.course.name
+
+  validate: (name) ->
+    error = CourseStore.validateCourseName(name, @props.course.name)
+    @setState({invalid: error?})
+    error
+
+  performUpdate: ->
+    if not @state.invalid
```
philschatz: can you use `unless` instead of `if not`?

src/components/course-settings/rename-course.cjsx
```diff
@@ -0,0 +1,94 @@
...
+    if @state?.invalid
+      formClasses.push('is-invalid-form')
+      disabled = true
+
+    <BS.Modal
+      {...@props}
+      title={'Rename Course'}
+      className='teacher-edit-course-modal'>
+
+      <div className={formClasses.join(' ')} >
```
philschatz: You should be able to use the `classnames` package from now on instead of concatenating strings

src/components/course-settings/rename-course.cjsx
```diff
@@ -0,0 +1,94 @@
...
+        <RenameCourseField
+        label="Course Name"
```
philschatz: Can you use single quotes instead of doubles (for consistency)?

PR #830, opened 2015-11-12
Fix for set periods when clicking on the all periods radio

jivewise:
Bug is described here: https://www.pivotaltracker.com/n/projects/1156756/stories/107536766

Seems like this was introduced a while ago here: https://github.com/openstax/tutor-js/commit/3b18b77fca7f6d2be113b630634d69d155416aeb#diff-5a158ad5bde62626d82249fc1b709b77R157


philschatz:
Oh wow, nice job hunting it down :smile:

PR #834, opened 2015-11-17
don't show namelink / due date when concept coach

darrenmason:
regarding pivotal tasks: 
https://www.pivotaltracker.com/story/show/107577060
https://www.pivotaltracker.com/story/show/107577064

When Concept Coach, don't show links on student name or due date row.

image

image

### DIFF
--- a/src/components/scores/index.cjsx
+++ b/src/components/scores/index.cjsx
@@ -101,7 +101,7 @@ Scores = React.createClass
       </span>
 
     sortingHeader = <SortingHeader type={heading.type} sortKey={i}
-      sortState={@state.sort} onSort={@changeSortingOrder}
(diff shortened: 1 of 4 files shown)


src/components/scores/index.cjsx
```diff
@@ -201,8 +201,13 @@ Scores = React.createClass
         d.last_name.toLowerCase()
     )
     { headings: scores.data_headings, rows: if sort.asc then sortData else sortData.reverse() }
+
   onColumnResizeEndCallback: (colWidth, columnKey) ->
     @setState({colResizeWidth: colWidth, colResizeKey: columnKey})
+
+  headerType: ->
+    if @props.isConceptCoach then 47 else 92
```
philschatz: Could you add a comment for why the height is different (I'm guessing it is because of the missing heading)?

PR #835, opened 2015-11-18
Replace spymode component with shared version

nathanstitt:
Note that this brings in quite a few other components as well since the components haven't been updated in awhile.  It all seems to work fine though, other than the fontawesome issue that I've fixed.


resources/styles/tutor.less
```diff
@@ -39,7 +39,6 @@
 @import './components/course-listings';
 @import './components/tutor-dialog';
 @import './components/task/index';
-@import './components/spy-mode';
 @import './components/task-step/index';
```
nathanstitt: I've removed the built in styles here, but the line below (you'll have to expand diff, it's line 46) that requires them from openstax-react-components seems incorrect.  There's not an assets/main.css file unless they're built using webpack.  Shouldn't we require the styles like how concept coach does by: `@import '~openstax-react-components/resources/styles/main';`

PR #847, opened 2015-11-25
concept coach logout link

darrenmason:
regarding task: https://www.pivotaltracker.com/n/projects/1156756/stories/108517582


philschatz:
Does tutor-server part work too (redirect to `cc.openstax.org`) when that param is set?

darrenmason:
tested with https://github.com/openstax/tutor-server/pull/823, not redirecting

nathanstitt:
:+1: Works with local accounts running.  Both `cc=true` or `cc=1` parameters are accepted.

philschatz:
yay!

PR #848, opened 2015-11-25
:art: find unused vars in cjsx files

philschatz:
`coffeelint` does not lint `cjsx` files and even for `coffee` files it does not find unused vars.

This adds a `npm run lint-cjsx` which compiles the `cjsx` files individually and then runs `eslint` on each.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,5 @@
+{
+  "rules": {
+    "no-unused-vars": 1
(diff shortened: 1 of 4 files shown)


nathanstitt:
:+1: Would be nice to keep me from doing things like leaving stray `console.log` stmts in code :grinning:

PR #849, opened 2015-11-25
Link CC scores to task review screen

nathanstitt:
Depends on https://github.com/openstax/tutor-server/pull/837

Two things I've noticed:

- [x] The continue button doesn't work, it's always disabled
- [x] There's no "back" button to go back to the scores report, other than re-selecting it from the menu.

screen shot 2015-11-30 at 2 08 49 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "mime-types": "^2.1.3",
     "moment": "^2.9.0",
     "moment-timezone": "^0.4.0",
-    "openstax-react-components": "openstax/react-components#b1d67bc",
+    "openstax-react-components": "openstax/react-components#a8a29638c07cdad0d693ac41da4a9957b922878a",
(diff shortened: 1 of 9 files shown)


philschatz:
maybe @pandafulmanda could help?

pandafulmanda:
Yes, i will look into this

nathanstitt:
Also note https://github.com/openstax/tutor-server/pull/837#issuecomment-159751034 regarding the endpoint that the task review button at the top of each column would need.  As things stand, I don't think that'll be easily implemented.

pandafulmanda:
I've checked this again with other existing assignments for non-concept coach tasks, and things still seem to be working, so i think is ready, @nathanstitt

nathanstitt:
Looks great!  Travis failure is due to first-of-month bug in specs

PR #852, opened 2015-11-27
Filter exercises by pool type and add edit link

nathanstitt:
screen shot 2015-11-27 at 4 09 21 pm


src/components/multi-select.cjsx
```diff
@@ -0,0 +1,44 @@
+React = require 'react'
+BS = require 'react-bootstrap'
+classnames = require 'classnames'
+
+Icon = require './icon'
+
+MultiSelect = React.createClass
```
philschatz: Ooh, nice widget!

PR #853, opened 2015-11-30
Concept coach progress bar style changes

jivewise:
screen shot 2015-12-02 at 7 46 53 pm
screen shot 2015-12-02 at 7 46 28 pm

Stemming from this story: https://www.pivotaltracker.com/story/show/108793570


philschatz:
Would it help to reuse (or pull out) the quick-look component directly? Also, I _think_ we did something about minimum length of the progress bar.

/cc @Fredasaurus thought?

Fredasaurus:
We aren't labeling directly on the progress bar. Also, we are only labeling % right.

http://3lwsnp.axshare.com/#p=teacher_dashboard_-_visual_design

jivewise:
It seems that Kathi mentioned something different in this pivotal ticket comments here: https://www.pivotaltracker.com/story/show/108794084

Fredasaurus:
I think this is a different one.

Fredasaurus:
OK, let's label in the bar, no need for the "completed" and "incorrect" labels.

kathi-fletcher:
Just FYI -- the only time that the completed or correct/incorrect appear directly in the bar is when there is a single color. Then you should get 100% correct or 100% incorrect or 0% complete, 100% complete. Note that complete works a little differently from correct/incorrect.

jivewise:
This should also include a fix for https://www.pivotaltracker.com/story/show/109174428

philschatz:
So, the quick-look component was different enough not to reuse?

jivewise:
Yea, it didn't really have the same data format, I tried massaging it, but I ended up reverting it as the code just didn't look great...

PR #856, opened 2015-11-30
Always edit exercises in new tab

nathanstitt:


philschatz:
Iconically this doesn't use the react component for creating links that open in a new tab :smile:

PR #858, opened 2015-12-01
show tabs when 'no assignments'

darrenmason:
regarding task: https://www.pivotaltracker.com/story/show/109169388

image


philschatz:
Can you put the "No Assignments Yet" below the period list so it looks closer to 
image

philschatz:
I'm guessing there will be some styling _suggestions_ from @Fredasaurus but merging now so things do not just explode :fire:

PR #859, opened 2015-12-02
Remove tiny "TM" from logo

nathanstitt:
It's too small to be read.  I didn't notice it at all last night on the laptop screen.  This morning when I plugged into my monitor it looked like a dirt spec.

Before
screen shot 2015-12-02 at 9 55 40 am

after:
screen shot 2015-12-02 at 9 51 08 am


Fredasaurus:
It has to be there. Even it you can't read it.

Fredasaurus:
We could make it bit larger.

philschatz:
Then I propose we make it transparent :smirk:

nathanstitt:
Hm,  How about if we leave it on, but size the background image so that it's not visible?

Fredasaurus:
Oooooh, it was there the whole time!

nathanstitt:
Done :)  Original logo is restored but looks better now

PR #862, opened 2015-12-02
"Change period" uses async loading status

darrenmason:
https://www.pivotaltracker.com/story/show/107220234

change-period-loader


src/flux/roster.coffee
```diff
@@ -11,6 +13,8 @@ RosterConfig = {
     @emitChange()
 
   save: (courseId, studentId, params) ->
+    @_asyncStatus[courseId] is SAVING
```
philschatz: I'm not sure what this line of code does....

PR #863, opened 2015-12-03
Use appearance code from BE for styling courses

nathanstitt:
As part of https://github.com/openstax/tutor-server/pull/853 the BE will be sending a `appearance_code` attribute on courses.  The code will default to the offering but can be set on a per-course basis if needed.

Should be merged in connection with BE pr.

### DIFF
--- a/resources/styles/components/course-listings.less
+++ b/resources/styles/components/course-listings.less
@@ -51,7 +51,7 @@
       }
     }
 
-    &[data-category]{
+    &[data-appearance]{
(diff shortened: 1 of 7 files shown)


philschatz:
There will be 8 appearance codes (1 for each book) because the dashboard and scores pages will need to have a little banner SVG in the background.

But otherwise, @jpslav merge away :+1:

jpslav:
Thanks @philschatz!

PR #868, opened 2015-12-04
Show message when no periods

darrenmason:
image

image

- [x] no periods boilerplate message on scores
- [x] finalized message text
- [x] no periods message on performance forecast
- [x] roster updated to handle no periods

### DIFF
--- a/src/components/no-periods.cjsx
+++ b/src/components/no-periods.cjsx
@@ -1,11 +1,14 @@
 _ = require 'underscore'
 React = require 'react'
+BS = require 'react-bootstrap'
 
 NO_PERIODS_TEXT = 'Please add at least one period to the course.'
 
 NoPeriods = React.createClass
 
   render: ->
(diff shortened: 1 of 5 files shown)


src/components/no-periods.cjsx
```diff
@@ -0,0 +1,21 @@
...
+
+NO_PERIODS_TEXT = 'Please add at least one period to the course.'
+
+NoPeriods = React.createClass
+
+  propTypes:
+    noPanel:   React.PropTypes.bool
+
+  render: ->
+    text = <span className='-no-periods-text'>{NO_PERIODS_TEXT}</span>
```
philschatz: Should this also have a link to the "Roster"? I'd probably still merge this in as-is and then have a separate PR to fix up the boilerplate text.

/cc @jason-holmes or @Fredasaurus

Fredasaurus:
Yeah, we can probably make the copy a little more friendly as well. We can look at it next week.

PR #869, opened 2015-12-04
Colors and styles for additional courses

nathanstitt:
If someone can figure out how to make the colors be an array that can be looped through with, that'd be awesome.  But I don't think it's possible...

screen shot 2015-12-04 at 5 35 10 pm


philschatz:
cool! @Fredasaurus said he'd try and send the SVG's for all the different courses tomorrow (Sat)

PR #873, opened 2015-12-06
Add booksplash header to dashboard

nathanstitt:
screen shot 2015-12-06 at 12 33 14 am

### DIFF
--- a/resources/styles/components/cc-dashboard/index.less
+++ b/resources/styles/components/cc-dashboard/index.less
@@ -1,4 +1,4 @@
-cc-dashboard {
+.cc-dashboard {
   .column-legend {
     span {
       color: @tutor-neutral;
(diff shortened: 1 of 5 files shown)


philschatz:
This was probably in the previous PR but I think the "Unable to display progress" case should just not show the row at all

nathanstitt:
Yes I thought of that, but then decided that teachers would be confused if the book appeared to be missing chapters.   They might think that CC wasn't enabled on them or something

philschatz:
Kathi brought it up as a way to reduce the noise if:

1. some students worked ahead
2. students accidentally worked problems in a section that was not assigned by the teacher.

added https://github.com/openstax/tutor-js/pull/874 . Can you check it over (maybe in the morning?)

PR #874, opened 2015-12-06
hide chapters without enough data

philschatz:
This should remove the "Unable to show progress because not enough data"

### DIFF
--- a/src/components/cc-dashboard/chapter.cjsx
+++ b/src/components/cc-dashboard/chapter.cjsx
@@ -16,15 +16,8 @@ DashboardChapter = React.createClass
     )
 
   renderSections: ->
-    if _.any @props.chapter.valid_sections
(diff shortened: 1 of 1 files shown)


nathanstitt:
It seems to me that the first thing a teacher is going to after signing up is to load CNX as a student and answer some questions.  They'll then go back to their dashboard and expect to see the scores.  Or maybe they'll have a few trial students go through it.  

If we don't at least indicate that something is going on, they will lose faith in the accuracy of the reporting.

With that said, it's not a big deal either way.  We can merge this and go back to the previous message very easily if needed.

nathanstitt:
I suppose a middle ground would be to put a legend at the bottom that says something to the effect of:

"Scores will be not be displayed until 10% of students have completed the section"

philschatz:
one minor tweak: ".... 10% of students in the [section/period] have completed ..."

PR #875, opened 2015-12-06
Bugfixes

philschatz:
- book does not always open up in a new tab every time a link is clicked
- “Get Student Enrollment Code” only shows up for `is_concept_coach` courses

### DIFF
--- a/src/components/navbar/book-links.cjsx
+++ b/src/components/navbar/book-links.cjsx
@@ -25,8 +25,6 @@ BookLinks = React.createClass
         </a>
       )
 
-    window.open(course.webview_url, 'cc-webview-link')
-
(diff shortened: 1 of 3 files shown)


src/components/navbar/book-links.cjsx
```diff
@@ -25,8 +25,6 @@ BookLinks = React.createClass
         </a>
       )
 
-    window.open(course.webview_url, 'cc-webview-link')
-
```
nathanstitt: Aha!  that's why I had the render returning null in booklinks.  I'd noticed the popup blocker saying windows were blocked and had put it on my to-investigate-later list.  That line was from the old method where it was using window.open in a method vs just using an `a` tag.

PR #880, opened 2015-12-07
Legend for hidden chapters/sections

jivewise:
screen shot 2015-12-08 at 4 09 46 am

Putting this at the bottom of the dashboard progress table, it shouldn't show up on the blank dashboard.


philschatz:
are the chapters and sections hidden? I thought only the sections were currently hidden (leaving orphaned chapters in the dashboard)

jivewise:
nope, i thought so too, but after some testing i found that Nathan added in a line where empty chapters are just hidden.

PR #882, opened 2015-12-07
Hide the remove teacher link if there's only one teacher in the list

### LINKED ISSUE #872: Should not be able to remove the last instructor from a course
This way there are no courses with 0 instructors

jivewise:
screen shot 2015-12-08 at 5 44 54 am

fixes #872


philschatz:
@jason-holmes said "Fire away!" We don't need to hand hold teachers

PR #883, opened 2015-12-07
Scores table - student name/id spacing, truncation, tooltip

darrenmason:
tooltip

non-CC:

image

### DIFF
--- a/resources/styles/components/scores/index.less
+++ b/resources/styles/components/scores/index.less
@@ -172,6 +172,10 @@
     }
   }
   .name-cell {
+    .name, a {
+      text-align: left;
+      padding-left: 8px;
+    }
     .student-name.fullwidth {
       width: 100%;
     }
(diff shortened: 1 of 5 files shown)


philschatz:
Can the non-CC scores not contain the extra student id column since High school teachers do not know about it?

resources/styles/components/scores/index.less
```diff
@@ -162,17 +162,20 @@
       display: inline-block;
       width: 50%;
       text-align: center;
+      overflow: hidden;
+    text-overflow: ellipsis;
```
philschatz: looks like these are not indented properly

resources/styles/components/scores/index.less
```diff
@@ -162,17 +162,20 @@
...
+      
       &:first-child {
         border-right: 1px solid @nav-tabs-border-color;
       }
     }
   }
-  .name-cell .name {
-    padding-left: 1rem;
-    text-align: left;
-    display: inline-block;
```
nathanstitt: The `text-align: left;` rule here was used to remove the default center alignment for the name column.  I think we probably want to retain that :grin:

philschatz:
Thanks for the updates!

PR #885, opened 2015-12-08
Booksplash styles

nathanstitt:
Finally figured out how to interpolate a selector in less.

This styles the following components

Desktop SVG for blank CC course
image

CC Dashboard
image

Course chooser
image

Student Dashboard:

image

### DIFF
--- a/resources/styles/components/cc-dashboard/index.less
+++ b/resources/styles/components/cc-dashboard/index.less
@@ -107,20 +107,10 @@
     height: @banner-height;
     &:before {
       height: @banner-height;
-      font-family: 'Helvetica bold', Lato;
-      font-size: 14rem;
-      background-size: auto 135px;
-      background-position: 60px 10px;
-      letter-spacing: -5px;
-      overflow: hidden;
-      color: #77af41;
+      font-size: 88px;
       text-align: left;
-      font-size: 16rem;
-      font-weight: 900;
-      z-index: -1;
-      letter-spacing: -6px;
-      padding: 60px 0 10px 42px;
-      line-height: 0;
+      padding-left: 150px;
+      line-height: @banner-height;
     }
   }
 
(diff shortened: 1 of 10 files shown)


nathanstitt:
@Fredasaurus this sets the CC dashboard to be 120px tall with 88px font.  I left the Tutor dashboard the same with the font at 35rem on a  24rem background (it overflows).

Stacking is defined as 'Helvetica Neue', Helvetica, Arial, sans-serif, with letter spacing at -6px to pull it in a bit.

resources/styles/components/cc-dashboard/index.less
```diff
@@ -159,50 +149,22 @@
...
+          .banner       { fill: @bg; }
+          .banner-light { fill: lighten(@bg, 10%); }
+          .course-name  { fill: @@fg; }
+        }
       }
-      &.anatomy_physiology {
-        .banner       { fill: @anatomy_physiology-bg; }
-        .banner-light { fill: lighten(@anatomy_physiology-bg, 10%); }
-        .course-name  { fill: @anatomy_physiology-fg; }
+      .for(@openstax-book-appearance-codes); .-each(@code) {
```
philschatz: Ah, you opted against `car` and `cdr` and using something like:

```less
.foo() { } // empty list case
.foo(@head; @tail...) {
   // stuff
  .foo(@tail...);
}
```

nathanstitt: Yeah, I ran across the `.for`mixin and think it expresses what's going on better.  Internally it's using that pattern of course,  as far as I know that's the only real way to loop in less.

philschatz: The `.for(@array)` implementation looks a bit odd but yeah, semantically they're (hopefully) the same :smile:

PR #887, opened 2015-12-08
Display ellipsis for course titles that do not fit on header

nathanstitt:
Bad
image

Ok*ish*:
image


Fredasaurus:
Can we make that font a little smaller? So that the space below the g and p is the same as the space above the A. Thanks.

PR #904, opened 2015-12-16
Limit completed percentage to 100

jivewise:
Relates to: https://www.pivotaltracker.com/n/projects/1156756/stories/109459654

Limits completed percentage to be no greater than 100.  I think the source of the error is on the backend though.  I'm just adding this front end piece to protect from the broken UI.


philschatz:
Jing: is there no other way you can think of this happening unless the BE data is incorrect?

@Dantemss @jpslav any idea why BE could be returning > 100%?

jivewise:
I don't think there's any other way.  If you look in the pivotal ticket on 200_percent.json on line 209, the backend is returning a -1 for a not_started in a section (line.  I think Dante's already looking into it.

PR #906, opened 2015-12-17
Link CC students to webview if they end up on Tutor

nathanstitt:
screen shot 2015-12-17 at 12 14 21 pm


gregfitch:
Does it show the course name or a CC string?

nathanstitt:
It's the course name.  The screen shot is just from the demo data

PR #907, opened 2015-12-17
Scores Table Width in IE

jivewise:
For ticket: https://www.pivotaltracker.com/story/show/110357606

screen shot 2015-12-17 at 3 53 54 pm

When the table first loads, because the container div is empty, the div has no width.  This PR adds a temporary spacer in there that gets removed once we resize the table.  Pretty hacky, but I couldn't think of a great css only solution.

### DIFF
--- a/src/components/scores/index.cjsx
+++ b/src/components/scores/index.cjsx
@@ -66,9 +66,7 @@ Scores = React.createClass
     @sizeTable()
 
   sizeTable: ->
-    _.delay( =>
-      @setState({tableWidth: @tableWidth(), tableHeight: @tableHeight(), resizeSpacerNeeded: false}) if @isMounted()
-    , 100)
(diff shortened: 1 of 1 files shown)


philschatz:
hmm, did #898 not work?

nathanstitt:
I wonder if the real culprit is that the table doesn't have a width until it renders with data.  If that's the case my 100ms delay would only work as long as the network request completed before then.

Perhaps a better solution might be to set the table width equal to `window.width - <XXpx>` ?

jivewise:
So the table and the container doesn't seem to have a width until there's something in there.  So even with the delay code there, when we check for the table width, it just always returns a small width at first.  

I put the spacer in there at first to set the container width, and then later, when the delay happens we get the actual width of the container.

philschatz:
Should we remove the delay added in #898 then?

jivewise:
Looking into it...

philschatz:
closing this in favor of #913

PR #908, opened 2015-12-18
updated maths to try katex first and then use mathjax when needed

pandafulmanda:
it's like so fast, as in i don't even see the flash of unrendered maths. and it works in Windows (IE included) with lots of maths on the same page.  soooooo it's kinda like...cool.

and also, also, because you can run katex without the DOM using `renderToString`, if we are really trying to pinch for speed, we can send katex processing easily to a web worker

If this is to be accepted, this change should be ported to the components repo as well

openstax/tutor-server#904

### DIFF
--- a/index.html
+++ b/index.html
@@ -3,6 +3,7 @@
   <head>
     <meta http-equiv="content-type" content="text/html; charset=UTF8">
     <link rel="stylesheet" href="/dist/tutor.css" />
+    <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/KaTeX/0.5.1/katex.min.css">
     <script src="//cdn.mathjax.org/mathjax/2.5-latest/MathJax.js?config=TeX-MML-AM_HTMLorMML-full&amp;delayStartupUntil=configured" async></script>
     <script src="/dist/tutor.js" async></script>
   </head>
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "font-awesome": "~4.3.0",
     "htmlparser2": "^3.8.3",
     "jquery": "^2.1.4",
+    "katex": "^0.5.1",
     "markdown-it": "^4.1.1",
     "mime-types": "^2.1.3",
     "moment": "^2.9.0",
--- a/resources/styles/global/maths.less
+++ b/resources/styles/global/maths.less
@@ -1,18 +1,12 @@
 // Math Rendering
 // Hides all data-math elements, but still reserves their space in the DOM.
 // when .MathJax renders them, it'll be inside a .MathJax element which is displayed
-[data-math]{
+[data-math]:not(.math-rendered) {
   visibility: hidden;
   .MathJax {
     visibility: visible;
   }
 }
-
-// KaTeX 1.0.2 generates both HTML and MathML
-//
-// Hide the HTML and use MathJax for rendering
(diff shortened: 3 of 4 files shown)


philschatz:
It's _both_ `Ready for Review` and `Holding Pen`? (I think I understand why :smile: )

philschatz:
It seems like `katex` does not address the root problem. Is the math being rendered multiple times? (if so, it should probably only be rendered once; cnx frequently renders pages with a bunch of math without problems)

thoughts @jivewise or @nathanstitt ?

nathanstitt:
Yes, I think the biggest performance gain we got on tutor was rendering the math only once, and then re-rendering only new nodes as needed.   It looks like that's what you're doing here as well so ++

pandafulmanda:
I noticed the `:not(.math-rendered)` in the selectors, but we weren't using them in the code so I add that as well.  That would make doubly sure that we aren't rendering already rendered nodes

pandafulmanda:
also, on the homework builder page, the maths doesnt show for some reason even on first load, before any extra re-rendering would happen.  The `ArbitraryHtmlAndMath` component only calls these math rendering functions when the HTML prop being passed is [has actually changed](https://github.com/openstax/tutor-js/blob/master/src/components/html.cjsx#L40-L44)

philschatz:
hmm, could `katex` be pulled into another PR and tested independently?

pandafulmanda:
that is the main reason and perf gains from this pr...do you mean pulling out adding the `math-rendered` class stuff as separate pr to test that?  I don't think that solves the problem alone, but sure, i can do that.

philschatz:
so, if the math is only rendered once, did the exercise picker never run fast in IE? (I noticed you also added `MathJax.Hub.Queue`)

/cc @jivewise

jivewise:
Hm, I'm not sure of the effects on the exercise builder.  I always thought it was only slow if you selected too many chapters.  I can run some tests on this branch to see if we see any performance increases though.

philschatz:
After selecting each chapter (1-10) in Physics and listing all the exercises, `katex` is able to render the math 96.5% of the time `3704/(132 + 3704)`.

jivewise:
It seems to work when using mathjax 2.4-latest, but takes a really long time to load all the mathjax elements if there's a lot of math elements on the page.  The page does become responsive again once all the math elements are loaded.

philschatz:
(Amanda is still looking but here are some options we could discuss Monday)

- It does not seem to be a regression. Jing & Amanda were both able to reproduce using Oct 28 code
- The MathJax version did not change since April and there is no newer version we could try instead

Things we (@pandafulmanda, @jivewise, and I) tried:

- MathJax 2.4 takes about as long as 2.5 but finishes
- use katex with MathJax as a fallback (97% of math in Physics exercises CH 1-10 renders using katex)
  - 3700 formulas total in CH 1-10 of Physics
  - need to eyeball-check only 1668 (removed duplicates) [gist of all is here](https://gist.github.com/philschatz/a7ffd5fd7c2523c85493)
  - list of supported formulas: https://github.com/Khan/KaTeX/releases

Some other options that _may_ work:

- Only render math when scroll into view
- Render each formula independently (no batching; slower overall)
- <s>render math using CNX's caching service</s> (the service uses mathml, not latex as input)

/cc @kathi-fletcher @gregfitch 


But Amanda is looking at maybe some styling being the culprit (the HW Review page does render math inside the blue table but not in the exercises below)

PR #912, opened 2015-12-18
maths no katex but with class marking rendered elements

pandafulmanda:
to exclude from additional renders, only hide unrendered maths with css

see #908


pandafulmanda:
closing in favor of #915

PR #913, opened 2015-12-18
Student Scores - fix IE table width on load

darrenmason:
fixes issue with IE not getting table width from dom ref on initial load


nathanstitt:
@darrenmason I thought you'd discovered that the use of Math.max was a bug? I think you're correct, it seems like that would lock it to no more than 400 px wide?

PR #914, opened 2015-12-19
Scores table - fix proptype warning.

darrenmason:
small fix to proptype warning.


nathanstitt:
@darrenmason I thought you'd discovered that the use of `Math.max` was a bug?  I think you're correct, it seems like that would lock it to no more than 400 px wide?

PR #916, opened 2015-12-19
build deployable version with sourcemaps

philschatz:
- All deployments now include the source map and `tutor-dev` can be built with React in non-production mode (so raised Errors are the original errors)
- adds an `npm run build-archive` script which will replace `gulp prod` (and thus `gulp`)

Feedback is welcome!!

Some questions:

1. Should React be run in "production" mode?
2. Should `recordo` always be on? or just if `?collect=true` is in the initial URL?


philschatz:
@pandafulmanda @darrenmason @jivewise @nathanstitt any thoughts on which features should be enabled on `tutor-dev`?

1. minified JS? with source maps? (as a separate file?)
  - if we included source maps as a separate file we _could_ include the source maps in production (tutor-js _is_ a public repo)
2. Should React be run in production mode? (I'm guessing no; the performance isn't _too_ bad for a dev machine)
3. Should `recordo` always be turned on for `tutor-dev`?

jivewise:
1. I don't think minification's necessary on dev is it?  I feel like it's a big easier to debug if it's off, in case we ever need it.
2. Nah
3. That's a good idea.  Out of curiosity is there a "VCR" feature where we can have recordo run through events that were recorded?

philschatz:
@jivewise RE # 1: hmm, it _would_ be significantly quicker to deploy and maybe source maps don't buy us much (they might in production though).

RE # 3: not yet, but off-hours PR's to `recordo` are **definitely welcome!** Something like the following could work:

```coffee
THE_LOG.forEach ([type, event, selector]) ->
  if type is 'BROWSER' and event is 'CLICK'
    document.querySelector(selector).click(); sleep(20)
``` 

Also, Selenium _does_ have a language for describing events (and playing them back in Selenium).

jivewise:
Aren't sourcemaps always downloaded?  Even if the dev console isn't open?  How would it affect loading?

philschatz:
I think that if the sourcemap is stored in a separate file it is only fetched when the dev console is opened. The minified code contains a comment like `"// sourceMap=file.map"` at the end of the file

philschatz:
This adds the command `npm run build-archive` and `npm run build-archive-with-source` for ansible to run when deploying. The `-with-source` is useful for the `dev` environment because it is a single JS file but is not minified making it easy to debug.

I was thinking production could have the source maps too since the only additional runtime is about 10chars (`// sourceMap=file.map`) and the repo _is_ open for anyone to look at.

Any thoughts @openstax/tutor-fe ?

nathanstitt:
I think it's a good idea since it wouldn't increase the build size much at all and would only take effect for users who had the developer console open to the source tab.

PR #921, opened 2015-12-28
Performance Forecast - Adjusted header wrap and spacing when screen width narrow.

darrenmason:
https://www.pivotaltracker.com/story/show/100222354
https://www.pivotaltracker.com/story/show/100222352

Adjusted header wrap and spacing for narrower screens.

image

image

### DIFF
--- a/resources/styles/components/performance-forecast/heading.less
+++ b/resources/styles/components/performance-forecast/heading.less
@@ -10,23 +10,22 @@
     .flex-display();
     .justify-content(flex-start);
     .align-items(center);
+    .flex-wrap(wrap);
 
     .guide-group-title {
       .flex(1);
+      font-size: 3rem;
+      display: inline-block;
+      min-width: 45%;
     }
     .info {
       .flex-display();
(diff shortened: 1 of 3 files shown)


resources/styles/components/performance-forecast/heading.less
```diff
@@ -17,12 +17,15 @@
     .info {
       .flex-display();
       .flex-flow(space-between);
+      flex-wrap: wrap;
```
philschatz: Could/Should we use the `.flex-wrap()` mixin (if there is one)?

resources/styles/components/performance-forecast/heading.less
```diff
@@ -17,12 +17,15 @@
...
       .btn { margin: 0; }
     }
+    flex-wrap: wrap;
   }
   .guide-group-title {
     font-size: 3rem;
     display: inline-block;
+    min-width: 377px;
```
philschatz: Is this number based on another breakpoint value?

PR #922, opened 2015-12-30
Course Roster - updated overlay to address positioning issue and close button

darrenmason:
https://www.pivotaltracker.com/story/show/109985726

https://www.pivotaltracker.com/story/show/105715954

image

image

image

Popovers were not positioning correctly in firefox or ie when page was scrolled.
Using modal instead also addresses ticket regarding having a confirm/close.

### DIFF
--- a/src/components/course-settings/change-period.cjsx
+++ b/src/components/course-settings/change-period.cjsx
@@ -34,12 +34,14 @@ module.exports = React.createClass
         <span>
           Move to <CourseGroupingLabel courseId={@props.courseId} lowercase/>:
         </span>
-    <BS.Popover className='change-period' title={title} {...@props}>
-      <BS.Nav stacked bsStyle='pills' onSelect={@updatePeriod}>
-        {for period in course.periods
-          @renderPeriod(period) unless period.id is @props.student.period_id }
-      </BS.Nav>
-    </BS.Popover>
+    <BS.Modal bsSize='small' className='change-period' title={title} {...@props}>
+      <div className='modal-body'>
+        <BS.Nav stacked bsStyle='pills' onSelect={@updatePeriod}>
+          {for period in course.periods
+            @renderPeriod(period) unless period.id is @props.student.period_id }
+        </BS.Nav>
+      </div>
+    </BS.Modal>
 
   render: ->
     # if we have only 1 period, it's imposible to move a student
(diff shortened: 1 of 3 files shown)


src/components/course-settings/change-period.cjsx
```diff
@@ -34,12 +34,14 @@ module.exports = React.createClass
...
           Move to <CourseGroupingLabel courseId={@props.courseId} lowercase/>:
         </span>
-    <BS.Popover className='change-period' title={title} {...@props}>
-      <BS.Nav stacked bsStyle='pills' onSelect={@updatePeriod}>
-        {for period in course.periods
-          @renderPeriod(period) unless period.id is @props.student.period_id }
-      </BS.Nav>
-    </BS.Popover>
+    <BS.Modal bsSize='small' className='change-period' title={title} {...@props}>
+      <div className='modal-body'>
```
philschatz: Any reason you didn't use `BS.Modal.Body` in here?

nathanstitt:
I'd have a slight preference for a popover because it 1) is immediately obvious what it's pointing to, and 2) requires less mouse movement after it's triggered.

With that said, I don't think a modal is a bad solution if we can't get the popovers to work properly.  We'll have to make sure the modal has enough information so it's obvious that it's being displayed as a result of an action.  Based on the screenshots I think that should be ok.

darrenmason:
react-bootstrap version .25 fixes the popover bug;  it looks like in order to upgrade, all deprecated elements would have to be replaced.  So far it seems this includes:

all NavItem elements will have to change to NavDropdown elements.
all TabArea elements

could be a few others.

darrenmason:
found that react-bootstrap version .24 also has the popover fix and requires fewer modifications to get working.  it seems to break modals on calendar and create assignments.  It could be related to failed propType on modals.

darrenmason:
Found the exact fix:

- [66e41a4](../../commit/66e41a4) [fixed] Fix scroll top calculation for overlays

present in react-bootstrap 0.23.0
Have been testing, don't see any issues so far.

https://github.com/openstax/tutor-js/pull/923

philschatz:
Any thoughts @jason-holmes @Fredasaurus on popup vs modal (given the amount of testing needed to update bootstrap to work with IE)?

Fredasaurus:
What Nathan said.

Having said that, how soon would we want/need to upgrade react bootstrap and how much work do we think it would be to replace the deprecated elements?

philschatz:
ok. @darrenmason can you open a new PR with the version update and close this one? If the new version of `react-bootstrap` causes too many problems in QA then we can reopen this.

philschatz:
oh, nevermind; you already did :smile:

PR #927, opened 2016-01-03
use date preserving helper function to handle calendar tests

pandafulmanda:
especially important around the first of the month since travis runs on a different timezone, that used to confused what month the tests should be rendering the calendar in when converting string dates back to moment date objects

Also includes commit for pinning `flux-react` to 2.6.6 so that travis can install to completion.

The tests will now pass on travis for this PR today, but still not all pass for travis on another branch around this time.

* `moment` upgrade from `2.10.6` to `2.11` is causing lots of problems.  This PR pins to `moment@~2.10` for now.
* `flux-react@2.6.7` requires `react@0.14.0` which we are not ready for either.  Pinning to `flux-react@2.6.6`
* various other test updates in this pr addresses the normal weird time dependent test failures we were getting

Looked into upgrading to `moment@2.11` locally...doesn't look fun if we want the course time stuff to work still, so leaving this PR like this and looking to make upgrading `moment` a separate issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "camelcase": "^1.0.2",
     "classnames": "^2.1.3",
     "es6-promise": "^2.3",
-    "flux-react": "~2.6.4",
+    "flux-react": "2.6.6",
     "font-awesome": "~4.3.0",
     "htmlparser2": "^3.8.3",
     "jquery": "^2.1.4",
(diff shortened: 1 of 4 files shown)


philschatz:
# Woot! Tests pass again! :balloon: :tada:

PR #923, opened 2015-12-31
Update react bootstrap

darrenmason:
This update fixes the popover positioning bug present in ie and firefox, pertaining to:
https://www.pivotaltracker.com/story/show/109985726

Requires that openstax-react-components peerDependencies be updated to 0.23.0

Have been testing and I found no issues due to changes in framework so far.

react-bootstrap changelog of issue:

- [66e41a4](../../commit/66e41a4) [fixed] Fix scroll top calculation for overlays

Refs #922

PR #928, opened 2016-01-04
Fix/pin packages

pandafulmanda:
Package version resolution was starting to cause problems with unintentional updates, causing build to fail and/or application to be broken.  See #927 for additional details


pandafulmanda:
Will need to update `react-bootstrap` as described in #923 after hotfix is deployed, and dependency openstax/react-components is updated

philschatz:
versions seem not-too-extreme :+1: Waiting to merge until hotfix is pushed to production

philschatz:
Since this hotfix is now released, could you resolve the conflicts?

PR #925, opened 2016-01-01
Update and fix shared components

pandafulmanda:
- [x] Clean up duplicate styles
- [x] Replace local copies of components with shared components
  * `ChapterSectionMixin`
  * `GetPositionMixin`
  * `ResizeListenerMixin`
  * `SmartOverflow`
- [x] update tests as needed
- [x] update `react-components` again after openstax/react-components#31 and openstax/react-components#33 are merged

PR #929, opened 2016-01-04
continue button fix no more disappearing

pandafulmanda:
This fix is ported over from openstax/react-components#31, and will basically be deprecated after #925 as that will move to using the shared `async-button` that will have the same fix.

There was intermittent cases in which the continue button would disappear during exercises.  This was due to a request taking too long to return (> 30 secs allotted by the `asyncButton`) and the button trying to render it's failed state, which it could not do without this fix.

### DIFF
--- a/src/components/buttons/async-button.cjsx
+++ b/src/components/buttons/async-button.cjsx
@@ -51,10 +51,11 @@ module.exports = React.createClass
     {isTimedout} = @state
 
     buttonTypeClass = 'async-button'
+    FailedState = @props.failedState
 
     if isFailed or isTimedout
       stateClass = 'is-failed'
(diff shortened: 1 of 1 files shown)


src/components/buttons/async-button.cjsx
```diff
@@ -51,10 +51,11 @@ module.exports = React.createClass
     {isTimedout} = @state
 
     buttonTypeClass = 'async-button'
+    FailedState = @props.failedState
 
     if isFailed or isTimedout
       stateClass = 'is-failed'
-      return <failedState {...failedProps}/>
```
philschatz: Is `failedState` still needed/used?

PR #930, opened 2016-01-05
Fix for long summaries in exercises table

jivewise:
For pivotal ticket: https://www.pivotaltracker.com/story/show/109743900

Before: 
screen shot 2016-01-05 at 7 54 29 pm

After:
screen shot 2016-01-05 at 7 53 48 pm

### DIFF
--- a/src/components/task-plan/homework/exercises.cjsx
+++ b/src/components/task-plan/homework/exercises.cjsx
@@ -183,11 +183,6 @@ ExerciseTable = React.createClass
   renderExerciseRow: (exerciseId, index, hasTeks) ->
     {section, lo, tagString} = ExerciseStore.getTagStrings(exerciseId)
     content = ExerciseStore.getContent(exerciseId)
-    contentDiv = document.createElement('div')
-    contentDiv.innerHTML = content
-    
-    if contentDiv.children.length > 1
-      content = contentDiv.children[0].innerHTML
 
     if (hasTeks)
       teksString = ExerciseStore.getTeksString(exerciseId)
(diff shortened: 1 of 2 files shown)


src/components/task-plan/homework/exercises.cjsx
```diff
@@ -184,6 +185,9 @@ ExerciseTable = React.createClass
     {section, lo, tagString} = ExerciseStore.getTagStrings(exerciseId)
     content = ExerciseStore.getContent(exerciseId)
 
+    if $("<div>#{content}</div>").children().length > 1
```
philschatz: Can you use `document.createElement('div').innerHTML = content` here for the test? instead of including jQuery

src/components/task-plan/homework/exercises.cjsx
```diff
@@ -183,6 +183,11 @@ ExerciseTable = React.createClass
   renderExerciseRow: (exerciseId, index, hasTeks) ->
     {section, lo, tagString} = ExerciseStore.getTagStrings(exerciseId)
     content = ExerciseStore.getContent(exerciseId)
+    contentDiv = document.createElement('div')
+    contentDiv.innerHTML = content
+    
+    if contentDiv.children.length > 1
```
philschatz: _Minor:_ hmm, sorry, looking at this a bit more closely. Could the same thing be accomplished by doing `.td > :not(::first-child) { display: none; }`? If so, maybe that would be quicker than creating an element, pulling out the 1st child, converting it back to a string, then having react convert it back to DOM nodes?

/cc @pandafulmanda @darrenmason @nathanstitt ; Thoughts on trade-offs?

darrenmason: We could use:
```
table.exercise-table {  
  td {
    * + * {
      display: none;
    }
  }
}
```

The * + * will select any elements except the first regardless of type.  Tested it on chrome, ie, firefox.

PR #932, opened 2016-01-06
Redirect to webview if a single course CC student

nathanstitt:
Note, this uses `.replace()` vs `window.location.href=` since `.replace()` doesn't leave an entry in the history.  I think that'd be the best behavior here.

### DIFF
--- a/src/components/course-listing.cjsx
+++ b/src/components/course-listing.cjsx
@@ -18,7 +18,10 @@ DisplayOrRedirect = (transition, callback) ->
     conceptCoach = courses[0].is_concept_coach
 
     if roleType is 'student'
-      if conceptCoach then callback() else view = 'viewStudentDashboard'
+      if conceptCoach
+        window.location.replace(course.webview_url)
+      else
+        view = 'viewStudentDashboard'
     else if roleType is 'teacher'
       view = if conceptCoach then 'cc-dashboard' else 'taskplans'
     else


nathanstitt:
tl;dr - specs on this one are hard/impossible

Ok, I've attempted to add specs to this and am failing at it.  

The code path we want to test calls `window.location.replace`.  Therefore it would seem easy to add a `spy` to that and test that it's called.  However the browser security model prevents replacing properties on `window` (thankfully).

As an alternate, I created a `src/helper/window` file with a redirect helper.  I then attempted to stub that and test if it's called.  But then I ran into issues with the file being required and cached.  To work around that, I attempted to manipulate web packs cache (see specs PR on CC).  But then that fails because the main file's required by the router.  When I uncache that I get a huge slew of errors :(

philschatz:
Are you not able to do something like the following (it tests everything except the actual click):

```
helper.window.cjsx:

export class Helper extends React.Element
  onSend: -> window.location.href = @props.href
  render: -> @onSend()

test.coffee:

X = require 'src/helper/window`
X::onSend = -> done()
```

nathanstitt:
I attempted something pretty similar:

```coffee
module.exports = {
  ReplaceBrowserLocation: (url) ->
    window.location.replace(url)
}
```

And then in the spec:
```coffee
    sinon.spy(WindowHelpers. 'ReplaceBrowserLocation')
    CourseListingActions.loaded([course])
    renderListing().then (state) ->
      expect(WindowHelpers.helpers.ReplaceBrowserLocation).to.have.been.calledWith('foo')

```
But the original ReplaceBrowserLocation fn is called, not the spy.  I think that's because webpack doesn't actually pass the same object around, instead it clones it as part of the `require` call.

philschatz:
Ah, i think 1 more level of indirection would fix it (ie with a `module.exports.CLASS_NAME::replaceBrowserLocation`)

nathanstitt:
In order to close the loop on the spec, I eventually figured it out in #937 The issue turned out to be that I was using `simon.spy` which calls through to the original method, which would redirect and cause Karma to loose track of the spec progress.  Replacing `spy` with `stub` corrected it.

PR #936, opened 2016-01-07
Fix CC help link and add specs to it

nathanstitt:


nathanstitt:
Looks like the invariant exception still occurs.  That's the only failure w/Travis now though

philschatz:
I think this can be merged in the morning if the `openstax-react-components` changes are removed from this PR. Those would increase the amount of testing considerably.

nathanstitt:
@pandafulmanda heads up that this removes the update to `openstax-react-components`.  Perhaps we can move that into a separate PR if needed.

philschatz:
It seems the link goes to tutor zendesk even though I have a CC course


image

nathanstitt:
I just tested with teacher01 (charles morris) and I get the CC zendesk as I should since I have at least one CC course.

Can you double check maybe?

PR #937, opened 2016-01-07
Fix specs and add a few to test CC redirect

nathanstitt:
The main thing causing them all to fail was the redirect on a CC course.  That was being triggered on specs and causing random failures.


philschatz:
Yay, Thanks @nathanstitt !

PR #938, opened 2016-01-08
Fix the specs so they pass

nathanstitt:
The main thing that was stopping them was we were calling `getDOMNode()` on an unmounted component in the `ResizeListenerMixin`  

After I fixed that, then two other tests on the practice widget were failing.  I wasn't sure what the cause of their failures were so I commented them out :(  My rational for doing so is that it ok*ish* to do so because they're going to be moving to react-components soon anyway, and will be easier to test there.

### DIFF
--- a/src/components/resize-listener-mixin.cjsx
+++ b/src/components/resize-listener-mixin.cjsx
@@ -46,6 +46,7 @@ module.exports =
     height: componentNode.offsetHeight
 
   setInitialSize: ->
+    return unless @isMounted()
     windowEl = @_getWindowSize()
     componentEl = @_getComponentSize()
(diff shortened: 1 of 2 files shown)


nathanstitt:
Hm. looks like this is more complicated than my notes above suggest.  It's still failing intermittently locally, and Travis also failed.

The investigation continues ...

nathanstitt:
It seems like the failures are occurring because the react root node isn't found when getDomNode() is called.  Not sure what's causing that yet though

nathanstitt:
Closing this PR since I changed #937 to not redirect under karma which was the main culprit.

PR #941, opened 2016-01-12
Student Scores - refactor for HS/CC separation

darrenmason:
- [x] initial split of table types
- [x] remove redundant methods in HS/CC
- [x] replace CC conditionals when possible, modify/add new components
- [x] refactor of css styling / cell types, separate structure of classnames

cc-table2
hs-table2


philschatz:
@nathanstitt any thoughts on this?

nathanstitt:
Hm, I'm wondering whats the difference between the two tables?  And if we could accomplish it in a different way other than duplicating the entire component?  If the only difference is the split cells, could we 

```coffee
   TableCell = if cc then CCTableCell else HSTableCell

   <TableCell {...props} />
```

Also, are we sure that the Tutor table should remain as-is?  I'm curious if UX might be ok with Tutor's scores table being styled like the nicer CC one?

darrenmason:
The discussion on this was that the Tutor table would remain the same while the CC table would continue to diverge.  This way the CC table can have a completely different rendering method and css.  Also there was the goal of minimizing the amount of CC conditions mixed into various components.

PR #942, opened 2016-01-12
Update CC enrollment code help text

nathanstitt:
To reflect the new CC login/enrollment code screen

image

### DIFF
--- a/resources/styles/components/course-settings.less
+++ b/resources/styles/components/course-settings.less
@@ -90,6 +90,13 @@
 }
 
 .teacher-enrollment-code-modal {
+  textarea {
+    width: 100%;
+    height: 42rem;
+    resize: none;
+    #fonts > .sans(1.5rem, 2rem);
+    padding: 1rem;
+  }
   .enrollment-code {
     .code {
       font-weight: 800;
@@ -104,15 +111,7 @@
     .callout {
       padding-top: 15px;
     }
-    .howto {
-      padding: 10px;
-      box-shadow: inset 1px 1px 1px 1px @tutor-neutral-bright;
-      background-color: @tutor-neutral-lightest;
-      ol span {
-        // li not used so that numbering is paste-able
-        margin-right: 4px;
-      }
-    }
+
   }
 }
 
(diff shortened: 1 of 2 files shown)


philschatz:
Should this be a hotfix?

nathanstitt:
Marketing would like it released as soon as possible, since the current instructions are inaccurate.

I'll re-create it as a hot fix against the current release sha

philschatz:
when copying the text (for my "LMS") this ends up on my clipboard:

```
1. Paste this link in your web browser to visit the class textbookBiologytextbook:
https://archive.cnx.org/contents/f10533ca-f803-490d-b935-88899941197f@3.1:3
2. Click on section 1.1 in the book. Scroll to the bottom of the section and click on the Concept Coach button.
3. On the right side of the screen, enter your two-word enrollment code <enrollment code>.coach awesome
4. On the next screen, click “Click to begin login.” Then click “Sign up” and follow the prompts to create your free account.
5. Continue to your Concept Coach questions!
```

It looks like `class textbook` should be removed, and ` <enrollment code>.` should switch to `:`

/cc @CourtneyRay

nathanstitt:
@phil good catch.  It occurs to me that given that they’re supposed to copy & paste the text, wouldn’t it be better to use a big texture and select the text when they focus into it?  

We could also auto-copy to clipboard like you do in the error reporting PR, but that's probably to large of a change to push as a hot fix

philschatz:
I think a textarea that's select-all'd on focus works (and makes it clearer that you're supposed to copy it to your clipboard). thoughts @Fredasaurus or @jason-holmes?

nathanstitt:
Updated to use select-on-focus textarea.  That has the additional benefit of "what you see is what you copy" since it's just plain text

philschatz:
can the textarea be disabled (so people don't edit the text because they might have the expectation that we are saving their edits somehow)?

nathanstitt:
Good point.  I've set it to read-only (apparently disabled doesn't support `.select()`).  read-only doesn't emit the `focus` event but it does `click` so I switched to that for the trigger.  It seems to work just as well.

PR #944, opened 2016-01-14
Section Performance Progress Bar Fix

jivewise:
For: https://www.pivotaltracker.com/story/show/111296178
Before:
screen shot 2016-01-15 at 12 47 39 am

After:
screen shot 2016-01-15 at 12 44 28 am


darrenmason:
Found that min-width on progress bar was preventing smaller bars from narrowing.  This also covered label when narrow, so made the label always above.
```
        &:not(.no-progress) {
          min-width: 0px;
        }

        span {
          position: relative;
          z-index: 2;
        }
```
image

Issue #931, opened 2016-01-05
Improve Confidence when Deploying


philschatz:
Since many of us work remote, an async discussion might be easier.

Thoughts on the easiest way to improve confidence when deploying?

Some options (add more!):

- add a coveralls.io webhook
  - easy to drop in
- selenium scripts
  - utility methods would help QA folks
  - would better test how code works between components
  - _downside:_ might require starting up `tutor` (too slow for travis?) or more mock JSON files (not too bad)
- Selenium + coveralls: see #924 for how-to

Here is the unit test code coverage as of @a68f082. By changing the reporter in `test/config/karma-coverage.config.coffee#L12` (adding `coverageReporter: { type: 'lcov', dir: './coverage'}` I got the following report: https://openstax.github.io/tutor-js/coverage/

@darrenmason @jivewise @nathanstitt @pandafulmanda ?

nathanstitt:
I like the approach you've setup in #924 (Selenium + coveralls)

How hard is it to spec network requests with Selenium?  Some light googling suggests its possible but I've not personally done that before.

I'm think it would be great if we could test "When this input sequence is conducted, a http PUT matching XXX is issued"  Selenium gives us the ability to craft the input sequences, but I'm not sure how to spec the network request.

gregfitch:
Technically, it doesn't...by design. Hackishly you can form the same requests against the running selenium server.

philschatz:
Yep, "recording" the requests would be a quick way to create the mock JSON for tests and could be easily updated. Ruby uses [vcr/vcr](https://github.com/vcr/vcr) (readme has a list of ports) and I found [linkedin/sepia](https://github.com/linkedin/sepia) for NodeJS.

For browsers, I created [sepia.js](https://github.com/philschatz/sepia.js) a while back to play back cassettes for [octokat.js](https://github.com/philschatz/octokat.js) but with some of the [recordo](https://www.npmjs.com/package/recordo) code it could record the requests too.

philschatz:
Maybe we could convert the conclusions in #951 into the `README` in the `./test-integration` dir?

philschatz:
From the discussion on selecting things via classes, `data-` attributes, or `:nth-of-type()`: from an accessibility angle, would it make sense to default to [`aria-`](http://www.w3.org/TR/wai-aria/states_and_properties) attributes when selecting things in Selenium?

Of course, those attributes are probably not enough to do things but it would be easier than having entirely separate helpers that navigate through the UI.

[...]

PR #946, opened 2016-01-15
Selenium: Scores table - webdriver test

darrenmason:
refs #931


test-integration/scores.coffee
```diff
@@ -0,0 +1,56 @@
...
+  @it 'loads HS scores table', ->
+    @login(TEACHER_USERNAME)
+    CourseSelect.goTo(@, 'PHYSICS')
+    @waitClick(css: '.calendar-actions a:nth-child(3)')
+    @sleep(1000)
+
+    elements = MAIN_ELEMENTS
+
+    for element in elements
+      @driver.findElement(css: element)
```
philschatz: I think this might need a `@driver.findElement(css: element).then -> console.log ....`

_Aside:_ ES7's `await` keyword might be easier/concise; I can take a look at adding babel (probably easier for greg too)

test-integration/scores.coffee
```diff
@@ -0,0 +1,127 @@
...
+  '.nav-tabs',
+  '.export-button',
+  '.course-scores-title',
+  '.course-scores-container'
+]
+
+
+class StudentScores
+
+  constructor: (@test, @login, @courseTitle) ->
```
philschatz: Why include login in the constructor? This seems like it would be difficult to do something like:

- login
- test that there are at least 2 courses
- pick a course
- go to student scores

test-integration/scores.coffee
```diff
@@ -0,0 +1,127 @@
...
+
+  generateExport: ->
+    @clickElement('.export-button','generate export')
+
+  doneGenerating: ->
+    @test.driver.wait =>
+      @test.driver.isElementPresent(css: '.export-button-buttons a')
+
+  downloadExport: ->
+    if @doneGenerating()
```
philschatz: I don't think this works. `@doneGenerating()` returns a promise. maybe something like (in ES7):

```js
isDone = yield this.doneGenerating()
if (isDone) {
  ...
}
```

test-integration/scores.coffee
```diff
@@ -0,0 +1,127 @@
...
+    nextPeriod = @test.driver.findElement(css: '.nav-tabs li:nth-child(2)')
+    if nextPeriod?
+      nextPeriod.click().then ->
+        console.log 'changed period/section'
+
+  generateExport: ->
+    @clickElement('.export-button','generate export')
+
+  doneGenerating: ->
+    @test.driver.wait =>
```
philschatz: can you use an existing `@wait*` method instead of calling the driver directly? That would help with keeping the timeouts within a minute or so

test-integration/scores.coffee
```diff
@@ -0,0 +1,127 @@
...
+      @test.driver.isElementPresent(css: '.export-button-buttons a')
+
+  downloadExport: ->
+    if @doneGenerating()
+      @clickElement('.export-button-buttons a','download')
+
+  goBack: ->
```
philschatz: Could this be somewhere more generic since I think this applies to any screen?

test-integration/scores.coffee
```diff
@@ -0,0 +1,127 @@
...
+      e = @test.driver.findElement(css: element)
+      e.getAttribute('class').then (element) ->
+        console.log "found element: #{element}"
+      e.isDisplayed().then (element) ->
+        if not element
+          throw Error('element is not visible')
+
+  clickElement: (className, message) ->
+    @test.driver.findElement(css: className).click().then ->
+      console.log "clicked: #{message}"
```
philschatz: Is this just for internal testing?

test-integration/scores.coffee
```diff
@@ -0,0 +1,127 @@
...
+
+module.exports = StudentScores
+
+
+
+describe 'Student Scores', ->
+
+  @it 'HS Scores workflow', ->
+
+    @scores = new StudentScores(@, TEACHER_USERNAME, 'PHYSICS')
```
philschatz: Why is this `@scores =` instead of `scores =`?

test-integration/scores.coffee
nathanstitt: Because this helper class is intended to test only the student dashboard. In order to open the student dashboard you need a login and a course, and I didn't want to have to include the three/four lines to navigate to it the spec.  If they were in the spec, it's possible they could be missed or buggy and then `StudentDashboard` wouldn't work properly.

I think the scenario you outlined would be better tested by using other helper classes.  The course chooser already exists and is used internally by the `StudentDashboard` class.

test-integration/scores.coffee
darrenmason: will do

test-integration/scores.coffee
darrenmason: its been working for me

test-integration/scores.coffee
darrenmason: ok ill move it

test-integration/scores.coffee
darrenmason: I thought it would be ok to log the different elements that are clicked.  I found it useful to watch as it runs through test.

test-integration/scores.coffee
darrenmason: just an oversight, will update

philschatz:
this probably has a conflict because everyone is appending the new test file to `all.coffee` can you merge conflicts?

darrenmason:
adding new selenium test to #958, so closing

PR #948, opened 2016-01-15
Selenium tests for student dashboard

nathanstitt:
Uses #931


test-integration/helpers/student-dashboard.coffee
```diff
@@ -0,0 +1,100 @@
...
+CourseSelect = require './course-select'
+
+# A base class for a "Panel" type DOM object
+# If anyone finds this useful, feel free to expand and extract to helpers
+class PanelElement
+
+  constructor: (test, el) ->
+    @_test = test
+    if _.isString(el)
+      el = @test.driver.findElement(css: el)
```
philschatz: Would this have to be re-run whenever the page URL changes?

nathanstitt: `PanelElement` is a base class that holds a reference to the element that defines itself.  The parent class will then use that element reference to locate sub-elements, `DashboardTask` below is a good example of how it's used.

test-integration/student/dashboard.coffee
```diff
@@ -0,0 +1,52 @@
+{describe, CourseSelect, Calendar, ReadingBuilder, StudentDashboard} = require '../helpers'
+{expect} = require 'chai'
+_ = require 'underscore'
+
+
+describe 'Student Dashboard', ->
+
+  @it 'Shows Performance Forecast', ->
+    @dash = new StudentDashboard(@, 'student01', 'Biology I')
```
philschatz: why `@dash =` instead of just `dash =`? Also, what's the advantage of instantiating a `new StudentDashboard` each time?

nathanstitt: Heh.  That's an oversight.  At one time I had a `beforeEach` with  `@dash ||= new StudentDashboard(...)`

That worked until I wrote the test for creating an assignment as teacher, then navigating back and checking it appeared.  At that point the @dashboard reference wasn't valid any longer since the DOM had been mutated.  So I moved the creation into each spec (and neglected to remove the `@`)

I'll remove the `@`

test-integration/helpers/student-dashboard.coffee
philschatz: so would the logic in the test need to be something like:

```
if !dash.valid()  # eep, could this be prefixed with `is` so it is not confusing whether it is a getter or performing an action
  dash = new StudentDashboard(@, 'student01', ...)
```

if the test logged in as a student, logged out and back as a teacher, and then logged out and back as a student?

philschatz:
waiting on decisions from #948

test-integration/helpers/test-element.coffee
```diff
@@ -51,7 +51,8 @@ Object.defineProperties TestItemHelper.prototype,
 
 
 class TestHelper extends TestItemHelper
-  constructor: (test, testElementLocator, commonElements = {}, options) ->
+  constructor: (test, testElementLocator, commonElements, options) ->
+    commonElements ||= _.result(@, 'elementRefs', {})
     defaultOptions =
```
nathanstitt: @pandafulmanda (and rest of FE team) what do you think about this?  It attempts to read the commonElements from an `elementRefs` property of the object.  I like setting them there better than an unattached constant from outside the class

test-integration/helpers/student-dashboard.coffee
```diff
@@ -0,0 +1,42 @@
...
+    visibleEvents:
+      css: '.tab-pane.active .task'
+
+  constructor: (test) ->
+    super(test, '.student-dashboard')
+
+  getVisibleEventHelpers: ->
+    makeEvent = _.partial(Event.fromElement, @test)
+    @el.visibleEvents.getAll().then( (tasks) ->
+      Promise.all( _.map(tasks, makeEvent ) )
```
pandafulmanda: should we be sticking to selenium promises i.e. `selenium.promise.all` as much as possible? or go convert to using regular promises (which i actually prefer) altogether?  more importantly though, i think we should try to be consistent unless we must use one or the other specifically

test-integration/helpers/test-element.coffee
pandafulmanda: okay, i don't see why not have both options.  i'd kept it out of the class because i felt it would keep it cleaner, but attaching it makes more sense for reference the locators throughout the helper later on if needed

test-integration/helpers/student-dashboard.coffee
nathanstitt: Yep, we should do that.  In this case it doesn't really matter because `Promise.all` just resolves all the existing (selenium) promises.  It's still a good idea to use the selenium though.  I'll switch it.

pandafulmanda: selenium's promises have an `.map` function as well, and the `.fullyResolved` function is interesting as well -- [example](https://github.com/openstax/tutor-js/blob/update/ref-book-test-helper/test-integration/helpers/reference-book.coffee#L56-L58) and [docs](http://selenium.googlecode.com/git/docs/api/javascript/namespace_webdriver_promise.html)

nathanstitt: Oh nifty!  Thanks for pointing that out.  I see there's also an interesting `fullyResolved` in there that will use an object with values.

philschatz: I thought the conclusion was to have these "context-sensitive" helpers either take the Selenium element as an arg to the method? That way the helpers always yield Selenium elements?

nathanstitt:
@philschatz, I can't find your comment any longer?

I actually started out doing it all in the spec and it got way to hairy so I moved some of the logic into the helper.  For instance this is what I started with:

```coffee
    @dash.el.visibleEvents.getAll()
      .then( (tasks)    -> Promise.all _.map tasks, (element) ->
        element.getAttribute('data-event-id') )
      .then( (eventIds) => _.map( eventIds, (eventId) =>
        new StudentDashboard.Event(@, eventId) ) )
      .then( (tasks)    -> Promise.all _.map tasks, (t) ->
        t.el.title.get().getText() )
      .then (titles)   ->
        for title in titles
          expect(title).to.match(/\w+/)
```

philschatz:
I guess the test could be written like the following. How would this read in a test?

```
// Option 1
elements = await this.dash.el.visibleEvents.getAll()
events = elements.map((el) => new StudentDashboard.Event(this, el))
events.forEach((event) => {
  title = await event.el.title.get()
  expect(title.getText()).to.match(/\w+/)
})
expect(events).to.have.length.gt(0)

// Option 2
elements = await this.dash.el.visibleEvents.getAll()
elements.forEach((el) => {
  event = new StudentDashboard.Event(this, el)
  title = await event.el.title.get()
  expect(title.getText()).to.match(/\w+/)
});
expect(elements).to.have.length.gt(0)
```

philschatz:
Aside from the reading builder comment, looks good to me!

/cc @openstax/test

nathanstitt:
@philschatz I see what you're saying about es7 being cleaner.  I just converted to that and it does read better.  However I think it still makes sense for the Dash helper to create and return the Tasks helpers, because now it's even cleaner

Now it's just:
```js
    const events = await this.dash.getVisibleEventHelpers({
      where: {title: title}
    });
    expect(events.length).to.equal(1)
```

nathanstitt:
@openstax/tutor-fe what think you about returning non-element objects from helpers?  I see the pattern also used in calendar as well.

nathanstitt:
Note, the example above is kinda contrived.  What I'm trying to avoid in the specs is to not have them instanciate the tasks directly because then the spec would need to either be responsible for the query logic or it would have to pass the task an element which the task would retain and possibly still use after its out of scope 

Plus it just feels like the "right way" to me, as fuzzy as that sounds.   If there's a reason not to do it, then I'm willing to reconsider though

test-integration/helpers/student-dashboard.coffee
```diff
@@ -0,0 +1,59 @@
...
+    locator = @getLocator().css or ''
+    title:
+      css: "#{locator} .title"
+    progress:
+      css: "#{locator} .feedback"
+
+  constructor: (test, eventId) ->
+    super(test, css: ".task[data-event-id='#{eventId}']")
+
+  @fromElement: (test, element) ->
```
philschatz: Nice static method!

philschatz:
Couldn't the signature of `getVisibleEventHelpers` remain the same but still return the Selenium Elements?

Internally, `Dashboard::getVisibleEventHelpers` _could_ use the `Event` class, but would not be required to.

nathanstitt:
@philschatz Ok, i've updated to return elements now.  The specs that use it can re-instantiate the objects if it needs to I guess

PR #949, opened 2016-01-16
Selenium: Test/fix current

pandafulmanda:
Includes fixes for
  * some warnings from calendar
  * calendar plan modal updating if plan needs to update without weirdly disappearing
  * calendar tests run
  * reference book test adjustments

tests are mad and timeout when running altogether :cry: but this has a few small fixes for calendar and reference book that would nice for master

refs #931


test-integration/reference-book.coffee
```diff
@@ -12,65 +13,91 @@ describe 'Reference Book Exercises', ->
...
-      for i in [1..SECTIONS_TO_TEST]
-        # Selenium sometimes keeps pressing the same next button (doneLoading doesn't seem to work 100%)
-        @driver.findElement(css: 'a.nav.next').getAttribute('href').then (oldHref) =>
-          @addTimeout(3)
+    checkEachPage = =>
+      hrefToCheck = null
+      nextNav = null
+      # Selenium sometimes keeps pressing the same next button (doneLoading doesn't seem to work 100%)
+      @driver.wait(doneLoading)
+        .then =>
```
philschatz: I don't think you need all these `.then`'s

pandafulmanda: i think when working with these promises, it makes it very clear that we are running the steps in series.  there were instances when things were running in parallel before and we were making up for it by adding timeouts.

pandafulmanda: @philschatz points out that selenium's promises que up before running in series: https://code.google.com/p/selenium/wiki/WebDriverJs#Control_Flows

src/components/course-calendar/plan.cjsx
```diff
@@ -125,6 +123,7 @@ CoursePlan = React.createClass
 
   componentWillMount: ->
     @subscribeToPublishing(@props.item.plan)
+    @checkRoute()
```
philschatz: _Minor:_ this name is confusing to me; it looked like this line did nothing (I thought `checkRoute()` should return something), not that it has side-effects :smile:

src/helpers/time.coffee
```diff
@@ -34,6 +34,11 @@ tzdetect =
 
 
 TimeHelper =
+
+  isPropAMoment: (props, propName, componentName) ->
```
philschatz: to match the naming of other `propTypes` could this be `TimeHelper.PropTypes.Moment`?

test-integration/helpers/describe.coffee
```diff
@@ -179,6 +179,22 @@ describe = (name, cb) ->
...
+      @forNTimesInSeries = (nTimes, fn, fn2) =>
+        arrayOfFns = []
+
+        for i in [1..nTimes]
+          arrayOfFns.push(new Promise(fn))
+
+        (args...) =>
+          @checkInSeries(arrayOfFns, fn2)(args...)
+
+      @checkInSeries = (arrayOfFns, fnBetween) =>
```
philschatz: Does Selenium not already do this since promises are resolved serially?

test-integration/reference-book.coffee
```diff
@@ -9,68 +10,111 @@ SECTIONS_TO_TEST = 10
...
+        .then(closeMenuAndResizeBook)
+        .then =>
+          # Perform `checkEachPage` in series for SECTIONS_TO_TEST times.
+          @forNTimesInSeries(SECTIONS_TO_TEST, checkEachPage)()
+
+    closeMenuAndResizeBook = =>
+      # Resize prevents need to click next again -- without this,
+      # if the window is too skinny/if the next button is right by the scroll bar,
+      # the page will accidentally get clicked to scroll as opposed to clicking next.
+      @driver.manage().window().setSize(1080, 1080)
```
philschatz: Could this be a helper? And instead of setting the size explicitly, have methods like `.setIPhone()`, `.setTablet()`, `.setNarrow()`, `.setLarge()` (not suggesting all would be implemented). That way we would have some "standard" sizes we'd be testing with?

test-integration/reference-book.coffee
```diff
@@ -9,68 +10,111 @@ SECTIONS_TO_TEST = 10
...
-
-      for i in [1..SECTIONS_TO_TEST]
-        # Selenium sometimes keeps pressing the same next button (doneLoading doesn't seem to work 100%)
-        @driver.findElement(css: 'a.nav.next').getAttribute('href').then (oldHref) =>
-          @addTimeout(3)
+    checkEachPage = =>
+      hrefToCheck = null
+      nextNav = null
+      # Selenium sometimes keeps pressing the same next button (doneLoading doesn't seem to work 100%)
+      @driver.wait(doneLoading)
```
philschatz: Could this be a common helper named something like `.waitFor(fn, failMsg=null)` ?

test-integration/reference-book.coffee
```diff
@@ -9,68 +10,111 @@ SECTIONS_TO_TEST = 10
...
+          console.info("Next href is #{oldHref}. Clicking next.")
+          hrefToCheck = oldHref
+          nextNav.click()
+
+      @driver.wait(doneLoading)
+      checkPageChanged(hrefToCheck)
+
+    doneLoading = =>
+      # sleep to make sure exercises have a chance to start loading before checking if page is indeed done loading
+      @sleep(100)
```
philschatz: Instead of sleeping for a fixed period of time (Sauce and BrowserStack may take longer so it's nice to change this globally) could you wait for an element to either be present or to not be present?

philschatz:
I have a feeling these sorts of _exhaustive_ tests are nice to have sometimes but might not be as crucial as translating the testrails tests over.

philschatz:
waiting on decisions from #948

pandafulmanda:
closing in favor of #964, #953, and #961

PR #953, opened 2016-01-25
Selenium: helpers for tests

pandafulmanda:
test pattern?


test-integration/calendar.coffee
```diff
@@ -14,7 +14,7 @@ describe 'Calendar and Stats', ->
     @login(TEACHER_USERNAME)
 
     _.each ['BIOLOGY', 'PHYSICS'], (courseCategory) =>
-      CourseSelect.goTo(@, courseCategory)
+      new CourseSelect(@).goTo(courseCategory)
```
philschatz: I thought this was going to be moved up into the `beforeEach`, no?

test-integration/helpers/calendar.coffee
```diff
@@ -1,6 +1,85 @@
...
+
+class PlanPopupHelper extends  TestHelper
+  constructor: (test, testElementLocator) ->
+
+    testElementLocator ?= '.plan-modal .panel.panel-default'
+    super(test, testElementLocator, COMMON_POPUP_ELEMENTS)
+
+  close: =>
+    @getCloseButton().click()
+    @test.sleep(2000)
```
philschatz: eep, could this wait for an element to be visible instead of a specific time?

test-integration/helpers/calendar.coffee
```diff
@@ -1,6 +1,85 @@
...
+  goPerformanceForecast: =>
+    @getForecastLink().click()
+
+  goOpen: (title) =>
+    # wait until the calendar is open
+    @waitUntilLoaded()
+    # TODO: Make this a `data-title` attribute
+    # HACK: Might need to scroll the item to click on into view
+    el = @test.waitAnd(css: "[data-title='#{title}']")
+    @test.scrollTo(el)
```
philschatz: `describe` already has a scrollToAndClick method (don't remember the exact name, maybe it's `waitClick`?) can you use that instead?

test-integration/helpers/describe.coffee
```diff
@@ -75,6 +75,29 @@ describe = (name, cb) ->
...
+          end = Date.now()
+          spent = end - start
+          diff = ms - spent
+          # console.log "Took #{spent / 1000}sec of #{ms / 1000}"
+          if spent > ms
+            throw new Error("BUG: Took longer than expected (#{spent / 1000}). Expected #{ms / 1000} sec")
+          @addTimeoutMs(-diff)
+          val
+        # Because of animations an element might be in the DOM but not visible
+        el = @driver.findElements(locator)
```
philschatz: Why find all but only test visibility of 1?

test-integration/helpers/reading-builder.coffee
```diff
@@ -1,108 +1,122 @@
...
-        alert.accept()
-
-      Calendar.verify(test, 60 * 1000)
-
-
-module.exports = {setDate, edit}
+  # Helper methods for dealing with the Reading Assignment Builder
+class ReadingBuilder extends TestHelper
+
+  constructor: (test) ->
```
philschatz: Why do these helpers take `test` instead of driver? If it is only because of the helpers in `describe.coffee` then would it make sense to pull them out of describe and revisit all these places?

nathanstitt: Yep, just chatted with @pandafulmanda about that.  We think we're going to extract the methods from describe and then pass the driver into the helpers

test-integration/helpers/test-element.coffee
```diff
@@ -0,0 +1,72 @@
+selenium = require 'selenium-webdriver'
```
philschatz: I was wondering if it would make sense to put this in a different directory than the other helpers because they are all expected to be called for doing things on a screen (to avoid confusion)

nathanstitt:
@philschatz, @pandafulmanda I've taken a shot at extracting the helpers from describe and then passing in the `@driver` to them vs `@test`.  I've hit a snag though.

`describe` replaces the `timeout` function with it's own implementation that can be extended.  I'd initially planned to leave that inplace since it's tightly coupled to the test.  However several of the helpers call `@addTimeoutMs(ms)`, so they'll need an instance of the `test`, not `driver` in order to work.  See the [extracted wait](https://github.com/openstax/tutor-js/blob/test/sel-test-helper/test-integration/helpers/wait.coffee#L12) for example.

I think we could work around that by extracting the addTimeout stuff into an object and then adding that onto the driver object in the custom `@__beforeEach` method.    I think it'd be cleaner to just pass the test to the helpers though.  Thoughts?

philschatz:
Ah, good snag! I think passing the test to the helpers makes sense then (since _everything_ calls `addTimeout`). Does that mean nothing can be instantiated in `beforeEach` ?

nathanstitt:
Oh, no we should still be able to use `beforeEach` fine.  We'll just have to still pass the test into each of the helpers

philschatz:
oh, I forgot `@` in `beforeEach` **was** the current test :smile:

nathanstitt:
@philschatz and @pandafulmanda I think this is ready for another round of reviews.  Take a look and see what you think.

test-integration/calendar.coffee
```diff
@@ -97,39 +94,37 @@ describe 'Calendar and Stats', ->
...
-        @waitAnd(css: '.chapter-panel.weaker, .no-data-message')
-        @waitClick(css: '.performance-forecast a.back')
-        @waitAnd(css: '.course-scores-wrap')
+        wait(@).for(css: '.chapter-panel.weaker, .no-data-message')
+        wait(@).click(css: '.performance-forecast a.back')
+        wait(@).for(css: '.course-scores-wrap')
 
       # only test the 1st row of each Student Response
-      @forEach '.fixedDataTableRowLayout_rowWrapper:nth-of-type(1) .task-result', (item, index, total) =>
+      forEach @, css: '.fixedDataTableRowLayout_rowWrapper:nth-of-type(1) .task-result', (item, index, total) =>
```
philschatz: Could this syntax be similar to `wait` in that the test arg is curried?

test-integration/helpers/wait.coffee
```diff
@@ -0,0 +1,67 @@
...
+      # console.log "Took #{spent / 1000}sec of #{ms / 1000}"
+      if spent > ms
+        throw new Error("BUG: Took longer than expected (#{spent / 1000}). Expected #{ms / 1000} sec")
+      @test.addTimeoutMs(-diff)
+      val
+    # Because of animations an element might be in the DOM but not visible
+    el = @test.driver.findElements(locator)
+
+    el.then (elements) =>
+      @test.driver.wait(selenium.until.elementIsVisible(elements[0]))
```
philschatz: Why find all elements and then only wait until the 1st is visible?

test-integration/helpers/window-position.coffee
```diff
@@ -0,0 +1,12 @@
+windowPosition = (test) ->
```
philschatz: It looks like other helpers use the class syntax while this one just relies on closures, thoughts on which method we should use in general?

test-integration/task-plan/publish.coffee
```diff
@@ -6,29 +6,38 @@ TEACHER_USERNAME = 'teacher01'
 
 describe 'Assignment Publishing Tests', ->
 
-  @it 'Publishes a Reading with opensAt to tomorrow and deletes (idempotent)', ->
-    title = @freshId()
-
-    @login(TEACHER_USERNAME)
+  beforeEach ->
+    @title = freshId()
+    new User(@, TEACHER_USERNAME).login()
```
philschatz: For consistency would it make sense to have login take the username argument?

test-integration/helpers/wait.coffee
nathanstitt: I dunno.  I just copied the codez in :)  Actually why are we waiting until they're visible anyway?  Do we really need to?  I could see that the first element might be visible but other's might be below the scroll.

test-integration/task.coffee
```diff
@@ -0,0 +1,50 @@
...
+
+  @it 'Can continue and go to expected steps (readonly)', ->
+
+    _.each ['BIOLOGY', 'PHYSICS'], (courseCategory) =>
+      appearance = courseCategory.toLowerCase()
+      wait(@).click(css: "[data-appearance='#{appearance}'] > [href*='list']")
+      wait(@).click(css: '.workable.task')
+
+      @task.waitUntilLoaded(2000)
+      @task.getEnabledContinueButton().click()
```
philschatz: was renaming this to `@task.els.enabledContinueButton.click()` or `@task.continue()` part of this work?

test-integration/task-plan/publish.coffee
nathanstitt: Yes, I considered that but decided against it because the User will have methods other than login at some point, like `displayProfile` or `selectMenuOption`.  When that happens then the username would need to be supplied as part of the constructor.

test-integration/task.coffee
nathanstitt: No, but that work has been completed.  Looks like this code didn't get updated.  Doing so now

test-integration/helpers/window-position.coffee
nathanstitt: Not really ;)  This and `wait` both have the same usage, i.e. `fn(test).action()`  They're just implemented differently.  My (fairly uninformed) thoughts were that the class syntax is better for larger helpers since it keeps the state enclosed, while the closure is better for simple ones like this one.  It doesn't really matter much to me though.

test-integration/task-plan/publish.coffee
philschatz: Why would `displayProfile` need to already have the username? It seems like `user.displayProfile()` would just need to click on the `<div class="profile">Profile</div>` element, no?

test-integration/helpers/reading-builder.coffee
philschatz: I think the `addTimeout` case you mentioned is a compelling reason for continuing to pass in the `test`, no?

test-integration/helpers/window-position.coffee
philschatz: My initial suggestion would be closures because it makes it clear there is no place to add additional state (the majority of helpers). If we wanted to store more state we could easily convert them to a class later.

nathanstitt: It's not so much the state that bothers me with (long) closures, it's the possibility for errors.  

With a closure you've got this "magic" variable(s) that each function has to take care not to modify.  It's pretty easy to accidentally create a variable with the same name as the closure value in a function and accidentally alter the closure value for all the other functions.  In a short 1/2 screen long block that's not so bad but when you have a longer one it's easy to forget that `test` is a reserved var.

In the object based setup the state is indicated by either a `@` or `this.`  That makes it readily apparent that that var is special and it won't be alter if you create a plain var with the same name.

I also suspect that the prototype based method will perform a bit better than re-creating an object each time, but that's speculation and it's quite possible the new fangled optimizers will treat each one the same.

test-integration/task.coffee
nathanstitt: Actually it looks like the `els` prefix isn't in place for the getters.  @pandafulmanda I'm trying to remember, but was that what we'd decided wasn't needed and we left it out, or was it something else?

test-integration/helpers/window-position.coffee
philschatz: Agreed on the clarity of naming vars. But what about naming it `theTest` (or something), or using `const` in ES6 to ensure the var is not renamed? I'm more worried about consistency and I guess just a slight preference for closures :smile:

test-integration/calendar.coffee
nathanstitt: It could be, but i'm having trouble envisioning what that would look like.  Maybe an `enum` interface that could be used like: `enum(@).each(css: '.foo')` ?

philschatz: I guess the suggestion would work best if all the methods on `wait` and `forEach`, etc were all on a common object (maybe `utils(@).click(...)/.forEach(...)`)?

test-integration/task.coffee
pandafulmanda: sorry, it's `el`

philschatz:
Stuff I think we could revisit in a separate PR (I think the comments got closed when indenting files):

- Why does `helpers/user` need the login in the constructor?
- Can we replace several of the `wait(@test)...` calls with `@test.utils.wait...`?

PR #955, opened 2016-01-25
Fixes to calendar and reference book

pandafulmanda:
Separate out code fixes from test updates in #949 

Includes fixes for
  * some warnings from calendar
  * calendar plan modal updating if plan needs to update without weirdly disappearing
  * calendar tests run
  * reference book test adjustments


philschatz:
a quick glance looks good to me. thoughts @nathanstitt ?

nathanstitt:
Looks good to me as well.  Really like the `TimeHelper.PropTypes.moment` propType validation.  The hover made me a bit nervous because of how it's implemented on touch UI's, but this looks like it should be fine.

PR #957, opened 2016-01-26
Selenium tests for exercise selection

jivewise:
A couple of homework builder tests.

I've pulled out the task plan date tests into another PR, still trying to get those cleaned up.


philschatz:
Why split up the `reading-builder` and `homework-builder`? I was hoping that people needing to build a reading but not interested in testing the `*builder` (`reading-builder` is a bad name IMO, maybe `plan-builder`?) would just call `.edit(...)` with the args they needed to populate

test-integration/helpers/exercise-selector.coffee
```diff
@@ -0,0 +1,43 @@
...
+  reviewExercisesButton:
+    css: '.exercise-summary .btn.-review-exercises'
+
+
+class ExerciseSelector extends TestHelper
+  constructor: (test, testElementLocator) ->
+    testElementLocator ?= css: '.add-exercise-list'
+    super test, testElementLocator, EXERCISE_SELECTOR_ELEMENTS, defaultWaitTime: 3000
+
+  selectExercises: (numExercises) ->
```
philschatz: Could this be something like `selectExercisesByNumber` or something similar to #987

jivewise: How about `selectNumberOfExercises`?

philschatz: ah, yep :smile: . I wonder if it would be more useful to have `isExerciseSelectedByNumber/ById` and `toggleExerciseByNumber/ById` but I can add those to the master list and just keep them unimplemented.

jivewise:
Sure, I can implement those in a separate PR.  I just wasn't sure yet how I'd get the ids to begin with.

test-integration/helpers/task-plan-builder.coffee
```diff
@@ -260,30 +228,60 @@ class ReadingBuilder extends TestHelper
 
     if sections
       # Open the chapter list by clicking the button and waiting for the list to load
-      @openSelectReadingList()
+      if numExercises
```
philschatz: Could this be agnostic to the type of Assignment? That way tests could click "Select Problems" (get to the problem listing screen) without actually having to select problems?

test-integration/helpers/task-plan-builder.coffee
```diff
@@ -260,30 +228,60 @@ class ReadingBuilder extends TestHelper
...
+
       # Make sure nav bar does not cover buttons
       @test.utils.windowPosition.scrollTop()
 
-      # Expand the chapter and then select the section
-      for section in sections
-        do (section) =>
-          section = "#{section}" # Ensure the section is a string so we can split it
-          @el.selectReadingsList.selectSection(section)
+      @el.selectReadingsList.selectSections(sections)
```
philschatz: I think the conclusion from https://github.com/openstax/tutor-js/pull/983#issuecomment-186302403 (Autogenerating documentation) was that the helpers would not be instantiated and stuck on `@el`. instead it would be something like `(new SelectReadingsList(@test)).selectSections` (or just define the nested helper in PlanBuilder's constructor

test-integration/helpers/select-readings-dialog.coffee
```diff
@@ -0,0 +1,51 @@
...
+  loadingLocator:
+    css: '.loadable.is-loading'
+
+  nextStepBtn:
+    css: '.panel-footer .btn.btn-primary'
+
+class SelectReadingsList extends TestHelper
+  constructor: (test, testElementLocator) ->
+    testElementLocator ?= css: ".select-reading-dialog#{OPENED_PANEL_SELECTOR}"
+    super test, testElementLocator, COMMON_SELECT_READING_ELEMENTS, defaultWaitTime: 5000
```
philschatz: Why the increase in the wait time? Can you do `utils.wait.giveTime ms, ->` just around the operation that is timing out?

jivewise: I increased it just because it was always timing out on loading that particular element, and this was the only place I saw that could do it, otherwise, we'll have to put in that timeout every time we load this element.

jivewise: Yep..loading the sections seemed to take longer than the timeout half the time.

philschatz: Can you try just removing the default? The default is 40sec.

jivewise: Sure.

jivewise: Looks like if I use the default, it works.  Good call.

test-integration/helpers/task-plan-builder.coffee
```diff
@@ -285,5 +250,30 @@ class ReadingBuilder extends TestHelper
       when 'CANCEL' then @cancel()
       when 'DELETE' then @delete()
 
+  selectNumberOfExercises: (numExercises=4) ->
+    @_exerciseSelector.waitUntilLoaded()
+    @_exerciseSelector.selectNumberOfExercises(numExercises)
+
+  verifySelectedExercises: (numExercises) ->
+    @el.selectedExercises.findElements().then (els) ->
+      expect(numExercises).to.be.equal(els.length)
```
philschatz: I was not expecting to see `expect` outside of a test. Maybe this could be a method defined inside the test? Thoughts @openstax/tutor-fe ?

pandafulmanda: I agree that i would expect the `expect` to be in the spec...i might expect to see this in there though, but it's not really needed:

```coffee
countSelectedExercises: ->
   @el.selectedExercises.findElements().then (els) ->
     els.length
```

then the spec might be
```coffee
   # stuffs
   @taskPlanBuilder.countSelectedExercises().then (numOfSelected) ->
      expect(numExercises).to.be.equal(numOfSelected)
```

jivewise: Ah, good point...I'll update this.

philschatz:

jivewise:
This should likely be closed in favor of: https://github.com/openstax/tutor-js/pull/1040

PR #960, opened 2016-01-28
convert helpers to es7

philschatz:
converts `wait` and `forEach` to ES6 files.

I changed the modules to export a single function instead of `export default () =>` because in coffee files you would have to do `wait = require('./wait').default` which seemed a bit more cumbersome (for now) than `{wait} = require './wait'`. 

<s>Still working on moving over to use the `await` keyword instead of promises.</s>

# TODO (DONE)

- [ ] use `await` syntax in those files instead of `Promise` (`.then()`)

Thoughts @pandafulmanda, @nathanstitt @darrenmason, @jivewise ?

# Conclusion

Turns out using `await` mixes the types of Promises floating around, which I think would be OK if we only used ES7 in helpers and tests but since we mix in coffeescript all of coffeescript would need additional `.then` calls (since we can't rely on Selenium's serialization anymore).

To migrate to ES7 we would have to:

1. rewrite all the tests to ES7 first
2. rewrite all the helpers that are not used by other helpers to ES7 (ie `reading-builder`)
3. ... (ie `user`)
4. rewrite all the helpers that do not use any other helpers to ES7 (ie `wait`)


philschatz:
closing in favor of refactoring the existing tests first to use ES7 and `await`.

PR #961, opened 2016-01-29
Selenium: continue helper-ish work

pandafulmanda:
we're here to helper!

hamburger helper that is

moar of #931, continuation of #953

### DIFF
--- a/test-integration/helpers/course-select.coffee
+++ b/test-integration/helpers/course-select.coffee
@@ -1,17 +1,38 @@
 selenium = require 'selenium-webdriver'
 {TestHelper} = require './test-element'
 
+
+COMMON_ELEMENTS =
+  courseLink:
+    locator: (appearance, isCoach = false) ->
+      dataAttr = 'data-appearance'
+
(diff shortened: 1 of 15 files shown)


test-integration/helpers/course-select.coffee
```diff
@@ -1,17 +1,37 @@
...
+
+COMMON_ELEMENTS =
+  courseLink: (appearance, isCoach = false) ->
+    dataAttr = 'data-appearance'
+
+    if appearance?
+      dataAttr += "='#{appearance}'"
+
+    if isCoach
+      teacherLink = "[#{dataAttr}] > [href*='cc-dashboard']"
```
philschatz: Could we add classes to the elements so it is easier to find in the DOM? It seems like they'd be less fragile too

pandafulmanda: Okay, we can look at this for a follow-up PR.  This pr has quite a few updates to the base `TestHelper` that would be useful in other PRs

PR #962, opened 2016-01-29
Concept Coach Dashboard Period Ordering

jivewise:
Before: 
screen shot 2016-01-30 at 12 58 14 am

After:
screen shot 2016-01-28 at 12 34 12 am

I think this PR is pretty urgent, sometimes, the dashboard will be showing the wrong data based on which period is selected initially.  See the hansolo prod user for the bug.


philschatz:
Thanks!

PR #963, opened 2016-02-01
Fix infinite continue button on an intro-only assignment

nathanstitt:
The actual bug is that if `steps.length is 1` then 0 is always returned, which causes `setStepKey` in [task/index](https://github.com/openstax/tutor-js/blob/master/src/components/task/index.cjsx#L37) to always display the first step.

After looking at the method closer it appears to me that most of it wasn't needed though.

`getCurrentStepIndex` method (at top of file) returns the index of the first non-completed step, or -1 if not found. It seems to me that that's exactly what we want to return here.  I've tested this and it seems to be correct.

Note that breadcrumbs are not displayed on a "intro only assignment" since it has only a single step. That's disabled by the check to `TaskStore.isSingleStepped(id)` in [task/index.cjsx](https://github.com/openstax/tutor-js/blob/master/src/components/task/index.cjsx#L225-L230)  not sure if that should be re-enabled?

@pandafulmanda thoughts?

### DIFF
--- a/src/components/task/index.cjsx
+++ b/src/components/task/index.cjsx
@@ -222,12 +222,11 @@ module.exports = React.createClass
     taskClasses += " task-#{panelType}" if panelType?
     taskClasses += ' task-completed' if TaskStore.isTaskCompleted(id)
 
-    unless TaskStore.isSingleStepped(id)
-      breadcrumbs = <Breadcrumbs
-        id={id}
-        goToStep={@goToStep}
-        currentStep={@state.currentStep}
-        key="task-#{id}-breadcrumbs"/>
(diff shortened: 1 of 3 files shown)


pandafulmanda:
This fix makes sense and I've tested in a number of cases and it seems to work! :smile:   I do think we should let the breadcrumbs show for even single stepped tasks, which are now two-stepped, by taking out the `TaskStore.isSingleStepped(id)` check.

philschatz:
is there a way to add a unit test for this?

nathanstitt:
Added some specs.  Took me a bit to figure out how to clear the store so it has only one record without trampling the import for the other specs ;)

PR #964, opened 2016-02-02
Selenium: make and use reference book test helper

pandafulmanda:
includes commit from #965, related to: #931


philschatz:
Looks good to me!

/cc @openstax/test

Issue #951, opened 2016-01-22
How best to write selenium helpers


nathanstitt:
Writing this as an issue vs a napkin note so so people can comment.

We can write selenium helpers as objects, or as a collection of structured functions.

An simplified version of an object oriented helper would be something like:

```coffee

class ProgressGuide extends PanelElement
  constructor: (test) ->
    @element = test.driver.findElement(css: '.student-dashbaord')

  getTitle: ->
    @element.findElement(css: '.title').then (el) ->
      el.getText()

class StudentDashboard
  constructor: (test) ->
    @element = test.driver.findElement(css: '.student-dashbaord')

  getProgressGuide: ->
    new Promise (resolve, reject) =>
      @element.findElement(css: '.progress-guide').then (el) =>
        resolve(new ProgressGuide(@test, el))

module.exports = StudentDashboard
```

To use this in a spec you could:

```coffee
StudentDashboard = require 'helpers/student-dashboard'

@it 'has the proper title', ->
    dash = new StudentDashboard(@)
    dash.getProgressGuide().then (guide) ->
      guide.getTitle().then (title) ->
        expect(title).to.equal('Progress Guide')
```

As you can see, the object oriented helper yields different things from it's promises. `StudentDashboard.getProgressGuide()` yields an instance of a `ProgressGuide` object, while the promise returned by `getTitle` yields a string.

Phil suggested we could employ a different method that might be less confusing.   In his suggestion, each getter method would return a selenium element object. That way there would be no confusion over what a helper would yield.

Code written in that manner would look more like:

```coffee

getProgressGuide = (test) ->
  test.driver.findElement(css: '.student-dashbaord')

ProgressGuide =
  getTitle: (el) ->
    @element.findElement(css: '.title')



module.exports = {getProgressGuide, ProgressGuide}

```

and could be used as:

```coffee
StudentDashboard = require 'helpers/student-dashboard'

@it 'has the proper title', ->
  StudentDashboard.getProgressGuide(@).then (guideEl) ->
    StudentDashboard.ProgressGuide.getTitle(guideEl).then (title) ->
      title.getText().then (title) -> # <- title is an Element, getText is selenium
        expect(title).to.equal('Progress Guide')
```

Thoughts on the two approaches?  @philschatz if I've misrepresented your approach please correct me

refs #931

pandafulmanda:
i think the object oriented version means that when we use that to write out tests, it follows the default selenium paradigm.  I am fiddling with something like this:

```
[...]

PR #965, opened 2016-02-02
Selenium: test helper docs

pandafulmanda:
some [docs](https://github.com/openstax/tutor-js/tree/update/test-helper-alias-rename/test-integration/helpers) and [auto-generated docs](http://openstax.github.io/tutor-js/docs/) here

related to #931 and #951 

@openstax/tutor-fe, @openstax/test -- [preview docs here](https://github.com/openstax/tutor-js/tree/update/test-helper-alias-rename/test-integration/helpers) and please let me know your thoughts!

Basically done i think? but awaiting some decisions from #1000  and some updates will need to be made after #992 is merged


test-integration/helpers/README.md
```diff
@@ -0,0 +1,246 @@
...
+    @courseSelect = new CourseSelect(@)
+
+    @user.login(TEACHER_USERNAME)
+
+  @it 'Shows stats for all published plans (readonly)', ->
+    COURSE_CATEGORY = 'BIOLOGY'
+    @courseSelect.goTo(COURSE_CATEGORY)
+    @calendar.waitUntilLoaded()
+
+    @calendar.el.publishedPlan.forEach (plan, index, total) =>
```
philschatz: It seems that for writing the integration tests in testrails `.forEach` is not really used.

Could it be moved into an "advanced" section?

test-integration/helpers/README.md
```diff
@@ -0,0 +1,246 @@
...
+    @user = new User(@)
+    @calendar = new CalendarHelper(@)
+    @courseSelect = new CourseSelect(@)
+
+    @user.login(TEACHER_USERNAME)
+
+  @it 'Shows stats for all published plans (readonly)', ->
+    COURSE_CATEGORY = 'BIOLOGY'
+    @courseSelect.goTo(COURSE_CATEGORY)
+    @calendar.waitUntilLoaded()
```
philschatz: Any reason `.goTo` does not already wait until loaded? It seems like adding `.waitUntilLoaded()` everywhere might get forgotten?

test-integration/helpers/README.md
```diff
@@ -0,0 +1,246 @@
...
+    @user.goHome()
+```
+
+# Helper Details
+
+From doing the above, if we instantiate CalendarHelper, we get back something that looks like this:
+
+```coffee
+calendarHelper =
+  test: testContext
```
philschatz: What about keeping the `README` to be examples of how to use the helpers and then editing them could be in a `./advanced.md` file?

For example, rather than describe the object this way (which is not easily copy/pasteable & exposes some implementation details), what about something like:

```js
calendar = new helpers.Calendar(this)

calendar.goForecast()
calendar.goStudentScores()
calendar.goPlanByTitle(title)
// This function will block all selenium based promises until
// the loading element is absent before continuing.
calendar.waitUntilLoaded(optionalWaitTime)

// These are elements you can use to test things about the calendar page.
// They are defined by the COMMON_ELEMENTS section in `helpers/calendar.coffee`
calendar.el.forecastLink      // <TestItemHelper>
calendar.el.studentScoresLink // <TestItemHelper>
calendar.el.addToggle         // <TestItemHelper>
```

I think the rest of the items might be too-overwhelming to get an idea of how to use the helpers (maybe a "For creating/editing helpers see `[./advanced.md](./advanced.md)`" link?)

```js
// "The rest of the items":

calendar.test // testContext
calendar.options // options set when the calendar was instantiated

// uses the loadingLocator option
calendar.el.loadingState      // <TestItemHelper>
// from explicitly calling setCommonHelper
calendar.el.planPopup         // <PlanPopupHelper>

// The following are used to attach the helpers that are now on CalendarHelper.el
// They are exposed here for customizing your helper as needed,
// as in this case with the planPopup.
// Usually, you can just rely on the commonElements object.
calendar.setCommonHelper(name, helper)
calendar.setCommonElement(locator, name)
```

philschatz:
Great job documenting how the helpers are structured! I was wondering if the `README.md` might be trying to satisfy 2 audiences:

- using the helpers
- writing helpers

Would it make sense to split the docs up (`README.md` for how to use and a separate file for how to create/edit one of these)?

pandafulmanda:
yes, that makes a lot of sense.  I was struggling with balancing the amount of info vs. what is useful to who. will do that soon

philschatz:
What is this blocked on?

pandafulmanda:
blocked because i think #1003, and #1000 should get merged first, and then I can make edits to this based on the results from those as needed.

philschatz:
Now that #1000 is merged I think this is unblocked, right?

PR #966, opened 2016-02-02
Scores export popup options for CC

darrenmason:
regarding new modal in: http://3lwsnp.axshare.com/#p=detailed_scores
(click on export)

- [x] add CC modal dialog for export
- [ ] build form
- [ ] send option parameters to export action


darrenmason:
this branch isn't necessary anymore since we are not doing export dialog

PR #967, opened 2016-02-03
Remember Free Response

jivewise:
Adding code to remember free response when user switches to a different exercises.  Also alerts the user if they have unsaved work.

This is for: https://www.pivotaltracker.com/story/show/108888036


jivewise:
Depends on: https://github.com/openstax/react-components/pull/41

pandafulmanda:
ooh cool!

PR #968, opened 2016-02-03
silent 404 for media preview links

pandafulmanda:
* silent 404 for media preview links
* make media selectors more picky
* pass through original href if needed

[ticket](https://www.pivotaltracker.com/story/show/113027561), worked with @inconceivablevizzini

screen shot 2016-02-03 at 5 45 44 pm

look, no 404s!

### DIFF
--- a/src/flux/media.coffee
+++ b/src/flux/media.coffee
@@ -3,7 +3,7 @@ htmlparser = require 'htmlparser2'
 {makeSimpleStore} = require './helpers'
 
 LINKS_BEGIN = ['#']
-LINKS_CONTAIN = ['/contents/', '.cnx.org']
+LINKS_CONTAIN = ['cnx.org/contents/']
 
 MEDIA_LINK_EXCLUDES = [
   '.nav'
--- a/src/flux/media.coffee
+++ b/src/flux/media.coffee
@@ -2,6 +2,26 @@ _ = require 'underscore'
 htmlparser = require 'htmlparser2'
 {makeSimpleStore} = require './helpers'
 
+LINKS_BEGIN = ['#']
+LINKS_CONTAIN = ['/contents/', '.cnx.org']
+
+MEDIA_LINK_EXCLUDES = [
+  '.nav'
+  '.view-reference-guide'
+  '[data-type=footnote-number]'
+  '[data-type=footnote-ref]'
+  '[data-targeted=media]'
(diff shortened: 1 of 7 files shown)


philschatz:
looks like an AssertionError in Travis (expected a link to exist) :cry:

philschatz:
Why not just not show the preview for external links (`/^https?\:/`)? It seems like silencing ones that _should_ work obscures the fact that we have a problem, no?

pandafulmanda:
well, it shouldn't throw a 404 that interrupts the user's flow, and the attempt to get the media from the BE will still be logged through BE, correct?

src/flux/media.coffee
```diff
@@ -2,6 +2,26 @@ _ = require 'underscore'
 htmlparser = require 'htmlparser2'
 {makeSimpleStore} = require './helpers'
 
+LINKS_BEGIN = ['#']
+LINKS_CONTAIN = ['/contents/', '.cnx.org']
```
philschatz: Any reason this isn't `['.cnx.org/contents/' ]` ? That way it doesn't try to fetch `http://www.cnx.org/search` or `http://jsmith.org/blog/contents/2016-01/`

philschatz: Also, what happens if the link is `http://cnx.org/contents/[UUID]`?

- this returns blank HTML that contains a `<script>` tag to load webview
- Not sure if browsers will like `http:` ajax requests from a `https:` site

pandafulmanda: I was thinking 'cnx.org/contents/' would be good enough...i was trying to be more flexible and less strict...but yeah i guess there's no real reason to allow those

we don't try to get it from cnx directly, it tries to get if from the tutor-be

philschatz: Huh RE "tries to get if from the tutor-be"? Do you mean the JS tries to get stuff from tutor-be unless it starts with `#` or `https?:` ? Or does JS parse the string `"http://cnx.org/contents/[UUID]"`, pull out `[UUID]`, and ask `/api/books/#/pages/[UUID]`?

pandafulmanda: JS parses the string `"http://cnx.org/contents/[UUID]"`, pull out `[UUID]`, and asks `/api/books/#/pages/[UUID]`

PR #969, opened 2016-02-07
Selenium: fix missing CSS file when building locally to run selenium

philschatz:
This was introduced in #916 . If `MINIFY=false npm run build` is run (for testing a local instance in Selenium) the `tutor.css` file was being saved as `tutor.min.css` and so it was missing.


pandafulmanda:
:+1:

PR #970, opened 2016-02-07
Selenium: add/fix features

philschatz:
- adds documentation for running Selenium against a remote server
- converts the `calendar` tests to be in ES7 instead of coffee
- creates `./test-integration/exhaustive` for the exhaustive `calendar` tests and the `reference-book` tests
- adds `VERBOSE=true` logging for easier debugging when a test times out so you know what it was doing right before the test failed
  - adds `utils.verboseWrap` for wrapping a set of async operations so you can see in the log when they start and when they finish. These messages are added to the controlFlow so they are outputted when they actually run
  - adds `utils.verbose` for logging instead of `console.log` which prints out immediately instead of when Selenium's `controlFlow` gets to the command
- pulls out `utils.wait.giveTime` which adjusts the test timeout (and gives back time when it succeeds  so commands other than `utils.wait.for` can use it
- changed the `task` tests to only run on 1 course for speed (but I guess we could add them back in)
- experimented with wrapping `selenium.WebElement::click` to log but ran into some snags (uncomment the code in `verbose.coffee` to see

refs #931

# TODO

- [ ] update the README once the helper docs (#965) are in

# Verbose logging example

Syntax:

```js
utils.verboseWrap('Go back home', () => user.goHome())
utils.verbose('Looking at HW', index, 'of', total)
```

Example output (a step towards getting _useful_ logging to see what is going on):
```
START: HELPER: edit
- START: Waiting until Loadable .is-loading is gone
- END  : Waiting until Loadable .is-loading is gone
- START: HELPER: setName
- - START: Waiting for {"css":"#reading-title"}
- - END  : Waiting for {"css":"#reading-title"}
- END  : HELPER: setName
- START: HELPER: setDate
- - START: HELPER: _setDate
- - - START: HELPER: openDatePicker
- - - - START: HELPER: _locator
- - - - END  : HELPER: _locator
- - - - START: Waiting for {"css":".-assignment-due-date .datepicker__input"}
```


philschatz:
The test files are now back to `.coffee` instead of `.js`. Anything else?

nathanstitt:
Looks good to me.  Like the wrapper stuff, hopefully it'll help us diagnose timeouts

test-integration/helpers/utils/verbose.coffee
```diff
@@ -0,0 +1,88 @@
...
+  else
+    _verboseEnd(test, msg)
+
+  promise
+
+module.exports = {verbose, verboseWrap, resetIndentationLevel}
+
+
+
+# selenium = require 'selenium-webdriver'
```
philschatz: @nathanstitt This was a wrapper I thought of including.

It wraps selenium's `.click` method directly but it's a little janky because in order to be useful you have to run `driver.findElement(locator)` which squirrels away the locator used to find the element.

This doesn't always happen so (in existing tests) you frequently get a `Clicking undefined` which is not useful. By not including this I think it encourages folks to use `Helper.els...click()` which does log and discourages writing tests that directly use CSS selectors.

Thoughts?

nathanstitt: I'm not 100% sure I understand the proposed wrapper, but I agree it sounds scary to wrap selenium's existing `.click()` method.  But I have encounter the `Clicking undefined`, and also `Unable to wait on undefined` (from passing an undefined locator to wait).  I think those are relatively easy to debug though.

PR #972, opened 2016-02-09
Selenium: replace .get().click() with just .click()

philschatz:
When paired with #970 this provides a bunch of extra logging.

The issue I was initially diagnosing was why publishing an assignment sometimes worked and sometimes did not.

The problem was that when you click an item in the calendar a popup may or may not appear depending on the state of the `task-plan`. The test was waiting for a button to appear but it never appeared.

Thoughts on this approach @openstax/tutor-fe ?

Atom search/replace string used: `\.get\((.*)\)\.click\(\)` and `.click($1)`

refs #931


gregfitch:
None...it improves readability.

PR #974, opened 2016-02-10
media preview sizing

pandafulmanda:
Image `max-width` was not working in [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=975632).  See #977 for full details.

This was causing two problems:
* image was floating outside of the width of popover sometimes or made the popover overflow unnecessarily
* popover would flash in and out because the `mouseEnter` and `mouseLeave` would get called repeated as the popover would cover the link where the mouse was hovering
Additionally, Maths was looking weird.  See [this ticket](https://www.pivotaltracker.com/story/show/112930795) for additional details.

This PR addresses the above issues by:
* making `span` styling more specific within the popover
* manually detecting if image's `width` should be adjusted to fit
* checking if mouse has left both the link and the popover on `mouseLeave`

# Before

## image width messed up in Firefox
screen shot 2016-01-29 at 3 15 33 pm

## Maths is wrong
screen shot 2016-01-29 at 3 13 46 pm

# After

## Image width fixed in Firefox
screen shot 2016-02-09 at 1 29 20 pm

## Maths fixed
screen shot 2016-02-09 at 7 00 26 pm

### DIFF
--- a/src/components/tutor-popover.cjsx
+++ b/src/components/tutor-popover.cjsx
@@ -58,6 +58,7 @@ TutorPopover = React.createClass
   setMaxImageWidth: (image, maxWidth) ->
     # unfortunately, css max-width does not work in firefox inside display: table elements
     # https://bugzilla.mozilla.org/show_bug.cgi?id=975632
+    # See https://github.com/openstax/tutor-js/issues/977 for full details.
     return unless image?
     {width} = image.getBoundingClientRect?()
     image.width = maxWidth if width > maxWidth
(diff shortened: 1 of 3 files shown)


src/components/tutor-popover.cjsx
```diff
@@ -52,6 +55,14 @@ TutorPopover = React.createClass
 
     @setState(imagesLoading: currentImageStatus)
 
+  setMaxImageWidth: (image, maxWidth) ->
+    # unfortunately, css max-width does not work in firefox inside display: table elements
+    # https://bugzilla.mozilla.org/show_bug.cgi?id=975632
+    # See https://github.com/openstax/tutor-js/issues/977 for full details.
```
philschatz: Thanks for linking to the issue in the code!

philschatz:
Hmm, might be unrelated but when loading a link to an external figure the "Loading..." text seems to remain on top of the image

image

philschatz:
The previous comment is not a regression but I think this fixes #977 so merging :smile:

PR #979, opened 2016-02-10
validate the versions of packages in package.json

philschatz:
Validates that we are using exact versions when including 3rd-party npm packages in the `package.json` file.

I was thinking of adding it to `gulp` but I'm hoping we can finally get rid of gulp. Maybe this should be ported to the other repos? (if merged)

Thoughts @openstax/tutor-fe ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,8 +4,9 @@
   "description": "Client for OpenStax Tutor",
   "main": "index.js",
   "scripts": {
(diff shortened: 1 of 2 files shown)


nathanstitt:
:+1: I first started to suggest that we verify the git sha's, but then realized that Travis will do that for us later when it attempts to install the packages.

PR #980, opened 2016-02-12
Student Scores: adds tooltips, ellipsis and scale/wrap to tabs

darrenmason:
tab-truncate

regarding task: https://www.pivotaltracker.com/story/show/112365737

added tooltips, ellipsis and scale/wrap to tabs to allow for longer names and more periods

### DIFF
--- a/resources/styles/mixins.less
+++ b/resources/styles/mixins.less
@@ -88,6 +88,7 @@
         display: block;
         white-space: nowrap;
         font-size: 1.3rem;
+        max-width: 100px;
       }
     }
 
--- a/resources/styles/mixins.less
+++ b/resources/styles/mixins.less
@@ -71,6 +71,8 @@
 
 .tutor-nav-tabs() {
   background: none;
+  display: flex;
+  flex-wrap: wrap;
   li {
     text-align: center;
     background: @tutor-neutral-lighter;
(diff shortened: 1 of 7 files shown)


philschatz:
Is the tab width (and ellipses) not possible using just CSS?

philschatz:
For example, something like:

```css
.period-tab {
  display: inline-block;
  max-width: 100px;
  overflow-x: hidden;
  text-overflow: ellipses;

  white-space: nowrap;
}
```


Otherwise, can you add tests?

resources/styles/mixins.less
```diff
@@ -80,6 +82,14 @@
     a {
       font-weight: 400;
       color: @tutor-neutral !important;
+      span {
```
philschatz: Could you add a class to this element so it is not selecting `span`? Maybe `react-bootstrap` does not have an easy way to do this though

resources/styles/mixins.less
```diff
@@ -71,7 +71,9 @@
 
 .tutor-nav-tabs() {
   background: none;
-  li {
+  display: flex;
+  flex-wrap: wrap;
+  li.tab-item-period {
     text-align: center;
```
nathanstitt: tutor-nav-tabs are used on other things than periods. ;)

PR #981, opened 2016-02-12
Selenium: fix tests to run remotely and more consistently

philschatz:
When trying to run the tests against `dev` I ran into some issues. This is what I changed to get them working (for me at least). refs #931

Follow-up to #970

### DIFF
--- a/test-integration/helpers/reading-builder.coffee
+++ b/test-integration/helpers/reading-builder.coffee
@@ -32,7 +32,7 @@ COMMON_ELEMENTS =
   selectReadingsButton:
     css: '#reading-select'
   addReadingsButton:
-    css: '.dialog:not(.hide) .-show-problems'
+    css: '.dialog:not(.hide) .-show-problems:not([disabled])'
   disabledAddReadingsButton:
     css: '.dialog:not(.hide) .-show-problems[disabled]'
 
@@ -205,7 +205,7 @@ class ReadingBuilder extends TestHelper
     # Wait up to 3min for publish to complete
     @test.utils.wait.giveTime (3 * 60 * 1000), =>
       @test.utils.verboseWrap 'Publishing', =>
-        @el.publishButton.click()
+        @el.publishButton.waitClick()
         # wait for
         @test.driver.wait =>
           @el.pendingPublishButton.isPresent().then (isPresent) -> not isPresent
@@ -215,7 +215,7 @@ class ReadingBuilder extends TestHelper
       (new Calendar.CalendarHelper(@test)).waitUntilPublishingFinishedByTitle(name)
 
   save: =>
-    @el.saveButton.click()
+    @el.saveButton.waitClick()
 
   cancel: =>
     # BUG: "X" close button behaves differently than the footer close button
@@ -228,7 +228,7 @@ class ReadingBuilder extends TestHelper
   delete: =>
     # Wait up to 2min for delete to complete
     @test.utils.wait.giveTime (2 * 60 * 1000), =>
-      @el.deleteButton.click()
+      @el.deleteButton.waitClick()
       # Accept the browser confirm dialog
       @test.driver.wait(selenium.until.alertIsPresent()).then (alert) ->
         alert.accept()
@@ -253,7 +253,7 @@ class ReadingBuilder extends TestHelper
     # hangs around for quite awhile.  Bump the wait time up to 4 seconds to work around
     @waitUntilLoaded(4000)
     @setName(name) if name
-    @setDate({opensAt, dueAt})
+    @setDate({opensAt, dueAt}) if opensAt or dueAt
 
     if sections
       # Open the chapter list by clicking the button and waiting for the list to load
@@ -269,12 +269,12 @@ class ReadingBuilder extends TestHelper
 
       if verifyAddReadingsDisabled
         # Verify "Add Readings" is disabled and click Cancel
-        @el.disabledAddReadingsButton.get().isDisplayed().then (isDisplayed) =>
+        @el.disabledAddReadingsButton.isPresent().then (isDisplayed) =>
           @cancel() if isDisplayed
       else
         # Click "Add Readings"
-        @test.sleep(100) # Not sure why this is needed
-        @el.addReadingsButton.click()
+        @test.sleep(500) # Not sure why this is needed
+        @el.addReadingsButton.waitClick()
 
     switch action
       when 'PUBLISH' then @publish(name)
--- a/test-integration/helpers/reading-builder.coffee
+++ b/test-integration/helpers/reading-builder.coffee
@@ -202,20 +202,23 @@ class ReadingBuilder extends TestHelper
     @el.requiredItemNotice.get(type).isDisplayed()
 
   publish: (name) =>
-    # Wait up to 3min for publish to complete
-    @test.utils.wait.giveTime (3 * 60 * 1000), =>
-      @test.utils.verboseWrap 'Publishing', =>
-        @el.publishButton.waitClick()
-        # wait for
-        @test.driver.wait =>
-          @el.pendingPublishButton.isPresent().then (isPresent) -> not isPresent
+    @el.publishButton.waitClick()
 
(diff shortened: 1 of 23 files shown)


test-integration/student/dashboard.js
```diff
@@ -62,7 +62,7 @@ describe('Student Dashboard', function(){
     expect(unstartedEvents).not.to.be.empty;
 
     // that's all folks
-    this.student.logout();
+    // this.student.logout();
```
nathanstitt: Should test be moved to the 'long tests' section instead of being `xit`. Shouldn't we still logout the user?

test-integration/scores.coffee
```diff
@@ -10,17 +10,17 @@ TEACHER_USERNAME = 'teacher01'
...
     @courseSelect = new CourseSelect(@)
     @user.login(TEACHER_USERNAME)
-    @courseSelect.goToCourseByName('Physics I')
+    @courseSelect.goTo('PHYSICS')
     @calendar.goStudentScores()
 
-  afterEach ->
-    @user.goHome()
+  # afterEach ->
+  #   @user.goHome()
```
nathanstitt: I like the generic `goTo` vs the `goToCourseByName`, should these stray comments be removed?

test-integration/helpers/user.coffee
```diff
@@ -60,6 +63,13 @@ class User extends TestHelper
         @logInLocal(username)
       else
         @logInDeployed(username, password)
+    # Inject a little CSS to unfix the pesky fixed navbar
+    .then =>
+      @test.driver.executeScript ->
+        hider = document.createElement('style')
+        hider.textContent = '.navbar-fixed-bottom, .navbar-fixed-top { position: initial !important; } body { padding: 0; }'
+        document.head.appendChild(hider)
+
```
nathanstitt: With theses styles will we no longer need to `scrollTop` for certain clicks?

test-integration/helpers/reading-builder.coffee
```diff
@@ -269,10 +273,11 @@ class ReadingBuilder extends TestHelper
           @cancel() if isDisplayed
       else
         # Click "Add Readings"
+        @test.sleep(100) # Not sure why this is needed
         @el.addReadingsButton.click()
```
nathanstitt: My guess for why some of the timeouts like this are needed is because the `async-button` will very briefly in the 'waiting' state.  If selenium's `click` is called at that instant then it's ignored.

test-integration/helpers/reading-builder.coffee
```diff
@@ -210,6 +210,10 @@ class ReadingBuilder extends TestHelper
         @test.driver.wait =>
           @el.pendingPublishButton.isPresent().then (isPresent) -> not isPresent
 
+      # Publishing can be async so wait until the assignment shows up as 'published' in the calendar
+      # Added here because folks won't remember to wait for this
+      (new Calendar.CalendarHelper(@test)).waitUntilPublishingFinishedByTitle(name)
+
   save: =>
```
nathanstitt: This is really good - I wrote the earlier `@el.pendingPublishButton.isPresent()` wait, but that'll fail when the publish is performed by a background job (as you discovered).

test-integration/helpers/reading-builder.coffee
philschatz: Oh, so I could maybe call `wait.click()` and have the selector use `:not(.is-waiting)` (or something)?

nathanstitt: Yeah, I *think* that works for some of them.  It's hard to debug though since the timeouts are pretty transient.  A 100ms timeout also works and is pretty painless.

test-integration/student/dashboard.js
philschatz: Ah, I `xit`'d this because the script times out when students have like a ton of assignments because of the nested selenium calls. Was going to see if you knew of a _quicker_ way to write the `findVisibleEvents`

test-integration/scores.coffee
philschatz: actually, I guess `goHome` is fine; This is probably an left-over artifact of trying to get this test to work (I'll uncomment)

test-integration/helpers/user.coffee
philschatz: yep, I can remove the `scrollTop` & see what happens. This probably needs to also include styling for other fixed elements on the page (like the bottom bar when doing an assignment)

test-integration/helpers/reading-builder.coffee
philschatz: My pref would be to add timeouts _very sparingly_ precisely because it introduces transient errors, no?

test-integration/student/dashboard.js
```diff
@@ -17,7 +17,7 @@ describe('Student Dashboard', function(){
     expect(this.dash.el.thisWeek).not.to.not.be.null;
   });
 
-  this.it('displays tasks that are assigned', async function(){
+  this.xit('displays tasks that are assigned', async function(){
```
philschatz: Ah, I `xit`'d this because the script times out when students have like a ton of assignments because of the nested selenium calls. Was going to see if @nathanstitt had suggestions on a quicker way to write the `findVisibleEvents`.

test-integration/helpers/utils/window-position.coffee
```diff
@@ -4,18 +4,22 @@ SIZE_PRESETS =
 windowPosition = (test) ->
 
   scrollTop: ->
-    # @driver.executeScript("arguments[0].scrollIntoView(true);", el)
-    test.driver.executeScript("window.scrollTo(0,0);")
-    test.sleep(200)
+    # test.utils.verbose('Scrolling to top')
+    # # @driver.executeScript("arguments[0].scrollIntoView(true);", el)
+    # test.driver.executeScript("window.scrollTo(0,0);")
+    # test.sleep(200)
```
nathanstitt: Can `scrollTop` be removed now since it's code's dead?

test-integration/helpers/test-element.coffee
```diff
@@ -48,26 +51,37 @@ class TestItemHelper
...
-    get: -> @_test
-  locator:
-    get: -> @_locator
-  element:
-    get: -> @get()
+  waitClick: (args...) =>
+    locator = @getLocator(args...)
+    @test.utils.verbose "Waiting for #{JSON.stringify(locator)}"
+    @test.utils.wait.for(locator)
+    @click(args...)
```
pandafulmanda: @philschatz since `@get` calls `@test.utils.wait.for` underneath, is `waitClick` then trying to wait for the element twice?  Are `click` and `waitClick` actually different?  Could we just add the logging to [`wait`](https://github.com/openstax/tutor-js/blob/master/test-integration/helpers/utils/wait.coffee#L33-L38)'s `click` and make one alias here like this?

```coffee
  waitClick: (args...) =>
    locator = @getLocator(args...)
    @test.utils.wait.click(locator)
```

philschatz: Yeah, I think they might be the same. I was wondering if the semantics for `click` should be that it does not wait. That way, if you are expecting an element to be present, Selenium will error immediately instead of waiting for the element to "eventually appear" and then time out. Also, that would help ensure that the test waited for the page to be loaded first. Thoughts?

pandafulmanda: hmm...is that case common enough? seems like it would be just as quick to do
```coffee
blahBlah.el.thing().findElement().click()
```
but i guess it would nice to have an immediate one to match the `waitClick`.

philschatz: I guess I figured most of the lines in tests would be `el.click()`, and `el.waitClick()` would be less common, but maybe still useful? Something like:

```coffee
{someButton, saveButton, anotherButton} = foo.el
calendar.goToFoo()  # which implicitly calls `waitUntilLoaded()` because people forget ; )
someButton.click()
anotherButton.click()  # With these being the common thing that happens
saveButton.waitClick()
```

pandafulmanda: okay, ill make adjustments in #1000.  noticed it as i was updating the docs

PR #983, opened 2016-02-14
Refactor how selenium TestElement elements are called

philschatz:
Changes the syntax to `courseSelect.el.courseByType('physics').click()` instead of `courseSelect.el.courseByType.click('physics')`

Keeps a deprecation warning in _just in case_.

Thoughts @openstax/tutor-fe @openstax/test ? refs #931

# This PR is not to master. refs #981

PR #985, opened 2016-02-15
Selenium use recordo for failed test

philschatz:
In addition to taking a screenshot for failed tests, this saves the `recordo` log (mostly to quickly get the server JSON) when `VERBOSE=true` environment var is set.

It uses the `window.__Recordo` global (if available) and has commented code to inject `recordo` instead. It works, but might need some refactoring to not include the UI.

Thoughts @openstax/tutor-fe?

# This PR is not to master . refs #983

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@
 /bower_components
 /.tmp
 /dist
-/test-failed-*.png
+/test-failed-*
 /coverage
 
 *.cjsx.js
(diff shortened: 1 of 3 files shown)


philschatz:
closing since it's probably not terribly useful to have recordo logs as well as the selenium logs. Also, the structure of selenium tests has changed a bit

PR #986, opened 2016-02-15
make crumbs not show for external tasks with only one step

pandafulmanda:
### Fixed for external
screen shot 2016-02-15 at 10 49 16 am

### Crumbs still correct for readings with only one step
screen shot 2016-02-15 at 10 49 03 am

Regression for external crumbs came from #963, this addresses [this ticket](https://www.pivotaltracker.com/story/show/113636529)


nathanstitt:
Hah, we (I) missed that external assignments do legitimately only have a single step when remove the check as part of https://github.com/openstax/tutor-js/pull/963

PR #989, opened 2016-02-16
Fix/teacher review coach

pandafulmanda:
See [ticket](https://www.pivotaltracker.com/story/show/110131688) for full details

### The review for coach functions like the one in actual coach
screen shot 2016-02-18 at 12 46 31 pm


philschatz:
Thoughts @jason-holmes or @Fredasaurus ?

pandafulmanda:
@jason-holmes @Fredasaurus -- the alternative would be to have the teacher's view of the students work link to a view actually inside webview, on coach.

philschatz:
Do the "Return to 1.1" or "Continue to" links do anything? If not, I'm wondering how important it is to make the view look _exactly_ like the student view

pandafulmanda:
They are disabled here so they don't do anything

pandafulmanda:
spoke with @jason-holmes and he agreed with you that the links are extraneous so I've made an update

PR #990, opened 2016-02-16
QA view: show all tags, two step filter, and question id

nathanstitt:
Needs to be merged & deployed along with BE PR: https://github.com/openstax/tutor-server/pull/986 

Notice that I'm displaying a pseudo text entry box (which is really just a styled div).  When I just hid the questions it wasn't easy to tell the 2-step exercises apart from one with a very short question.

image

### DIFF
--- a/api/ecosystems.json
+++ b/api/ecosystems.json
@@ -1,0 +1,1 @@
+[{"id":3,"books":[{"id":3,"uuid":"334f8b61-30eb-4475-8e05-5260a4866b4b","title":"Physics","version":"7.21"}]},{"id":2,"books":[{"id":2,"uuid":"f10533ca-f803-490d-b935-88899941197f","title":"Mini CC Biology Tes Coll","version":"3.1"}]},{"id":1,"books":[{"id":1,"uuid":"d52e93f4-8653-4273-86da-3850001c0786","title":"Biology For AP® Courses","version":"9.1"}]}]
--- a/api/exercises.json
+++ b/api/exercises.json
@@ -2,6 +2,7 @@
     "total_count": 5,
     "items": [{
         "id": "1616",
+        "pool_types": ["reading_dynamic", "practice_widget", "all_exercises"],
         "url": "https://exercises-dev.openstax.org/exercises/1663@1",
         "content": {
             "uid": "1663@1",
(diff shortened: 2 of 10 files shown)


philschatz:
Is there a way we could reuse the logic for when to display free-response already in the exercise widget?

nathanstitt:
Heh, it's actually not a textarea.  The exercise widget doesn't actually display free response, since it's just used on the teacher's HW builder.  Rather than extend the it to display a textarea that'll never actually be used, I just styled the div as to appear like one when the questions are hidden and it's displayed on QA view.

philschatz:
Ah, `@kathi-fletcher` and @kevinburleigh75 pointed out that the logic for determining when an exercise gets a free response is not trivial ( @pandafulmanda had to jump through some hoops to get it right) and they were suggesting that this view jump through the same hoops to not accidentally show @openstax/test the wrong thing

nathanstitt:
I'm not sure what you're referring to when you say it's not trivial.  I spoke with @pandafulmanda and @Dantemss to determine how to figure out if an exercise has a free-response or not.  They both indicated that I should check the `formats` array for the presence of `free-response`.  I added a method [hasQuestionWithFormat](https://github.com/openstax/tutor-js/pull/990/files#diff-ce4cb3be778fd8ed5807fb351cf7485aR104) to the flux store that's used for that.

Is that not the correct pattern?  Or did I mis-understand your question?

philschatz:
I think that's probably good enough (although it seems like there might be 2 places that "check the `formats` array for the presence of `free-response`" since I only saw 1 place that calls `hasQuestionWithFormat` (maybe I missed the other place?)

nathanstitt:
I found one other spot that could use the helper method in (TaskTeacherReviewExercise, in `components/task-teacher-review/exercise.cjsx`)  I've converted it to use it.

The student side doesn't examine the exercise for deciding when to display the task step, it looks at the task step.  See the [TaskStep store](https://github.com/openstax/tutor-js/blob/master/src/flux/task-step.coffee#L110-L114) for it's method.

philschatz:
Looks good! Waiting for https://github.com/openstax/tutor-server/pull/986 to be reviewed /cc @jpslav

jpslav:
That PR is merged and ready for y'all to merge when you are ready / it is non-WIP

nathanstitt:
Yep, the BE is ready to go, sorry I forgot to remove the WIP prefix

PR #991, opened 2016-02-16
for reading exercise, if can only continue, button should say continue and not 'Next Question'

pandafulmanda:
Will need an update after openstax/react-components#45 is merged.

See [ticket](https://www.pivotaltracker.com/story/show/113635573) for more details

- [x] add a quick test

exercise

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "mime-types": "2.1.7",
     "moment": "2.10.6",
     "moment-timezone": "0.4.1",
-    "openstax-react-components": "openstax/react-components#e285f711d94dff48094f8da2c0556b58151a3771",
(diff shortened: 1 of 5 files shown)


philschatz:
I was a little confused about the button text saying "Submit" when I didn't change any of the multiple choice options but otherwise looks ok to me.


/cc @openstax/ux

pandafulmanda:
Hmm...I will make the screenshots more clear

philschatz:
I was testing it on dev & noticed if I revisit a question and don't change anything the button still says "Submit" instead of "Continue" which made me wonder "Why say Submit at all instead of just continue"? But I guess it's not clear to the student if they select an option and close their browser that their choice has not been recorded yet.

Issue #987, opened 2016-02-16
Selenium: Suggestion for common names for screen commands


philschatz:
Here's a _suggestion_ for predictable names for commands that can be done on all the screens in tutor. I went through all the screens and tried to list off all the things a person can click. What do folks think?

/cc @openstax/tutor-fe @openstax/test 

# TODO

- [ ] add these to a documentation file

# Teacher

These are largely teacher-specific screens

## Calendar

`assignmentType`: `'HOMEWORK' | 'READING' | 'EXTERNAL' | 'EVENT'`
`relativeDate`: `'TODAY' | 'TOMORROW' | 'FIRST'`

- [x] `goToBook()`
- [x] `goToForecast()`
- [x] `goToScores()`
- [ ] `goToAddByType(assignmentType)`
- [ ] `goToAddByTypeFromNow(assignmentType, relativeDate)`
- [ ] `goToPreviousMonth()`
- [ ] `goToNextMonth()`
- [x] `goToOpenByTitle(title)`
- [ ] `goToEditByTitle(title)`
- [ ] `goToReviewByTitle(title)`

### Popup

- [ ] `selectPeriodByIndex(num)`
- [ ] `selectPeriodByTitle(title)`
- [x] `goToReview()`
- [x] `goToEdit()`

## CC-Dashboard

- [x] `goToScores()`
- [x] `goToBook()`
- [ ] `selectPeriodByIndex(num)`
- [ ] `selectPeriodByTitle(title)`

## Builder

- [x] `edit({settings})`

## Scores

- [ ] `selectDisplayAsNumber()`
- [ ] `selectDisplayAsPercent()`
- [ ] `selectBasedOnPossible()`
- [ ] `selectBasedOnAttempted()`
- [ ] `selectPeriodByIndex(num)`
- [ ] `selectPeriodByTitle(title)`
- [ ] `sortName()`
- [ ] `goToAssignmentByIndexes(row, column)`
- [ ] `downloadExport()`

CC Commands:

- [ ] `sortAssignmentScoreByIndex(index)`
- [ ] `sortAssignmentCompletedByIndex(index)`
- [ ] `sortAssignmentScoreByTitle(title)`

HS Commands:

- [ ] `goToForecastByName(studentName)`
- [ ] `goToReviewByIndex(index)`
- [ ] `goToReviewByTitle(title)`

### Review

[...]

PR #992, opened 2016-02-17
Selenium: clean up helper commands

philschatz:
Addresses naming convention described in #987. refs #931

### DIFF
--- a/test-integration/README.md
+++ b/test-integration/README.md
@@ -75,7 +75,7 @@ So, the test would look something like:
 
 ```coffee
 App.login(TEACHER_USERNAME)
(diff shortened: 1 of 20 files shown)


jivewise:
Oh, why did you want to take that out?

nathanstitt:
It shouldn't be needed now that @philschatz made the earlier PR to lock the navbar to the top.  Previously we had to scroll to the top so that items weren't hidden beneath it.

jivewise:
Gotcha, I remember using it in a previous PR at one point...thanks!

nathanstitt:
Looks good to me

PR #993, opened 2016-02-17
CC Student Scores - progress info hover

darrenmason:
scores-info-tooltip2

dependent on https://github.com/openstax/tutor-server/pull/985
regarding story: https://www.pivotaltracker.com/n/projects/1156756/stories/111067052

### DIFF
--- a/src/components/scores/concept-coach-cell.cjsx
+++ b/src/components/scores/concept-coach-cell.cjsx
@@ -15,9 +15,27 @@ ConceptCoachCell = React.createClass
       Math.round((@props.task.correct_exercise_count / @props.task.exercise_count) * 100)
     tooltip =
       <BS.Tooltip>
-        <div>
-          Date Last Worked: <Time format='MMM. D' date={@props.task.last_worked_at} />
-        </div>
+        <BS.Table>
+          <thead>
+            <tr>
+              <th>Correct</th>
+              <th>Attempted</th>
+              <th>Total possible</th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr>
+              <td>{@props.task.correct_exercise_count}</td>
+              <td>{@props.task.completed_exercise_count}</td>
+              <td>{@props.task.exercise_count}</td>
+            </tr>
+            <tr>
+              <td colSpan="3">
+                Date Last Worked: <Time format='MMM. D' date={@props.task.last_worked_at} />
+              </td>
+            </tr>
+          </tbody>
+        </BS.Table>
       </BS.Tooltip>
 
     <div className="cc-cell">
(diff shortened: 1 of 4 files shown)


src/components/scores/concept-coach-cell.cjsx
```diff
@@ -10,15 +10,49 @@ ConceptCoachCell = React.createClass
 
   mixins: [CellStatusMixin] # prop validation
 
+  getInitialState: ->
+    placement: 'right'
+
+  setPopoverPosition: ->
```
philschatz: @openstax/tutor-fe : Would this be useful in other popovers? I recall places like the roster moving the position over to the left because the links were on the right side of the page and the link hover popups doing something similar

philschatz:
Can you add a selenium test to get the popup to show up?

src/components/scores/concept-coach-cell.cjsx
philschatz: Also, I was only ever able to get the popover to show up to the left of the icon, never to the right. When is this extra code needed (since we would always have student names to the left)?

darrenmason: scores-info-tooltip-flip

The initial solution was for the popover to always be on the left so it would not go off the table when hovering over item near the right edge of the table.  Then spoke with Robert, he wanted the flyover to not cover the other information in that row/column, which led to current solution.

philschatz: I was unable to get the popover the show up to the right of the icon though.

darrenmason: strange, might be a bug, I will look at it.

pandafulmanda: oh...we do have some existing code that tries to set popover position-- [TutorPopover](https://github.com/openstax/tutor-js/blob/master/src/components/tutor-popover.cjsx) component

darrenmason:
will do

src/components/scores/concept-coach-cell.cjsx
philschatz: Any updates to this? (RE reusing `TutorPopover` instead of using `BS.Popover`?)

pandafulmanda: Looking closer, it seems like `TutorPopover` is quite different from what this component needs

philschatz: okies! Do we still need this code to guess whether to appear on the left or the right? It seems to only show up on the left but that seems ok since the student names will always provide enough space to the left of any assignment

darrenmason: I'll take it out and making popover on the left, that is how it was originally.

PR #995, opened 2016-02-18
Selenium: Simplify helper namespace

philschatz:
Sometimes a helper would export:

- just the helper (ie `Calendar`)
- `{Calendar}`
- `{CalendarHelper}`
- `{CalendarHelper, CalendarPopupHelper, verify}`.

This makes it export just the helper and the subhelper (if you need it it can be referred to by `new Calendar.Popup(test)`) and removes the `Helper` part in the class name

### DIFF
--- a/test-integration/README.md
+++ b/test-integration/README.md
@@ -147,11 +147,11 @@ this.waitClick('.button').then(function() { console.log('debugging println'); })
 Example from [./task-plan/draft.coffee](./task-plan/draft.coffee)
 
 ```js
-var helpers = require('../helpers');
-var describe = helpers.describe;
-var CourseSelect = helpers.CourseSelect;
-var Calendar = helpers.Calendar;
-var ReadingBuilder = helpers.ReadingBuilder;
+var Helpers = require('../helpers');
+var describe = Helpers.describe;
+var CourseSelect = Helpers.CourseSelect;
+var Calendar = Helpers.Calendar;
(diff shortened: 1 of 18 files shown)


pandafulmanda:
I tend to like having `Helper` be part of the name because it makes it really clear what is it, and also, I think that exporting an object means flexibility in the future if we want to for example, export a static version, or helper functions from the files, but I don't feel strongly about either thing.  I see how exporting the helper itself will decrease repetitive code

philschatz:
This stemmed from my hacky script to document all the helpers ; )

What about:

```coffee
Helpers = require './helpers'

  before: ->
    @calendar = new Helpers.Calendar(@)
```

instead of:

```coffee
{Calendar} = require './helpers'
{CalendarHelper} = Calendar

  before: ->
    @calendar = new CalendarHelper(@)
```

This PR _does_ add static methods to the classes (ie `Calendar.verify()` but I think that code has bit-rotted). I'm not sure what the flexibility adds (except more places to add inconsistencies)

Also, by doing something like `Calendar.Popup` instead of `@calendar.planPopup` you can generate the documentation :smile:

pandafulmanda:
hmm yeah I think the `new Helpers.Calendar` syntax makes sense

jivewise:
I agree with Amanda, I like the `Helpers.Calendar` syntax.

PR #997, opened 2016-02-22
CC Student Scores - removed basedOn toggle 

darrenmason:
made removal of basedOn toggle permanent instead of temp (commented)


philschatz:
Can you remove the selenium test for basedOn as well?

philschatz:
Thanks!

PR #999, opened 2016-02-24
Selenium: test helper code clean up

pandafulmanda:
companion to #965 

related to #931 and #951


pandafulmanda:
could not see any conflicts testing merge with #965 locally.  made a PR with this against master instead in #1000

Issue #919, opened 2015-12-25
Reduce Technical Debt


philschatz:
# Upgrade Dependencies
- [ ] `react`
  - [ ] [`react-router@1.0`](https://github.com/reactjs/react-router/blob/master/upgrade-guides/v1.0.0.md)
    - we can do dynamic chunks really easily: https://github.com/rackt/react-router/blob/master/docs/guides/advanced/DynamicRouting.md
    - uses `history` package instead of `Navigation`
  - [x] `react-bootstrap`
    - OverlayTrigger: https://github.com/react-bootstrap/react-bootstrap/pull/855
    - trigger='manual' removed
    - NavBrand formatting
    - DropdownMenu, MenuItem update jsx format / props
  - [ ] `fixed-data-table`
  - [ ] `react-calendar`
- [x] standardize selenium helper names
- [x] remove jQuery
- [ ] update `phantomjs` to 2.x
- [ ] replace `underscore` with `lodash`
- [x] upgrade time stuffs -- #1055 
  - `moment`
  - `twix`
  - `react-datepicker`
# Refactor Code
- [ ] use `classnames`
- [ ] remove `bootstrap-material-design`
- [ ] remove lint warnings from cjsx files
- [ ] `moment`
- [ ] increase code coverage
- [ ] replace flux
- [ ] ES6?!?!?!?!??!?!?!

pandafulmanda:
does this mean, `react@0.14` as well?

philschatz:
ah yep; added. Feel free to add more to the list :smile:

philschatz:
Example of upgrading `react`, `react-router`, and `react-bootstrap`: https://github.com/philschatz/gh-board/compare/2b12616...4bcf49

pandafulmanda:
might removing dependency on `jQuery` be something we do as part of this?

nathanstitt:
I agree, would be nice to remove jQuery.  @philschatz I think you mentioned at one time that you know of or had written a plain xhr alternative ?

pandafulmanda:
i liked when we used [axios](https://github.com/mzabriskie/axios) in openstax/concept-coach but that depended on promises :(

pandafulmanda:
would having a style guide be a part of this?

philschatz:
Yep, added `jQuery` to the list! The "plain xhr alternative" is at [octokat.js/src/request.coffee](https://github.com/philschatz/octokat.js/blob/master/src/request.coffee#L12-L43)

PR #1000, opened 2016-02-24
Update/test code cleanups

pandafulmanda:
companion to #965 

related to #931 and #951

oh i got #1000 !!!

will need updates/rebasing after #992 is merged, and updates for `el`s to be called as functions after

trying the new editor in philboard


philschatz:
# :tada: :balloon: :sparkles: #1000 !!!! :sparkles: :balloon: :tada:

package.json
```diff
@@ -40,6 +40,7 @@
     "font-awesome": "4.4.0",
     "htmlparser2": "3.8.3",
     "jquery": "2.1.4",
+    "lodash.curry": "4.0.0",
```
philschatz: Could you just `require 'lodash/function/curry'`?

test-integration/calendar.coffee
```diff
@@ -1,5 +1,6 @@
 Helpers = require './helpers'
 {describe} = Helpers
+selenium = require 'selenium-webdriver'
```
philschatz: Is this needed?

test-integration/helpers/ui/test-element.coffee
```diff
@@ -28,14 +32,28 @@ class TestItemHelper
     else
       @locator
 
-  get: (args...) =>
+  aliasWait: (waitFunction, args...) =>
     locator = @getLocator(args...)
-    @test.utils.wait.for(locator)
+
+    waitTime = _.last(args)
```
philschatz: I'm a bit worried about having too many places to provide timeout data. Why not just use `giveTime` if needed?

test-integration/helpers/utils/dom.coffee
```diff
@@ -0,0 +1,18 @@
+selenium = require 'selenium-webdriver'
+
+class DOM
+  constructor: (test) -> @test = test
+
+  _getParent: (currentElement) ->
```
philschatz: Why not include this on `TestElement`?

test-integration/helpers/utils/wait.coffee
```diff
@@ -14,29 +17,36 @@ class Wait
...
+      el = @for(locator, ms)
+      # Scroll to element so that it's clickable
+      @test.utils.windowPosition.scrollTo(el)
+      @test.utils.verbose "Clicking #{JSON.stringify(locator)}"
+      # el click promise returns from verbose wrapping and can be chained
+      el.click()
 
   # Adjusts the test timeout for a function (which returns a Promise) time to execute
-  giveTime: (ms, fn) ->
+  giveTime: (ms, fn) =>
```
philschatz: Why use the fat arrows syntax? I noticed some files use `->` and some use `=>`.

test-integration/helpers/ui/test-element.coffee
```diff
@@ -94,10 +97,12 @@ class TestHelper extends TestItemHelper
       defaultWaitTime: 40 * 1000 # TODO: Letting tests define their own wait time is dangerous. tutor-dev takes > 10sec to delete a task-plan
       # defaultWaitTime: 20 * 1000 # 20sec seems to be enough for deployed code but not local
 
+    @test = test
     @_options = _.assign {}, defaultOptions, options
     @_el = {}
 
     commonElements.loadingState = @options.loadingLocator
+    commonElements.self = testElementLocator
```
philschatz: I like the `el.self()` syntax. Could it just happen to be one of the locators passed in (simplifying the constructor args) rather than special-casing it in here?

package.json
pandafulmanda: yes, will do

test-integration/helpers/ui/test-element.coffee
pandafulmanda: I think we should be able to instead of re-writing existing functions with one variable changed.  since these are aliases, it seems like args should be able to pass through.  Also, by collecting these things into one `wait.for` we ensure that all the logging from that doesn't get neglected.

test-integration/helpers/utils/dom.coffee
pandafulmanda: as opposed to have a separate `DOM` utils? okay, will move

test-integration/helpers/ui/test-element.coffee
pandafulmanda: Hmm...but then would `loadingState` also be passed in through `COMMON_ELEMENTS`?  It seems to me that since these -- `self` and `loadingState` -- are required, they should be different from the others, as opposed to saying that `COMMON_ELEMENTS` must have certain locators.  This way, there's no demands on `COMMON_ELEMENTS` to have any specifically named locators, and the potential for them to be misnamed.

test-integration/helpers/utils/wait.coffee
pandafulmanda: All functions on this `wait` rely on being bound on this context regardless of the context it's being called in.  This way, regardless of how the `wait` utils get attached/called elsewhere, the `wait`'s will be bound the correct context.

test-integration/calendar.coffee
pandafulmanda: okay, will take out.

test-integration/helpers/ui/test-element.coffee
philschatz: Are `self` and `loadingState` always required? I guess it seems to me like the syntax used for `TestElement` (passing in locators, having a `.el.`, having command functions on it) are things that are useful for any helper, but having self/loading isn't (ie the `User` helper)

pandafulmanda: [`waitUntilLoaded`](https://github.com/openstax/tutor-js/pull/1000/files#diff-8644634b42f6a903538586ecef0c1451L105) uses `loadingState`.  `self` isn't really right now, but the idea was for `waitUntilLoaded` to check for it's presence as well

philschatz: I don't think every helper has a `loadingState` (& therefore a working `waitUntilLoaded` method), right?

Another advantage of having `self` & `loadingState` in the config (I think) is that the documentation might be a bit clearer than having to look for CSS selectors in 3 places (instead of 1)

Thoughts @openstax/tutor-fe ?

pandafulmanda: every helper has `loadingState`, it defaults to `.is-loading`.  Part of `self`'s role would be for validation -- i.e. to check that `self` is present after the current logic of `waitUntilLoaded`.

package.json
```diff
@@ -40,6 +40,7 @@
     "font-awesome": "4.4.0",
     "htmlparser2": "3.8.3",
     "jquery": "2.1.4",
+    "lodash": "4.5.1",
     "markdown-it": "4.4.0",
```
nathanstitt: If we're bringing in lodash (which I'm in favor of), shouldn't we remove underscore?

nathanstitt: Actually nm.  that'd be way to large of a change to introduce in this PR.

philschatz: Maybe add it to #919 (under "Refactor Code")?

test-integration/helpers/utils/wait.coffee
nathanstitt: I was also curious about the fat arrows.  Are they so you can pass the method into a Promise or something like (in pseudo code):

```javascript
cal = new Calendar
cal.finishLoading().then(cal.el.foo.click)
```

test-integration/helpers/ui/test-element.coffee
philschatz: OK. I guess I liked having all the selectors in 1 place and simple constructors. In an HTML comment in this comment I tried writing the helpers without the `.is-loading` but realized it would be crazy to have almost every helper have to define that element (even if `Helpers.User` didn't use it).

test-integration/helpers/ui/test-element.coffee
```diff
@@ -62,31 +83,12 @@ class TestItemHelper
...
-    # @isDisplayed(args...).then (isDisplayed) =>
-    #   unless isDisplayed
-    #     el = @findElement(args...)
-    #     @test.utils.windowPosition.scrollTo(el)
-    @test.utils.verboseWrap "Clicking #{JSON.stringify(locator)}", =>
-      el = @get(args...)
-      @test.utils.windowPosition.scrollTo(el)
-      el.click()
+  getParent: (args...) =>
+    @findElement(locator).findElement(selenium.By.xpath('..'))
```
philschatz: How does this line work? If it is not used, can it be removed?

test-integration/helpers/utils/wait.coffee
```diff
@@ -4,8 +4,11 @@ class Wait
   constructor: (test) -> @test = test
 
   # TODO reduce the copy pasta between for and forMultiple
-  forMultiple: (locator, ms = 60 * 1000) ->
+  forMultiple: (locator, ms = 60 * 1000) =>
     locator = @test.utils.toLocator(locator)
+    locator.shouldBeVisible ?= true
```
philschatz: Was there discussion on this somewhere else (maybe in a diff that was closed)? I thought the conclusion was to not add fields to the locator, no?

test-integration/helpers/ui/user.coffee
```diff
@@ -104,8 +104,7 @@ class User extends TestHelper
       hider = document.createElement('style')
       hider.textContent = '''
         .navbar-fixed-bottom, .navbar-fixed-top,
-        .pinned-on .pinned-header,
-        .pinned-footer {
+        .pinned-on .pinned-header {
```
philschatz: Why was `.pinned-footer` removed? I just included all the selectors where `position: fixed` was set (rather than trying to decide which ones are ok to remain pinned and which were not ok)

test-integration/helpers/ui/calendar.coffee
```diff
@@ -118,6 +129,12 @@ class Calendar extends TestHelper
     @test.utils.wait.giveTime PUBLISHING_TIMEOUT, =>
       @test.driver.wait((=> @el.publishedPlanByTitle(title).isPresent()), PUBLISHING_TIMEOUT)
 
+  canReviewPlan: (plan) =>
+    @test.utils.dom._getParent(plan).getAttribute(@el.canReview().locator.attr)
```
philschatz: Were you going to remove the `dom` helper?

test-integration/helpers/utils/wait.coffee
pandafulmanda: I took it out in #1003

test-integration/helpers/ui/user.coffee
pandafulmanda: it affected whether the continue button could be found in completing student tasks I think.  I dunno, i think we should manipulate the UI as little as possible for testing

test-integration/helpers/ui/calendar.coffee
pandafulmanda: Oh! yes! I did so I need to update this, good catch

package.json
pandafulmanda: just added

test-integration/helpers/utils/wait.coffee
pandafulmanda: yes it helps with that.k

test-integration/helpers/ui/test-element.coffee
```diff
@@ -62,71 +88,68 @@ class TestItemHelper
...
-    @click(args...)
+    @test.utils.dom.getParent(locator)
 
 
-class TestHelper extends TestItemHelper
+class TestHelper
   constructor: (test, testElementLocator, commonElements, options) ->
-    super(test, testElementLocator)
     commonElements ||= _.result(@, 'elementRefs', {})
+    commonElements = _.extend({}, DEFAULT_ELEMENTS, commonElements)
```
pandafulmanda: @philschatz how about this?  this makes it so that `loadingState` in on the `commonElements` argument, has a default value of `css: .is-loading`, and can be overridden with custom `loadingState` if needed as part of the `commonElements` argument

test-integration/helpers/ui/calendar.coffee
philschatz: Ah ok, makes sense :smile:

PR #1001, opened 2016-02-24
Teacher questions library

nathanstitt:
Creates a new screen for teachers to review and exclude exercises.  Goes along with https://github.com/openstax/tutor-server/pull/994  BE PR needs to be merged before this one 

TODO:
 * [x] Blocked until BE PR has changes for BigLearn incorporated
 * [x] Add bottom panel to exercise cards with toggle and trouble link.
 * [x] Mark exercises as excluded when clicked
 * [x] Load and Save exclusion list to BE. 

## warning dialog

screen shot 2016-02-29 at 4 56 55 pm

## Selected & hover
screen shot 2016-02-29 at 5 01 55 pm

## Selection screen on widescreen monitor at 1800px wide
screen shot 2016-02-26 at 9 22 58 am

## Teacher HW picker using new exercise selector and cards
screen shot 2016-02-26 at 9 23 20 am

## section selector
screen shot 2016-02-24 at 4 54 01 pm

[mockups](http://siqif5.axshare.com/#p=question_library)


src/api.coffee
```diff
@@ -152,10 +156,18 @@ start = (bootstrapData) ->
...
+  # This one loads using an ecosystemId
   apiHelper ExerciseActions, ExerciseActions.load,
     ExerciseActions.loaded, 'GET', (ecosystemId, pageIds, requestType = 'homework_core') ->
-      page_id_str = pageIds.join('&page_ids[]=')
-      url: "/api/ecosystems/#{ecosystemId}/exercises/#{requestType}?page_ids[]=#{page_id_str}"
+      url: "/api/ecosystems/#{ecosystemId}/exercises/#{requestType}?#{arrayToParams(pageIds, 'page_ids')}"
+  # And this one loads using a courseId
+  apiHelper ExerciseActions, ExerciseActions.loadForCourse,
+    ExerciseActions.loadedForCourse, 'GET', (courseId, pageIds, requestType = 'homework_core') ->
+      url: "/api/courses/#{courseId}/exercises/#{requestType}?#{arrayToParams(pageIds, 'page_ids')}"
```
nathanstitt: This is loading exercises from two different endpoints, but saving them into the same store. I think that's ok because the store only stores them by the pageIds and doesn't actually have any course/ecosystem logic.  If it does turn out to later,  then we should split them apart.

src/api.coffee
```diff
@@ -45,6 +45,10 @@ IS_LOCAL = window.location.port is '8000' or window.__karma__
 # Make sure API calls occur **after** all local Action listeners complete
 delay = (ms, fn) -> setTimeout(fn, ms)
 
+# Join an array
+arrayToParams = (vals, name) ->
+  _.map(vals, (value) -> "#{name}[]=#{value}").join('&')
```
pandafulmanda: could we either use [qs](https://www.npmjs.com/package/qs) or jQuery's [`.param`](http://api.jquery.com/jquery.param/) instead?

nathanstitt: The `qs` lib looks good to me, I wasn't aware of it's exsistance.  I'll add it.

test/components/scroll-spy.spec.coffee
```diff
@@ -0,0 +1,61 @@
...
+  render: -> React.createElement('div')
+
+RootComponent = React.createClass
+  render: ->
+
+    React.createElement('div', {},
+      React.createElement(ScrollSpy, @props,
+        React.createElement(NestedComponent, onScreenElements:[])
+      )
+    )
```
pandafulmanda: Should we make this file a `.cjsx` instead and use the syntax we are using everywhere for React elements?

nathanstitt: no, I don't think so. It's a spec file, and all the other's are `.coffee`.  This one sticks out though because it's got a bit of the `React.createElement` in it

src/api.coffee
```diff
@@ -152,10 +157,23 @@ start = (bootstrapData) ->
...
     url: "/api/plans/#{id}/stats"
 
-  apiHelper ExerciseActions, ExerciseActions.load,
-    ExerciseActions.loaded, 'GET', (ecosystemId, pageIds, requestType = 'homework_core') ->
-      page_id_str = pageIds.join('&page_ids[]=')
-      url: "/api/ecosystems/#{ecosystemId}/exercises/#{requestType}?page_ids[]=#{page_id_str}"
+  # Note: the below exercise endpoints share the same store.
+  # The contents of the json payload is identical, except the second includes an is_excluded flag
+  # since it operates at the course level
+  #
```
pandafulmanda: cool! nice comments :smile:

test/components/scroll-spy.spec.coffee
pandafulmanda: there are a few ones in the [`./test/components`](https://github.com/openstax/tutor-js/tree/master/test/components)

screen shot 2016-03-02 at 4 52 10 pm

nathanstitt: Well, how about that. I didn't know were were doing that. I'll redo as cjsx then, that'll make it much clearer what's going on.

Dantemss:
Uh we have nothing saying "need more than 5 questions" in the BE btw... should we? We only require 1 question per slot.

nathanstitt:
No idea why 5 was selected, that's just what the epic doc and mockups said should be used as the trigger point for displaying a warning.  I think that possibly proposed by @jpslav ?

src/components/tri-state-checkbox.cjsx
```diff
@@ -0,0 +1,32 @@
+_     = require 'underscore'
+React = require 'react'
+
+Icon  = require './icon'
+
+ICON_TYPES =
+  partial:   'check-square-o'
+  checked:   'check-square'
+  unchecked: 'square-o'
+
```
pandafulmanda: c:sunglasses::sunglasses:l!

src/components/questions/sectionizer.cjsx
```diff
@@ -0,0 +1,55 @@
...
+
+  scrollToSection: (section) ->
+    @scrollToSelector("[data-section='#{section}']")
+
+  scrollIndex: ->
+    @props.chapter_sections.indexOf(@props.chapter_sections)
+
+  goBack: ->
+    index = @scrollIndex()
+    @scrollToSection(sections[ if index > 1 then index - 1 else 0])
```
pandafulmanda: `sections` is `undefined` when trying to use the `goBack`, `goNext`

src/components/questions/sectionizer.cjsx
```diff
@@ -0,0 +1,55 @@
...
+  render: ->
+    <div className="sectionizer">
+      <div
+        className={cn('prev', disabled: 0 is @scrollIndex())}
+        onClick={@goBack}>❮</div>
+      {for cs in @props.chapter_sections.sort()
+        <div key={cs}
+          onClick={_.partial(@scrollToSection, cs)}
+          className={'active' if cs is _.first(@props.onScreenElements)}
+        >{cs}</div>}
```
pandafulmanda: hmmmm, i've been using `_.map` for this kind of thing, maybe at some point we would come up with an official style guide and do this one way throughout.  probs no need to change this for this PR though

src/components/questions/sectionizer.cjsx
pandafulmanda: stylistic thing, no need to change, should add to the style guide discussion -- inline vs. a separate var for the index in this

resources/styles/components/exercise-card.less
```diff
@@ -18,51 +14,55 @@
...
+          padding-top: 5px;
         }
       }
     }
-  }
-
-  &:hover {
-    .panel-heading {
-      display: block;
+    &.is-selected .toggle-mask .icon::before {
```
pandafulmanda: `&:hover` might be needed?  the "Exclude this Question" icon is the `@fa-var-plus` for some reason...also should `.icon` be `.message`?  i.e.

```less
    &.is-selected,
    &:hover {
      .toggle-mask .message::before {
        content: @fa-var-minus;
      }
    }
```

src/components/questions/sectionizer.cjsx
nathanstitt: Great catch, thanks!  That was missed when I extracted the "Sectionizer" component out from where it used to be inline on the controls.

src/components/questions/sectionizer.cjsx
nathanstitt: Yeah, to be honest, I go back and forth on both ways.  I like the inline method because it keeps the HTML all in one spot so you can see exactly what it's doing.  

However, it's also easy to miss that it's a loop and/or to miss where the loop terminates. 

I guess I'm gradually coming to the conclusion that inline is preferred for a one-element loop, but `_.map` assigned to a variable would be better for a block of elements.  

And then the block should be extracted into it's own component if it has it's own logic.  That's roughly the process I went through with writing the `Sectionizer` component at least.

resources/styles/components/exercise-card.less
nathanstitt: Another great catch.   `:hover` wasn't needed but it should have been `.message` instead of `.icon` (`.icon` was the old element, before the text message was added).

However, they're actually both incorrect :(  When I re-looked at the mockup the styles are switched from what they are on HW builder.  It should show the minus when they're selected since a selection means they're being removed, and then a plus on de-select.
 
I think i'll expand the `.exercise-card-hover-colors` mixin to be a general `.exercise-card-hover-styles` that will also accept icon arguments (with defaults.

resources/styles/components/exercise-card.less
```diff
@@ -1,11 +1,7 @@
-.exercise.card {
-  position: relative;
-  max-width: 1200px;
-  cursor: pointer;
 
-  &:hover {
-    background-color: @tutor-info-bg;
-  }
+.exercise-card {
+  &:extend(.card);
```
pandafulmanda: Cool! Didn't know about [extend](http://lesscss.org/features/#extend-feature)

src/components/questions/index.cjsx
```diff
@@ -0,0 +1,26 @@
+React = require 'react'
+
+{TocStore, TocActions} = require '../../flux/toc'
+{CourseStore} = require '../../flux/course'
+LoadableItem = require '../loadable-item'
+
+
+Dashboard = require './dashboard'
```
pandafulmanda: I think name this as `Dashboard` can lead to confusion, especially since we have another component we refer to as `Dashboard`...I think in general, each component should on be referred to with one name -- the one used to define it -- and be unique from each other.  Could this be `QuestionsDashboard` instead?

pandafulmanda:
looks :+1:

philschatz:
Merge Conflicts! can you rebase master?

nathanstitt:
Yep, rebased

nathanstitt:
BE is merged so this can be as well. @pandafulmanda has reviewed

philschatz:
This has merge conflicts. Could you rebase master into this?

nathanstitt:
@philschatz rebased

philschatz:
Aha! Selenium is failing because the HW builder no longer has checkboxes; it has that tri-state graphic. Could you update the Helper?

nathanstitt:
Sure - actually I think this PR might be older than the selenium specs..  Not sure which came first anyway :)

philschatz:
Looks awesome!

PR #1003, opened 2016-02-24
Selenium: reduce for's repeated code

pandafulmanda:
# This is not a PullRequest to Master

@philschatz, @openstax/tutor-fe  heheh is this too much?  I kinda like that it means the structure of the `for`'s will remain in sync but it make's it less readable.  Although, with `_for`, pulling out the `shouldBeVisible` from the `locator` and making a `forHidden` would be easier and less copy and paste

base branch is #1000

### DIFF
--- a/test-integration/helpers/ui/test-element.coffee
+++ b/test-integration/helpers/ui/test-element.coffee
@@ -87,6 +87,10 @@ class TestItemHelper
     locator = @getLocator(args...)
     @test.utils.dom.getParent(locator)
 
+  getParent: (args...) =>
+    locator = @getLocator(args...)
+    @test.utils.dom.getParent(locator)
+
 
 class TestHelper
   constructor: (test, testElementLocator, commonElements, options) ->
--- a/test-integration/helpers/ui/test-element.coffee
+++ b/test-integration/helpers/ui/test-element.coffee
@@ -87,10 +87,6 @@ class TestItemHelper
     locator = @getLocator(args...)
     @test.utils.dom.getParent(locator)
 
-  getParent: (args...) =>
-    locator = @getLocator(args...)
-    @test.utils.dom.getParent(locator)
-
 
 class TestHelper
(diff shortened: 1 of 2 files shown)


pandafulmanda:
any thoughts @openstax/tutor-fe ?

philschatz:
I'm not sure how often loops would be used in selenium tests; I mostly introduced them as a proof-of-concept to do the _exhaustive_ examples, but I think it may be moot if looping tests just used ES7 :smile:

pandafulmanda:
? this is not for the loops, this is for the `wait.for`'s

philschatz:
Ah, the 1st line in the diff was changing how `.forMultiple` worked so I thought that was the focus of this PR. I think 

- the readability might dissuade others from editing the file
- the `ms` arg might not actually be used anywhere (I'm also worried that having that arg available might have people use it less-sparingly)

Thoughts others?

pandafulmanda:
`ms` is being fed into [`giveTime` here](https://github.com/openstax/tutor-js/pull/1003/files#diff-7269abbae560fa8a2181f547641cd6cbR28).  With regards to `giveTime` and varying `ms` -- since we are exposing it, then I think it should be easy to use a different `ms` if needed.  It would not make sense to copy and paste code in order to use a different `ms`.

philschatz:
I just think that it should be used sparingly; the only instances that I ran into it being necessary were for publishing or deleting a plan locally.

While getting the tests to run against `tutor-dev` I did end up having to remove custom timeouts in **a lot** of places (hence my fear of overusing the feature)

But, I guess I don't feel too strongly about the `ms` arg.

Thoughts @openstax/tutor-fe ?

nathanstitt:
I think I'm in slight favor of the changes.  I agree re the readability but think it's worthwhile because (hopefully) it makes it easier to use the helper.

As for `ms` arg, it seems like the usage stays roughly the same, no?

PR #1004, opened 2016-02-29
Teacher Calendar: Fix publishing status

pandafulmanda:
* match updated data structure where `publish_job` is an object, and the `publish_job_uuid` is `publish_job.id`
* add a way to customize API fails
  * 404 of jobs should not trigger modal
  * should load as the latest state and show on calendar as failed

matches BE openstax/tutor-server#997

### DIFF
--- a/src/api.coffee
+++ b/src/api.coffee
@@ -202,6 +202,7 @@ start = (bootstrapData) ->
 
   apiHelper JobActions, JobActions.load, JobActions.loaded, 'GET', (id) ->
     url: "/api/jobs/#{id}"
+  , displayError: false
 
   apiHelper EcosystemsActions, EcosystemsActions.load, EcosystemsActions.loaded, 'GET', ->
     url: "/api/ecosystems"
--- a/src/helpers/job.coffee
+++ b/src/helpers/job.coffee
@@ -67,11 +67,15 @@ JobListenerConfig = (checkIntervals, checkRepeats) ->
       # whenever this job status is updated, emit the status for this wrapper
       updateJobStatus = @_updateJobStatusFor(id)
       JobStore.on("job.#{jobId}.*", updateJobStatus)
-      JobStore.off("job.#{jobId}.final", updateJobStatus)
+      JobStore.once("job.#{jobId}.final", @stopListening(updateJobStatus))
 
     stopChecking: (id) ->
       jobId = @_getLatestJob(id)
       JobActions.stopChecking(jobId) if jobId?
+
+    stopListening: (updateJobStatus) ->
+      (jobData) ->
+        JobStore.off("job.#{jobData.id}.*", updateJobStatus)
 
     _getJobs: (id) ->
       _.clone(@_job[id])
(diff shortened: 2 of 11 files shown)


pandafulmanda:
on hold until openstax/tutor-server#997 is finished being reviewed

src/api.coffee
```diff
@@ -202,6 +202,7 @@ start = (bootstrapData) ->
 
   apiHelper JobActions, JobActions.load, JobActions.loaded, 'GET', (id) ->
     url: "/api/jobs/#{id}"
+  , displayError: false
```
nathanstitt: Nice! Another use for our new `displayError: false` flag

src/helpers/job.coffee
```diff
@@ -66,12 +67,16 @@ JobListenerConfig = (checkIntervals, checkRepeats) ->
...
+      JobStore.once("job.#{jobId}.final", @stopListening(updateJobStatus))
 
     stopChecking: (id) ->
       jobId = @_getLatestJob(id)
       JobActions.stopChecking(jobId) if jobId?
 
+    stopListening: (updateJobStatus) ->
+      (jobData) ->
+        JobStore.off("job.#{jobData.id}.*", updateJobStatus)
+
```
nathanstitt: I see what you're doing here, but it took me a bit to realize that `stopListening` is a factory method that returns a function which stops binding the updateJobStatus when it's called.

Maybe we could call it `buildStopListeningHelper` ?  Not super happy with that either, but maybe you can come up with something better, `stopListening` just sounds like it's going to immediately stop listening, not return a function that will eventually do so.

pandafulmanda: @nathanstitt good point, I didn't even really like it when I was writing it, and I wanted to do it like 94d996d but it was being weird so I just went with that for now.  Does 94d996d work?

pandafulmanda: aw darn...looks like i caused a test failure with that, ill fix that tomorrow, esp if you think 94d996d is the way to go

src/helpers/job.coffee
```diff
@@ -63,10 +67,16 @@ JobListenerConfig = (checkIntervals, checkRepeats) ->
...
+      # make functions for updating the status for what started the job
+      # and for stopping updates
+      updateJobStatus = _.bind(@_updateJobStatusFor, @, id)
+      stopListening = _.bind(@_stopListeningToJob, @, jobId, updateJobStatus)
+
+      # on all job updates, update the job status for jobId
       JobStore.on("job.#{jobId}.*", updateJobStatus)
-      JobStore.off("job.#{jobId}.final", updateJobStatus)
+      # on a final job status, stop listening for updates
+      JobStore.once("job.#{jobId}.final", stopListening)
```
nathanstitt: Cool, yes, I like that pattern.

nathanstitt:
PR looks good, ready to merge once specs are fixed up.

pandafulmanda:
heheh, it was a timing error, passed now

PR #1008, opened 2016-03-04
Load notifications from server and display them

nathanstitt:
Goes along with BE PR: https://github.com/openstax/tutor-server/pull/1011

The notices are fixed to the navbar, but that's not an ideal experience because it may be hidden if the breadcrumbs or other bar is  locked to the top. 

screen shot 2016-03-07 at 5 39 00 pm


pandafulmanda:
:+1:

philschatz:
ooh, Since this polls, I wonder if it could be a good place to handle disconnections (`xhr.status is 0`)

philschatz:
Does devops plan on removing the message after the outage occurs?

nathanstitt:
@philschatz Yes, I'm guessing so.   Your point about ` (xhr.status is 0)` is good, we probably should just mark the request as `displayError: false` so it won't bother the user every 5 minutes if the request fails.

PR #1011, opened 2016-03-08
Fix overflow on exercises table

jivewise:
For: https://www.pivotaltracker.com/n/projects/1156756/stories/109743900

I've made the image fit the width of the box, I think it's better than hiding the image.  Also, the overflow for text is fixed as well.

Before:
https://www.pivotaltracker.com/file_attachments/57298699/download
After: 
screen shot 2016-03-08 at 7 56 20 am

### DIFF
--- a/resources/styles/components/task-plan/homework/exercises.less
+++ b/resources/styles/components/task-plan/homework/exercises.less
@@ -54,6 +54,16 @@ table.exercise-table {
     * + * {
       display: none;
     }
+
+    img {
+      display: block;
+      width: 100%;
+      padding: 20px;
+    }
+
+    span p:first-child {
+      display: inline;
+    }
   }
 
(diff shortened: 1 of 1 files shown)


nathanstitt:
Looks good @jivewise, thanks

resources/styles/components/task-plan/homework/exercises.less
```diff
@@ -54,6 +54,16 @@ table.exercise-table {
     * + * {
       display: none;
     }
+
+    img {
+      display: block;
+      width: 100%;
```
nathanstitt: I don't think `display: block` is strictly needed here.  I tested locally and images still displayed scaled in the table with the rule disabled at least.  Maybe there's other rules you're working around that I don't see though

PR #1014, opened 2016-03-09
quick relative path update for index.html for gh-pages because why not?

pandafulmanda:


philschatz:
:+1: can't think of why they wouldn't have been relative before (& I'm usually a stickler for relative paths)

nathanstitt:
I'd forgotten we were even using gh-pages :)

PR #1019, opened 2016-03-16
HW Builder: Remove images in summary table

jivewise:
https://www.pivotaltracker.com/story/show/109743900

This PR removes images from the summary table.

screen shot 2016-03-18 at 1 48 28 am

### DIFF
--- a/src/components/task-plan/homework/exercises.cjsx
+++ b/src/components/task-plan/homework/exercises.cjsx
@@ -183,10 +183,13 @@ ExerciseTable = React.createClass
 
   renderExerciseRow: (exerciseId, index, hasTeks) ->
     {section, lo, tagString} = ExerciseStore.getTagStrings(exerciseId)
-    content = $("<span>#{ExerciseStore.getContent(exerciseId)}</span>")
+    content = document.createElement("span")
+    content.innerHTML = ExerciseStore.getContent(exerciseId)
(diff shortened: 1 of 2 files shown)


src/components/task-plan/homework/exercises.cjsx
```diff
@@ -182,7 +183,10 @@ ExerciseTable = React.createClass
 
   renderExerciseRow: (exerciseId, index, hasTeks) ->
     {section, lo, tagString} = ExerciseStore.getTagStrings(exerciseId)
-    content = ExerciseStore.getContent(exerciseId)
+    content = $("<span>#{ExerciseStore.getContent(exerciseId)}</span>")
```
philschatz: Can you use `document.createElement('span')` and `.innerHTML = ` instead of jQuery?

src/components/task-plan/homework/exercises.cjsx
```diff
@@ -182,7 +183,10 @@ ExerciseTable = React.createClass
...
+
+    _.each(content.find('img'), (img) -> $(img).before("<span>[image] #{img.alt}</span>"))
```
philschatz: I'm not sure if it is clear to the user that the text is actually alt text and not part of the question.

thoughts @openstax/ux ?

jason-holmes:
I don't recommend using the alt text, just [image] text to indicate that there's an image that would appear here in the real question.

PR #1025, opened 2016-03-23
Selenium: rename ReadingBuilder to TaskBuilder

nathanstitt:
Since the helper creates both Reading and Homeworks it's better named as Task

Note, there's currently one spec failure:
```
  1) CC Student Scores hovers tooltip info popover:
     AssertionError: expected '' to include 'Correct Attempted Total possible'
```

I'm not sure if that's a side effect of my courses loaded or what, but don't see how it could be related to the rename.


philschatz:

nathanstitt:
Yep, rebased!

PR #1028, opened 2016-03-30
Switch from using pool_types to 'type:' and 'filter-type:' tag for filtering on QA view

nathanstitt:
screen shot 2016-04-01 at 1 35 33 pm

### DIFF
--- a/resources/styles/qa.less
+++ b/resources/styles/qa.less
@@ -60,9 +60,12 @@
         padding: 0 5px;
         border-radius: @border-radius-base;
       }
-      .is-ignored { text-decoration: line-through; }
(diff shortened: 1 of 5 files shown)


nathanstitt:
*Note!* Before testing you'll have to re-import your content so that the new tag types show up.

philschatz:
Eep, that color scheme looks really busy. maybe lighten the blue?

nathanstitt:
Yeah, I think I'll just forgo the stripes.  It kind of makes sense since it's both conceptual and recall so I thought I'd be clever and share those colors - I didn't really care for the look, but thought maybe it was just me :grin:

PR #1030, opened 2016-04-01
Selenium: update user login helper to new selector

pandafulmanda:


philschatz:
Oops, I think I reinvented this fix before looking at this PR #1034 . Thanks for finding it first!

PR #1036, opened 2016-04-11
Course Settings - Undrop students

darrenmason:
Regarding story: https://www.pivotaltracker.com/story/show/115538039

undrop

### DIFF
--- a/src/flux/roster.coffee
+++ b/src/flux/roster.coffee
@@ -31,10 +31,20 @@ RosterConfig = {
       students.splice(index, 1) unless -1 is index
     @emitChange()
 
+  # studentUndrop: () ->
+
+
+
+  # studentUndropped: () ->
+
+
   exports:
 
     getActiveStudentsForPeriod: (courseId, periodId) ->
       _.where(@_get(courseId).students, period_id: periodId, is_active: true)
+
+    getDroppedStudents: (courseId, periodId) ->
+      _.where(@_get(courseId).students, period_id: periodId, dropped: true)
 
 }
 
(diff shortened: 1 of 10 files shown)


src/flux/roster.coffee
```diff
@@ -24,18 +24,35 @@ RosterConfig = {
...
+
+  undrop: (unused, studentId) ->
+
+  undropped: (unused, studentId) ->
+    # set active
+    for courseId, roster of @_local
+      students = roster.students
+      student = _.findWhere(students, id: studentId)
+      student.is_active = true
     @emitChange()
```
nathanstitt: You're looping over each course and looking for a student by id, and then setting `is_active` on it without checking if it was found.  If multiple courses are loaded and the student only belongs to one, then the `findWhere` will return null.  Ideally

I think you should pass a courseId in and only search that course.  It'll be surprising behavior for a teacher if they've dropped a student from multiple courses and then attempt to re-add him to one course and it does it to all courses.

The student may still be not found though, it'd probably be good defensive programming to do something like: 
`student?.is_active = true`

philschatz:
Can you add a selenium test (or add to an existing selenium test) to undrop a student?

test-integration/roster.coffee
```diff
@@ -0,0 +1,26 @@
...
+    @user.login(TEACHER_USERNAME)
+    @courseSelect.goToByType('PHYSICS')
+    @calendar.waitUntilLoaded()
+    @roster.goToRoster()
+    @roster.waitUntilLoaded()
+
+  @it 'drops student', ->
+    @roster.dropStudent()  
+
+  @it 'undrops student', ->
```
philschatz: It seems this test is not idempotent; can you combine the tests into 1 so `undrops student` does not depend on `drops student` running first?

PR #1038, opened 2016-04-14
Student Scores - updated layout/design and late work display/accept

darrenmason:
image

image

- [x] updated pie progress
- [x] combined hs/cc table component
- [x] new cc table layout
- [x] new hs table layout
- [x] added new average data to tables
- [x] progress cell color/style if late
- [x] progress cell chart data for readings
- [x] latework popover / cell flag
- [x] update latework avgs on fe
- [x] api accept latework
- [x] update test


philschatz:
Any updates on this? Also, can you replace the `wip` in the Pull Request body with screenshots showing the progress?

philschatz:
Hmm, does this depend on a specific version of tutor-server? I'm currently seeing:

image

while running https://github.com/openstax/deploy-manifests/blob/ade42aed6ba011031c00757c1032e4089409d9f7/tutor

darrenmason:
It works with latest tutor-server; also you probably have to run rake demo tasks again.

philschatz:
Hmm, the test seems to hang (and then timeout) when students have not completed anything. Can you update the test to still work in that case? (you can point Selenium to `tutor-dev`)

image

resources/styles/components/scores/table.less
```diff
@@ -0,0 +1,472 @@
...
+    }
+  }
+
+}
+
+.late-work-info-popover {
+  height: 98px;
+  min-width: 315px;
+  border-radius: 0;
+  border: 1px solid @tutor-neutral-light;
```
nathanstitt: I'm confused by why the dimensions are so precise?  Is there a reason for the `98px` and `315` vs just say `100` and `300` I'd find it clearer if we could either comment them, or assign it to a master variable and then do math as needed to express why it's 98px for instance.

resources/styles/components/scores/table.less
```diff
@@ -0,0 +1,472 @@
...
+      }
+    }
+    .late-button {
+      padding-left: 2px !important;
+      padding-right: 2px !important;
+      width: 48%;
+      text-transform: none !important;
+      font-size: 1.1em !important;
+      float: right;
+    }
```
nathanstitt: completely personal preference on my part, but I like `flexbox` a lot more than good 'ole `table-cell` and friends.

resources/styles/global/typography.less
```diff
@@ -63,6 +63,9 @@
     font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;
     font-weight: 900;
   }
+  .scores-cells(){
+    font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;  
+  }
 }
```
nathanstitt: Pretty sure you're just keeping the existing font here, but shouldn't we be using Lato? I'm guessing probably not though

PR #1046, opened 2016-04-20
Feature - Multipart questions for students and teachers

pandafulmanda:
## Teacher assigning multipart
screen shot 2016-05-06 at 7 52 53 pm

## Student doing multipart
multi1

## teacher reviewing multipart
screen shot 2016-05-06 at 5 38 39 pm

* depends on openstax/tutor-server#1047
- [x] teacher review of multiparts is connected
- [x] students works multiparts on a single page
- [x] qa and content libs also have multiparts in single card


src/components/task-step/ends.cjsx
```diff
@@ -49,14 +50,14 @@ PracticeEnd = React.createClass
 
     # custom footer for practices
     footer =
-      <div className='-practice-end'>
+      <div className="-#{type}-end">
```
philschatz: Could/should these be `data-type={type}`? That way we're not dynamically creating class names? (greppability). Maybe I'm the only one though ; (

philschatz:
Can you add selenium helpers for answering multipart questions?

src/components/task-step/ends.cjsx
pandafulmanda: okay, makes sense

pandafulmanda:
ah yes to selenium tests. heh. you caught me

philschatz:
moved to `needs-code` since it's missing Selenium 😄

src/components/task-step/exercise.cjsx
```diff
@@ -1,67 +1,205 @@
...
+      @setState(@getPartsInfo(nextProps))
+
+  stopListeningForProgress: (props) ->
+    props ?= @props
+    {taskId} = props
+
+    TaskProgressStore.off("update.#{taskId}", @setCurrentStepFromProgress)
+
+  startListeningForProgress: (props) ->
+    props ?= @props
```
nathanstitt: how would `@props` be null ?

resources/styles/components/task-step/ends.less
```diff
@@ -40,15 +40,19 @@
     .homework-review-problem-enter-active {
       .card-body {
         .transition(~"max-height 0.5s ease-in, padding 0.5s ease-in, margin 0.5s ease-in");
-        max-height: 999px;
+        max-height: 3000px;
         padding: @tutor-card-body-padding;
```
nathanstitt: Have you considered switching between `transform: scaleY(0);` to `transform: scaleY(1);` ?  That avoids the `max-height` issue, but isn't quite the same effect.  https://jdsteinbach.com/css/holy-grail-css-animation/

nathanstitt:
When switching steps, I get an exception cannot read `.name` of undefined here:  https://github.com/openstax/tutor-js/blob/master/src/helpers/policies/index.coffee#L52

screen shot 2016-05-19 at 10 23 04 pm

It does go ahead and switch though, but the step doesn't seem to be recorded as completed.  The breadcrumbs don't update to appear completed, but at the end it says I only completed 2 of the 5 questions.

src/router.cjsx
```diff
@@ -34,7 +34,7 @@ routes = (
         <Router.DefaultRoute handler={TeacherTaskPlans}/>
 
         <Route path='list/?' name='viewStudentDashboard' handler={StudentDashboardShell} />
-        <Route path='tasks/:id/?' name='viewTask' handler={SingleTask}/>
+        <Route path='tasks/:id/?' name='viewTask' handler={SingleTask} ignoreScrollBehavior/>
         <Route path='tasks/:id/steps/:stepIndex/?'
```
nathanstitt: Neat.  Didn't know bout `ignoreScrollBehavior`

nathanstitt:
I also get a bunch of  `Invariant Violation: findComponentRoot` exceptions, but I suspect that they're related to the above bug

nathanstitt:
I think the code looks pretty good though - if we can get the bug fixed I'd be ok with merging.  Probably need some selenium updates for MPQ though, right?

src/components/task-step/exercise.cjsx
pandafulmanda: when `@ stopListeningForProgress` is called without an argument, the`props` is `null` and then defaults to `@props`.  this makes it easy to call the same function with `nextProps` or just leave it empty so that it uses `@props`

resources/styles/components/task-step/ends.less
pandafulmanda: ah! neat idea, will check it out

pandafulmanda:
in reference to [this comment](https://github.com/openstax/tutor-js/pull/1046#issuecomment-220508688) i believe #1066 will fix this

pandafulmanda:
are we going with selenium update within this PR or after this merges, as separate one @philschatz / @nathanstitt ?

philschatz:
yep, separate PR for Selenium fixes makes sense. @nathanstitt can you merge if there's nothing else?

pandafulmanda:
i think there are changes i need to make in the `react-components` PR before this one is ready

philschatz:
removing `needs review` since it sounds like it's not ready anymore 😄

nathanstitt:
@pandafulmanda I've discovered a bug, sorry 😦   

If you create an assignment with all of section 1.2 of biology, then work it as a student, it proceeds through the first multi-part just fine.  

Then on the second multi-part (breadcrumb step 5), it answerers the first part fine, and scrolls down after you click "next question",  but the second part isn't answerable.  It only has two options, A & B, but neither of them can be clicked.  

However the "Next Question" button in the bottom navbar is enabled and can be clicked to advance to breadcrumb step 6. Step 5 remains white and the step isn't marked as complete though.

nathanstitt:
Since this one is pretty large and holding other things up, I'm fine with merging and working on the  bug as a follow up though.  I'm looking through the code and haven't spotted what's going wrong yet.

pandafulmanda:
@nathanstitt, that's due the the question being a new `true-false` format, which was discussed [here](https://github.com/openstax/tutor-js/pull/1046#issuecomment-220744524) and fixed in #1066.  I will rebase this with master and push again so we can test this all the way through

src/components/task-step/exercise.cjsx
```diff
@@ -52,48 +113,129 @@ module.exports = React.createClass
...
+
+    TaskProgressStore.on("update.#{taskId}", @setCurrentStepFromProgress)
+
+  canAllContinue: ->
+    {parts} = @state
+
+    _.reduce parts, (previous, part) ->
+      previous and canOnlyContinue(part.id)
+    , true
+
```
nathanstitt: Not a big thing, but this and the `_.reduce` on line 153 below could use `_.every`.  that's a tiny bit more efficient since it'll abort the loop early as soon as something returns `false`.

src/components/task-step/exercise.cjsx
```diff
@@ -52,48 +113,129 @@ module.exports = React.createClass
...
+      disabled={TaskStepStore.isSaving(part.id)}
+      isContinueEnabled={StepPanel.canContinue(part.id)}
+
+      canOnlyContinue={canOnlyContinue}
+      getWaitingText={getWaitingText}
+      getCurrentPanel={getCurrentPanel}
+      getReadingForStep={getReadingForStep}
+      setFreeResponseAnswer={TaskStepActions.setFreeResponseAnswer}
+      onFreeResponseChange={@onFreeResponseChange}
+      setAnswerId={TaskStepActions.setAnswerId}/>
```
nathanstitt: I like how you abstracted the scrolling into the `ExerciseWithScroll`

src/components/task-teacher-review/crumb-mixin.cjsx
```diff
@@ -66,16 +66,29 @@ module.exports =
...
 
 
   _getCrumbsForHomework: (stats) ->
     exercises = @_getExercisesFromStats(stats)
     crumbs = _.chain(exercises)
-      .map(@_makeCrumb)
-      .each(@_indexCrumb)
+      .map((data) =>
+        {questions} = JSON.parse(data.content)
+        stepCrumbs = []
```
nathanstitt: Hmm.  Do you really need `JSON.parse` ?  I stuck a console.log in here and it looks to me like `data.content` is already an object.  That was when loading the metrics for an assignment as a teacher.

src/components/task-step/exercise.cjsx
pandafulmanda: okay

src/components/task-teacher-review/exercise.cjsx
```diff
@@ -60,21 +64,75 @@ TaskTeacherReviewExercise = React.createClass
...
+  render: ->
+    {content} = @props
+    {questions, uid} = content
+
+    exercise = _.map questions, @renderQuestion
+    {stimulus_html} = content
+
+    stimulus = <ArbitraryHtmlAndMath
+      {...htmlAndMathProps}
+      className='exercise-stimulus'
```
nathanstitt: `htmlAndMathProps` is undefined, is causing the assignment metrics screen to fail 😿

src/components/task-teacher-review/crumb-mixin.cjsx
pandafulmanda: screen shot 2016-05-27 at 1 07 56 pm

src/components/task/index.cjsx
```diff
@@ -1,11 +1,12 @@
-React = require 'react'
+React = require 'react/addons'
 BS = require 'react-bootstrap'
```
nathanstitt: is `addons` needed?  I don't see anywhere in the changes that you're using any of 'em.  I've probably just missed it though

nathanstitt:
Looks good other than the undefined var bug I noted above.

src/components/task-teacher-review/crumb-mixin.cjsx
nathanstitt: yeah that's true. If it wasn't JSON we'd have exceptions.  Here's what I see though:
screen shot 2016-05-27 at 1 31 21 pm

nathanstitt:
Looks good @pandafulmanda - let's merge it 😁

pandafulmanda:
thanks! pending merging/decision from this [discussion](https://github.com/openstax/react-components/pull/62#issuecomment-222208436)

PR #1047, opened 2016-04-26
Filter exercise tags on QA view

nathanstitt:
Implements a filtering function that the exercise data is passed through before being displayed on the QA view preview cards.

No user-visible changes, other than QA exercise cards should no longer show LO and tags that do not belong to current book.


src/components/qa/exercise-data-filter.cjsx
```diff
@@ -0,0 +1,29 @@
+_ = require 'underscore'
+
+{EcosystemsStore} = require '../../flux/ecosystems'
+
+BOOK_MAPPINGS =
+  '02040312-72c8-441e-a685-20e9333f3e1d': 'stax-soc'
```
philschatz: This part seems duplicated; could it be in 1 place and just pulled in from the flux store?

nathanstitt: Hm, I'm not aware of any place that the book UUID's are mapped to tag id's (`stat-bio`, etc.)

Are they?

nathanstitt: arg, nm.  This file shouldn't even be here!  It was here originally, then I moved it to the store.  I'll remove it.

PR #1054, opened 2016-05-06
display LMS assignment link

nathanstitt:
Needs BE PR: https://github.com/openstax/tutor-server/pull/1079

image

image

### DIFF
--- a/resources/styles/components/course-calendar/plan-details.less
+++ b/resources/styles/components/course-calendar/plan-details.less
@@ -53,4 +53,13 @@ body.modal-open {
       .flex(1);
     }
   }
+  .sharable-link{
+    .description{
+      margin-top: 10px;
+      padding: 0px 5px 0px 0px;
+      min-height: 50px;
+      max-height: 200px;
+      overflow-y: auto;
+    }
+  }
 }
--- a/resources/styles/components/course-calendar/plan-details.less
+++ b/resources/styles/components/course-calendar/plan-details.less
@@ -45,9 +45,12 @@ body.modal-open {
     .flex-display();
     .align-items(baseline);
     .left-buttons {
-      .flex(1);
       .flex-display();
       .align-items(baseline);
+    }
+    .lms-info {
+      text-align: right;
+      .flex(1);
     }
   }
 }
--- a/resources/styles/components/course-calendar/plan-details.less
+++ b/resources/styles/components/course-calendar/plan-details.less
@@ -39,4 +39,15 @@ body.modal-open {
 
     .panel-body { padding: 0; }
   }
+
+  .modal-footer {
+    text-align: inherit;
+    .flex-display();
+    .align-items(baseline);
+    .left-buttons {
+      .flex(1);
+      .flex-display();
+      .align-items(baseline);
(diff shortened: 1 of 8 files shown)


nathanstitt:
@openstax/ux 

Note: I took a few (small) liberties on this.  The mockup showed a 'Summary', the due date, title, and description under the link. 

@kathi-fletcher asked to remove the Summary, so I did so.

The FE code does not have the description included in the output the BE sends, and even if it did, I don't think long descriptions would really work in the constrained size.  We can add it to the BE data feed if needed though.

I omitted the date because there's not just one date, each period has their own.  We could display the date if there was only a single common one, but then I fear teachers would be confused as to why the date would sometimes appear and sometimes not.

Then after I removed both of those fields the title felt weird sitting below the input box, so I moved it on top.

philschatz:
@openstax/ux : Any thoughts on this? It's a bit different than what was shown in the mockups

rvnewell:
Jason has the last word on this, as it's his mockup.  I don't believe he accesses github, so it would be better to ask him on the pivotal ticket.  My two cents:

1) The url should be highlighted as if selected.  This tells the user the next step is to copy what has been selected (matches sharing on GoogleDocs.)
2) "Read Chapter" and link as well as due date and description are important to the design because instructors want to do one or both of two things 1) get url only and/or 2) get assignment metadata with link.  They would paste these into different parts of their LMS.  Including all information ready to be copied would make this more efficient for them.
3) I don't know how Jason would respond to the date issue, so you will need to ask him. He is out sick, but should be back online tomorrow.

philschatz:
Thanks! I think the other sites you've cited select the entire text, not just part of it. Would that work?

Regarding the description: Teachers enter Markdown for the description which is converted to HTML. (This is probably mostly only useful to be able to easily enter multiple paragraphs of text and have it display reasonably).

If we did include the description in the snippet, should it be markdown or HTML? I'm guessing ultimately it depends on the LMS that they would be pasting into.


Example Markdown I think teachers would type:

```
Try to do these without looking at the book.

They are just for practice for the exam.
```

... which turns into `<p>Try to do...</p> <p>They are just....</p>`

rvnewell:
I would say to have the URL auto-selected, since we're guessing 80% of the time that's what instructors will want to copy (or copy first, separately).  They then would have the option of click-dragging over the rest of the text and expanding the selection, or selecting to copy again.  This is different than other sites because in those cases they are sharing a link only.  

I'm not sure on markdown vs HTML, but my first thought is markdown, since they would just be pasting into a text field.  Jason has the LMS experience (at least with Blackboard), though, and should decide this. 

Nathan's point about description getting lengthy and unruly is well taken.  I will pass this on to Jason and we will look into cases where this might happen.

jason-holmes:
I agree with Robert on these issues...autoselect just the URL and use markdown for descriptions.  If they get long, I guess give the description a fixed height and then scroll it.

philschatz:
Can you add a selenium test that adds the helper to click the link?

philschatz:
Didn't test with Selenium but merged since people were waiting on it

PR #1060, opened 2016-05-20
Progress bar for readings

jivewise:
New progress bar for readings, see screenshots below.  I still need to finish tests, and I'm putting the keyboard events in another PR.

screen shot 2016-05-20 at 8 18 54 am

screen shot 2016-05-20 at 8 20 03 am

### DIFF
--- a/resources/styles/components/task/index.less
+++ b/resources/styles/components/task/index.less
@@ -1,4 +1,5 @@
 @import './details';
+@import './progress';
 
 .task {
   max-width: @tutor-task-width;
@@ -30,4 +31,14 @@
       padding-bottom: 1.5 * @tutor-navbar-height;
     }
   }
+}
+
+&.instructions {
+	//Product name and content for instructions
+	.product-name:after {
+		content: 'Tutor';
+	}
+	.feedback-type:after {
+		content: 'personalized feedback';
+	}
 }
--- a/resources/styles/components/task/progress.less
+++ b/resources/styles/components/task/progress.less
@@ -1,0 +1,32 @@
+.task-view .task-reading .pinned-header {
+	padding: 0;
+
+	.reading-task-progress {
+		height: 4px;
+		overflow: hidden;
+		margin-bottom: 0;
+	}
+}
+
+.arrow {
+	top: 40%;
+	cursor: pointer;
+	position: fixed;
+
+	&.left {
+		left: 50px;
+	}
+
+	&.right {
+		right: 50px;
+	}
+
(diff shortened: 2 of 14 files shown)


pandafulmanda:
hmmm, i guess i have a questions for @openstax/ux ...should the progress bar be based on where the student is in the reading currently or be based on last worked step?  for example, if i go to a reading and like step 8 of 10 and then decide to nav back to step 4 of 10, should the progress bar shrink from 80% to 40% as i navigate back?  otherwise this is looking pretty good

Fredasaurus:
@pandafulmanda @jason-holmes Where the student is currently is what I was thinking. It is not as accurate as last worked, but gives the advantage of having one step already filled on the initial screen, which shows progress relative to the size of the assignment. Otherwise this would be blank.

jason-holmes:
agree with Fred, the bar should indicate current location

jivewise:
:+1: @jason-holmes @Fredasaurus

pandafulmanda:
then should the first step show as some progress, as opposed to 0?

jivewise:
I'm fixing a case where the spacer for some readings is messing up the progress bar arrows.

Fredasaurus:
@pandafulmanda Yes, the progress should be commensurate with the completion of the first step.

philschatz:
Do the selenium helpers not need to be updated? I think they had code for clicking on the breadcrumbs

jivewise:
Good point, I'm updating them currently, and they revealed a couple of bugs that I'm fixing with the task review page.

pandafulmanda:
@jivewise, by the way, the previous arrows gets covered when the screen is narrow

jivewise:
Thanks @pandafulmanda , I fixed it in the latest update along with some selenium tests.

philschatz:
The selenium tests time out for me:

```
[59s] START: Waiting for {"css":".continue"} to show
[59s] END  : Waiting for {"css":".continue"} to show returning=true
[9s]
[9s] START: Wait until scrolling completes
[9s] END  : Wait until scrolling completes returning=true
[60s] START: Looking for {"css":".pinned-container .arrow.right"}

  HS Student Scores
    2) "before all" hook

  16 passing (6m)
  2 pending
  2 failing

  1) Student performing tasks Can continue and go to expected steps (readonly):
     Error: timeout of 68176ms exceeded. Ensure the done() callback is being called in this test.
```

It looks like it happens when a teacher is reviewing the HW:

image

PR #1063, opened 2016-05-20
example edits

pandafulmanda:


nathanstitt:
looks good, thanks for fixing up the stuff i missed in #1059

PR #1064, opened 2016-05-20
Question library/HW Builder fixes

nathanstitt:
Depends on openstax/react-components#70 (merged) and https://github.com/openstax/react-components/pull/72
 
 * Add details view to HW builder
  * **warning** This turned into a major refactor
 * Change HW builder control bar to include "sectionizer" and view toggles.
 * Moves exercise to top of details view instead of vertically centering
 * Pulls in new exercise preview with updated "badge" icons
 * uses pure-render mixin on sections to speed up rendering
 * Updates name in menu per UX feedback
 * Sort sections in store - fixes bug where sectionizer and cards would be sorted as `10.1`,  `1.2`, `3.1`
 * Hides "Add/Remove" control from HW review
 * adjusts z-index of controls mask to be under top control bar

screen shot 2016-05-27 at 10 56 52 am

screen shot 2016-05-20 at 3 33 35 pm


src/components/exercises/details.cjsx
```diff
@@ -0,0 +1,151 @@
...
+    onExerciseToggle:      React.PropTypes.func.isRequired
+    onShowCardViewClick:   React.PropTypes.func.isRequired
+    topScrollOffset:       React.PropTypes.number
+    getExerciseActions:    React.PropTypes.func.isRequired
+    getExerciseIsSelected: React.PropTypes.func.isRequired
+
+  mixins: [ScrollTo]
+  scrollingTargetDOM: -> document
+  getDefaultProps: ->
+    topScrollOffset: 40
```
philschatz: Can you add a comment on why this is 40px or if it's used elsewhere can it be a const that is pulled in to this file?

philschatz:
Given the number of changes needed to make this PR work, these could be fixed later as bugs, but adding screenshots now:

### When in single card view it seems both single-card and table view are disabled/selected (not sure which but it does not match the styling for the list of sections just to the left)

image


### When viewing a single card the current section is not highlighted

image

### User can scroll up which is probably not very lightboxy

image

### The arrows move up and down depending on the height of the exercise; maybe only the exercise should scroll vertically?

image

philschatz:
I think the only thing missing on this PR is getting the existing Selenium tests to run. It seems the "Tutor Selections" +- buttons are missing from the DOM


image

nathanstitt:
@philschatz Fixed the display indicator and the +/- tutor selection buttons. 

The selenium specs appear to be failing on calendar load for me. I'll see if I can figure them out tomorrow.

PR #1026, opened 2016-03-23
Selenium: modal is considered closed too early in chrome. 

pandafulmanda:
need to wait until the modal is not present at all.  For use in `openstax/test-automation`

Tests have been fixed/updated as much as possible.  Current failures in this branch from running tests will be due to:
* Teacher task review being unable to render
  * will be fixed when #1046 is merged
* assignment open/due date bugs
  * will be fixed when #1067 is merged
* assignment feedback immediate bug
  * will be fixed when #1069 is merged

PR #1067, opened 2016-05-23
various small fixes related to open/due time changes

pandafulmanda:
- [x] after task save, url for calendar date should be date only
- [x] time input masking behavior should be more intuitive and allow values like `7:00 am` as opposed to `07:00 am`

Note: there's a slightly less annoying behavior where if you press `1` in the first position for a time that is < `10`, the mask will expand and the cursor will move to the end.  If you go back to the proper position and continue typing your time, things will behave as expected -- a `:` will shrink the mask back to `1:__` while a number between `0` and `2` will input a double digit hour as expected.


src/components/tutor-input.cjsx
```diff
@@ -212,15 +217,37 @@ TutorDateInput = React.createClass
...
-      m: '[0-9]'
-      P: '(A|P|a|p)'
-      p: '(M|m)'
     fromMomentFormat: 'HH:mm'
-    toMomentFormat: 'hh:mm a'
+    toMomentFormat: 'h:mm a'
+    formatCharacters:
+      i:
+        validate: (char) ->
+          /([0-2]|:)/.test(char)
```
philschatz: Could these lines just be `validate: /([0-2]|:)/.test` ?

philschatz:
Did the tests not need to be updated?

pandafulmanda:
they do a little bit, but i think we need some other PRs to be merge first, and then #1026 with the multiple test updates to be merged

philschatz:
yay it works!

PR #1068, opened 2016-05-25
make checking for log in link a wait get, is displayed off the elemen…

pandafulmanda:
…t directly only checks for if it's immediately displayed

### DIFF
--- a/test-integration/helpers/ui/user.coffee
+++ b/test-integration/helpers/ui/user.coffee
@@ -94,7 +94,7 @@ class User extends TestHelper
     @el.loginSubmit().click()
 
   login: (username, password = 'password') =>
-    @el.loginLink().isDisplayed().then (isDisplayed) =>
+    @el.loginLink().get().isDisplayed().then (isDisplayed) =>
       unless isDisplayed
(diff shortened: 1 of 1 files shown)


test-integration/helpers/ui/user.coffee
```diff
@@ -94,7 +94,7 @@ class User extends TestHelper
     @el.loginSubmit().click()
 
   login: (username, password = 'password') =>
-    @el.loginLink().isDisplayed().then (isDisplayed) =>
```
philschatz: Is there a case for using `foo().isDisplayed()` instead of `foo().get().isDisplayed()`? If not, maybe it's worth removing?

pandafulmanda: `isDisplayed` checked for immediate display which is needed for times when the `foo` should be there immediately and not have any waits

philschatz: Hmm, ok. I guess I didn't realize that difference between `.isDisplayed` and `.get().isDisplayed`. Maybe @openstax/test will run into the same problem?

PR #1072, opened 2016-05-27
Fix lint warnings & errors

nathanstitt:


src/components/scores/pie-progress.cjsx
```diff
@@ -1,3 +1,5 @@
+# coffeelint: disable=max_line_length
```
philschatz: Can this be disabled/enabled only around the offending lines?

src/components/scores/export.cjsx
```diff
@@ -1,3 +1,5 @@
+# coffeelint: disable=max_line_length
```
philschatz: Can this be disabled/enabled only around the offending lines?

PR #1077, opened 2016-05-31
Scores late work popup dialog

nathanstitt:
Goes with BE PR: https://github.com/openstax/tutor-server/pull/1088

Before Accepting:

screen shot 2016-06-01 at 7 05 38 pm

And after (note increased averages):

screen shot 2016-06-02 at 12 04 48 pm

And then after student does more work after an accept:
screen shot 2016-06-02 at 11 56 49 am


src/components/scores/homework-cell.cjsx
```diff
@@ -15,14 +15,14 @@ HomeworkCell = React.createClass
   getScore: (isAccepted) ->
     {task} = @props
     if isAccepted
```
jpslav: Can `getScore` and `getProgress` ditch the `isAccepted` argument? and just return the `...on_time... + ...accepted_late...` sum?

src/components/scores/late-work.cjsx
```diff
@@ -1,113 +1,120 @@
...
+class LateWork
+  constructor: (@task) ->
+    @isAccepted = @task.is_late_work_accepted
+    @status = if @isAccepted then 'accepted' else 'pending'
+
+    @displayValue =
+        if @isAccepted
+          @task.correct_on_time_exercise_count + @task.correct_accepted_late_exercise_count
+        else
+          @task.correct_on_time_exercise_count
```
jpslav: This should also reduce to line 15 I think

src/components/scores/late-work.cjsx
```diff
@@ -1,113 +1,120 @@
...
+  componentWillMount: ->
+    Content = if @props.task.type is 'homework' then HomeworkContent else ReadingContent
+    @setState(
+      content: new Content(@props.task)
+    )
 
-  isUpdatingLateStatus: ->
-    ScoresStore.isUpdatingLateStatus(@props.task.id)
+  onButtonClick: ->
+    if @state.content.isAccepted
```
jpslav: I think when you have the "Additional late work" popover, this check will not be sufficient (because you'll be in the `isAccepted == true` state but wanting to call `acceptLate` again)

src/flux/scores.coffee
```diff
@@ -10,105 +11,111 @@ REJECTING = 'rejecting'
...
     .value()
 
+getTaskInfoById = (taskId, data) ->
+  taskId = parseInt(taskId, 10)
+  for courseId, period of data
+    for period, periodIndex in data[courseId]
+      for student, studentIndex in period.students
+        for task in student.data
+          if task.id is taskId
+            return {task, courseId, period, periodIndex, studentIndex}
```
jpslav: If a class has 200 tasks and 500 students, is this nested iteration going to be too slow?  If so, do a one-time, possibly-laze build of a hash mapping `taskId` to `{task, courseId, period, periodIndex, studentIndex}`?

src/flux/scores.coffee
```diff
@@ -10,105 +11,111 @@ REJECTING = 'rejecting'
...
-    currentValue = currentValue / 100
-    acceptValue = acceptValue / 100
 
+adjustTaskAverages = (data, taskInfo) ->
+  {task} = taskInfo
+  oldScore = task.score
+  task.score = Math.round((
+    (task.correct_on_time_exercise_count + task.correct_accepted_late_exercise_count ) /
+      task.exercise_count
+  ) * 100 ) / 100
```
jpslav: man it is so unfortunate that you can't put methods on `task`!

src/flux/scores.coffee
```diff
@@ -10,105 +11,111 @@ REJECTING = 'rejecting'
...
+      ( task.score / studentCount )
 
-    period?.data_headings[columnIndex]?.average_score = assignmentAverage
+  taskCount = 0
+  for student in course.students
+    taskCount += 1 for studentTask in student.data when studentTask.is_included_in_averages
 
+  course.overall_average_score =
+    (course.overall_average_score - ( oldScore / taskCount ) ) +
+      ( task.score / taskCount )
```
jpslav: Around here is probably where you need a `student_average` calculation (should be three averages updating per score change)

src/flux/scores.coffee
nathanstitt: Yeah, I thought about doing that, but didn't do so be cause that'd require setting up the cache overtime the performance report loads, and it probably won't actually be used too often, since the task lookup only happens if the teacher accepts late work.

However - your comment's made me think on it a bit more and what we actually want is a cache that is initialized on first access, then later accepts will be fast without impacting the first-display time.   I'll do that

src/flux/scores.coffee
nathanstitt: there are 3 updates.  This is for course, the heading (assignment) happens on line 42, and task is above the earlier comment on line 33.

src/flux/scores.coffee
```diff
@@ -10,105 +11,111 @@ REJECTING = 'rejecting'
...
+ScoresConfig = {
 
-    currentStudentAverage = taskStudent?.average_score
+  _asyncStatus: {}
 
-    # number of included homework tasks for student
-    numTasksStudent = _.filter(taskStudent.data, {is_included_in_averages: true}).length
+  hasLateWork: (task) ->
+    task.completed_step_count > task.completed_on_time_step_count +
+      task.completed_accepted_late_step_count
```
jpslav: @rvnewell has said that reading progress and homework progress should just based on exercise counts, not generic step counts.  So I think this function should change to reference exercise counts.  Even though I provide the generic counts, I don't think they should be used anywhere in the scores report.

src/flux/scores.coffee
```diff
@@ -10,105 +11,111 @@ REJECTING = 'rejecting'
...
 
-    studentAverage =
-      (currentStudentAverage - (currentValue / numTasksStudent)) +
-      (acceptValue / numTasksStudent)
```
jpslav: And here is the missing average calculation :-)

src/flux/scores.coffee
nathanstitt: hm, that method is also called by readings though.  maybe I should make two methods, one that reads _exercise_, and the other steps

nathanstitt:
@jpslav ready for another review

PR #1080, opened 2016-06-02
Updated Assignment Settings

jivewise:
Updates which settings are editable on the front end after an assignment is published and is open.

screen shot 2016-06-02 at 11 51 45 pm
screen shot 2016-06-02 at 11 51 36 pm

Also includes changes to student dashboard, showing deleted tasks and hide task buttons:

pasted_image_at_2016_06_10_04_11_am

For: https://www.pivotaltracker.com/story/show/115959401


jivewise:
Before merging this, merge #1094 into here.

philschatz:
After changing the feedback on an assignment that has already closed I got:

image


Also, after deleting an assignment and going back to the URL for the assignment I see a 404 (maybe that's ok)

image

jivewise:
Hm, @openstax/ux should the feedback field be updatable if the assignment is past due?  If not, I can disable it.  Otherwise, it will have to be a backend fix.

jason-holmes:
Feedback field should not be updatable after the assignment is due.  Please disable.

philschatz:
merging now for testing but still need @openstax/ux feedback and maybe we can create a new PR

philschatz:
Does @openstax/tutor-be give enough information to tell that an assignment's due date has already passed? Aside from doing date-math in the browser?

jivewise:
I believe we have an isPastDue helper on the flux store that we can use, I'll add that in a separate PR.

PR #1082, opened 2016-06-07
HW / QL details view tweaks

nathanstitt:
This:

  * Keeps controls visible but allows them to scroll when scrolling up to top
  * Removes the details / card view toggle from control bars
  * Colors "Back to card view link grey"
  * Fixes regression where the controls bar background was transparent instead of white
  * Fixes sectionizer controls in HW builder so it updates and can jump to a section
  * Pulls in exercise preview update from react-components for the feature flags

@rvnewell is preparing a few more tweaks to the icons, but I think they'll be fairly stand alone, we can go ahead and merge this 

scroll

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "mime-types": "2.1.7",
     "moment": "2.13.0",
     "moment-timezone": "0.5.4",
-    "openstax-react-components": "openstax/react-components#22bb557becc27bfccb9bb274dd3f47397f5b0d51",
+    "openstax-react-components": "openstax/react-components#455a284003968deca6fa5fba923ef94b3151901b",
     "pluralize": "1.2.1",
     "promise-loader": "1.0.0",
     "qs": "6.1.0",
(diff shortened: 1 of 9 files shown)


philschatz:
When I scroll down to look at questions and then click the "Question Details" view the top of the exercise is cut off:

image

nathanstitt:
@philschatz - just pushed an update to set bar height at 100% when not scrolling up, that should fix the offset issue.

I'm unclear on how you ended up with the top of the exercise cut off though.  I tried to replicate by:
 * load a large list of exercises
 * scrolled to the bottom
 * Clicked details on an exercise.

When it displayed details view it correctly scrolled to the top of the exercise in details view.  It looks like that failed for you for some reason.  Do you have any more info on what happened ?

philschatz:
I don't know that I did anything special. Running `Chrome 52.0.2716.0`. I'll merge for now & we can revisit if QA duplicates the problem

PR #1083, opened 2016-06-08
Implement due dates and other fixes to LMS info link

nathanstitt:
When all periods have same due date:
screen shot 2016-06-08 at 4 39 34 pm

With different due dates for periods:
screen shot 2016-06-08 at 5 29 08 pm

### DIFF
--- a/api/courses/1/dashboard.json
+++ b/api/courses/1/dashboard.json
@@ -181,7 +181,13 @@
         {
           "target_id": "1",
           "target_type": "period",
-          "opens_at": "2015-03-31T11:40:23.796Z",
+          "opens_at": "2015-02-31T11:40:23.796Z",
+          "due_at": "2015-04-05T10:00:00.000Z"
+        },
+        {
+          "target_id": "2",
+          "target_type": "period",
+          "opens_at": "2015-02-31T11:40:23.796Z",
           "due_at": "2015-04-05T10:00:00.000Z"
         }
       ]
@@ -194,7 +200,13 @@
         {
           "target_id": "1",
           "target_type": "period",
-          "opens_at": "2015-03-31T11:40:23.796Z",
+          "opens_at": "2015-01-31T11:40:23.796Z",
+          "due_at": "2015-04-05T10:00:00.000Z"
+        },
+        {
+          "target_id": "2",
+          "target_type": "period",
+          "opens_at": "2015-03-30T11:40:23.796Z",
           "due_at": "2015-04-05T10:00:00.000Z"
         }
       ]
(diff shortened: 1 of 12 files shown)


philschatz:
Clicking the link selects the link but clicking the rest of the text does not select it. Also, if you happen to click the title of the assignment it is _actually_ a link and takes you out of the calendar and into the "Edit Assignment" page. I'm guessing that is not desired behavior. Could we just make the whole thing a textarea that, when clicked, selects everything (like Google Share links)? /cc @openstax/ux

nathanstitt:
Not sure of the purpose of the hyphen, that was to match mockup. I **think** it's to demarcate the sections.

The task said to make title a link - should it not be?

The rest I'll leave to @openstax/ux

Fredasaurus:
I think @rvnewell worked on this one.

rvnewell:
Jason designed this, I am just trying to clarify the design. The intent behind the title as link was that the summary information would be copied and pasted into a different part of the LMS than the URL.  Instructors might want to paste it into an assignment summary or description text field.  The title link within that summary info would be useful for students in that location in their LMS.  

The URL was intended to be pasted into a specific hyperlink field almost always present in an LMS.  This is likely the most common instructor behavior.  

This means that the instructors may want to copy only the URL, only the summary information, or both, depending on the context.  Clicking and copying all won’t give them that option. 

Intended behavior is — instructor clicks "Get assignment link” > the popup appears and displays all of the information in the text area window, with the URL already highlighted.  They control + C or copy from edit dropdown (without needing to select) to copy the link. This matches GoogleDoc behavior when you want to share a link.  

If the instructor wants the URL + the summary info, or just the summary info, they can click-drag to select what they want to copy.  

If the title-as-link creates an issue, then it may only work as non-linked text.

philschatz:
ok, so then the changes to the screenshot would be:

- convert the `-` to a `<hr/>`
- replace the link in the title to just be the title

right?


For what it's worth, here's what I was suggesting:

image

And if you focused on the bottom `<textarea>` (ie clicking, or tabbing to it) the entire text would be highlighted (but you couldn't edit it here)

rvnewell:
Okay, I see what you were getting at now.  I think both would basically work.  Your one-click option does make selecting simpler.  But I would still prefer to honor Jason's design-- it does allow the teacher to copy part of the summary information if they don't need all of it (e.g. the assignment description could be really long and/or they may only want use title/section times.)

kathi-fletcher:
I think phils suggestion might be good because out allows both options - just the url or all the info. I assume you can deselect some info if you don't want it all. 

Also i think it is fine to lose the word "summary". I think the ticket said that also. 

This comment field does a thing where double clicking selects some stuff and then pops up a menu so you can copy or select all. May be an option for mobile.

PR #1086, opened 2016-06-09
update version of recordo to log the current URL

philschatz:
for when a snapshot was taken. It currently only logs when the URL changes.

So, if you load a page while recording, click clear, and then do something, the current URL is not in the log.

Also, I was thinking of adding another feature to recordo: a checkbox to exclude AJAX responses. Currently, when you load the Student Scores, it's a big honking JSON response which isn't terribly useful for devs.

Thoughts?


nathanstitt:
love it. this'll be useful

PR #1094, opened 2016-06-13
Set Period Default Plan

jivewise:
For: https://www.pivotaltracker.com/story/show/121087051

This adds a parameter to the set period function in the flux store.  This way, when the dates are updated, the defaultPlan of the store is only updated when the component is loaded, not when it's updated.

This is why the plan wasn't marked as changed, and the confirm dialog was not showing when cancelling the task plan changes.


jivewise:
This gets merged into #1080

PR #1096, opened 2016-06-13
Auto-save included/excluded exercises

nathanstitt:
Missed doing this as part of https://github.com/openstax/tutor-js/pull/1082

Also corrects the excluded mask color to match mockup


philschatz:
Wow! If I disconnect, click-to-exclude, and then try to reload the page I get an "Unsaved Changes" prompt. nice!

PR #1101, opened 2016-06-14
Use SVG icons for exercise action overlay, fix action display in IE

nathanstitt:
Deploying to tutor-dev for verification

screen shot 2016-06-22 at 10 28 59 am
screen shot 2016-06-22 at 10 28 48 am


rvnewell:
The Question Library looks great.  The intent was for the assets to be used in HW Builder as well.  How much of a challenge would it be to implement in HW Builder?

rvnewell:
One more question:  In the first QL screenshot, is the exclude question button in a hover state? It's appearing as a little larger than the question details button with what looks like a drop shadow.  Just want to confirm, since both assets need to be the same size. Will send another exclude question button if it's larger.

rvnewell:
Just to clarify:  they don't need to be same size when in hover state, just the assets need to be the same size in normal state.

rvnewell:
One more thing as a nice to have, but not critical : Could the action buttons in the details view left align with the left border of the question?  They are flush in HW builder, but not QL.

rvnewell:
@nathanstitt I actually prefer the smaller size of your details view action buttons.  I've adjusted my svg's down to 65px width and placed the new ones in the assets folder:  https://drive.google.com/drive/folders/0BxTx2rMemrLSNGU0aV9DY0lJNDA

My requests then are:
1) Replace details view buttons with new "_65px" button svg's in assets folder
2) Implement all new assets in HW Builder
3) Left align details view buttons with left border of question

nathanstitt:
@rvnewell - I'm unclear on what you mean by the smaller size?  The assets are SVG so they can scale to whatever size.  The only reason we need two different versions is because the text wraps in the smaller ones.

I was more looking for guidance on if I used the correct icon in the correct spot. I think I've got screenshots of each spot, the question library and hw builder both card and details view.

rvnewell:
Yes, the icons are in the correct spot in the Question Library.  The reason why I've supplied separate assets for the details view icons (smaller) and the card view icons (bigger), and now supplied a smaller version of the details view icons, is, because there are both text and graphic elements in the icons, each element needs to be sized differently when scaling up or down.  If you scaled the first svg's I supplied down, the font would become too small to be legible.  For the new smaller versions ("_65px") I scaled down the background and white circle behind the icon, but kept the font size and icon itself the same size.  I then repositioned everything manually.  I think this looks a lot better than just a scale up or down. 

Also, the HW builder screenshots do show the older action buttons, not the assets, so I'm hoping to get the assets in there as well.

philschatz:
Why is there text in the SVG? If the text was removed from the SVG would we still need 2 sizes for icons? (i18n becomes a problem if icons have text in them)

If it still is necessary, then should we use a convention like `mega-` for the large icons. ie https://octicons.github.com/icon/issue-closed/

rvnewell:
We will need two button sizes in any case:  The bigger ones for the card view (where the buttons are much larger and in the center of the question card) and the smaller ones for the details view (where they are positioned to the side of the scrollable question).  They are sized to match two very different layouts. The assets I supplied are buttons -- icons with text.  The icons need text to accompany them so that they make sense.  (e.g An eye icon accompanies "Preview feedback" in the preview feedback button.  An eye icon by itself will be confusing.) 

I supplied the entire button as an svg, thinking that 1) it might be easier to implement and 2) I can fine-tune the sizing and placement of the icons and text within the buttons.  However, if that's problematic I can simply supply the icon only.

rvnewell:
@philschatz I'm afraid I may have misunderstood your question.  In short, text is in the svg's because I thought the buttons themselves as two separate sizes of .svg's would be what's needed, because we're dealing with 2 sets of buttons with 2 slightly different layouts. This is because of the two different views (card and details).  (e.g. scaling down the exclude question button in the card view will not look the same as the exclude question button in the details view.)  If I only supply the icons, then, yes, I'd think there's no need to have 2 different .svg's for each button.  However, font, icon, and button size and dimensions would need to match the mockup for both the card and question details view.

I'll pull the icons out of the button .svg's and upload them to the assets folder separately.

nathanstitt:
@philschatz / @rvnewell let's discuss the priority of this a bit more before you go to the work of re-doing the SVG. If FE is in bug-fixing only mode, then I'm not sure we'll be able to do the update at this point.

rvnewell:
Understood that this isn't a priority given other tickets, but I have uploaded new SVG's to the assets folder (only took about 15 min).  Two sets -- icons with the white circular backgrounds and icons without the backgrounds.  I'm not sure which are the best to use here, but I would be for whichever set comes closest to matching the mockups and visual spec.

PR #1102, opened 2016-06-15
HW Builder bug fixes

nathanstitt:
Reorganizes hw builder to have components in separate files and fixes:

 * [x] Excluded exercises were still offered as options
 * [x]  `undefined id` bug on review page
 * [x]  clicking "choose more" has no effect

... and maybe more as I look through the bugs


pandafulmanda:
that scroll to exercises after loading is reaaaal nice @nathanstitt

PR #1104, opened 2016-06-15
Student Scores reading links

darrenmason:
regarding: https://www.pivotaltracker.com/story/show/121453633 

Reading cells can link back to reading as before, via popover.

reading-link

### DIFF
--- a/resources/styles/components/scores/table.less
+++ b/resources/styles/components/scores/table.less
@@ -383,7 +383,7 @@
         height: @completed-popover-row-height;
         vertical-align: middle;
         display: table-cell;
-        padding-left: 20px;
+        text-align: center;
       }
     }
   }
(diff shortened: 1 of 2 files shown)


philschatz:
So, clicking the pie chart itself doesn't open the reading/homework? /cc @openstax/ux

rvnewell:
I'm thinking we confirmed this in ux-review on Slack, but just in case:  Correct, clicking pie chart does not open reading or homework. Hover changes to tap behavior on tablets.  

As you may know, this is a compromise.  In earlier iterations, we displayed scores for readings.  In those versions, clicking scores for both reading and hw took users straight to the assignment.  We tested this with FAB, and all of them said they would click the score if they wanted to go to the assignment.  Clicking the score, as tested, seems to be the most intuitive option for going to an assignment.  More recently, the requirements changed and we had to remove reading scores from the UI. We reviewed multiple options with the PO and, as a compromise of consistency and reducing redundancy (first design option was to replace reading score with a Review link), PO requested and approved this design.

PR #1105, opened 2016-06-15
Fixed HW spinning forever after ecosystem update

Dantemss:
https://www.pivotaltracker.com/story/show/121532729

Needs https://github.com/openstax/tutor-server/pull/1142

This should take care of HW spinning forever after ecosystem updates. Might need a different PR to prevent people from selecting exercises then unselecting their pages, which I think will also mess up the display.


Dantemss:
My local testing confirms that this is fixed for ecosystem updates but is still a problem if I select exercises, unselect their pages, save and then force reload everything.

PR #1109, opened 2016-06-16
Datepicker fix label overlap when invalid date manually entered

darrenmason:
regarding: https://www.pivotaltracker.com/story/show/121087871

tutor-date-input allowed bypassing the datepicker with manual input, which broke the animated label state.


philschatz:
Nice! Only problem I see now is that if you tab through the dates you can have 2 calendars open at the same time but maybe that's a separate bug?

PR #1111, opened 2016-06-17
Completion Report for Immediate/Delayed Feedback 

jivewise:
For: https://www.pivotaltracker.com/story/show/121083971

Immediate feedback report: 
screen shot 2016-05-19 at 9 00 17 am

Delayed feedback report:
screen shot 2016-06-15 at 9 09 23 am

### DIFF
--- a/resources/styles/components/task-step/ends.less
+++ b/resources/styles/components/task-step/ends.less
@@ -12,14 +12,14 @@
     .tutor-completed-message();
     .size-exercise-card();
 
-		ul li {
-			text-align: left;
-		}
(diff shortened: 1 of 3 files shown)


resources/styles/components/task-step/ends.less
```diff
@@ -11,6 +11,15 @@
   .completed-message {
     .tutor-completed-message();
     .size-exercise-card();
+
+		ul li {
```
philschatz: /me wonders if we should include a `.editorconfig` file... http://editorconfig.org/

PR #1114, opened 2016-06-17
Fix/is loading timeouts

pandafulmanda:
and some refactoring teehehehehe and making the app store collect a little more info.


philschatz:
The refactor looks promising 😄 . Are there any references to jQuery left in the code?

pandafulmanda:
I believe...we...dont have any, i dont think

PR #1115, opened 2016-06-17
add editorconfig for formatting files

philschatz:
Most text editors support this file. See https://atom.io/packages/editorconfig


jivewise:
👍  I think the default formatting in my .vimrc has been messing everyone else up :(

PR #1117, opened 2016-06-17
the plan's server returned is_feedback_immediate being overwritten

pandafulmanda:
…by the default setting of is_feedback_immediate when dropdown select is untouched.  however, server side was returning the expected `is_feedback_immediate` value for whatever the last save was

Amendment to fix for [Unable to change when feedback appears](https://www.pivotaltracker.com/n/projects/1156756/stories/121620619)

the problem with #1112 is that if you never touch the dropdown, the `is_feedback_immediate` defaults to `true` despite the UI suggesting that it will be set to `false`.  This means if that you do not touch the dropdown, all homeworks will default to  `is_feedback_immediate: true`.

screen shot 2016-06-17 at 4 10 17 pm
screen shot 2016-06-17 at 4 10 55 pm

Why does this need to be set on `_change`?  because of this: https://github.com/openstax/tutor-js/blob/master/src/api.coffee#L138
The payload when `TaskPlanActions.save` is triggered is `_change`.  At some point we should clean up the `task-plan` flux and make it behave like our other standard crud fluxes.

See #1065 and #1069 for additional history

### DIFF
--- a/src/flux/task-plan.coffee
+++ b/src/flux/task-plan.coffee
@@ -374,7 +374,6 @@ TaskPlanConfig =
 
     isFeedbackImmediate: (id) ->
       plan = @_getPlan(id)
-      console.info('is fb immediate',  plan?.is_feedback_immediate)
       plan?.is_feedback_immediate
 
     getEcosystemId: (id, courseId) ->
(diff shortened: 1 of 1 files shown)


src/flux/task-plan.coffee
```diff
@@ -374,6 +374,7 @@ TaskPlanConfig =
 
     isFeedbackImmediate: (id) ->
       plan = @_getPlan(id)
+      console.info('is fb immediate',  plan?.is_feedback_immediate)
```
philschatz: Can you remove the `console.log`?

pandafulmanda: ooh oops, yes

PR #1119, opened 2016-06-17
Fix clicking scrolling to section on sectionizer click in HW builder

nathanstitt:
Scrolling was extremely sluggish when the control was clicked.  I traced it back to the card view was re-rendering excessively because the pure render wasn't having any effect.

Using `_.isEqual` seems to work well and improves performance considerably.

### DIFF
--- a/src/components/task-plan/homework/add-exercises.cjsx
+++ b/src/components/task-plan/homework/add-exercises.cjsx
@@ -10,6 +10,7 @@ ExerciseHelpers  = require '../../../helpers/exercise'
 ExerciseControls = require './exercise-controls'
 ExerciseDetails  = require '../../exercises/details'
 ExerciseCards    = require '../../exercises/cards'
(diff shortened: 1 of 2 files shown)


src/components/task-plan/homework/add-exercises.cjsx
```diff
@@ -81,8 +82,10 @@ AddExercises = React.createClass
     TaskPlanStore.hasExercise(@props.planId, exercise.id)
 
   setCurrentSection: (currentSection) ->
+    @scrollToSelector("[data-section='#{currentSection}']")
```
philschatz: Hmm, I wonder if this dependency could/should be documented in the Component that generates the `data-section` attribute; so we don't forget that there's this dependency

nathanstitt: yes, good idea.  Just updated `src/components/exercises/cards.cjsx` to note the importance of the `data-section` attribute

PR #1125, opened 2016-06-23
style table captions to match CNX

nathanstitt:
Before:
screen shot 2016-06-23 at 1 21 04 pm

After:
screen shot 2016-06-23 at 1 20 52 pm

### DIFF
--- a/resources/styles/mixins.less
+++ b/resources/styles/mixins.less
@@ -233,44 +233,6 @@
   }
 }
 
-.tutor-tables() {
-  table {
-    width: 100%;
-    margin-bottom: 32px;
-    #fonts > .sans(1.75rem, 3.6rem);
-    padding: 20px 0;
-    line-height: 2.5rem;
-    thead {
-      th {
-        padding: 8px 10px;
-        border-bottom: solid 2px @tutor-neutral-light;
-      }
-      tr:first-of-type:not(:last-child) th {
-        padding-left: 0;
-      }
-    }
-    tr:nth-child(odd) {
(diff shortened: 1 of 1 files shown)


philschatz:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;Mz5xIzhL&#x2F;1-aksjhdaksjhd"> aksjhdaksjhd</a>

philschatz:
It seems like `Examples ... Variables` should be *bold* but not the rest; is that possible?

nathanstitt:
Yes, we could use `first-word` to do so.   CNX doesn't do that though and I attempted to match their styles more or less.  See https://cnx.org/contents/r-QzKsl_@7.23:QMLV7p9L@4/Research-Methods for an example. Search for `Main Sociological Research Methods`  to find table.

philschatz:
Hmm, I was ambiguous... it seems like the 1st part of the table caption (`Examples ... Variables`) is actually the Table Title while the rest (`Typically ... way.`) is the caption. Are they somehow marked up as separate fields in the table caption so that the Table Title is bold while the caption is not bold?

nathanstitt:
Ah! I see now.  Yes, the first bit does seem to have a `data-type='title'` attribute.  We could style it as bold by targeting that.  Note that CNX isn't doing so though, see: https://cnx.org/contents/2zcpYNr4@4/Approaches-to-Sociological-Res for the matching example against the screenshots.  

I've done so and it's now:

screen shot 2016-06-27 at 4 56 43 pm

PR #1122, opened 2016-06-22
Various builder bugs

pandafulmanda:
### Fixes include

- [x] "Set as default" will show as expected for valid times
- [x] "Default set" message will show for a bit after default set
- [x] cursor does not jump when typing time
- [x] default times fill in as default times as expected when making new assignment and switching between all vs individual periods
- [x] show individual periods by default if default times for periods are different from each other
- [x] only subscribe to job status when publish is in progress
- [x] cancel will better track if there's been changes
- [x] if new assignment being created without default due date, due date will remain blank until deliberately selected
- [x] datepicker blur will return input label to expected location

PR #1127, opened 2016-06-23
address comments from #1122

pandafulmanda:


nathanstitt:
Very nice!

src/components/task-plan/builder/date-time.cjsx
```diff
@@ -80,7 +80,7 @@ DateTime = React.createClass
     @state?.date? and _.isEmpty(@refs?.date?.state?.errors)
 
   isTimeValid: ->
-    @state?.time? and _.isEmpty @refs?.time?.refs?.timeInput?.state?.errors
+    @state?.time? and _.isEmpty(@refs?.time?.refs?.timeInput?.state?.errors)
```
nathanstitt: Hey, I just looked at this a bit closer and noticed that `@ref.time` is set to the `TutorTimeInput` which looks like it will always render.  What do you think about adding a `isEmpty` method to  `TutorTimeInput`.  Then this could be just `@state?.time? and @refs.time.isEmpty()`

jivewise:
These changes look good to me.

PR #1132, opened 2016-06-27
Fix/deep server task plan copy

pandafulmanda:
To fix a bug i made recently to prevent posting of unchanged tasking-plans, especially important for when due dates are untouched but in the past.

this will keep that fix, but also fix a bug in which the "server_copy" of the task plan was being unintentionally mutated and making the change comparisons wrong


nathanstitt:
Looks good, I was able to save an un-edited assignment without errors

PR #1133, opened 2016-06-28
Prevent duplicate loads when reload is triggered

nathanstitt:
This was discovered by new code in `TeacherTaskPlanListing` which has a `LoadableItem` for the dashboard. The Dashboard's props are updated quite a few times (300+) when it's loaded directly. 

`LoadableItem` in turn has a `componentDidUpdate` that calls it's `reload` method every time, which blindly fires a load request.  This adds a check to only fire the load if it's not already loading.

I **think** this is safe but should be reviewed carefully since it does change the behavior of `LoadableItem`, which is used pervasively throughout our code-base.

Intended to prevent shenanigans like this:

screen shot 2016-06-28 at 10 11 29 am

### DIFF
--- a/src/components/loadable-item.cjsx
+++ b/src/components/loadable-item.cjsx
@@ -40,7 +40,7 @@ module.exports = React.createClass
 
     load ?= actions.load
     isLoading ?= store.isLoading
-    unless store.isNew(id, options) or isLoading(id, options)
+    unless store.isNew(id, options) and not isLoading?(id, options)
       load(id, options)
 
   render: ->
--- a/src/components/loadable-item.cjsx
+++ b/src/components/loadable-item.cjsx
@@ -31,7 +31,7 @@ module.exports = React.createClass
   componentDidUpdate: (oldProps) -> @reload(oldProps)
 
   reload: (oldProps) ->
-    {id, store, load, actions, options} = @props
+    {id, store, load, actions, options, isLoading} = @props
     return unless id?
 
     # Skip reloading if all the props are the same (the case in the Calendar for some reason)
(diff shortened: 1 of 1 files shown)


philschatz:
Nice! Is this indicative of a bigger bug? Like why would the screen be re-processing (recreating the VirtualDOM) hundreds of times per second?

pandafulmanda:
i dont think this happens anywhere but on local right? but looks like a good fix either way

nathanstitt:
I'm not sure if it's a local-only bug or not.  Would like to test on -dev, but it's currently down :(

src/components/loadable-item.cjsx
```diff
@@ -31,15 +31,16 @@ module.exports = React.createClass
...
     return unless id?
 
     # Skip reloading if all the props are the same (the case in the Calendar for some reason)
     return if oldProps.id is id and oldProps.store is store and oldProps.actions is actions and
       oldProps.load is load and _.isEqual(oldProps.options, options)
 
     load ?= actions.load
-    unless store.isNew(id, options)
+    isLoading ?= store.isLoading
+    unless store.isNew(id, options) and not isLoading?(id, options)
```
pandafulmanda: hmmm...i see how this is the same but would `unless isLoading?(id, options) or store.isNew(id, options)` be more readable?

nathanstitt: great point.  your version does read better.  I've updated to do that

nathanstitt: Yeah... so I tried that and it didn't work out well.  there's code in the task plans that check if the stores are reading & homework specs that expect the store to load unless it's new.  My moving the isLoading condition first the isNew status isn't checked.  

I could fixup the specs but am worried that there might be other untested areas that have the same logic.  Probably best to leave as is despite the better readability of your version

nathanstitt:
Closing in favor of #1139 which fixes this issue by doing the check in api.coffee

PR #1134, opened 2016-06-28
Exercise Preview Wrapper

jivewise:
Adding in a preview wrapper to the exercise cards that will re-render the exercise card instead of all the cards that are rendered.

### DIFF
--- a/src/components/exercises/cards.cjsx
+++ b/src/components/exercises/cards.cjsx
@@ -13,6 +13,39 @@ ScrollTo = require '../scroll-to-mixin'
 ChapterSection = require '../task-plan/chapter-section'
 Icon = require '../icon'
 
+PreviewWrapper = React.createClass
+
+  propTypes:
+    exercise:               React.PropTypes.object.isRequired
+    onShowDetailsViewClick: React.PropTypes.func.isRequired
+    onExerciseToggle:       React.PropTypes.func.isRequired
+    getExerciseIsSelected:  React.PropTypes.func.isRequired
+    getExerciseActions:     React.PropTypes.func.isRequired
+    watchStore:             React.PropTypes.func
+    watchEvent:             React.PropTypes.string
+
+  componentWillMount: ->
+    { watchStore, watchEvent, exercise } = @props
+    watchStore?.on("#{watchEvent}#{exercise.id}", @update)
+  componentWillUnmount: ->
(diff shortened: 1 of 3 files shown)


src/components/task-plan/homework/add-exercises.cjsx
```diff
@@ -98,6 +98,8 @@ AddExercises = React.createClass
 
     sharedProps =
         exercises: exercises.all
+        watchStore: TaskPlanStore
+        watchEvent: 'change-exercise-'
         onExerciseToggle: @onExerciseToggle
```
nathanstitt: Can we move the `watch..` props from `sharedProps` and directly onto `ExeriseCard` below?  That will make it more clear that they are `ExerciseCard` specific and not used by `ExerciseDetails`.

nathanstitt: I'm hopeful that this will allow us to remove my hacky `selectCount={SELECT_COUNT} # re-render flag` stuff on line 121 below.  Can you remove that and see if HW builder still allows you to select exercises?

nathanstitt:
Nice idea @jivewise!  I like this a ton better than my `SELECT_COUNT` method.

src/components/task-plan/homework/add-exercises.cjsx
jivewise: Will do, I'll remove from `sharedProps`.  I already tested without the select_count stuff there, and it seemed to be fine.

src/components/exercises/cards.cjsx
```diff
@@ -13,6 +13,39 @@ ScrollTo = require '../scroll-to-mixin'
...
+      key={exercise.id}
+      className='exercise-card'
+      isInteractive={false}
+      isVerticallyTruncated={true}
+      isSelected={@props.getExerciseIsSelected(exercise)}
+      exercise={exercise}
+      onOverlayClick={@onExerciseToggle}
+      overlayActions={@props.getExerciseActions(exercise)}
+    />
+
```
pandafulmanda: looks a lot like `BindStoreMixin` which can be customized to listen to custom events

nathanstitt: I agree, seems like we could use `BindStoreMixin` here, no?

PR #1135, opened 2016-06-28
fix not being able to make new assignments

pandafulmanda:
fix the bug from #1132 so that new assignments  dont trip up the get original


nathanstitt:
👍  @pandafulmanda

PR #1136, opened 2016-06-28
plan was saving as draft...

pandafulmanda:
but changing from month to month wasnt updating the publish status

### DIFF
--- a/src/components/course-calendar/plan.cjsx
+++ b/src/components/course-calendar/plan.cjsx
@@ -55,21 +55,31 @@ CoursePlan = React.createClass
     activeHeight: 35
 
   getInitialState: ->
-    {item} = @props
+    @getStateByProps()
+
+  getStateByProps: (props) ->
+    props ?= @props
+
+    {item} = props
     {plan} = item
 
     publishStatus = PlanPublishStore.getAsyncStatus(plan.id)
 
-    isViewingStats: @_doesPlanMatchesRoute()
+    isViewingStats: @_doesPlanMatchesRoute(props)
     publishStatus: publishStatus
     isPublishing: PlanPublishStore.isPublishing(plan.id)
     isHovered: false
     isPublished: plan.is_published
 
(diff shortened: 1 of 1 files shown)


src/components/course-calendar/plan.cjsx
```diff
@@ -120,15 +130,22 @@ CoursePlan = React.createClass
...
+
     location = @context.router.getLocation()
     location.addChangeListener(@syncRoute)
 
   componentWillReceiveProps: (nextProps) ->
     if @props.item.plan.id isnt nextProps.item.plan.id
-      @subscribeToPublishing(nextProps.item.plan)
+      publishState = PlanHelper.subscribeToPublishing(nextProps.item.plan, @checkPublishingStatus)
       @stopCheckingPlan(@props.item.plan)
+      state = _.extend(@getStateByProps(nextProps), publishState)
```
nathanstitt: If a prop other than plan was updated, `nextProps.plan` will be the same as `@props.plan`.

In that case, this will subscribe and then immediately unsubscribe from it when `@stopCheckingPlan` is called.  Maybe `stopCheckingPlan` should be called first?

pandafulmanda: it subscribes to the new plan, and then unsubscribes to the old plan, and only if it passes through [`@props.item.plan.id isnt nextProps.item.plan.id`](https://github.com/openstax/tutor-js/pull/1136/files#diff-60a357506dabb2d734615a51de235068L129)

nathanstitt: ah, yep that makes sense.  I hadn't factored that line in my internal logic.   I still think it'd be a good idea to flip the `subscribe` and `stopChecking` calls though, unless they're in that order for a reason.  That way 1) other's won't make the same mistake I just did, and 2) it'll be a bit more future proof if the `plan.id` check is ever removed.

pandafulmanda:
ah...i was going to change this, i guess i make a tiny pr for this

PR #1138, opened 2016-06-28
Fix exercise selection slowness, apply exclusions before save

nathanstitt:
When exercise expulsions were changed, the highlights weren't applied until the cards were rendered twice.  once for when they save started and once after it completed.  Since that's a slow operation, testers would often double-click.  

The second click would be applied before the store was updated, causing multiple identical saves and a BE 500 error when it attempted to mark an already excluded exercise as excluded once again.


nathanstitt:
@pandafulmanda can you review please?

src/components/bind-store-mixin.cjsx
```diff
@@ -1,12 +1,14 @@
+_ = require 'underscore'
+
 module.exports =
 
   # can modify which event you want to bind on as needed.
   _bindEvent: ->
-    @bindEvent or @props.bindEvent or 'change'
+    _.result(@, 'bindEvent', @props.bindEvent or 'change')
```
pandafulmanda: COOL! didnt know about `_.result`

src/flux/exercise.coffee
```diff
@@ -109,7 +109,7 @@ ExerciseConfig =
   saveExerciseExclusion: (courseId, exerciseId, isExcluded) ->
     @_exclusionsAsyncStatus = SAVING
     @_unsavedExclusions[exerciseId] = isExcluded
-    @emitChange()
+    @emit("change-exercise-#{exerciseId}")
```
pandafulmanda: yayyy specific events

pandafulmanda:
waaay smoother, almost like butter

PR #1139, opened 2016-06-28
Fix/api prevent multiple same requests

pandafulmanda:
Alternative to #1133, handles and prevents multiple for all requests as opposed to just those in loadable

fix is in ea53f8c, some refactoring in 35accc7, can be reset if needed.

### DIFF
--- a/src/helpers/api.coffee
+++ b/src/helpers/api.coffee
@@ -33,7 +33,9 @@ makeLocalRequest = (requestConfig) ->
     url_parts = ["#{uri}.json", params]
   else
     url_parts = ["#{uri}/#{requestConfig.method}.json", params]
+    requestConfig.mockMethod = requestConfig.method
     requestConfig.method = 'GET'
+
 
   requestConfig.url = _.compact(url_parts).join('?')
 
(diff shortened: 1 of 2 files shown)


src/helpers/api.coffee
```diff
@@ -33,42 +33,100 @@ makeLocalRequest = (requestConfig) ->
...
+
+    Promise.reject(errorResponse)
+
+  handleLocalErrors: (errorResponse) ->
+    {status, statusText, config} = errorResponse
+    {method, mockMethod} = config
+
+    mockMethods = ['PUT', 'PATCH']
+
+    # Hack for local testing, fake successful PUT and PATCH
```
pandafulmanda: I kinda think it's better not to have this now that `makeRequestLocal` handles this, or this should return as a different error that makes it clear where the fake `json` should be made to handle the fake request.

src/helpers/api.coffee
```diff
@@ -33,42 +33,100 @@ makeLocalRequest = (requestConfig) ->
...
+    if _.contains(mockMethods, mockMethod or method)
+      data = JSON.parse(errorResponse.config.data)
+      return Promise.resolve({data})
+
+    # Hack for local testing. Webserver returns 200 + HTML for 404's
```
pandafulmanda: I don't think our FE throws this error anymore?

nathanstitt: I agree - I think this is unused

nathanstitt:
I like this better than my #1133 - it also matches what we did on CC.   If @philschatz agrees i'll close the other

nathanstitt:
I likes, It fixes the dashboard triggering 300+ loads.  Am merging so it can be in next round of testing

PR #1140, opened 2016-06-28
Scores - fixed absent cell type displaying incorrectly

darrenmason:
regarding: https://www.pivotaltracker.com/story/show/121986485

when student assignment data is null (if a student moves and is unassigned), absent cell will render with either split (hw) or single (external, reading), based on the heading type.


nathanstitt:
thanks, looks good @darrenmason

PR #1141, opened 2016-06-29
Fixing source maps

jivewise:
This PR fixes source maps so they can be parsed, I believe it's actually a chrome error.  From this conversation here: https://github.com/webpack/webpack/issues/2145, I've changed webpack to use: `devtool: '#inline-source-map'` to fix the issue.

### DIFF
--- a/webpack.config.coffee
+++ b/webpack.config.coffee
@@ -20,7 +20,7 @@ else
 module.exports =
   cache: true
 
-  devtool: 'source-map' # Always include source maps!
+  devtool: '#inline-source-map' # Always include source maps!
 
   entry:
     tutor: [


webpack.config.coffee
```diff
@@ -20,7 +20,7 @@ else
 module.exports =
   cache: true
 
-  devtool: 'source-map' # Always include source maps!
+  devtool: '#inline-source-map' # Always include source maps!
```
nathanstitt: I don't think we can use `inline-source-map`.  https://webpack.github.io/docs/configuration.html says they'll be included as a `DataUrl`, which I believe means they're base64 encoded and included in the actual file.

If we do that then I **think** production will include the source maps in the build for all users, making the download size larger.  Currently it's supposed to be a URL to the source map, which is only download when the dev tools source console is opened.

Maybe I'm incorrect though.  We should probably deploy to dev to test

jivewise: Hm, okay, I'll test on dev.  If that's true though, would we rather disable source maps in other environments other than dev, or keep them, and just not use chrome to debug?  I believe the source maps are fine in firefox.

philschatz: Yep, agreed on not including them inline (because of file size). From the webpack PR it looks like Chrome Canary might have fixed it? https://github.com/webpack/webpack/issues/2145#issuecomment-224368361 or chunking the files into smaller pieces ; (

I think it's important to try to keep the source maps in production though

jivewise: Okay, I think we can probably close this PR then.

PR #1142, opened 2016-06-29
Fixes "Bug: Probing an Electric Field sim doesn't fit in its frame"

pwolfert:
Fixes the problem where the embed frame gets wrapped around interactives when it shouldn't.  This makes the condition based on the type of step instead of trying to figure out what it is based on the URL.  There is an additional sister commit being made in react-components.


test/components/task-content.spec.coffee
```diff
@@ -0,0 +1,43 @@
...
+
+  it 'should render a frame for non-interactives', ->
+    # render the step and then make sure it doesn't have a frame around it
+    props = {
+      id: interactiveStepId
+      stepType: 'interactive'
+    }
+
+    Testing.renderComponent(StepContent, props: props).then ({dom}) ->
+      expect(dom.querySelector('.frame-wrapper')).to.not.be.null
```
nathanstitt: A huge nitpick: I like using `exist`, since it matches both `null` and `undefined`:
`expect(dom.querySelector('.frame-wrapper')).to.exist`

src/components/task-step/step-with-reading-content.cjsx
```diff
@@ -42,11 +42,14 @@ StepContent = React.createClass
   getCnxId: ->
     TaskStepStore.getCnxId(@props.id)
 
+  shouldExcludeFrame: ->
+    (@props.stepType == 'interactive')
+
```
nathanstitt: We typically prefer to use: `is` vs `==`, and the parens aren't really needed.  Could be:
```
shouldExcludeFrame: ->
  @props.stepType is 'interactive'
```

test/components/task-content.spec.coffee
```diff
@@ -0,0 +1,43 @@
+# Tests for homework specific tasks
+
+# -----------------------------------------------------------------------------------------------------
+# TODO: For some reason the StepContent is undefined in the test, and therefore this code doesn't work.
+# -----------------------------------------------------------------------------------------------------
+
```
nathanstitt: Is this comment accurate?  Travis seems to think the spec is passing?

nathanstitt:
Nice @pwolfert !  Glad you figured out the confusing Test stuffs :)

PR #1143, opened 2016-06-29
Set defaults when switching from periods to combined

nathanstitt:
This fixes a bug where the HW builder would have invisible errors when validating.  The sequence to trigger was:

* Have some differing default times set for periods
* Load the builder 
* Note that individual periods are displayed at the start
* switch to "All Periods" and set the end date
* set title and add some exercises (this can be done at any time)
* attempt to publish

Nothing would happen because the opens_at wasn't set on the taskings, because that date was never manipulated.

### DIFF
--- a/src/flux/task-plan.coffee
+++ b/src/flux/task-plan.coffee
@@ -605,7 +605,6 @@ TaskPlanConfig =
 
     getEnabledTaskings: (id) ->
       plan = @_getPlan(id)
-      console.log "Get Enabled", plan
       plan?.tasking_plans
 
     isStatsLoading: (id) -> @_asyncStatusStats[id] is 'loading'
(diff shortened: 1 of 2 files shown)


src/flux/task-plan.coffee
```diff
@@ -168,6 +172,12 @@ TaskPlanConfig =
     tasking_plans = @_getDefaultTaskingTimes(id, courseId, periods)
     @_change(id, {tasking_plans})
 
+  # set taskings that lack an opens_at to the default
+  setBlankOpensAtDateForCourse: (id, courseId, opensAtDate) ->
```
pandafulmanda: this doesnt seem to be being used?

nathanstitt: yeah, that was an initial stab at setting the date, then I noticed the stuff in the component was doing it

PR #1146, opened 2016-06-30
Hide missing note title pseudo element

nathanstitt:
Previously we decided against this since the weird blank element makes it obvious that the content needs updating, but Sociology is missing them all, making it impossible to fix by release.

before:
screen shot 2016-06-30 at 9 21 40 am

after:
screen shot 2016-06-30 at 9 21 18 am

### DIFF
--- a/resources/styles/book-content/note.less
+++ b/resources/styles/book-content/note.less
@@ -9,6 +9,11 @@
     display: inline-block;
     margin: @book-content-note-with-background-margin;
   }
+  // hide note title if data-label attribute is blank or missing
+  &[data-label=""],
+  &:not([data-label]) {
+    &:before{ display: none; }
+  }
 
   :last-child {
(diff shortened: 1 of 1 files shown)


resources/styles/book-content/note.less
```diff
@@ -9,6 +9,11 @@
     display: inline-block;
     margin: @book-content-note-with-background-margin;
   }
+  // hide note title if data-label attribute is blank or missing
```
pandafulmanda: why not move this up to wrap the `&:before` above so that it's like this:

```css
&[data-label]:not([data-label=""]) {
  &:before {
    blah: blahblah;
  }
}
```

nathanstitt: Yes, great point.  I've done that now.  Thanks!

PR #1148, opened 2016-06-30
Better wording for the note about scores past the due date

pwolfert:
https://www.pivotaltracker.com/n/projects/1156756/stories/123189821


pwolfert:
Now I'm able to test locally, and it works.

PR #1152, opened 2016-07-01
Adds in message for when no exercises are found

jivewise:
Pivotal: https://www.pivotaltracker.com/story/show/121524179

screen shot 2016-07-01 at 2 03 17 am

### DIFF
--- a/resources/styles/components/exercises.less
+++ b/resources/styles/components/exercises.less
@@ -78,6 +78,9 @@
 
 }
 
+.no-exercises-found {
+	text-align: center;
+}
 
 .exercise-controls-bar {
   @padding: 20px;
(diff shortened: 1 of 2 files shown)


resources/styles/components/exercises.less
```diff
@@ -78,6 +78,9 @@
 
 }
 
+.no-exercises-found {
+	text-align: center;
```
philschatz: seems like this is missing a space character 😄

src/components/exercises/cards.cjsx
```diff
@@ -68,12 +69,20 @@ ExerciseCards = React.createClass
...
+      <SectionsExercises key={cs}
+        {...@props}
+        chapter_section={cs}
+        exercises={@props.exercises.grouped[cs]}
+      />
+    )
+
+    sections = <p className="no-exercises-found">
+      No exercises found in the selected sections.
+    </p> if not chapter_sections.length
```
philschatz: Could these be `if .... else ...` to make the control flow more explicit?

PR #1153, opened 2016-07-01
This allows the same date to be set in the builder

jivewise:
Pivotal: https://www.pivotaltracker.com/story/show/123488831

This needs some UX love, if we want the same date, we'll need to show error message for when they enter the wrong time.  See screenshots below.

screen shot 2016-07-01 at 1 09 20 am
screen shot 2016-07-01 at 1 08 52 am

### DIFF
--- a/resources/styles/components/task-plan/builder.less
+++ b/resources/styles/components/task-plan/builder.less
@@ -75,6 +75,13 @@
   }
 }
 
+.due-before-open {
+	text-align: right;
+	line-height: 1.2rem;
+	margin: -10px 0 30px;
+}
+
+.due-before-open,
 .readings-required,
 .problems-required {
   color: @form-error-color;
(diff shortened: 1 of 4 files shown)


jivewise:
What does @openstax/ux think?

jason-holmes:
I think this is the way it needs to be...it's a field level error and the user's action is clear to fix the error.  It's good.

jivewise:
Okay, thanks @jason-holmes

pandafulmanda:
why is there so much space between the bottom of the fields and warning?  is it because that leaves space for the warnings about the time being misformatted/the time set default time label?

jason-holmes:
@pandafulmanda Yes, in Axure, it's sometimes hard to move stuff around (like inserting warning under the fields) so the prototype leaves more space than is necessary.

pandafulmanda:
Also, should the message be 'Due time cannot be before open time'.  i.e. in screenshot, would user be confused by 'Due date cannot be before open date` since the dates are the same there and they actually need to change the time?

jason-holmes:
Yes. "Due time cannot be before open time." is a better error message.

jivewise:
I needed to leave space below the input boxes for the set default time link, and the other error messages.  The margin bottom are on the tutor input containers, so I'm not sure we'll be able to get rid of that space.

I have updated the error messages though.

PR #1155, opened 2016-07-01
Fixes Loadable indicator so it re-renders less often (formerly "hey...so...")

pandafulmanda:
loadable item was triggering load AFTER `componentDidUpdate` and then only triggering a load when certain props had been loaded.  it seems like this trigger should happen sooner -- i.e. right when the component will receive the props.  this change should make things feel snappier and also decrease extra re-renderings due to first rendering since props updated and then rendering again once the store is loaded and emits a change

there are also stylistic updates -- putting methods on `@` and using `_.partial` from common practices that have developed since this was first written

_Hey, we release soon,
so this is crazy.
But here's this pr.
So merge it, maybe?_

### DIFF
--- a/src/components/loadable-item.cjsx
+++ b/src/components/loadable-item.cjsx
@@ -28,57 +28,75 @@ module.exports = React.createClass
   getDefaultProps: ->
     bindEvent: 'change'
 
-  componentDidMount: -> @reload({})
-  componentDidUpdate: (oldProps) -> @reload(oldProps)
+  componentWillMount: ->
+    {id, store, options} = @props
+    @load(id, options) if id? and not store.isNew(id, options)
 
-  reload: (oldProps) ->
-    {id, store, load, actions, options} = @props
+  componentWillReceiveProps: (nextProps) ->
+    @reload(@props, nextProps)
+
+  arePropsSame: (prevProps, nextProps) ->
+    {id, store, load, actions, options} = nextProps
+
+    prevProps.id is id and
+      prevProps.store is store and
+      prevProps.actions is actions and
+      prevProps.load is load and
+      _.isEqual(prevProps.options, options)
+
+  reload: (prevProps, nextProps) ->
+    {id, store, load, actions, options} = nextProps
     return unless id?
 
     # Skip reloading if all the props are the same (the case in the Calendar for some reason)
-    return if oldProps.id is id and oldProps.store is store and oldProps.actions is actions and
-      oldProps.load is load and _.isEqual(oldProps.options, options)
+    return if @arePropsSame(prevProps, nextProps)
+    @load(id, options) unless store.isNew(id, options)
 
+  load: (args...) ->
+    {load, actions} = @props
     load ?= actions.load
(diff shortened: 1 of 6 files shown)


philschatz:

pandafulmanda:
@philschatz, yup done!

src/components/loadable-item.cjsx
```diff
@@ -28,57 +28,75 @@ module.exports = React.createClass
...
+
+  arePropsSame: (prevProps, nextProps) ->
+    {id, store, load, actions, options} = nextProps
+
+    prevProps.id is id and
+      prevProps.store is store and
+      prevProps.actions is actions and
+      prevProps.load is load and
+      _.isEqual(prevProps.options, options)
+
```
nathanstitt: I'm a little confused by the explicit `xxx is yyy` comparisons.  Why not just use `_.isEqual(prevProps, nextProps)` here?  That'd be small enough that you could just do it inline vs having the `arePropsSame` method. 

I'm pretty sure there's a good reason since you went to the trouble though.  Maybe a good spot for an explanation comment 😁

jivewise:
👍  This makes the UI so much better.  Links and buttons used to move around after they got mounted and re-rendered, this makes it a lot better.

src/components/loadable-item.cjsx
```diff
@@ -28,57 +28,75 @@ module.exports = React.createClass
...
+    else if @isLoaded(id, options)
+      false
+    else if store.isUnknown(id, options) or store.reload(id, options)
+      true
+    else if store.isNew(id, options) and store.get(id, options).id and saved
+      # If this store was just created, then call the onSaved prop
+      saved()
+    else
+      false
+
```
nathanstitt: Might be a good spot to use a `switch` without value statement. http://coffeescript.org/#switch

```
switch 
  when store.get(id, options)     then false
  when if @isLoading(id, options) then true
  when @isLoaded(id, options)    then  false
  when ( store.isUnknown(id, options) or store.reload(id, options) ) then true
....
  else
      false
```

nathanstitt:
👍 from me - I think if this surfaces any bugs then it's probably the fault of the calling code and should be corrected there

pandafulmanda:
@nathanstitt, do these changes make sense?

src/components/loadable-item.cjsx
```diff
@@ -28,57 +27,71 @@ module.exports = React.createClass
...
-    {id, store, load, actions, options} = @props
+  componentWillReceiveProps: (nextProps) ->
+    @reload(@props, nextProps)
+
+  arePropsSame: (prevProps, nextProps) ->
+    {id, store, load, actions, options} = nextProps
+
+    propsToCheck = ['id', 'store', 'load', 'actions', 'options']
+    _.isEqual(_.pick(prevProps, propsToCheck), _.pick(nextProps, propsToCheck))
+
```
nathanstitt: I like - that's a really nice and concise way to check only the props you care about.

nathanstitt:
Looks great to me @pandafulmanda

PR #1158, opened 2016-07-05
date time separation from task-plans

pandafulmanda:
Specs remaining:
- [ ] finishing tasking flux
- [ ] tasking component
- [ ] tasking date time component
- [ ] date time component

Fixes:
* various bugs related to default times
* datepicker sometimes picking previous date from what's clicked
* due date and open date re-publishings weren't being submitted to the BE
* and probably lots more stuffs just hiding around

### DIFF
--- a/src/components/bind-stores-mixin.coffee
+++ b/src/components/bind-stores-mixin.coffee
@@ -1,0 +1,28 @@
+_ = require 'underscore'
+
+module.exports =
+
+  # can modify which event you want to bind on as needed.
+  _defaultEvent: 'change'
+
+  _defaultUpdate: ->
+    @setState({})
+
+  _addListener: (bind) ->
+    bind.store.on(bind.listenTo or @_defaultEvent, bind.callback or @_defaultUpdate)
+
+  _removeListener:  (bind) ->
+    bind.store.off(bind.listenTo or @_defaultEvent, bind.callback or @_defaultUpdate)
+
+  _addListeners: ->
+    bindEvents = @_getBindEvents?()
+    return if _.isEmpty(bindEvents)
+    _.each bindEvents, @_addListener
+
+  _removeListeners: ->
+    bindEvents = @_getBindEvents?()
+    return if _.isEmpty(bindEvents)
+    _.each bindEvents, @_removeListener
+
+  componentWillMount:   -> @_addListeners()
+  componentWillUnmount: -> @_removeListeners()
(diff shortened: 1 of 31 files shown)


src/components/bind-stores-mixin.coffee
```diff
@@ -0,0 +1,28 @@
...
+
+module.exports =
+
+  # can modify which event you want to bind on as needed.
+  _defaultEvent: 'change'
+
+  _defaultUpdate: ->
+    @forceUpdate()
+
+  _addListener: (bind) ->
```
philschatz: Would it be clearer if this was `_addListener: ({store, listenTo, callback}) ->`?

src/components/bind-stores-mixin.coffee
```diff
@@ -0,0 +1,28 @@
...
+    bind.store.on(bind.listenTo or @_defaultEvent, bind.callback or @_defaultUpdate)
+
+  _removeListener:  (bind) ->
+    bind.store.off(bind.listenTo or @_defaultEvent, bind.callback or @_defaultUpdate)
+
+  _addListeners: ->
+    bindEvents = @_getBindEvents?()
```
philschatz: It seems `_getBindEvents` must be defined on the `Component` that uses this mixin; if so, then should this not have the underscore (since it is not a private method)?

src/components/task-plan/builder/tasking-date-times.cjsx
```diff
@@ -16,20 +16,17 @@ TaskingDateTimes = React.createClass
...
-    return false unless date?
-    TimeHelper.makeMoment("#{date} #{time}").isBefore(TimeHelper.makeMoment())
+  getError: ->
+    return false unless @refs?.due?.hasValidInputs() and @refs?.open?.hasValidInputs()
+
+    {id, period} = @props
+
+    return false if TaskingStore.isTaskingValid(id, period)
+
+    _.first(TaskingStore.getTaskingErrors(id, period))
```
philschatz: When would there be more than 1 error?

pandafulmanda: the first one is the right one to display as you would not expect the user to try to address all errors at once.  errors could include misformatted date, misformatted time, and time range being wrong.

jivewise:
One thing I noticed when testing out the builder:
1. Enter all other information (name, date, readings/exercises, etc.)
2. Select "Individual Periods"
3. Unselect all periods
4. Click save

Notice that even though an error message displays, it still tries to save, and results in a 422 error.

pandafulmanda:
@jivewise good one, i can fix that quick

pandafulmanda:
okay @jivewise i've pushed in a fix, let me know if you have a chance to see if that fixes it

jivewise:
Will do.

PR #1159, opened 2016-07-05
change label from "Average" to "Percent Complete" on teacher review

nathanstitt:
Since average isn't really what it's calculating and it didn't match student scores.  Adds a info icon with additional explanation.

Quick metrics:
screen shot 2016-07-05 at 2 04 56 pm

full metrics version:
screen shot 2016-07-05 at 2 05 41 pm

### DIFF
--- a/src/components/icon.cjsx
+++ b/src/components/icon.cjsx
@@ -24,6 +24,7 @@ module.exports = React.createClass
   render: ->
     classNames = classnames('tutor-icon', 'fa', "fa-#{@props.type}", @props.className, {
       'fa-spin': @props.spin
+      'clickable': @props.tooltip and @props.tooltipProps.trigger is 'click'
     })
 
     icon = <i {...@props} className={classNames} />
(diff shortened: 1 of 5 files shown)


src/components/icon.cjsx
```diff
@@ -24,6 +24,7 @@ module.exports = React.createClass
   render: ->
     classNames = classnames('tutor-icon', 'fa', "fa-#{@props.type}", @props.className, {
       'fa-spin': @props.spin
+      'clickable': @props.tooltip and @props.tooltipProps.trigger is 'click'
```
philschatz: woah, pretty slick solution! I wonder how that affects other places where we use `<Icon>`...

nathanstitt: Hm, well it will add the `cursor: pointer` style, but I think that's a "good thing".  I attempted to find if there were spots in the styles where we were already setting it manually so they could be removed by using `grep -r -C5 cursor ./resources/styles*`, but didn't spot any.

PR #1160, opened 2016-07-05
Embedding Interactives

pwolfert:
This adds css rules for what to do when an interactive iframe is not inside an interactive-step.  In this case the content will be inset with thick padding in the card, and so an iframe would be wider than the content.  This makes sure it centers itself in the whole card's width.

To test, one can change `stimulus_html` in `step-id-9-1.json` to have this value:
`"<div data-type=\"note\" data-has-label=\"true\" id=\"hookes_law_5_3_2\" class=\"note watch-physics\" data-label=\"\">\n<div data-type=\"title\" class=\"title\">Watch Physics: 5.3 Elasticity:  Stress, and Strain</div>\n<div data-type=\"media\" id=\"fs-id1169085779740\" data-alt=\"Link to lab simulation tool on masses and springs.\">\n<iframe width=\"660\" height=\"371.4\" src=\"https://www.openstaxcollege.org/l/02sprigmasst\"></iframe>\n</div>\n</div>"`

This also requires new react-components code:
https://github.com/openstax/react-components/pull/91

### DIFF
--- a/resources/styles/components/task-step/all-steps.less
+++ b/resources/styles/components/task-step/all-steps.less
@@ -26,6 +26,15 @@
     }
   }
 
+  .task-step {
+    :not(.interactive-step) {
+      iframe.interactive {
+        margin-left: -120px; // The rest of the card has a lot of padding, and it won't fit unless we shift it over so it's centered
+      }
+    }
+
+  }
+
   .openstax-exercise-card {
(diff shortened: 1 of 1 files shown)


philschatz:
marking this as `blocked` because it's waiting on openstax/react-components#91

PR #1161, opened 2016-07-06
display an empty string instead of "Undefined"

nathanstitt:
When sections are not found.  This should never happen, but may if the ecosystem map fails.

Previously it would display as:

screen shot 2016-07-05 at 8 55 37 pm

Now it'll render as:
screen shot 2016-07-05 at 8 50 51 pm


philschatz:
There was some discussion in chat about what to show in this case; could/should it be `?.?`?

PR #1162, opened 2016-07-06
Reading Scroll To Top

jivewise:
Pivotal ticket: https://www.pivotaltracker.com/story/show/124223571

Scrolls the reading to the top when going to the next or previous:
reading-scroll


philschatz:
cool! Does this work in Safari/IE?

jivewise:
Yep, I tested in IE11, Safari, Firefox, and Chrome.  I only tested the last 3 on my mac though.  Maybe someone with a windows machine can verify FF and Chrome?

PR #1163, opened 2016-07-06
Worked Example styling on additional book content

darrenmason:
regarding: https://www.pivotaltracker.com/story/show/121780773

worked-example
worked-example-video

regarding: https://www.pivotaltracker.com/story/show/121781491

image


philschatz:
hmm, is there a way to only put the changes in 1 file? Having to update 3 files to style 1 feature seems to indicate a problem with how the CSS is organized generally. If we need to add more custom styling for each of the books then it seems like this could get hairy quickly

darrenmason:
I'll try combining them and see what happens; there might be a reason they were separate, but maybe not

philschatz:
Thanks! I think this is quite a bit easier to find what to edit, though the changes to `resources/styles/book-content/theming-mixins.less` seems to indicate that more could be pulled out (ideally the text `[data-element-type="check-understanding"]` should only occur in one file (probably `note.less`)

philschatz:

PR #1173, opened 2016-07-08
Book content - multiple choice styling - fixed classnames

darrenmason:
image

### DIFF
--- a/resources/styles/book-content/base.less
+++ b/resources/styles/book-content/base.less
@@ -84,7 +84,7 @@ figure {
   img { width: 100%; }
 }
 
-section {
+.multiple-choice, .section-quiz {
   counter-reset: exerciseCount;
   [data-type="exercise"] {
     [data-type="problem"],


philschatz:
It seems like the previous code:

1. reset exercise numbering whenever a `<section>` was encountered
1. added alpha numbering for all lists inside a CNXML exercise 

while this change:

1. resets the exercise whenever `.multiple-choice, .section-quiz` is encountered
2. adds alpha numbering to only `.multiple-choice, .section-quiz` elements that contain an exercise in them

If that's correct, then I think the previous code is "more correcter" ™️ and more versatile (it handles any CNXML exercise with a list of multiple choice options at the expense of correctly labeling an arbitrary CNXML list inside of an exercise)

/cc @openstax/ux @reedstrm (because he talked recently about auto-importing CNXML `<exercise>` elements into Exercises)

darrenmason:
The previous code added numbering and lettering to sections that for sure did not need them.  We use data-types of exercise, problem, and solution in sections that we do not necessarily want the new styling.

I confirmed with @kjdav that .multiple-choice, .section-quiz are the sections that need numbering/lettering:

https://www.pivotaltracker.com/story/show/121781627/comments/143441849

philschatz:
I'm opting for using the original #1169 (any exercise with a numbered list is lettered) so we do not have to remember to continue special-casing exercises whenever we add a new book to tutor. Thanks for sleuthing & hunting down the set of exercises we currently use!

PR #1174, opened 2016-07-08
Redirect a CC student to web view url of book

nathanstitt:
Blocked until @jivewise's PR #1172 lands

screen shot 2016-07-08 at 2 20 42 pm

### DIFF
--- a/src/components/countdown-redirect.cjsx
+++ b/src/components/countdown-redirect.cjsx
@@ -9,6 +9,7 @@ CountdownRedirect = React.createClass
     secondsDelay: 10
     message: "Redirecting"
     windowImpl: window
+    redirectType: 'replace'
 
   getInitialState: ->
     secondsRemaining: @props.secondsDelay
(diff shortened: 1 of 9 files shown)


src/components/countdown-redirect.cjsx
```diff
@@ -0,0 +1,45 @@
...
+
+  onCounterTick: ->
+    secondsRemaining = @state.secondsRemaining - 1
+    if secondsRemaining <= 0
+      @props.windowImpl.location[@props.redirectType](@props.destinationUrl)
+    else
+      @setState({secondsRemaining})
+
+  componentWillMount: ->
+    @setState( updateInterval: @props.windowImpl.setInterval(@onCounterTick, @props.secondsDelay * 100) )
```
philschatz: eip! I'm a little worried about sprinkling `setInterval` into the code (instead of having a central place that does "ticks"). Also:

- is the interval always cleared?
- Does react like to call `setState` just before mounting?

src/components/navbar/logout.cjsx
```diff
@@ -8,6 +8,12 @@ CSRF_Token = CurrentUserStore.getCSRFToken()
 
 LogoutLink = React.createClass
 
+  propTypes:
+    label: React.PropTypes.string
+
+  getDefaultProps: ->
+    label: 'Log Out'
```
philschatz: When would this not be "Log Out"? oic, slick.

src/components/countdown-redirect.cjsx
nathanstitt: Sorry - missed this previously.  
 * Yes, the interval is cleared on unmount, even tested it :)
 * Calling `setState` in `willMount` is actually the preferred place to call it. React batches that together and will only call render once.  See https://facebook.github.io/react/docs/component-specs.html#mounting-componentwillmount

PR #1175, opened 2016-07-08
Fix LMS info popup appearance

nathanstitt:
Removes strange dash and "Summary" heading.  Removes strange line-height that was added to close buttons.

screen shot 2016-07-08 at 3 30 03 pm


philschatz:
Y U NO USE `<TEXTAREA>`?

nathanstitt:
I thought about it -- time mainly, but also wasn't entirely sure we'd decided on that.. and didn't have time to discuss with UX :)

PR #1180, opened 2016-07-11
Book Content - splash videos not visible

darrenmason:
regarding https://www.pivotaltracker.com/story/show/124155601

image

splash-video(s) in readings and refbook were not visible due to 0 width.


nathanstitt:
Marking as blocked so this isn't merged until showstoppers get in

PR #1182, opened 2016-07-12
Updating role checking on course listing

jivewise:
We were only checking the first role before, which was causing the error seen here: https://www.pivotaltracker.com/story/show/126045715

Now we check all roles, and give priority to the teacher.

### DIFF
--- a/src/components/course-listing.cjsx
+++ b/src/components/course-listing.cjsx
@@ -5,6 +5,7 @@ Router = require 'react-router'
 WindowHelpers = require '../helpers/window'
 
 {CourseListingActions, CourseListingStore} = require '../flux/course-listing'
+{CourseStore} = require '../flux/course'
 {RefreshButton} = require 'openstax-react-components'
 EmptyCourses    = require './course-listing/empty'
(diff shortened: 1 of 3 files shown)


src/components/course-listing.cjsx
```diff
@@ -5,6 +5,7 @@ Router = require 'react-router'
 WindowHelpers = require '../helpers/window'
 
 {CourseListingActions, CourseListingStore} = require '../flux/course-listing'
+{CourseStore} = require '../flux/course'
 {RefreshButton} = require 'openstax-react-components'
```
nathanstitt: I thought that we were going to make a `allCoursesWithRoles` or something and use that here.  That way the picker will not show for courses where the period is archived.  Or is this addressing a different bug?

jivewise: This was addressing a different, but similar one, I wasn't aware of adding that function in when we talked about it last.

nathanstitt: NP - I was just confused on the bug this addressed.  Looks good!

PR #1184, opened 2016-07-12
Fixing more embedded sim styling

pwolfert:
When an interactive iframe is a child of `.interactive-content > .note`, it gets offset by 160px instead of just 120px, because the note adds an extra 40px of padding.  

This was the screenshot from QA: https://www.pivotaltracker.com/file_attachments/66030327/download?inline=true

### DIFF
--- a/resources/styles/components/task-step/all-steps.less
+++ b/resources/styles/components/task-step/all-steps.less
@@ -36,6 +36,11 @@
       }
     }
 
+    .interactive-content > .note {
+      iframe.interactive {
+        margin-left: -160px;
+      }
+    }
   }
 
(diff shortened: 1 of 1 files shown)


resources/styles/components/task-step/all-steps.less
```diff
@@ -36,6 +36,11 @@
       }
     }
 
+    .interactive-content > .note {
+      iframe.interactive {
+        margin-left: -160px;
```
philschatz: Could you calculate this offset based on variables for the note width or put in a comment to clarify why this number was chosen so someone else looking at it a month from know would know why?

PR #1185, opened 2016-07-12
Book Title - background banner font/size

darrenmason:
regarding: https://www.pivotaltracker.com/story/show/109905656

Tested on knowable titles Physics I, Biology I and Sociology.  Fixed Physics having the same color as Bio. Truncation is still possible if the book name is really long.

break1

break2

break3

### DIFF
--- a/resources/styles/global/tutor-booksplash-background.less
+++ b/resources/styles/global/tutor-booksplash-background.less
@@ -12,12 +12,29 @@
     top: 60px;
     left: 0px;
     right: 0px;
-    height: 23rem;
-    line-height: 20rem;
+
+    height: 9rem;
+    line-height: 9rem;
+    font-size: 12rem;
+    @media (min-width: @screen-sm-min) {
+      height: 11rem;
+      line-height: 11rem;
+      font-size: 14rem;
+    }
+    @media (min-width: @screen-md-min) {
+      height: 14rem;
+      line-height: 13rem;
+      font-size: 18rem;
+    }
+    @media (min-width: @screen-lg-min) {
+      height: 16rem;
+      line-height: 15rem;
+      font-size: 21rem;
+    }
+
     overflow: hidden;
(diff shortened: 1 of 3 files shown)


nathanstitt:
This may need @openstax/ux to look at. If I recall correctly, the original implementation was specced so the text would slightly extend past the background.  i.e.  "Physics I" in the screenshot would extend a bit past the top/bottom of the green banner. 

If that's still the intention, then we'll need to also decrease the banner height and re-position the text a bit.

jason-holmes:
I think we're still miscommunicating on this...see attached.
course title book title

philschatz:
@darrenmason from the screenshots it looks like there is still more text than just "Physics"

jason-holmes:
I'm going to make an assumption...that the course picker dashboard (when you have more than course) is driving this problem.  They look the same and I'm guessing (assuming) are based on the same code.  On the course picker page, you need this background to be the course title because that's what you need to distinguish the classes.  But once inside the class, like on the student dashboard, you don't want the course title as the background image, but the book title (as I outline previously).  These two things (course picker title and book title for background image) need to be separated because they are used differently. @darrenmason @nathanstitt

darrenmason:
@jason-holmes I see what you mean, we have shared code that is showing course title for both: `getCourseDataProps`.

darrenmason:
right now we have `salesforce_book_name` in course data which could be added to `getCourseDataProps`.  Would that work for book title or should they be separate?

nathanstitt:
If you merge `master` you should be able to use the new `CourseStore.getBookName(courseId)` method to get the book name. It's calculated from the `appearance_code`.  Note that it may return be `undefined` if the code isn't set or is set to an invalid value.

nathanstitt:
Looks good @darrenmason - thanks for your patience with the change-up of the `physics` code mid-pr 😁

PR #1186, opened 2016-07-12
fixes assignments not publishing for certain courses

pandafulmanda:
archived periods need to be ignored, otherwise trying to check for validity gets confused because it tries to validing the tasking for the archived period as well

addresses [Cannot publish readings](https://www.pivotaltracker.com/story/show/126168299)


nathanstitt:
Looks good @pandafulmanda - tests on store are failing though.

pandafulmanda:
ah, @nathanstitt specs updated now

PR #1189, opened 2016-07-14
Add check for due time past in the builder

jivewise:
Pivotal Ticket: https://www.pivotaltracker.com/story/show/126241577

Since we updated the builder to allow same dates for the open date and due date, we weren't checking if the due time was had already passed before.  This PR adds that check, along with an error message, and some unit tests.  

@openstax/ux see screenshot below for error message.  Please let me know if that wording is acceptable.
screen shot 2016-07-13 at 8 28 16 pm


jason-holmes:
Discussed with Jing on Slack.  Looks good.

nathanstitt:
Tested locally and seems to work well.

PR #1191, opened 2016-07-14
Display "No Exercises" message when loading fails

nathanstitt:
Current error dialog is below, @openstax/ux please review and advise if it should be changed

Goes with BE PR: https://github.com/openstax/tutor-server/pull/1192, but can be merged independently 

screen shot 2016-07-18 at 10 49 38 am

### DIFF
--- a/src/components/navbar/server-error-monitoring.cjsx
+++ b/src/components/navbar/server-error-monitoring.cjsx
@@ -98,7 +98,7 @@ ERROR_HANDLERS =
       router.transitionTo('viewStudentDashboard', {courseId})
       Dialog.hide()
     dialog:
-      title: 'Unable to practice topic'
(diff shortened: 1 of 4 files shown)


jason-holmes:
This will show anytime there are not problems to display (i.e., homework, practice, reading whatever) right?  If so, then a more generic message like: "There are no problems to show for this topic." [OK]

nathanstitt:
Spec failures are unrelated - HW builder specs appear to be testing using date math that's failing

pandafulmanda:
ugh will have to look into that at some point.

nathanstitt:
oh, I was actually incorrect on the spec failures - sorry!  It was the error handler spec that wasn't clearing.  I fixed that in the last commit

PR #1192, opened 2016-07-14
New Question Controls Design

pwolfert:
https://www.pivotaltracker.com/n/projects/1156756/stories/120227453

- [x] Fixed/changed proportions, dimensions, borders, and margins for both
    - [x] Question Library page and
    - [x] Homework Builder
- [x] Change text from Include/Exclude to Add/Remove
- [x] Change Homework Builder's "Add Problems" header to "Add Questions"
- [x] Change Question Library's selected exercise overlay color to that light blue
- [x] Increase border for detail view's exercise cards
- [x] Change "Tutor Selection" buttons
    - [x] Switch to chevron icons instead of arrows and hide the circleness by default
    - [x] Keep the number always centered where things don't shift when one of the buttons disappears
- [x] Make the left and right buttons in detail view match those found in reading assignments
- [x] Center the section info block horizontally in the Homework Builder toolbar
- [x] Change "Add More..." button text to "+ Add More Sections"
- [x] Restrict the width of the question selection block to match the section selection block above it and to maintain consistency with the rest of Tutor.

Question Library After:
screen shot 2016-07-14 at 4 39 49 pm

Question Library Before:
screen shot 2016-07-14 at 4 37 19 pm

Homework Builder After:
screen shot 2016-07-15 at 2 37 09 pm

Homework Builder Before:
screen shot 2016-07-14 at 4 38 07 pm

Question Library Detail View After
screen shot 2016-07-15 at 2 40 23 pm

Question Library Detail View Before:
screen shot 2016-07-14 at 4 41 33 pm


philschatz:

nathanstitt:
I see the paging controls now use the "chevrons" design which I like, but it no longer has the nice wide click target on the gutters.  You used to be able to hover and click anywhere on the sides of the details view to navigate, now you can only click on the chevron.  That makes you have to move your mouse up/down to find it in the center, vs just moving it a bit to the right/left.

Is that by design?

nathanstitt:
The selection mask color in the details view of QL is now blue, the same as the HW builder.  It should be grey, right?  I tried to read through the pivotal and didn't see any mention of changing it, but I might have missed that.

nathanstitt:
I **thought** you fixed the broken line between controls and top task bar thing?  Maybe that was a different PR though?
I see:
screen shot 2016-07-21 at 4 52 21 pm

resources/styles/components/exercises.less
```diff
@@ -39,11 +55,42 @@
   }
 
   .page-navigation {
-    .page-navigation-arrows(800px);
+    // .page-navigation-arrows(800px);
```
nathanstitt: if unused, we should probably delete the line ;)

resources/styles/components/exercises.less
```diff
@@ -39,11 +55,42 @@
...
+        left: 40px;
+      }
+
+      &.next {
+        right: 40px;
+      }
+    }
+
+    i.fa {
+      color: #e5e5e5;
```
nathanstitt: are we using the `Icon` component ?  If so it'll have class `tutor-icon` which might be better to target

nathanstitt: These colors look identical to ones defined in `variables/colors.less`.  `#e5e5e5` is `@tutor-neutral-light`.  You could probably use that and a `darken(@tutor-neutral-light, 10%)` for the hover effect

resources/styles/components/exercises.less
```diff
@@ -104,21 +138,27 @@
...
+        margin-left: -1px;
+
+        &:first-child {
+          margin-left: 0;
         }
       }
       &.active {
         background-color: @tutor-neutral-light;
+        z-index: 10;
       }
```
nathanstitt: Hm, what's the `z-index` needed for?  I **seems** like an active sections shouldn't be floated above anything

resources/styles/components/exercises.less
```diff
@@ -130,8 +170,14 @@
...
       margin-right: 0;
+      margin-left: -1px;
+
+      &:first-child {
+        margin-left: 0;
+      }
       &.active {
         background-color: @tutor-neutral-light;
+        z-index: 10;
       }
```
nathanstitt: Same comment as above, unclear on why the `z-index` is for

resources/styles/components/exercises.less
```diff
@@ -130,8 +170,14 @@
       border: 1px solid @tutor-neutral;
       height: 40px;
       margin-right: 0;
+      margin-left: -1px;
+
```
nathanstitt: Is the `-1px` to remove the border between buttons?  If so, can you just set border-left to `0` for `button+button`

resources/styles/components/questions-dashboard.less
```diff
@@ -23,6 +23,11 @@
     }
   }
 
+  .exercises-display {
+    max-width: @tutor-max-panel-width;
+    margin: 0 auto;
+  }
+
```
nathanstitt: Hm, is this intentional?  When we choose the `card`  style layout we choose the column layout specifically so it would expand for users who have large monitors so they could be more productive and view more exercises at once.

resources/styles/components/task-plan/homework.less
```diff
@@ -20,11 +28,35 @@
...
+    margin: 0;
+    padding: 0;
+
+    // Right now the up chevron doesn't really look vertically centered in its circle even though
+    //   it lines up with the down chevron.  To fix this, we could put a margin-top of 4px on the
+    //   right (up) button and a margin-top of -2px on its icon child element, but this makes the
+    //   circles offset from each other vertically when hovering over a button.  I'm going to
+    //   leave it the way it is for now because that's probably a worse problem than the icon not
+    //   being centered vertically in the button.
+  }
```
nathanstitt: Thanks for the comment 😁 !

resources/styles/components/task-plan/homework/exercise-summary.less
```diff
@@ -1,5 +1,5 @@
 .homework-plan .exercise-controls-bar {
-
+  height: 90px;
   flex-wrap: wrap;
```
nathanstitt: 90px?  That seems (and looks) really big to me.  It's also causing some issues with the section titles being barely visible.  If 90 is really desired, then I think we'll also need to add some margin-top on the section so the title is more visible. Right now it looks like (also not a fan of the huge margin on sides):

screen shot 2016-07-21 at 5 11 47 pm

resources/styles/components/task-plan/homework/exercise-summary.less
```diff
@@ -48,14 +61,17 @@
     }
     &.total h2 { color: @tutor-quaternary; }
   }
+
   .tutor-added-later {
-    color: @tutor-neutral;
-    width: 200px;
+    color: #999999;
+    width: 232px;
```
nathanstitt: hm, the `#999999` seems like it might be a useful grey in between `@tutor-neutral` and `@tutor-neutral-light`.  No idea on what to call it though; if you can think of something maybe make that a var in `colors.less`

src/components/exercises/details.cjsx
```diff
@@ -146,7 +146,7 @@ ExerciseDetails = React.createClass
         />
       </div>
       {<div className="page-navigation next" onClick={@onNext}>
-        <div className='triangle' />
+        <Icon type='angle-right' />
        </div> if moves.next}
```
nathanstitt: I think this is the cause of the missing gutters.  Pretty sure the `.triangle` class set them. Take a look at `mixins/page-navigation-arrows.less`

Fredasaurus:
tutor-neutral-lite

src/components/task-plan/homework/exercise-controls.cjsx
```diff
@@ -68,24 +68,28 @@ ExerciseControls = React.createClass
     else if @props.canAdd
       <BS.Button bsStyle="default"
         className="-add-exercises"
-        onClick={@props.addClicked}>Add More...</BS.Button>
+        onClick={@props.addClicked}>+ Add More Sections</BS.Button>
     else
```
nathanstitt: We can't use the bare word `Sections` or `Periods`.  You can use the `CourseGroupingLabel` to switch between the two depending on the course type.

src/components/task-plan/homework/exercise-controls.cjsx
```diff
@@ -68,24 +68,28 @@ ExerciseControls = React.createClass
...
-    return null unless @canChangeTutorQty() and TaskPlanStore.canDecreaseTutorExercises(@props.planId)
-    <BS.Button onClick={@removeTutorSelection} className="btn-xs circle">
-      <Icon type='arrow-down' />
-    </BS.Button>
+    if @canChangeTutorQty() and TaskPlanStore.canDecreaseTutorExercises(@props.planId)
+      <BS.Button onClick={@removeTutorSelection} className="btn-xs hover-circle">
+        <Icon type='chevron-down' />
+      </BS.Button>
+    else
+      <span className="circle-btn-placeholder"></span>
```
nathanstitt: A nitpick, but React allows you to just do  XML style empty tags `<span className="circle-ban-placeholder"/>`

src/components/task-teacher-review/index.cjsx
```diff
@@ -108,7 +108,7 @@ TaskTeacherReview = React.createClass
         title={task.title}
         courseId={courseId}
         key="task-#{id}-breadcrumbs"/>
-
+    
     <PinnedHeaderFooterCard
```
nathanstitt: whitespace must die! (please:)

nathanstitt:
Wow!  I really like the improved appearance of the navbar alignment and colors.  Great job implementing this, I think most of my comments are really better directed to UX.

resources/styles/components/exercises.less
pwolfert: This is because the borders for the buttons are overlapping.  I found that that gave the best effect.  It was the cleanest look.  I could do some really complicated things to change the borders when their siblings are active and stuff, but I thought this was the simplest solution and one that I used for other button groups elsewhere.

resources/styles/components/exercises.less
pwolfert: I don't want to remove the border, because I want the one that is active to be the one with the border that shows up over the other borders because the border colors are different.

resources/styles/components/questions-dashboard.less
pwolfert: We discussed this in the story and Kathi gave Robert the goahead on this design change.

src/components/exercises/details.cjsx
pwolfert: I purposely took it out because the gutters are not part of the new design.

resources/styles/components/questions-dashboard.less
nathanstitt: Ok, it sounds like @kathi-fletcher approved so I'm ok with it.  Just think it looks not-so-great on wider displays and isn't as useful.

nathanstitt:
@pwolfert Can you merge/rebase master once more? The diff looks really weird, I'm not sure what's going on.

nathanstitt:
Looks good, thanks for your patience with my unrelated comments @pwolfert !

PR #1193, opened 2016-07-15
hide exercises instead of the "no exercises" message when sections change

nathanstitt:
No real UI changes other than the "No exercise found" message doesn't appear


philschatz:
I still see it!

hw-builder

nathanstitt:
Hm, based on the gif, it looks like that message is legitimate.  

It appears to be attempting to load and then doesn't find any.  The bug this is fixing was that it would immediately display "No exercises found" as soon as you selected a section, before clicking the "Show Problems" button.

philschatz:
oh, you're right; tutor-dev doesn't have later chapters ; (

PR #1198, opened 2016-07-18
Fixes the main nav bar's bottom border not drawing all the way across.

pwolfert:
Before:
screen shot 2016-07-18 at 11 18 49 am

After:
screen shot 2016-07-18 at 11 18 20 am

### DIFF
--- a/resources/styles/global/navbar.less
+++ b/resources/styles/global/navbar.less
@@ -35,9 +35,7 @@
   // override the bootstrap defaults with tutor defaults
   &.navbar.navbar{
     border-bottom: 1px solid @border-color;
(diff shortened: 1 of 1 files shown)


darrenmason:
@pwolfert I think this one was caused by me :) when fixing another bug: https://github.com/openstax/tutor-js/pull/1089

However you may see that when the window width < ~1000px, the hamburger dropdown is transparent.  I think `navbar-collapse` needs to be `@tutor-white` without affecting the rest.

darrenmason:
this works:
```
    &, .navbar-collapse.collapsing, .navbar-collapse.in {
    }
```

PR #1202, opened 2016-07-19
Calculate height of days based on max plan order

nathanstitt:
Check the order of each plan and sets the height of week's days based on
the higest plan order. 

Turns this:

screen shot 2016-07-18 at 8 09 53 pm

Into:
screen shot 2016-07-18 at 8 10 15 pm


philschatz:
While not as good as https://github.com/nathanstitt/dayz , it _seems_ like it'll work; thoughts @pandafulmanda ?

pandafulmanda:
looks great to me

PR #1210, opened 2016-07-21
Better nav menu for small screens

pwolfert:
screen shot 2016-07-21 at 9 34 47 am

screen shot 2016-07-21 at 9 35 02 am

### DIFF
--- a/resources/styles/global/navbar.less
+++ b/resources/styles/global/navbar.less
@@ -6,7 +6,7 @@
 @grid-float-breakpoint: @screen-md-min;
 
 .navbar-fixed-top {
-	height: @tutor-navbar-height;
+  height: @tutor-navbar-height;
   // override bootstrap material design media query for (max-width: 1199px)
   .navbar-brand {
     height: 60px;
@@ -60,6 +60,7 @@
     .tutor-background-image('rice-logo.png');
   }
   .book-links {
+    border-right: 1px solid @tutor-neutral-light;
 
     a {
       height: @tutor-navbar-height;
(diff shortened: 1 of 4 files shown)


src/components/navbar/index.cjsx
```diff
@@ -2,6 +2,7 @@ React = require 'react'
 BS = require 'react-bootstrap'
 Router = require 'react-router'
 _ = require 'underscore'
+classnames = require 'classnames'
```
philschatz: Is this variable used?

src/components/navbar/user-actions-menu.cjsx
```diff
@@ -75,9 +76,11 @@ UserActionsMenu = React.createClass
 
 
   render: ->
+    course = CourseStore.get(@props.courseId)
+
     <BS.NavDropdown
       eventKey={1}
-      className='-hamburger-menu'
+      className={classnames('-hamburger-menu', 'is-concept-coach': course?.is_concept_coach)}
```
philschatz: Ooh, cool! I think this could be used to show the student/teacher that they are in a concept coach (vs tutor) course which would be useful for pilot teachers that used CC last year but are trying Tutor this year

resources/styles/global/navbar.less
```diff
@@ -69,14 +70,76 @@
...
+        display: block;
+      }
+
+      &.is-concept-coach {
+        > .dropdown-menu {
+          border-top: 1px solid @tutor-neutral-light;
+          padding-top: 8px;
+          margin-top: 0;
+        }
+      }
```
nathanstitt: are these `is-concept-coach` styles specced anywhere?  I'm curious why we're styling the drop-down menu differently for those users?  Seems like if the additional styles are good, then shouldn't they apply to all users?

nathanstitt: Ah, I see @philschatz may have answered my question below.  Maybe it's so teachers can tell the course types apart?

pwolfert: If it's concept coach, then the list above it is not empty, which means we need to put spacing in between.  There's no way in css of finding out if the child of the child of the element's sibling has specific contents or not, so I had to do it this way.  It can also be used for more useful things, but this was just a hack.  I can't even use the `:empty` selector in the section above it because it's not that it has no children but that it doesn't include children that aren't the `.navbar-brand`.  If someone has a selector for that, I'm all ears.

src/components/navbar/index.cjsx
pwolfert: Nope, I was using it and then forgot to remove it when I stopped using it.

resources/styles/global/navbar.less
nathanstitt: ah, yep makes perfect sense.  Thanks!

PR #1211, opened 2016-07-21
"(See Figure)" instead of "(Figure)" for links

pwolfert:


nathanstitt:
Nice and simple! glad you tracked 'em down

PR #1216, opened 2016-07-26
Fixes uncentered sims in book content

pwolfert:
Sets the right negative margin for interactives in the book so they're centered.

### DIFF
--- a/resources/styles/components/task-step/reading/interactive.less
+++ b/resources/styles/components/task-step/reading/interactive.less
@@ -1,7 +1,7 @@
 .task-reading {
   .interactive-step {
     iframe {
-      width: 960px;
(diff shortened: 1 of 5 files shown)


resources/styles/components/task-step/reading/interactive.less
```diff
@@ -1,7 +1,7 @@
 .task-reading {
   .interactive-step {
     iframe {
-      width: 960px;
+      width: @tutor-interactive-iframe-width;
       height: 560px;
       margin-left: -(@tutor-card-body-padding-horizontal - 20px);
```
nathanstitt: Nice catch, it's great to update these!

nathanstitt:
Looks great @pwolfert !

PR #1221, opened 2016-07-26
start up a mock server based on the API blueprint

philschatz:
This starts up a mock server instead of relying on the API JSON files. Why?
1. the API is documented and kept up to date
2. allows for conditional requests (based on headers) which are useful for Selenium tests
3. response headers that contain the server time
4. properly supporting non-GET requests
5. Running Selenium tests in Travis without having to start up `tutor-server`

To see it in action, run `npm start` which starts up the mock server and runs `gulp dev`, and then go to http://localhost:8000 (or http://localhost:3000/api/user)


nathanstitt:
Thanks @philschatz but I think this one is languished for too long - closing it so we can attempt to reach "OPEN PR ZERO"

PR #1222, opened 2016-07-27
Course Listing - visual distinction between Tutor/CC

darrenmason:
regarding https://www.pivotaltracker.com/story/show/127076359

listbreak2

capture

### DIFF
--- a/src/components/course-listing.cjsx
+++ b/src/components/course-listing.cjsx
@@ -61,24 +61,14 @@ CourseListing = React.createClass
       isTeacher = _.find roles, (role) -> role.type is 'teacher'
 
       courseType = if isConceptCoach then 'Concept Coach' else 'Tutor'
-
-      if courselink?
-        courseTypeFlag = "(#{courseType})"
-      else
-        courseTypeFlag = <div className='course-type-flag'>{courseType}</div>
+      courseTypeFlag = <div className='course-type-flag'>{courseType}</div>
 
       if isTeacher
-        if courseLink?
-          altLink = <Router.Link
-            className='tutor-course-alt-link'
-            to='taskplans'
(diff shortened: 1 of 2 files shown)


src/components/course-listing.cjsx
```diff
@@ -60,12 +60,19 @@ CourseListing = React.createClass
       isStudent = _.find roles, (role) -> role.type is 'student'
       isTeacher = _.find roles, (role) -> role.type is 'teacher'
 
+      courseType = if isConceptCoach then 'Concept Coach' else 'Tutor'
+
+      if courselink?
+        courseTypeFlag = "(#{courseType})"
+      else
+        courseTypeFlag = <div className='course-type-flag'>{courseType}</div>
+
```
nathanstitt: It seems to me that something weird's going on with the `courseLink` var.  It's not assigned until way below in the else statement. Or am I missing something?  I wonder if that's an artifact from how the listing used to work before we had archived periods to worry about?  Seems like this entire block could be re-written to be a bit clearer.

darrenmason: Yeah I noticed the same thing.  If you reverse the two `if courselink?` conditions you can see it is being used for `alt-link` which seems to be just a text only version of the course list.  Not sure we need this; it seems `if courselink?` is never true until courselink is defined anyway, I was just matching what was there.  It may just be old code...

nathanstitt: Yeah, I suspect it's old code as well.  git blame suggests that @pandafulmanda added it  last summer.  Why don't you take out the courseLink conditionals and see how it works.

nathanstitt:
Looks good, thanks for the cleanup!

PR #1223, opened 2016-07-27
Use college specific icons

nathanstitt:
Depends on https://github.com/openstax/react-components/pull/102

Adds the 'is-college' and 'is-high-school' classes on root element for switching icons and styles
is no longer based off production

screen shot 2016-07-27 at 10 54 48 am

screen shot 2016-07-27 at 12 17 39 pm

### DIFF
--- a/src/components/app.cjsx
+++ b/src/components/app.cjsx
@@ -1,11 +1,12 @@
 React = require 'react'
+classnames = require 'classnames'
 
 {HistoryLocation, History, RouteHandler} = require 'react-router'
 
 Navbar = require './navbar'
 Analytics = require '../helpers/analytics'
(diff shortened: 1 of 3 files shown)


philschatz:
@openstax/ux Any thoughts on the yellow circles? Nathan mentioned that they might not match but it's all grey to me 😄

philschatz:
Ah, @Fredasaurus changed them in the screenshot to just be white

nathanstitt:
Note! I was unable to find any sims or video reading steps in my college content to test with.  I **think** they should all work, but someone should double check them

pandafulmanda:
just wanted to check, @Fredasaurus, is the shift of the circle icons from outside of the box to inside intentional for college vs hs?

PR #1224, opened 2016-07-27
Fixes embedded videos not fitting in their wrappers

pwolfert:
https://www.pivotaltracker.com/n/projects/1156756/stories/127106185

screen shot 2016-07-27 at 9 00 19 am

### DIFF
--- a/resources/styles/components/task-step/all-steps.less
+++ b/resources/styles/components/task-step/all-steps.less
@@ -3,7 +3,7 @@
 
 .task {
   iframe {
-    width: 960px;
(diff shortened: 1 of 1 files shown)


resources/styles/components/task-step/all-steps.less
```diff
@@ -6,7 +6,7 @@
     width: 960px;
     height: 500px;
```
nathanstitt: Not really related to this pr, but you might update the 960 to your new variable (that I can't remember the name of) :)

nathanstitt:
Awesome - thanks @pwolfert

PR #1226, opened 2016-07-27
Student Scores: past-due shows score and is blue

philschatz:
Every homework that has been started and is past due shows a score, and every progress pie that is past due is blue.

screen shot 2016-07-27 at 2 49 40 pm


src/components/scores/homework-cell.cjsx
```diff
@@ -71,7 +71,7 @@ HomeworkCell = React.createClass
               isConceptCoach={isConceptCoach}
               size={20}
               value={TH.getCompletedPercent(task)}
-              isLate={TH.isLate(task)}
+              isLate={TH.isDue(task)}
```
philschatz: Should this parameter be renamed from `isLate=` to `isDue=`? (can happen after all the 🔥 subsides 😄 )

PR #1227, opened 2016-07-28
Fix for mobile dropdown nav menu not hiding after click

pwolfert:
https://www.pivotaltracker.com/n/projects/1156756/stories/127202601


nathanstitt:
Yeah, seems like about as good as we can do until we can upgrade react-bootstrap :(

PR #1232, opened 2016-07-29
Fix calendar events escaping the week they should be in

nathanstitt:
There were different edge cases that would cause problems seen [here](https://www.pivotaltracker.com/story/show/127275267) and [here](https://www.pivotaltracker.com/story/show/127915967).

Specifically, when multiple plans overlap on common day, kinda transitively, like this:

Plan 1 - Sunday, **Monday**, **Tuesday**
Plan 2 - **Monday**, **Tuesday**, _Wednesday_
Plan 3 - _Wednesday_, Thursday

the calendar was not accounting for when groups of plans like this amounted to more than 5 rows and needed more space on the calendars.
 
## Before
screen shot 2016-08-11 at 1 26 11 am
screen shot 2016-08-11 at 1 25 57 am

## After

screen shot 2016-08-11 at 1 21 37 am
screen shot 2016-08-11 at 1 24 02 am

### DIFF
--- a/tutor/src/components/course-calendar/duration.cjsx
+++ b/tutor/src/components/course-calendar/duration.cjsx
@@ -108,7 +108,7 @@ CourseDuration = React.createClass
 
         _.chain(plans)
           .sortBy((plan) ->
-            -1 * plan.rangeDuration.start.valueOf()
+            plan.rangeDuration.count('days')
           )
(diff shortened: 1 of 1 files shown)


nathanstitt:
Setting as `code` because @pandafulmanda found another case that this doesn't handle.

PR #1234, opened 2016-07-29
Fix exercise exclusions to only look at loaded sections

nathanstitt:


jivewise:
Yay, @nathanstitt you got 1234: https://www.youtube.com/watch?v=ABYnqp-bxvg

PR #1239, opened 2016-08-01
Fixes mobile nav menu problems

pwolfert:
- Fixes problem where users could not click on elements to the left of the dropdown menu.
- Fixes problem where menu wouldn't close after clicking "My Account" or "Get Help".

### DIFF
--- a/tutor/resources/styles/global/navbar.less
+++ b/tutor/resources/styles/global/navbar.less
@@ -78,6 +78,10 @@
   }
 
   @media screen and (max-width: (@screen-md - 1px)) {
+    .container-fluid {
+      height: @tutor-navbar-height
+    }
+
     .navbar-collapse {
       max-width: 230px;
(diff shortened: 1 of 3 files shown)


src/components/navbar/account-link.cjsx
```diff
@@ -9,5 +9,5 @@ module.exports = React.createClass
   render: ->
     return null unless CurrentUserStore.getProfileUrl()
     <li>
-      <a href={CurrentUserStore.getProfileUrl()} target='_blank'>My Account</a>
+      <a href={CurrentUserStore.getProfileUrl()} target='_blank' onClick={if @props.onClick then @props.onClick}>My Account</a>
     </li>
```
nathanstitt: In React, setting a prop to an undefined value is the same as not setting it at all. Accordingly this could just be `onClick={@props.onClick}`

PR #1241, opened 2016-08-02
Fixes calendar button's tooltip rendering underneath the navbar

pwolfert:
https://www.pivotaltracker.com/n/projects/1156756/stories/127581441

screen shot 2016-08-02 at 2 37 56 pm

### DIFF
--- a/tutor/resources/styles/global/navbar.less
+++ b/tutor/resources/styles/global/navbar.less
@@ -228,6 +228,6 @@
   }
 }
 
-.fixed-tooltip {
+.tooltip.on-navbar {
   z-index: (@zindex-navbar-fixed + 1);
 }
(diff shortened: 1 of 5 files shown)


nathanstitt:
Hm, I think I might have caused this with an earlier commit that set the z-index for tooltips so they'd go under the navbar for the QL/HW builder.  @pandafulmanda what do you think, should we do what @pwolfert is doing here, or maybe we should just apply a class to the navbar on the QL that would raise that specific navbar?

nathanstitt:
I thought maybe my earlier changes to the z-index for tooltips caused this, but that appears unrelated.  I renamed the css class to be "on-navbar" though since I found that a bit easier to understand it's purpose.

Good 🔎 detective work on finding the spots to update for the icon

PR #1242, opened 2016-08-02
Fixes "All" button not showing selection in the Question Library's nav bar

pwolfert:
https://www.pivotaltracker.com/n/projects/1156756/stories/127578009

screen shot 2016-08-02 at 3 00 43 pm


nathanstitt:
Nice use of `classnames`, I have no idea why the previous committer (me) tried to do it using sting concatenation 😊

PR #1246, opened 2016-08-03
Fixes weird tabs in Performance Forecast and implements new design

pwolfert:
Fixes the bug by implementing a slightly new tab design that should be consistent across all these kinds of tabs.

https://www.pivotaltracker.com/n/projects/1156756/stories/127656737

screen shot 2016-08-03 at 4 04 36 pm
screen shot 2016-08-03 at 4 04 45 pm
screen shot 2016-08-03 at 4 04 51 pm
screen shot 2016-08-03 at 4 17 54 pm

Verify no change for:
screen shot 2016-08-03 at 10 09 36 am

### DIFF
--- a/tutor/resources/styles/components/course-settings.less
+++ b/tutor/resources/styles/components/course-settings.less
@@ -33,10 +33,6 @@
   .nav-tabs {
     .tutor-nav-tabs();
     span { max-width: 100px; }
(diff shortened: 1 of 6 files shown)


nathanstitt:
Very nice! It'll be great to get these all uniform

PR #1248, opened 2016-08-05
Escape css selectors 

nathanstitt:
The mediaId may have slashes in it. Physics 11.3 has id's like: `ost/api/ex/k12phys-ch11-ex051`


nathanstitt:
Closing in favor of the better fix in #1255

PR #1249, opened 2016-08-05
Fixes things that were broken as part of the monorepo conversion

philschatz:
Feel free to push other fixes here as they are found


tutor/src/components/task-teacher-review/exercise.cjsx
```diff
@@ -16,7 +16,7 @@ TaskTeacherReviewQuestion = React.createClass
 
   onChangeAnswerAttempt: (answer) ->
     # TODO show cannot change answer message here
-    console.log('You cannot change an answer on a problem you\'ve reviewed.', 'TODO: show warning in ui.')
+    alert('You cannot change an answer on a problem you\'ve reviewed.')
```
nathanstitt: this make me nervous since an `alert` message is so ugly and we didn't display anything previously.  

@pandafulmanda do you have any idea how likely it is that this will be displayed?

pandafulmanda: the alert looks scary...could we take out both the alert and console log for now, `return null` from this, and update the TODOs to say, `clean up onChangeAnswerAttempt`?  That feature never got fully implemented and is not relevant anymore, so we can remove some dead code around that

philschatz: If it's dead code then the alert shouldn't be scary, right? 😉  I'll switch it to `return null` and add a TODO

PR #1251, opened 2016-08-05
Hide LO tags for college courses

nathanstitt:
Before:
screen shot 2016-08-05 at 5 12 30 pm

after:
screen shot 2016-08-05 at 5 12 11 pm


kathi-fletcher:
Looks nice. The tutor-content views will still show LO's, right?

nathanstitt:
@kathi-fletcher QA view will now 😉   Thanks for pointing out that case.

PR #1253, opened 2016-08-08
Fix scores review link period

pandafulmanda:
[Bug: No period tab selected when you try to review an assignment ](https://www.pivotaltracker.com/story/show/127280523)

### DIFF
--- a/tutor/src/components/task-teacher-review/index.cjsx
+++ b/tutor/src/components/task-teacher-review/index.cjsx
@@ -34,26 +34,16 @@ TaskTeacherReview = React.createClass
     crumbKey = if sectionIndex then parseInt(sectionIndex) - 1 else defaultKey
     @setState(currentStep: crumbKey)
 
-  getPeriodIndex: ->
-    {id, periodId} = @context.router.getCurrentParams()
-
-    if @state.isReviewLoaded
-      plan = TaskPlanStatsStore.get(id)
-      index =
-        _.findIndex(plan.stats, (stat) -> stat.period_id is periodId)
-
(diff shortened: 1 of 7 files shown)


tutor/src/components/task-teacher-review/index.cjsx
```diff
@@ -33,21 +34,16 @@ TaskTeacherReview = React.createClass
     crumbKey = if sectionIndex then parseInt(sectionIndex) - 1 else defaultKey
     @setState(currentStep: crumbKey)
```
nathanstitt: I know you didn't write this, but `parseInt`  must always have it's radix specified, otherwise older browsers will use octal for numbers that lead with a `0` http://stackoverflow.com/questions/6611824/why-do-we-need-to-use-radix

tutor/src/flux/task-plan-stats.coffee
```diff
@@ -8,15 +8,18 @@ TaskPlanStatsConfig = {
...
           is_trouble: stat.is_trouble
-
-        )
-        .tap(PeriodHelper.sort)
         .value()
 
+      # tapping into the sort wasn't actually returning sorted
+      # periods for some reason
+      PeriodHelper.sort(periods)
+
```
nathanstitt: Pretty sure that's because `.tap` doesn't actually use the return value of the function. It's intended to allow a function to manipulate the values, but not change them to a new thing.  If you wanted to set a new order for periods, I think you could just use `
`.sort(PeriodHelper.sort)`  This works though - is equally good to me

PR #1254, opened 2016-08-08
Fix/confirm hide task

jivewise:
Pivotal Ticket: https://www.pivotaltracker.com/story/show/115959401

Adds in confirmation dialog for removing withdrawn events.  Also a bit of cleanup.

screen shot 2016-08-08 at 11 51 51 pm

### DIFF
--- a/shared/src/components/surety-guard.cjsx
+++ b/shared/src/components/surety-guard.cjsx
@@ -5,7 +5,9 @@ SuretyGuard = React.createClass
 
   propTypes:
     onConfirm:  React.PropTypes.func.isRequired
-    message:    React.PropTypes.string.isRequired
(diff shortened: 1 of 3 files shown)


shared/src/components/surety-guard.cjsx
```diff
@@ -5,7 +5,9 @@ SuretyGuard = React.createClass
 
   propTypes:
     onConfirm:  React.PropTypes.func.isRequired
-    message:    React.PropTypes.string.isRequired
+    message:    React.PropTypes.oneOfType([
+      React.PropTypes.string, React.PropTypes.element
+    ]).isRequired
     placement:  React.PropTypes.string
```
nathanstitt: Fixed the prop type so React doesn't complain

PR #1255, opened 2016-08-08
Fix media link select

pandafulmanda:
alternative to #1248.  fixes the problem by:
  * being stricter about what types of links are considered media links
    * this will prevent selectors like `#hello/you/thing` from being considered media to process. Therefore, `\` should no longer get into the selector in the `getMedia` function
  * wraps the `querySelector` function in a try catch to prevent any other invalid selectors from throwing uncaught exceptions
  * include `.ost-exercise` as an exercise selector for exercise rendering/embedding

### DIFF
--- a/tutor/src/components/book-content-mixin.cjsx
+++ b/tutor/src/components/book-content-mixin.cjsx
@@ -67,7 +67,13 @@ LinkContentMixin =
 
   getMedia: (mediaId) ->
     root = @getDOMNode()
-    root.querySelector("##{mediaId}")
+    try
+      root.querySelector("##{mediaId}")
+    catch error
+      # silently handle error in case selector is
+      # still invalid.
+      console.warn(error)
+      false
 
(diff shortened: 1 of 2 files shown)


nathanstitt:
I think this is a good approach, but wonder if  there will ever be a case where content teams may want to use a selector that contains characters that need to be escaped? Note it's a pretty big list and I only handled a few of them in #1248 https://mathiasbynens.be/notes/css-escapes

If we do, then I'd propose we combine the two PR's so we still have the `dom.escapseCss` method.

pandafulmanda:
i dont think the content teams will ever want to, at least not in this context -- this is for media preview and figures -- stuff that should have valid selectors and `#___` hash links to begin with.  the problem was, this media logic was trying to parse links that werent media links.

nathanstitt:
ok, looks good, thanks. I'll close #1255 then

PR #1256, opened 2016-08-08
Fix developer mode source maps, optimize webpack

nathanstitt:
I considered adding this to #1249 but decided against it since I believe the bugs with source maps predates the monorepo.

Previously I would occasionally get a slew of warnings about incorrect souremaps formatting and then they wouldn't show the coffee script source, just the raw JS.

Plus the new `cheap-module-eval-source-map` seems about twice as fast, which makes hot-reloading much nicer.


jivewise:
:+1: for the faster source maps!

nathanstitt:
Weirdly, it now seems like the phantoms shim isn't loading for tests.  Will debug later

configs/webpack/base.coffee
```diff
@@ -10,7 +10,7 @@ _ = require 'lodash'
 # base config, true for all builds no matter what conditions
 BASE_CONFIG =
   cache: true
-  devtool: 'source-map'
+  devtool: 'cheap-module-eval-source-map'
```
philschatz: Does this still keep the minified JS file small? The only documentation I've seen on this is https://webpack.github.io/docs/configuration.html and it's sparse

nathanstitt: yeah, this is only for dev mode.  Production sets it to `source-map` (line 136 below)

nathanstitt: I tested building production to make sure the `.map` files are still built

PR #1261, opened 2016-08-09
No periods/section handling in calendar dashboard

pandafulmanda:
Addresses [No error message if teacher tries to create an assignment before creating at least 1 section](https://www.pivotaltracker.com/story/show/127943403)

Also, doesn't attempt to load plans from BE when there are no sections, making the load time of the dashboard appear much quicker.

pop-ups are clickable and will like to course settings

screen shot 2016-08-10 at 10 35 51 am
screen shot 2016-08-10 at 10 35 33 am
screen shot 2016-08-09 at 4 39 19 pm

all instances of the "Please add at least one period/section to the course." links to course settings except for when on course settings page.

screen shot 2016-08-09 at 4 23 42 pm


jason-holmes:
If we can't make a HS/college specific message to account for the period/section text, we need to change to these messages.  "Please add a period/section...".  
Also, can the text "course settings" be linked to the course setting page?

resources/styles/components/course-calendar/header.less
```diff
@@ -33,6 +26,25 @@
         border-top-right-radius: 0;
       }
     }
+
+    .-no-periods-text {
+      display: inline-block;
+      margin: 16px 0;
+    }
+
```
nathanstitt: I kinda thought we weren't supposed to use `-` prefixed classes for styling.  They're for test selection only right, and if we use them for styling we're supposed to remove the `-`

src/components/no-periods.cjsx
```diff
@@ -9,14 +10,34 @@ NoPeriods = React.createClass
...
+
+  getDefaultProps: ->
+    link: true
+
+  getMessage: ->
+    [
+      'Please add at least one '
+      <CourseGroupingLabel lowercase courseId={@props.courseId} />
+      ' to the course.'
+    ]
```
nathanstitt: React complains because these don't have `keys`.  I think we'll have to make the bare strings into `<span key="stuff">`, and add a key to CourseGroupingLabel to make it happy, React will insert span tags anyways so it's really the same.

src/helpers/period.coffee
```diff
@@ -33,6 +33,9 @@ PeriodHelper =
   activePeriods: (course) ->
     _.filter(course.periods, isActiveCheckFn)
 
+  hasPeriods: (course) ->
+    not _.isEmpty(@activePeriods(course))
+
   archivedPeriods: (course) ->
```
nathanstitt: nice addition 😁

test/components/helpers/calendar/checks.coffee
```diff
@@ -149,6 +149,24 @@ checks =
...
+  _checkNoPeriodsOnAddAssignmentWarning: ({div, component, state, router, history, courseId}) ->
+    addDropdown = div.querySelector('.add-assignment .dropdown-menu')
+
+    expect(addDropdown.style.display).to.not.equal('none')
+    expect(addDropdown.innerText).to.contain('Please add')
+      .and.to.contain('in \nCourse Settings before\nadding assignments.')
+
+    {div, component, state, router, history, courseId}
+
   _checkIsTodayClickable: ({div, component, state, router, history, courseId}) ->
```
nathanstitt: 👍 nice 👓

pandafulmanda:
@jason-holmes, they are college/hs specific.  currently, the whole dropdown/sentence is linked.  should it be limited to "course settings"?

jason-holmes:
Oh, great about hs/college specific.  I like that the whole thing is linked, but there' not visual affordance, so maybe put link styling on "course settings" text.

resources/styles/components/course-calendar/header.less
pandafulmanda: okay yah, good point.  i dont know how much we've been sticking to that, but yah we should fix if we see it

pandafulmanda:
okay, @jason-holmes and @nathanstitt, pushed in stuffs to address the stuffs

PR #1262, opened 2016-08-09
Add ui-settings model, use it to display one-time help 

nathanstitt:
needs to be merged with: https://github.com/openstax/tutor-server/pull/1230

Concept coach:
screen shot 2016-08-10 at 4 33 19 pm

Tutor
screen shot 2016-08-10 at 4 36 57 pm


philschatz:
Should people have an option to "Don't show me this again" rather than assume that they don't want to see it again? /cc @openstax/ux

rvnewell:
This is only supposed to be displayed before the first two-step question a user sees, so they only see it once.  It should not be displayed before any successive two-step questions in the course. So there shouldn't be a need to for "don't show" messaging.

nathanstitt:
Yeah, that's why we have the accompanying BE PR.  It adds a `ui_settings` endpoint that we store a "has seen two step help" variable, so the user should only see  it once.

PR #1269, opened 2016-08-11
continuing to mono and remove dead files, particularly things...

pandafulmanda:
related to gulp

Also, coach release script, with colors!

screen shot 2016-08-12 at 12 19 27 pm


nathanstitt:
All those commits mentioning color - we must have screenshots!

PR #1270, opened 2016-08-12
Bootstrap instead of Bootrap, and initialize maybe helps with the specs

pandafulmanda:
quick follow up on #1262


nathanstitt:
But.. but.. maybe I really like to rap about boots?

pandafulmanda:
do you also like boo traps?

nathanstitt:
👻 🍯

PR #1271, opened 2016-08-12
Coach Enrollment Error Message

jivewise:
Pivotal Ticket: https://www.pivotaltracker.com/story/show/127655219

screen shot 2016-08-12 at 10 36 01 pm

- [x] Updates coach enrollment error message
- [x] Updates less errors with demo
- [x] Updates react bootstrap nav component


nathanstitt:
The code looks good and it works fine,  I think this is unrelated but the drop-down menu isn't appearing correctly for me. Do you also see this:

screen shot 2016-08-12 at 11 51 22 am

PR #1277, opened 2016-08-12
create rev-manifest.json file with mapping

nathanstitt:


nathanstitt:
@pandafulmanda - we've tested this and seems to work now.

PR #1278, opened 2016-08-15
add support for jsx files

philschatz:
This adds support for `.jsx` files while still supporting the existing `.cjsx` files. As a proof-of-concept, I converted `http://localhost:8000/sandbox`

Also, it should probably be tested locally and deployed before merging just to make sure everything is ok


nathanstitt:
Looks like travis failed with:
```
Module not found: Error: Cannot resolve 'file' or 'directory' ./sandbox in /home/travis/build/openstax/tutor-js/tutor/src
 @ ./tutor/src/router.cjsx 43:10-30f
```

Other than that, code & configs look great!

tutor/src/sandbox.jsx
```diff
@@ -0,0 +1,25 @@
+const React = require('react');
+const BS = require('react-bootstrap');
+
+const Sandbox = React.createClass({
+  render: () => {
+    return (
```
nathanstitt: If/when we move to es6, we should discuss fat arrow usage.  For instance here, it seems like it'd be unwanted here since it binds to `this`, which `React.createClass` is already doing for us.  Neither binding is needed for `render`, it's not called from a callback.  I bet we're essentially nesting 3 functions deep here.

philschatz:
I was hoping the failure was only in my local `npm start tutor` but I guess not ; ( Hmm, any ideas why webpack would complain but still compile the file?

pandafulmanda:
it's because of our configs, about to push in a fix

configs/webpack/base.coffee
```diff
@@ -1,16 +1,66 @@
...
+
+STATICS =
+  json: RESOLVABLES.json
+  image: { test: /\.(png|jpg|svg|gif)/, loader: LOADERS.file }
+  font: { test: /\.(woff|woff2|eot|ttf)/, loader: LOADERS.url }
+
+BASE_BUILD =
+  # TODO
+  #   Loading .js through babel is not yet working.
+  #   It doesn't seem to exclude node_modules,
```
philschatz: I've added an `exclude: [/node_modules/]` entry to the loader in https://github.com/philschatz/gh-board/blob/master/webpack.config.js#L27 and that seems to work

pandafulmanda: hmm, will have to try it again...maybe `exclude` has to be an array

PR #1282, opened 2016-08-15
Fix/plan spacing

pandafulmanda:
Preserves fixes from #1232 while deleting code (-7 lines difference!) and using previous logic that can prevent those problems, while fixing extra spacing caveat from #1232 fix.

## Fix: queries full month bounds
Also queries the full week for the month bounds as opposed to just the day before and after month bounds.

## Before, extra spacing from #1232 
screen shot 2016-08-15 at 5 25 48 pm

## Now
screen shot 2016-08-15 at 5 25 26 pm

Also, a nice little modification for changing "Loading Calendar..." to:

screen shot 2016-08-15 at 6 39 20 pm

### DIFF
--- a/tutor/src/components/course-calendar/duration.cjsx
+++ b/tutor/src/components/course-calendar/duration.cjsx
@@ -63,14 +63,13 @@ CourseDuration = React.createClass
       # TODO these parts actually seem like they should be in flux
       .each(@setDuration(viewingDuration))
       .filter(@isInDuration(viewingDuration))
-      .sortBy((plan) ->
-        moment(plan.published_at).valueOf()
-      )
-      .sortBy((plan) ->
-        plan.duration.end().valueOf()
-      )
-      .sortBy((plan) ->
-        plan.duration.start().valueOf()
+      .sortBy((plan) =>
+        [
+          plan.duration.start().valueOf()
+          plan.duration.end().valueOf()
+          @_getEarliestOpensAt(plan).valueOf()
+          moment(plan.last_published_at).valueOf()
+        ].join()
       )
       .value()
 
--- a/tutor/src/components/course-calendar/duration.cjsx
+++ b/tutor/src/components/course-calendar/duration.cjsx
@@ -81,29 +81,17 @@ CourseDuration = React.createClass
       .value()
 
   calcDurationHeight: (rangeData) ->
-    overlappingPlans = []
-    planStacks = []
-
-    _.each(rangeData.plansByDays, (plansByDay) =>
-      @setPlanRelativeOrder(plansByDay)
-
-      planIds = _.map(plansByDay, (planInfo) -> planInfo.plan.id)
-      planOrder = _.map(plansByDay, (planInfo) -> planInfo.relativeOrder)
-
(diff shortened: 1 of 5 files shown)


tutor/src/components/course-calendar/duration.cjsx
```diff
@@ -81,25 +80,17 @@ CourseDuration = React.createClass
       .value()
```
nathanstitt: I'm not sure using the joined array to sort will do what you want.   For instance picture a array of values such as this (with values representing what's returned by the `valueOf`):

| values| | after join()|
|------------|---|--------:|
| `[ 10, 1, 3, 101 ]` | => | `1013111` |
| `[ 11, 1, 3, 99 ]`|   => | `111399` |

The second should be first because it's `published_at` is 11, but it'll come last because the # of digits aren't as long.

The "right" way to do this is to multiple the numbers by their place, then add them.

| values| | equals|
|------------|---|--------:|
| `(10*1000) + (1*100) + (3*10) + 101` | => | `10231` |
| `(11*1000) + (1*100) + (3*10) + 99`|   => | `11229` |

tutor/src/components/course-calendar/duration.cjsx
```diff
@@ -81,25 +80,17 @@ CourseDuration = React.createClass
...
+      .map (plansByDay) =>
+        @_setPlanRelativeOrder(plansByDay)
+        # use plan relative order to calculate plan "height"
+        _.map(plansByDay, (plan) -> -1 * plan.relativeOrder + 1)
+      # flatten for all heights in week
+      .flatten()
+      # union with a 0 height, for durations with no plans
+      .union([0])
+      .max()
+      .value()
```
nathanstitt: Nice cleanup 😁

tutor/src/components/course-calendar/duration.cjsx
pandafulmanda: `valueOf` returns the same number of digits for each

nathanstitt: Hm, I think valueOf returns the # of milliseconds since epoch  http://momentjs.com/docs/#/displaying/unix-timestamp-milliseconds/  Which will most likely be the same number of digits, but won't always be.  Think the case where some of the dates straddle the point where the milliseconds wrap to the next digit.

nathanstitt: Ok, I've thought about it a bit more and think we're both correctish.  I still think my method (multiply by the decimal place value) is the "correct" way to do it, but yours is fine because the # of milliseconds since the epoch won't roll over for a looong time.  Right now it's at: `1471448051077` and won't add another digit until sometime in the year 2300.  😁    So it's safe to do that for sure.

However, it did last add a digit in the year 2000.  So any dates that are before that have the potential for losing digits and sorting incorrectly.  That's still very unlikely though.

Note: I tested using: `moment().valueOf()` => `1471448219520` and then inverted it by: `moment(1471448219520).toString()`

nathanstitt:
Nice, appears to solve the "too much padding when stacking" display bug

PR #1283, opened 2016-08-16
eh, pretty cool, webpack-dashboard.  worth it?

pandafulmanda:
What do you think?

Looks pretty cool, easier to read errors i think?

screen shot 2016-08-16 at 9 58 41 am


philschatz:
It looks like the speed issue is a common concern: https://github.com/FormidableLabs/webpack-dashboard/issues/18 https://github.com/FormidableLabs/webpack-dashboard/issues/30 but no fix yet

pandafulmanda:
oh, apparently they fixed it, will update

nathanstitt:
I tested this and seems about as fast to me.  I vote merge, thoughts @philschatz & @pandafulmanda ?

configs/webpack/base.coffee
```diff
@@ -181,6 +183,12 @@ makeDevelopmentBase = (projectConfig) ->
   publicPath = "#{servePath}/dist/"
   outputPath = "#{projectConfig.basePath}/"
 
+  if projectConfig.webpackDashboardSocketPort?
+    dashboardConfig =
+      port: projectConfig.webpackDashboardSocketPort
+
+  console.info("HALLO", dashboardConfig)
```
philschatz: Seems like we should elevate the fail-Travis-when-console-messages-are-present ; )

bin/serve
```diff
@@ -7,4 +7,10 @@ bin/checkinputs "$@"
 export OX_PROJECT=$1
 echo Serving: $OX_PROJECT
 
-webpack-dev-server --progress --config webpack.config.js
+dashboardSocket=$(grep 'webpackDashboardSocketPort' $OX_PROJECT/configs/base.coffee | sed 's|.*: \(.*\)|\1|')
```
philschatz: Is there a better way to handle this? grepping files seems very fragile . Could the port be defined in this file and then passed into `base.coffee` as an ENV variable?

bin/serve
```diff
@@ -7,4 +7,10 @@ bin/checkinputs "$@"
...
+
+if [ ${#dashboardSocket} -eq 4 ]; then
```
philschatz: Why 4? Could you add a comment?

configs/webpack/base.coffee
pandafulmanda: oops.

bin/serve
pandafulmanda: yes, to validate kinda, that it's a reasonable-ish value for a socket

bin/serve
pandafulmanda: yeah it's pretty gross right? apparently it has to run in the command, and can just be in the coffee, so i suppose we could run a node script that loads the env...yah ill look into that. i've been wanting to get [dotenv](https://www.npmjs.com/package/dotenv) in here anyways to make things nicer/more configurable

pandafulmanda:
@philschatz / @nathanstitt what do you guys think of using `#{project}/.env`s as opposed to the `#{project}/configs/base.coffee`?

nathanstitt:
Hm, it does't really matter to me.  It seems like it might be difficult to express all the nested data structures using env vars though, but you know how that's setup better 😁

philschatz:
Also, no objections on using `.env` but could that be a separate PR? That way it could be tested independently (also, it seems like changing to `.env` would change the scope of this PR

pandafulmanda:
heh. oops i broke everything.

PR #1284, opened 2016-08-16
Remember tab position in URL

nathanstitt:
Creates a `Tabs` component that records it's position using the `tab` query parameter.

It's used to remember tab selections on the follow screens:
 * Performance Review
 * CC dashboard
 * Student dashboard
 * Teacher's Student Scores
 * Teacher's quick metrics view when clicking an assignment from dashboard
 * Teacher's course settings

Each screen should remember the tab that is selected and return to it when the back button is pressed, then return to the original tab on forward.

It should also return to the last selected tab when a link goes away from the screen and the back button is used to return.

It does not go to the last viewed tab when links are used to visit a screen.  For instance the student dashboard will always start at the "ThisWeek" tab.


pandafulmanda:
i have some questions about this that would be easier to discuss in a call, hopefully we'll get to that tomorrow!

tutor/src/components/tabs.cjsx
```diff
@@ -0,0 +1,89 @@
...
+        {tab}
+      </a>
+    </li>
+
+  render: ->
+    <nav className={classnames('tutor-tabs', @props.className)}>
+      <ul role="tablist" className="nav nav-tabs">
+        {_.map @props.tabs, @renderTab}
+      </ul>
+      {@props.children}
```
pandafulmanda: nice, makes a lot of sense outside of the list

PR #1286, opened 2016-08-16
Fix/handle no personalized

pandafulmanda:
Address [FE needs to handle no personalized question error gracefully](https://www.pivotaltracker.com/n/projects/1156756/stories/128339225), as mocked up [here](http://mmwerz.axshare.com/#g=1&p=no_personalized_question_messaging&c=1)

- [x] No more error modal/stuck FE state when completing core steps
- [x] Update messaging on personalized placeholder

For another PR:
- [ ] ~~Update styling on personalized placeholder~~
- [ ] ~~Fake placeholder for homeworks that have deleted placeholders~~

### DIFF
--- a/tutor/src/flux/helpers.coffee
+++ b/tutor/src/flux/helpers.coffee
@@ -8,6 +8,14 @@ SAVING  = 'saving'
 DELETING = 'deleting'
 DELETED = 'deleted'
 
+STATES =
+  LOADING : LOADING
+  LOADED  : LOADED
+  FAILED  : FAILED
+  SAVING  : SAVING
+  DELETING: DELETING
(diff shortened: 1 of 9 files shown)


tutor/src/flux/helpers.coffee
```diff
@@ -8,6 +8,14 @@ SAVING  = 'saving'
 DELETING = 'deleting'
 DELETED = 'deleted'
```
nathanstitt: Since we're defining and exporting the new `STATES`, shouldn't we remove these?

nathanstitt:
Looks good @pandafulmanda 
screen shot 2016-08-17 at 1 00 06 pm

PR #1290, opened 2016-08-18
Tutor Change Student Id

jivewise:
Pivotal: https://www.pivotaltracker.com/story/show/127931475

Adds in the Change Student Id feature into tutor.  Moved the Change Student Id form into `shared/` along with corresponding tests.

screen shot 2016-08-18 at 5 31 17 pm
screen shot 2016-08-18 at 5 30 58 pm


tutor/src/api.coffee
```diff
@@ -172,6 +173,11 @@ start = (bootstrapData) ->
   apiHelper RosterActions, RosterActions.load, RosterActions.loaded, 'GET', (id) ->
     url: "/api/courses/#{id}/roster"
 
+  apiHelper StudentIdActions, StudentIdActions.load, StudentIdActions.loaded, 'GET', (courseId) ->
+    url: "/auth/status"
+  apiHelper StudentIdActions, StudentIdActions.save, StudentIdActions.saved, 'PATCH', (courseId, params) ->
```
nathanstitt: `/auth/status` is a CC endpoint.  While Tutor can technically use it (obviously), there's not any reason to.  Tutor get's the same data as part of it's bootstrap data, and it's stored in `CurrentUserStore` and `CourseStore`.  

For instance you can get the student ID from there by looking at `CourseStore.get(<courseId>).students`

Something I just realized.   There's no way to find the current student in `students`.  I see you're looking at the first one below in your new `StudentIdStore` but that's not going to work for the case of a teacher who's also a student.  In that case they'll be multiple students in the array.

I think we'll have to add the student id onto the BE's user model so that you can call `CurrentUserStore.getId()` like you can now call `CurrentUserStore.getName()`.  Then use that ID to find the correct student in `CourseStore.get(<courseId>).students`

I'll investigate this on the BE and make a PR there if needed

tutor/src/flux/student-id.coffee
```diff
@@ -0,0 +1,39 @@
...
+    @emitChange()
+
+  exports:
+    getId: (courseId) -> @_local[courseId]?.student_identifier
+    getErrors: (courseId) -> @_errors[courseId] or []
+}
+
+extendConfig(StudentIdConfig, new CrudConfig())
+{actions, store} = makeSimpleStore(StudentIdConfig)
+module.exports = {StudentIdActions:actions, StudentIdStore:store}
```
nathanstitt: Given my comments above about not needed `/auth/status`, I'd suggest getting rid of the `StudentIdActions`/`StudentIdStore` and just adding an action  `CourseActions.updateUsersStudentId(<newId>)` that would hit the `/api/user/courses/#{courseId}/student` endpoint and then update the `students` array once it completes.

tutor/src/helpers/time.coffee
```diff
@@ -14,6 +13,7 @@ TIME_LINKS =
   'US/Eastern': 'Eastern Time (US & Canada)'
   'US/East-Indiana': 'Indiana (East)'
   'Canada/Atlantic': 'Atlantic Time (Canada)'
+  'Asia/Shanghai': 'Shanghai'
```
nathanstitt: These populate the dropdown on the course settings. Do we want to add 'Asia/Shanghai' to them, or was this a debug thing that got left in ?

tutor/src/flux/helpers.coffee
```diff
@@ -47,10 +47,10 @@ CrudConfig = ->
     dontReload: (id) ->
       @_asyncStatus[id] is LOADED and @_HACK_DO_NOT_RELOAD
 
-    FAILED: (status, msg, id) ->
+    FAILED: (status, msg, data, id) ->
       @_asyncStatus[id] = FAILED
       @_errors[id] = msg
-      @_failed?({status, msg}, id)
+      @_failed?({status, msg, data}, id)
       unless status is 0 # indicates network failure
```
nathanstitt: is data being used here? If not, why dereference it from the object?

tutor/src/helpers/api.coffee
```diff
@@ -159,12 +159,12 @@ apiHelper = (Actions, listenAction, successAction, httpMethod, pathMaker, option
...
 
       rejected = (response) ->
-        {status, statusText, statusMessage, handled} = response
+        {status, statusText, statusMessage, handled, data} = response
         onRequestError(response, requestConfig)
         requestConfig.handleError?(response, args...)
         return if handled
 
-        Actions.FAILED(status, statusMessage, args...)
+        Actions.FAILED(status, statusMessage, data, args...)
```
nathanstitt: Same question as above re `data`

nathanstitt:
The panel looks really nice Jing, I like how the component is now shared between Coach and Tutor. I'll investigate the student id thing

tutor/src/api.coffee
jivewise: Yea, it'd be preferable to have it in the CurrentUserStore.  I was looking there, but couldn't find anything.

tutor/src/flux/student-id.coffee
jivewise: Agreed, I'll move this.

tutor/src/helpers/time.coffee
jivewise: Yep, just a debugging thing I left in there, I'll take this out.

tutor/src/flux/helpers.coffee
jivewise: Yes, we need this since the backend will return an error key that is displayed when the student id is already taken by another student.

nathanstitt:
I investigated the student id thing a bit and discovered that the BE does send the `role_id` which is equally good for finding the student record.

as a test, I added this to CourseStore:
```coffee
    getCurrentUsersStudentRecord: (courseId) ->
      course = @_get(courseId)
      role = _.findWhere(course.roles, type: 'student')
      if role then _.findWhere(course.students, role_id: role.id) else null
```

I think that should work well for you 😁

jivewise:
I'll take a look and update.

tutor/src/components/change-student-id.cjsx
```diff
@@ -0,0 +1,104 @@
...
+    {courseId} = @context.router.getCurrentParams()
+
+    if not newStudentId
+      StudentIdActions.addError(courseId, code:'blank_student_identifer')
+      @setState({})
+      return
+
+    StudentIdStore.on('student-id-saved', @saved)
+    StudentIdStore.on('student-id-error', @update)
+    StudentIdActions.save(courseId, {
```
nathanstitt: I think using explicit `on` here is going to cause errors, because they're being bound each time submit is called.  

So if there's an error that that's fixed and re-submitted, the methods will be bound twice, causing the callbacks to fire multiple times.

To fix that you could use `once`, or `bindStoreMixin` with `asyncButton`

tutor/src/components/change-student-id.cjsx
```diff
@@ -0,0 +1,104 @@
...
+  contextTypes:
+    router: React.PropTypes.func
+
+  onCancel: -> @goBack()
+
+  componentWillMount: ->
+    {courseId} = @context.router.getCurrentParams()
+    CourseStore.on('course.loaded', @update)
+    CourseActions.load(courseId)
+
```
nathanstitt: I don't see CourseStore being unbound anywhere.  You probably don't need to load it though, the course data is already loaded by the bootstrap data.

tutor/src/components/change-student-id.cjsx
```diff
@@ -0,0 +1,104 @@
...
+{ StudentIdStore, StudentIdActions } = require '../flux/student-id'
+
+ERROR_MAP = {
+  student_identifier_has_already_been_taken:
+    'The provided student ID has already been used in this course. ' +
+    'Please try again or contact your instructor.'
+  blank_student_identifer:
+    'The student ID field cannot be left blank. Please enter your student ID.'
+}
+
```
nathanstitt: I'd prefer the `ERROR_MAP` to live in the store.  Then it could fire the appropriate error based on server error response, and `blank_student_identifer` could be emitted during the set inside the store's action, rather than in the component

tutor/src/components/change-student-id.cjsx
```diff
@@ -0,0 +1,104 @@
...
+        { errorLi }
+      </ul>
+    </div>
+
+  renderSuccess: ->
+    <BS.Panel bsStyle='primary' className="change-id-panel">
+      <h3>You have successfully updated your student identifier.</h3>
+      <a className="btn btn-default">
+        <i className="fa fa-angle-left" /> Go Back
+      </a>
```
nathanstitt: Consider using `<Icon type="angle-left" />` component ?

tutor/src/components/change-student-id.cjsx
jivewise: Gotcha, moved.

tutor/src/components/change-student-id.cjsx
jivewise: Changed it to use `<Icon />` component.

tutor/src/components/change-student-id.cjsx
jivewise: We do need that there, otherwise, it won't load the new student ID if they go and edit it again.

tutor/src/components/change-student-id.cjsx
jivewise: Good point, changed those to use `.once`.

tutor/src/flux/student-id.coffee
```diff
@@ -0,0 +1,39 @@
...
+  blank_student_identifer:
+    'The student ID field cannot be left blank. Please enter your student ID.'
+}
+
+
+StudentIdConfig = {
+
+  _failed: (error, id) ->
+    @_errors[id] = error?.data?.errors
+    AppActions.resetServerErrors()
```
pandafulmanda: are you doing this to prevent the default error handler?  we've been doing that like [this](https://github.com/openstax/tutor-js/blob/master/tutor/src/api.coffee#L193), where `displayError` prevents the server error modal, and `handleError` allows for endpoint specific error handling

tutor/src/components/change-student-id.cjsx
```diff
@@ -0,0 +1,100 @@
...
+module.exports = React.createClass
+  contextTypes:
+    router: React.PropTypes.func
+
+  onCancel: -> @goBack()
+
+  componentWillMount: ->
+    {courseId} = @context.router.getCurrentParams()
+    CourseStore.once('course.loaded', @update)
+    CourseActions.load(courseId)
```
pandafulmanda: From what I can see, there is nothing that will get the `CourseStore` to reload it's student id once `student-id-saved`  fires.  It looks like it will only get updated if the student comes back to update their id again.

tutor/src/components/change-student-id.cjsx
```diff
@@ -0,0 +1,100 @@
...
+    @setState({})
+
+  goBack: ->
+    historyInfo = TransitionStore.getPrevious(@context.router)
+    url = historyInfo.path or 'dashboard'
+    @context.router.transitionTo(url)
+
+  onSubmit: (newStudentId) ->
+    {courseId} = @context.router.getCurrentParams()
+
```
pandafulmanda: any chance either here, in the two `onSubmit`s being fed from `tutor` and `coach` to check for where it's the same id before submitting? or in the `onSubmit` in the `shared` component, although that would make it tricky for any other side-effect from a "successful" submit to be called

tutor/src/components/change-student-id.cjsx
```diff
@@ -0,0 +1,100 @@
...
+    <BS.Panel bsStyle='primary' className="change-id-panel">
+      <BS.Row>
+        <ChangeStudentIdForm
+          label="Enter your school issued ID:"
+          title="Change your student ID"
+          onCancel={@onCancel}
+          onSubmit={@onSubmit}
+          saveButtonLabel="Save"
+
+          isBusy={StudentIdStore.isSaving()}
```
pandafulmanda: not critical, but this doesn't ever seem to be `true` when saving to trigger the disabled async state of the button

pandafulmanda:
screen shot 2016-08-24 at 11 21 31 am
may have to get this wording checked at some point, but will go ahead and merge.

PR #1291, opened 2016-08-18
IE11 Mathjax - use classNames instead of classList

darrenmason:
IE11 is buggy with / does not have support for classList.

http://caniuse.com/#feat=classlist
http://stackoverflow.com/questions/8098406/code-with-classlist-does-not-work-in-ie

rendering now, though still has some alignment issues.

mathjax-workin

### DIFF
--- a/shared/src/helpers/mathjax.coffee
+++ b/shared/src/helpers/mathjax.coffee
@@ -30,8 +30,9 @@ typesetDocument = (windowImpl) ->
 
   windowImpl.MathJax.Hub.Queue ->
     # Queue a call to mark the found nodes as rendered so are ignored if typesetting is called repeatedly
+    # uses className += instead of classList because IE
     for node in latexNodes.concat(mathMLNodes)
-      node.classList.add(MATH_RENDERED_CLASS)
+      node.className += " #{MATH_RENDERED_CLASS}"
 
 # Install a debounce around typesetting function so that it will only run once
 # every Xms even if called multiple times in that period


shared/src/helpers/mathjax.coffee
```diff
@@ -30,8 +30,9 @@ typesetDocument = (windowImpl) ->
 
   windowImpl.MathJax.Hub.Queue ->
     # Queue a call to mark the found nodes as rendered so are ignored if typesetting is called repeatedly
+    # uses className += instead of classList because IE
     for node in latexNodes.concat(mathMLNodes)
-      node.classList.add(MATH_RENDERED_CLASS)
+      node.className += " #{MATH_RENDERED_CLASS}"
```
nathanstitt: we're operating on "real" DOM Elements here, so `className` isn't a property, we'll need to use the `class` attribute directly

philschatz:
Should we be selecting MathML or SVG nodes? I thought we'd only be selecting the `<span data-math>` elements. If we are selecting the MathML or SVG nodes, is that a bug?

darrenmason:
@philschatz I don't think we are...  I should restate the description :)

philschatz:
In that case, should the fix be to not select the MathML/SVG nodes in the first place? It seems continuing to select the MathML/SVG nodes would just patch the existing problem and maybe cause other problems in the future?

shared/src/helpers/mathjax.coffee
darrenmason: I thought we could use this? https://developer.mozilla.org/en-US/docs/Web/API/Element/className

Seems to be working

darrenmason:
This fix should work with mathml or svg should they be selected in the future

nathanstitt:
I'm going to merge since it's conceptually the same as using `classList` and will hopefully fix the land-standing IE bugs, since we do know that IE's `classList` support isn't the best

PR #1292, opened 2016-08-18
cache failures to create practices and make practice button matching …

pandafulmanda:
## Before

you can get stuck in a loop if you really want to practice something.

practice-broken

## After

the practice button disables once it errors.

practice-fixed

Also, pulls code together a little more for practice buttons for hopefully switching to async creating of practice (#1310) and then only going to practice page when it successfully creates


tutor/test/components/performance-forecast/progress-bar.spec.coffee
```diff
@@ -1,25 +1,51 @@
...
+  params =
+    page_ids: pageIds
+
+  CoursePracticeActions.create(COURSE_ID, params)
+  CoursePracticeActions._failed({}, COURSE_ID, params)
+
+didRouterGoToPractice = ->
+  expect(Testing.router.transitionTo).to.have.been.calledWith( 'viewPractice',
+    { courseId: COURSE_ID }, { page_ids: ['2', '3'] }
+  )
```
nathanstitt: Nice way to test the router :)

tutor/src/components/performance-forecast/practice-button.cjsx
```diff
@@ -1,46 +1,39 @@
...
-    else
-      button
+    classes = classnames 'practice', @props.practiceType
+
+    <Practice
+      courseId={courseId}
+      page_ids={page_ids}>
+      <ButtonWithTip id={id} className={classes} getTip={@getTip} placement='top'>
+        {@props.title}<i />
+      </ButtonWithTip>
```
nathanstitt: Just curious but what's the deal with the blank `<i />` ?

pandafulmanda: I actually am not sure, i've just left that in for now from some previous version of things?  It's probably safe to remove right?

PR #1295, opened 2016-08-19
Removing sections from readings

jivewise:
PIvotal: https://www.pivotaltracker.com/story/show/127926577

There was a javascript error when removing sections from readings, this fixes that.

remove-reading


nathanstitt:
Good catch @jivewise !

PR #1296, opened 2016-08-19
adjustments to #1288 decrease use of delay and defer

pandafulmanda:
Not against `master`, against #1288


nathanstitt:
awesome improvements :)

PR #1297, opened 2016-08-19
Filter out dropped students from scores and speed it up a bit

nathanstitt:
Goes along with BE: https://github.com/openstax/tutor-server/pull/1235, but can be merged independently 

This changes the logic around a bit on the scores table so that it filters out dropped students and no longer re-sorts on every render.

No user-visible changes


tutor/src/components/scores/index.cjsx
```diff
@@ -25,209 +26,83 @@ Scores = React.createClass
...
-    # The period may not have been selected. If not, just use the 1st period
-    {sort, period_id, firstDataColumn, displayAs} = @state
-
-    data = ScoresStore.get(@props.courseId)
-    scores = if period_id
-      _.findWhere(data, {period_id})
+    asc = if @state.sort.key is key then (not @state.sort.asc) else false
+    @updateStudentData({sort: {key, asc, dataType}})
+  selectPeriod: (period) -> @updateStudentData({period_id: period.id})
+  setPeriodIndex:  (key) -> @updateStudentData({periodIndex: key + 1})
```
pandafulmanda: selecting different period tabs isn't working because `selectPeriod` gets called just before `setPeriodIndex`.  Just before `selectPeriod` has a chance to update the `period_id` on `@state`.  This meant that even though `selectPeriod` gets the proper score for the period, when `setPeriodIndex` gets called, it would be using the `@state` from before the `selectPeriod` `setState` gets called, using the old period :( .  This is basically my fault from a long time ago, and I think we should just get rid of `handleKeyUpdate` from `CoursePeriodsNavShell` and just pass the key into `handleSelect` as well, like [this](https://github.com/openstax/tutor-js/compare/dropped-scores...update/dropped-scores).  We were not really using `handleKeyUpdate` anyways.  What do you think?

bin/karma
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e
+
+bin/checkinputs "$@"
+
+export OX_PROJECT=$1
+echo Karma: $OX_PROJECT
+
+karma start test/karma.config.js --auto-watch --no-single-run
```
pandafulmanda: now that we have this, i wonder if we should just call this from with the `bin/tdd`?

tutor/src/components/scores/student-data-sorter.coffee
```diff
@@ -0,0 +1,66 @@
...
+      switch dataType
+        when 'score'
+          score = record.correct_exercise_count
+          if displayAs is 'number'
+            score or 0
+          else
+            percent(score, record.exercise_count) or 0
+        when 'completed'
+          progress = record.completed_exercise_count
+          percent(progress, record.exercise_count)
```
pandafulmanda: AWESOME.  nice. yay!

tutor/src/components/scores/student-data-sorter.coffee
```diff
@@ -0,0 +1,66 @@
...
+
+
+
+StudentDataSorter = ({sort, displayAs}) ->
+
+  return (a, b) ->
+    a = getSortValue(a, sort.key, sort.dataType, displayAs)
+    b = getSortValue(b, sort.key, sort.dataType, displayAs)
+    cmp = if a < b then -1 else if a > b then 1 else 0
+    return if sort.asc then cmp else (cmp * -1)
```
pandafulmanda: is "native" sort faster or `_.sortBy`?  maybe here it won't matter too much, and I couldn't find a straightforward answer comparing the two online, but I did find [`_.sortBy`'s](http://underscorejs.org/docs/underscore.html#section-41) code, which makes me think that'll be faster since it'll calculate the sort criteria just once per item, as opposed twice.

tutor/src/components/scores/index.cjsx
```diff
@@ -25,209 +26,83 @@ Scores = React.createClass
...
+  changeDisplayAs:(mode) -> @updateStudentData({displayAs: mode})
```
pandafulmanda: very clean, but it seems like trying to change the `mode` will call a lot of additional logic like `merge`, `sort`, `extend`, that doesn't seem like will change any other `state` property other than `displayAs` itself.

tutor/src/components/scores/table.cjsx
```diff
@@ -22,29 +24,58 @@ ColumnGroup = FixedDataTable.ColumnGroup
...
     dataType: React.PropTypes.string
     isConceptCoach: React.PropTypes.bool.isRequired
 
+
+  getInitialState: ->
+    tableWidth: 0
+    tableHeight: 0
+
+  componentDidMount: -> @sizeTable()
+  _resizeListener:   -> @sizeTable()
```
pandafulmanda: c:sunglasses:l yah, this logic makes so much more sense here that in `scores`

pandafulmanda:
really great improvements!

tutor/src/components/scores/student-data-sorter.coffee
nathanstitt: Yeah, great question!  I'd assumed that `Array.sort` would have to be faster since that's what `_.sortyBy` uses underneath.  Your point about calculating the value multiple times gave me pause though.  Unfortunately my go-to benchmarked jsperf.com is down :(

I whipped up: https://github.com/nathanstitt/jsbenchmarks which says:
```
> node benchmark.js
Array.sort x 8.17 ops/sec ±0.52% (24 runs sampled)
_.sortBy x 167 ops/sec ±1.05% (81 runs sampled)
Fastest is _.sortBy
```

So _.sortBy is indeed faster.  I'll switch to that, it's clearer anyways.

pandafulmanda: yah, im sad about jsperf being down :(

nathanstitt: Ok, I switched it and then realized another advantage of native sort.  With that you can switch the order inline by flipping the returned value.    With `_.sortyBy` you have to call `reverse()`  My benchmark suggests that `_.sortBy` is still cleaner so I'm leaving it that way, but code isn't quite as clean :(

tutor/src/components/scores/index.cjsx
nathanstitt: yeah but it has to.  `displayAs` influences how values are sorted.  also, merge should be pretty cheap since it's not a complex deeply nested structure.

nathanstitt:
K, back over to you @pandafulmanda - good suggestion on the diff between `Array.sort` and `_.sortBy`

PR #1298, opened 2016-08-19
Update icons

nathanstitt:
Also removes a bunch of unused icon files.   I believe it's safe to remove them since `less` will complain about missing asset files that are pointed to by `url`.

screen shot 2016-08-22 at 10 19 06 am


pandafulmanda:
are these really supposed to be different shades of yellow?

screen shot 2016-08-22 at 12 54 22 pm

nathanstitt:
hm, I hadn't noticed that.  The issue is that the "Visual Connections" is an SVG, while the other is just text with yellow background.  @Fredasaurus should we change the background color to match the SVG?

nathanstitt:
I went ahead and changed the section circles to use the new orangish yellow, I hereby dub it `openstax-tangerine`

screen shot 2016-08-22 at 4 45 48 pm

PR #1310, opened 2016-08-25
Fix/practice async

pandafulmanda:
to be considered after #1292 gets merged.  Not tested, tests not yet updated, and styling is not yet polished (for example, a failed/disabled button should have some explanation) but this is the basic idear, where the first one is a practice that cannot be created, and the second can be, and will complete creating before going to the task page:
## Before

practice-before
## After

practice-async


pandafulmanda:
waiting for more design

PR #1312, opened 2016-08-25
Don't mark tasks as not late when work is preformed

nathanstitt:
Uses the task helper for determining lateness.

The earlier check was marking exercises as non-late as soon as any work was performed

### DIFF
--- a/tutor/src/components/student-dashboard/event-info-icon.cjsx
+++ b/tutor/src/components/student-dashboard/event-info-icon.cjsx
@@ -3,6 +3,7 @@ BS       = require 'react-bootstrap'
 S = require '../../helpers/string'
 {TimeStore} = require '../../flux/time'
 moment   = require 'moment'
+TH = require '../../helpers/task'
 
 module.exports = React.createClass
   displayName: 'EventInfoIcon'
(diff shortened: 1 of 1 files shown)


tutor/src/components/student-dashboard/event-info-icon.cjsx
```diff
@@ -11,19 +11,22 @@ module.exports = React.createClass
...
+    dueAt = moment(event.due_at)
+    isIncomplete = event.complete_exercise_count isnt event.exercise_count
+    pastDue      = event.type is 'homework' and dueAt.isBefore(now, 'd')
+    workedLate   = moment(event.last_worked_at).isAfter(dueAt)
+
+    return null unless workedLate or (pastDue and isIncomplete)
 
     # use 'day' granularity for checking if the due date is today or after today
-    status = if due.isSame(now, 'd') then 'incomplete' else 'late'
+    status = if dueAt.isSame(now, 'd') then 'incomplete' else 'late'
```
pandafulmanda: I wonder if we'll need to move from `day` granularity soon, since we do have due times now, but maybe another pr down the orad

nathanstitt: Hm, good point, we maybe will have to

PR #1314, opened 2016-08-25
dont try to render questions when there are no stats for it

pandafulmanda:
Will need to clarify with @openstax/tutor-be as to why this might happen, but for some reason, for this [Bug: Reviewing assignment not loading for some assignments](https://www.pivotaltracker.com/n/projects/1156756/stories/128086791), this particular `review` has`content` where `question.id` is `52229`, but there is no matching `question_stats` where `questions_id` is `52229`.

## Before

screen shot 2016-08-25 at 4 08 18 pm

## After

screen shot 2016-08-25 at 4 03 08 pm


nathanstitt:
No idea how to test this, but looks like the correct approach 😁

PR #1319, opened 2016-08-29
Fix teacher review styles on tablets

jivewise:
Pivotal: https://www.pivotaltracker.com/story/show/128105343

Before:
image

After:
screen shot 2016-08-29 at 10 06 51 pm

### DIFF
--- a/tutor/resources/styles/components/task-step/index.less
+++ b/tutor/resources/styles/components/task-step/index.less
@@ -11,7 +11,7 @@
   .exercise-typography();
   padding: @tutor-card-body-padding;
   @media (max-width: @tutor-card-padding-collapse-breakpoint) {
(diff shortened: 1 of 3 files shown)


nathanstitt:
It looks to me like this isn't an "tablet" bug really, but more a "width less than around 1000px" bug.

Your fix does work, but it seems to me like we're wasting a lot of space on the left of the exercise. 

Instead of adding more padding to the left, can we make the correct/incorrect and counts smaller?  As an example, your fix with iPad width of 768 px looks like:

screen shot 2016-08-31 at 12 27 03 pm

jivewise:
Okay, updated the icon and text to take up less space, as in screenshot below:
screen shot 2016-09-02 at 7 35 54 am

nathanstitt:
Thanks @jivewise  - I think that'll work great

PR #1320, opened 2016-08-29
Unified page navigation

nathanstitt:
Introduces a new `PagingNavigation` component which handles forward / back paging for
 * Exercise Details view on QL & HW builder
 * Reference book
 * Readings

It has unified behavior:
 * clickable gutter area
 * chevron arrows instead of solid triangle arrows 
 *  gutter area does not change color on hover; instead, chevron arrow darkens

screen shot 2016-08-29 at 4 49 43 pm
screen shot 2016-08-29 at 4 52 45 pm

screen shot 2016-08-29 at 4 48 52 pm


pandafulmanda:
still reviewing, but this is so awesome.

tutor/src/components/reference-book/reference-book.cjsx
```diff
@@ -18,7 +18,7 @@ MENU_VISIBLE_BREAKPOINT = 1350
 module.exports = React.createClass
   displayName: 'ReferenceBook'
 
-  mixins: [ResizeListenerMixin, ChapterSectionMixin]
+  mixins: [ResizeListenerMixin]
```
pandafulmanda: `sectionFormat` is being used from `ChapterSectionMixin`, ref book errors without it

tutor/src/components/reference-book/page-navigation.cjsx
```diff
@@ -5,33 +5,55 @@ React = require 'react'
 classnames = require 'classnames'
 Router = require 'react-router'
 
-PageNavigation = React.createClass
+PagingNavigation = require '../paging-navigation'
+{ReferenceBookStore} = require '../../flux/reference-book'
+{ChapterSectionMixin} = require 'shared'
+
+RefenceViewPageNavigation = React.createClass
```
pandafulmanda: `ReferenceViewPageNavigation`

tutor/src/components/reference-book/reference-book.cjsx
```diff
@@ -7,7 +7,7 @@ classnames = require 'classnames'
 
 NavBar = require './navbar'
 Menu = require './slide-out-menu'
-{ChapterSectionMixin, ResizeListenerMixin} = require 'shared'
+{ResizeListenerMixin} = require 'shared'
 PageShell = require './page-shell'
 PageNavigation = require './page-navigation'
```
pandafulmanda: hmm, i think it would be more clear here to use this as `ReferenceViewPageNavigation` vs `PageNavigation`

tutor/resources/styles/components/paging-navigation.less
```diff
@@ -0,0 +1,31 @@
...
+    .tutor-icon {
+      z-index: 1; //appear above content when page is extremely narrow
+      position: fixed;
+      top: 50vh;
+      color: @tutor-neutral-light;
+      font-size: 150px;
+      .transition(color 150ms linear);
+    }
+    &[disabled] {
+      cursor: not-allowed;
```
pandafulmanda: I can see this being slightly confusing to a user.  To the user, they are hovering their mouse over the side of the page where it looks like a normal background/a continuation of the rest of the page, but they see a not-allowed cursor.

tutor/resources/styles/components/paging-navigation.less
```diff
@@ -0,0 +1,31 @@
+.tutor-paging-navigation {
+  .flex-display();
+  .justify-content(center);
+  .paging-control {
+    .flex(1);
+    .flex-display();
```
pandafulmanda: yay for big hit-area! :)

pandafulmanda:
weird sizing on spacer type reading steps: 
screen shot 2016-08-31 at 4 21 13 pm

pandafulmanda:
overall looks awesome.  paging by keyboard works great as well!

Fredasaurus:
Looking good! A couple things from Robert's design that I want to make sure don't get overlooked. See attachement.
screen shot 2016-08-31 at 4 43 06 pm

tutor/src/components/reference-book/reference-book.cjsx
nathanstitt: Ah good catch.  I'll add it

tutor/src/components/reference-book/page-navigation.cjsx
nathanstitt: Ug. Auto-complete is so nice until you speeell something wrong once and then manage to do the same thing all over the place.  Thanks!

tutor/resources/styles/components/paging-navigation.less
nathanstitt: Yeah, i kinda wondered about that.  What I was trying to prevent was confusion over "where'd the arrows go"  Originally I had them still showing, but greyed out very light.  Then I hid them all together, but neglected to remove the cursor style

nathanstitt:
@pandafulmanda Yeah, the arrows do appear funny on cards that are short, but that's how they are currently as well.  However that's something I'm going to tackle as part of fixing the styles on QA/HW details view.

@Fredasaurus Good points.  The arrows are currently centered inside the gutter area.  They used to be a fixed pixel away from the edge of the screen.  I think centered is better, but neither meets the spec that you point out.  I'll see what I can come up with on that.

To have the "legs" with flat ends, we'll need to use an SVG.  Currently it's font-awesome icon.  I can probably make one of those as well.

pandafulmanda:
@nathanstitt, it's not about them being short, it's about them being skinnier than the other cards.  due to some of the flex stuffs i think

nathanstitt:
@Fredasaurus - I've limited the expansion of the arrows so they remain approx 60 px from the edge, but collapse inward as needed and switched to SVG icons.

Left arrow is hovered below:
screen shot 2016-08-31 at 6 14 23 pm

nathanstitt:
Looks like the last round of changes has a strange IE bug I'm tracking down

nathanstitt:
@pandafulmanda should be good to go now. It's currently deployed it to `tutor-dev`  where I was testing IE stuffs if you wanna look at it there.

PR #1321, opened 2016-08-30
Handle duplicate id errors when restoring dropped students

nathanstitt:
Goes with BE https://github.com/openstax/tutor-server/pull/1261

screen shot 2016-08-30 at 10 38 31 am


pandafulmanda:
something maybe to ask UX later -- how to handle, once errored, a user maybe wanting to bring back the original value

PR #1324, opened 2016-08-30
listen for logout.received when opening from a coach view route.

pandafulmanda:
Fixes [logging out from coach does not close coach](https://trello.com/c/OJkEmiFh/22-concept-coach-bug-log-out-doesn-t-work).

Specifically, as @nathanstitt found, this happened whenever the webview was loaded with a coach route suffix.  When this happens, the `cc.open` function where we were adding the listeners for `logout.received` does not get called.  This PR moves adding the listeners on `cc.initialized`, which will always get called, regardless of webview loading with or without coach route suffix.

### DIFF
--- a/coach/src/concept-coach/index.cjsx
+++ b/coach/src/concept-coach/index.cjsx
@@ -123,13 +123,16 @@ class ConceptCoachAPI extends EventEmitter2
       componentModel.channel.emit('close.clicked')
 
     @close = props.close
+
+    # Needs to be added on initialize since opening from a coach path calls
+    # initialize, and not open.  On will handle multiple logging in and out.
+    # 
+    # Wait until our logout request has been received and the close
+    User.channel.on('logout.received', @close)
+
     @component = coachWrapped.render(mountNode, props)
 
   open: (props) ->
-    # wait until our logout request has been received and the close
-    User.channel.once 'logout.received', =>
-      @close()
-
     openProps = _.extend({}, props, open: true)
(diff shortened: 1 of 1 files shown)


coach/src/concept-coach/index.cjsx
```diff
@@ -123,13 +123,16 @@ class ConceptCoachAPI extends EventEmitter2
...
 
     @close = props.close
+
+    # Needs to be added on initialize since opening from a coach path calls
+    # initialize, and not open.  On will handle multiple logging in and out.
+    # 
+    # Wait until our logout request has been received and the close
+    User.channel.on('logout.received', @close)
+
     @component = coachWrapped.render(mountNode, props)
```
nathanstitt: I seem to remember that at one time we had problems with webview calling initialize being multiple times, which is why we added the `remove()` call on line 112 above.  

Should we add a `User.channel.off('logout.received', @close) to the `remove` method so they're balanced?

pandafulmanda: [stopModelChannels](https://github.com/openstax/tutor-js/blob/master/coach/src/concept-coach/index.cjsx#L68-L70) offs the listeners.  Webview calls the `destroy` as the opposite of `initialize`.  I'll check a little more carefully.  I thought about the `off` when changing from `once` to `on`, but it needed the `on`, for like, if you logout, login, and then logout again.

nathanstitt: ok, cool.  just wasn't sure if you'd considered that 😁

PR #1326, opened 2016-08-31
CC Dashboard: Formatting issues on android tablet

darrenmason:
regarding: https://www.pivotaltracker.com/story/show/127123201

updated button spacing and icon position

screenshot from 2016-09-02 11 42 12

### DIFF
--- a/tutor/resources/styles/components/cc-dashboard/index.less
+++ b/tutor/resources/styles/components/cc-dashboard/index.less
@@ -3,8 +3,15 @@
     span {
       color: @tutor-neutral;
       #fonts .sans(1.2rem, 1.2rem);
-      font-weight: 300;
+      font-weight: 400;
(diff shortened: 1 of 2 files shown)


nathanstitt:
Hm, that does fix the "Detailed Scores" button, but the wording beside the icons still doesn't look very good.  Is there anyway to keep the icon from butting up against the text, and fixing the line-height of the labels when they wrap?  

It also seems like the icon should appear after the text since that's how all the other tooltip info icons are displayed elsewhere.

darrenmason:
I moved the icon after the text and changed the text to make it more readable and I shortened it slightly, let me know if that works.  (The word "Performance" is implied and can be read in the tooltip)

nathanstitt:
Looks good @darrenmason, thanks!

PR #1327, opened 2016-08-31
Display data-type=abstract elements in content

nathanstitt:
From https://trello.com/c/Asuv9A1Y/69-bug-abstracts-div-data-type-abstract-not-being-displayed-in-college-books-readings-or-reference-view

This styles abstracts like learning-objectives.  The only difference is that they do not have the separating line.  I had to disable that because all abstracts do not have a "In this chapter you will learn about" pre-amble.  Without that, the line looks weird because it doesn't separate anything.

HS Physics abstract with pre-amble:

screen shot 2016-09-06 at 12 00 42 pm

College physics without one:
screen shot 2016-09-06 at 12 01 28 pm

### DIFF
--- a/tutor/resources/styles/book-content/learning-objectives.less
+++ b/tutor/resources/styles/book-content/learning-objectives.less
@@ -1,4 +1,6 @@
-.learning-objectives {
+.learning-objectives,
+[data-type=abstract]
+{
   &.note:before {display: none;}
   #fonts > .sans(1.8rem, 1.8rem);
   width: 100%;
@@ -9,6 +11,7 @@
   margin-top: -@tutor-card-body-padding-vertical;
   padding: 10px 20px;
   border: none;
+  &:empty { display: none; }
   // hide the "Section learning Objectives" phrase
   > div[data-type=title] {
     display: none;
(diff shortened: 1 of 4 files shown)


pandafulmanda:
should this be limited to non-intro abstracts?

screen shot 2016-08-31 at 5 04 02 pm

nathanstitt:
Hm, the abstracts I found did not have the "in this chapter you will learn" For those we shouldn't say "Abstract"  if we do want to display them.

Fredasaurus:
They look good. There shouldn't be a space between the abstract and the image.

nathanstitt:
Thanks @Fredasaurus -  the margin between the abstract and splash image is because I used the existing styling for `learning-objective` elements, which had it.  

Should the margin be removed from the `learning-objective` heading as well?

nathanstitt:
I went ahead and removed the bottom margin.  I think that looks best in all cases.  Screenshots above are updated with latest.

tutor/resources/styles/book-content/learning-objectives.less
```diff
@@ -1,14 +1,16 @@
...
   #fonts > .sans(1.8rem, 1.8rem);
   width: 100%;
   min-height: 100px;
-  margin-bottom: 40px;
   .book-content-full-width();
   .flex-display(inline-flex);
   margin-top: -@tutor-card-body-padding-vertical;
   padding: 10px 20px;
   border: none;
+  &:empty { display: none; }
```
pandafulmanda: c:sunglasses:l

PR #1328, opened 2016-08-31
Add section & name that reading step comes from

nathanstitt:
Extracts the "Comes from section/ title" from exercises and shares it with readings.

screen shot 2016-08-31 at 4 42 57 pm


pandafulmanda:
code looks good to me, any thoughts @openstax/ux ?

PR #1329, opened 2016-09-01
Tappable Icons with Tooltips

jivewise:
Pivotal: https://www.pivotaltracker.com/story/show/128099687

With this fix:
screen shot 2016-09-01 at 11 50 00 am

### DIFF
--- a/tutor/src/components/icon.cjsx
+++ b/tutor/src/components/icon.cjsx
@@ -30,10 +30,11 @@ module.exports = React.createClass
       'clickable': @props.onClick or (@props.tooltip and @props.tooltipProps.trigger is 'click')
     })
 
-    if not @props.tooltip
+    unless @props.tooltip
       return <i {...@props} className={classNames} />
 
     icon = <button {...@props} className={classNames} />
+    classNames =
     tooltip = <BS.Tooltip id={@state.uniqueId}
       className={classnames({'on-navbar': @props.onNavbar})}
       >{@props.tooltip}</BS.Tooltip>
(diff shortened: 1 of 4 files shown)


tutor/src/components/icon.cjsx
```diff
@@ -30,13 +30,13 @@ module.exports = React.createClass
...
+    if not @props.tooltip
+      return <i {...@props} className={classNames} />
+
+    icon = <button {...@props} className={classNames} />
+    tooltip = <BS.Tooltip id={@state.uniqueId}
+      className={classnames({'on-navbar': @props.onNavbar})}
+      >{@props.tooltip}</BS.Tooltip>
+
+    <BS.OverlayTrigger {...@props.tooltipProps} overlay={tooltip}>{icon}</BS.OverlayTrigger>
+
```
nathanstitt: Good idea to use button!  I think the non-existent "OX Styleguide" says to use `unless` instead of `if not` (line 33) though 😁

tutor/test/all-tests.coffee
```diff
@@ -5,6 +5,9 @@
 # TODO figure out why.
 #
 # Components
+require './components/icon.spec'
+
+###
 require './components/course-listing.spec'
```
nathanstitt: This is also required on line 46 below.  Maybe this was left in by mistake ?

jivewise: Yep, let it in by mistake, taking it out now.

nathanstitt:
@jivewise Sorry, but I didn't notice that you weren't removing the button styling from the icons.  They look like this to me:

screen shot 2016-09-01 at 12 12 25 pm


I think we should remove the border and background color so the appearance of all our icons doesn't change.  Adding:

```
    border: 0;
```

To the `tutor-icon` styles seems to fix it for me.

nathanstitt:
Looks great @jivewise - thanks!

PR #1330, opened 2016-09-01
Readings: Course name and assignment name overlap in a reading assignment on tablet

darrenmason:
regarding: https://www.pivotaltracker.com/story/show/127732387

There is an edge case when a reading name is long but the header has not switched to compact mode for certain tablet resolutions/orientations.  In this case the course title and reading title may overlap. 

screenshot from 2016-09-01 11 45 52

screenshot from 2016-09-01 11 45 47

text ellipsis way w/ tooltip:

screenshot from 2016-09-02 10 55 40

### DIFF
--- a/tutor/resources/styles/global/navbar.less
+++ b/tutor/resources/styles/global/navbar.less
@@ -159,6 +159,9 @@
     display: flex;
     font-size: 1.8rem;
     font-weight: bold;
+    background: @tutor-white;
+    z-index: 1;
+    box-shadow: 0 0 5px 10px @tutor-white;
 
     .fa {
         color: #888;


nathanstitt:
Hm, that's a neat idea to have the assignment title overlap and obscure the course title by setting a  back-ground color on it.

Did you try to set the course title to be overflow text-ellipsis ?  That seems like it might be a better approach since it would visually indicate that the course title was partially hidden.

Maybe that didn't work for some reason though, in which case this is a good approach.

darrenmason:
Ellipsis is possible as well.  I added a commit with the text-ellipsis way.  In order to avoid overlapping the course title which also has ellipsis, the maximum length of the reading title is about 14 characters at current font size.

nathanstitt:
@darrenmason Sorry - I'd meant if we could apply ellipsis  to the course name when it's being overwritten.  I think it's more important to show the complete assignment name, and we definitely shouldn't limit that to 14 chars.  However I'm not sure offhand if it'll be possible to use ellipsis on the course name or not, and if we can't then I think the "white background overlay" approach is good.

darrenmason:
updated this to use background method.

nathanstitt:
Looks good @darrenmason.  Thanks for investigating using ellipsis even though it didn't pan out.

PR #1333, opened 2016-09-02
grouped figures should be allowed to float naturally next to each toher

pandafulmanda:
## Before
screen shot 2016-09-02 at 4 25 35 pm

## After

screen shot 2016-09-02 at 4 23 53 pm

1.2 History of Sociology


nathanstitt:
Awesome!  I noticed that earlier on that section also, but hadn't poked at it yet.

Fredasaurus:
Yay. I'm hoping in the future we can spec the art so that it is one image instead of several.

PR #1331, opened 2016-09-01
note stylings

pandafulmanda:
For several cards related to reading content styling:

* [note stylings through different books](https://trello.com/c/astbP3CN)
* Also addresses [Sims sizing](https://trello.com/c/JPCLeC5s)
* [soc colors](https://trello.com/c/sUILjntp)
* [splash opacity](https://trello.com/c/ow6tQsKF)
* [splash padding](https://trello.com/c/IZ6jxO7V)

- [x] fix the notes
- [x] organize/make easier to add new note selectors in future
- [x] check reading step stylings
- [x] read through old discussions to make sure styles match intent
- [x] updated the individual colors for the different books
- [x] document changes
- [ ] look into funny numbering stuffs that happens.
- [ ] check with ux
  * will do this later and have a follow up PR

# Accommodates both types of feature labels:

These different HTML for note feature labels can be seen within Biology HS or between different books.

Where they are defined in the HTML as

## `<div class="note" data-label="I am the feature label">`

screen shot 2016-09-02 at 4 51 49 pm

## `<div class="title" data-type="title">I am the feature label</div>` is the first-child in the note

when `data-label=""`/is missing

screen shot 2016-09-02 at 4 52 42 pm

## Works in assigned Readings as well

screen shot 2016-09-02 at 4 55 18 pm

## Sociology Theme

screen shot 2016-09-02 at 4 46 31 pm

screen shot 2016-09-02 at 4 56 08 pm

## Fixes Sims in readings and ref view

screen shot 2016-09-02 at 3 03 36 pm

PR #1334, opened 2016-09-06
Book Content: College Physics section 2.5 has incorrectly sized sim

darrenmason:
regarding: https://www.pivotaltracker.com/story/show/127010335

screenshot from 2016-09-06 12 01 08
screenshot from 2016-09-06 12 00 43


pandafulmanda:
The original ticket refers to problems with sims in notes which has been fixed in #1331.  There may be new problems now with sims due to responsiveness added from #1320's flex usage.  We may need to come up with a different solution -- sims have not been tested in widths < `960`

darrenmason:
Maybe it's okay if they stick out the sides?  Since they aren't necessary designed for a narrow viewport

PR #1335, opened 2016-09-06
Follow Up reading content styling details

pandafulmanda:
follow up to #1331

## Show worked example solutions

Also, get rid of extraneous exercise numbers

### Before
screen shot 2016-09-07 at 5 15 01 pm

### After
screen shot 2016-09-07 at 5 22 16 pm

## Fix imported exercise rendering

Make styled (almost) the same as embedded exercises, fix numbering

### Before
screen shot 2016-09-07 at 5 14 43 pm

### After
screen shot 2016-09-07 at 5 08 18 pm

### Embedded exercises before
screen shot 2016-09-07 at 5 15 35 pm

## Single stepped notes have clear background

screen shot 2016-09-07 at 5 26 33 pm

### Before
screen shot 2016-09-07 at 5 14 30 pm

### After
screen shot 2016-09-07 at 5 08 07 pm

### DIFF
--- a/shared/resources/styles/components/question.less
+++ b/shared/resources/styles/components/question.less
@@ -22,10 +22,13 @@
 @feedback-vertical-spacing: @feedback-horizontal-spacing;
 @feedback-horizontal-buffer: 2 * @feedback-horizontal-spacing;
 
-.openstax-bubble(@diameter, @border-size){
+@openstax-bubble: {
+  @diameter: @answer-bubble-size;
+  @border-size: 2px;
+
   width: @diameter;
   height: @diameter;
-  border-radius: 2 * @diameter;
+  border-radius: @diameter / 2;
   border-width: @border-size;
   border-style: solid;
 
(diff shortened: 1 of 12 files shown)


nathanstitt:
Looks great @pandafulmanda 

I was positive that there were abstracts that had the "preamble" text directly inside the main div. i.e.:
```html
<div data-type="abstract">
  By the end of this section you will learn things
  <ul>...</ul>
</div>
```

I can't find any to test the new styles on though, so maybe I just imaged that.  If there are any, I'm pretty sure they won't display correctly.   However there probably isn't too many (give that I can't find them again).

pandafulmanda:
there were in physics hs...that's why there's [dom manipulation here](https://github.com/openstax/tutor-js/pull/1335/files#diff-c58689bba914b367fa406e97cfd2aca3R176).  I saw that we were already doing that with the `splash` so i was like...eh

PR #1336, opened 2016-09-06
Don't warn about students working when deleting unopened assignments

nathanstitt:
Work was done via the the "get the code to a place where you can make the easy change, then make it" philosophy.

Also, I'm surprised we don't have a "isPlanOpened" helper somewhere.  Maybe I just missed it though?

screen shot 2016-09-06 at 4 44 14 pm

### DIFF
--- a/tutor/src/components/task-plan/reading.cjsx
+++ b/tutor/src/components/task-plan/reading.cjsx
@@ -132,6 +132,7 @@ ReadingPlan = React.createClass
 
     footer = <PlanFooter
       id={id}
+      isVisibleToStudents={@state.isVisibleToStudents}
       courseId={courseId}
       onPublish={@publish}
       onSave={@save}
--- a/tutor/src/components/task-plan/reading.cjsx
+++ b/tutor/src/components/task-plan/reading.cjsx
@@ -132,11 +132,11 @@ ReadingPlan = React.createClass
 
     footer = <PlanFooter
       id={id}
-      isVisibleToStudents={@state.isVisibleToStudents}
       courseId={courseId}
       onPublish={@publish}
       onSave={@save}
       onCancel={@cancel}
+      isVisibleToStudents={@state.isVisibleToStudents}
       getBackToCalendarParams={@getBackToCalendarParams}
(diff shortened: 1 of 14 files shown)


tutor/src/helpers/plan.coffee
```diff
@@ -27,4 +27,9 @@ PlanHelper =
     {isPublishing, publishStatus}
 
 
+  isPlanOpen: (plan) ->
+    now = moment(TimeStore.getNow())
+    !! _.find plan?.tasking_plans, (tasking) -> moment(tasking.opens_at).isBefore(now)
```
pandafulmanda: `isPlanOpen`/something similar does exist.  It's just my fault that it's in a weird place [in the `PlanMixin`](https://github.com/openstax/tutor-js/blob/master/tutor/src/components/task-plan/plan-mixin.cjsx#L25-L27).  This does get pulled in to the parents of the `PlanFooter`, so the `state.isVisibleToStudents`  could be passed in to `PlanFooter` and `DeleteLink` as a `prop`.

pandafulmanda:
yayyy for breaking the footer into smaller components!

tutor/src/helpers/plan.coffee
nathanstitt: Yeah, I'd prefer to have all the components in the tree use the same logic.  I'll pass `isVisibleToStudents` into Footer and and then onto the child components.

pandafulmanda: if this is no longer used, could we remove?

nathanstitt: heh, yeah. That was an oversight - I meant to remove when I switched to the `isVisibleToStudents` prop.  Thanks!

tutor/src/components/task-plan/reading.cjsx
```diff
@@ -132,6 +132,7 @@ ReadingPlan = React.createClass
 
     footer = <PlanFooter
       id={id}
+      isVisibleToStudents={@state.isVisibleToStudents}
```
pandafulmanda: Should these get passed into external and/or event footers as well?

PR #1340, opened 2016-09-07
Performance Forecast tooltip on tablet

jivewise:
Fixes performance forecast tooltip not showing up on tablets

Pivotal: https://www.pivotaltracker.com/story/show/128026799
performance-forecast-tablet

### DIFF
--- a/tutor/src/components/performance-forecast/info-link.cjsx
+++ b/tutor/src/components/performance-forecast/info-link.cjsx
@@ -1,5 +1,6 @@
 React = require 'react'
 BS = require 'react-bootstrap'
+Icon = require '../icon'
 
 MESSAGES = {
   student: [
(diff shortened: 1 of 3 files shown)


tutor/src/components/performance-forecast/info-link.cjsx
```diff
@@ -37,5 +37,5 @@ module.exports = React.createClass
     </BS.Tooltip>
 
     <BS.OverlayTrigger placement='right' overlay={tooltip}>
-      <span className='info-link'></span>
+      <button className='info-link'></button>
     </BS.OverlayTrigger>
```
nathanstitt: Hm, looking at this code, I think it pre-dates the `Icon` component which has built-in tooltip capability.  Since you converted it to use a `<button>` internally, could we swap all the entire render here with something like:

`<Icon type='info-circle' tooltip={MESSAGES[@props.type]} />`

If that works we could also shed all (or most) of the custom `info-link` styles.

nathanstitt:
@jivewise I also see that @pandafulmanda has #1342 open, which will set the styles you need

jivewise:
@nathanstitt updated to remove the styles and use the Icon component.

nathanstitt:
Looks great @jivewise, but I think there's also some styles in https://github.com/openstax/tutor-js/blob/master/tutor/resources/styles/components/performance-forecast/heading.less#L29-L34  and https://github.com/openstax/tutor-js/blob/master/tutor/resources/styles/components/performance-forecast/index.less#L8-L13 that could be cleaned up as well.

I'm not sure why the second style link is there, it doesn't seem to be needed at all and isn't being used.

nathanstitt:
Thanks @jivewise - good call on using `<button>` for the icon for mobile support.

PR #1342, opened 2016-09-07
tutor-icon followup

pandafulmanda:
remove hanging classNames assignment, move button de-styling to icon.less so that buttons tutor-icons not in dashboard will revert to original style

## Before

screen shot 2016-09-07 at 6 14 57 pm

## After

screen shot 2016-09-07 at 6 14 27 pm

### DIFF
--- a/tutor/resources/styles/components/cc-dashboard/index.less
+++ b/tutor/resources/styles/components/cc-dashboard/index.less
@@ -13,10 +13,7 @@
         color: @tutor-neutral-darker;
       }
     }
-    button {
-      border: 0;
-      background-color: transparent;
-    }
+
     .tutor-icon {
       margin-left: 0.5rem;
     }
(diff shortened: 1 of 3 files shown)


nathanstitt:
Thanks @pandafulmanda!

I looked into where that came from and it looks like it was introduced in #1329.  That PR set the `border: 0; background-color: transparent;` styles only on https://github.com/openstax/tutor-js/blob/master/tutor/resources/styles/components/cc-dashboard/index.less#L16-L19  (on my suggestion 😞)   When it probably should have done what you just did.

Can we remove the now unneeded  styles that were added to `cc-dashboard/index.less` ?

pandafulmanda:
yes and they are removed from [`cc-dashboard/index.less`](https://github.com/openstax/tutor-js/pull/1342/files#diff-437dafb3bdda92448884cf722b272844L16)

nathanstitt:
😆 oh wow. so they are (RIGHT ABOVE MY COMMENT)

PR #1346, opened 2016-09-08
handle sim width in 3 cases -- base, skinny, and wide, and note vs non-note

pandafulmanda:
Sims will be
* `100%` of content width on small screens
* `960px`, centered, for sizes above `960px`
  * squishy margin for sizes below `1020px`
  * fixed margin, centered within content, bulging out of the content, for sizes above `1020px`

### Ref View
responsive-ref

### Task View
responsive-task

Take note that this particular sim happens to be responsive.  Most of our sims are not.  They will center and size down as window size shrinks.  This is a temporary solution, as sims will not be workable below a certain width.

### DIFF
--- a/tutor/resources/styles/book-content/questions.less
+++ b/tutor/resources/styles/book-content/questions.less
@@ -23,7 +23,7 @@ section {
     }
 
     // make fully imported questions imitate embedded exercises
-    ol {
+    > ol:not(:first-child) {
       counter-reset: lowerAlpha;
       list-style: none;
       padding-left: 3rem;
--- a/tutor/resources/styles/book-content/questions.less
+++ b/tutor/resources/styles/book-content/questions.less
@@ -23,6 +23,8 @@ section {
     }
 
     // make fully imported questions imitate embedded exercises
+    // If the list follows something else in the problem,
+    // kinda assume it is most likely multiple-choices.
     > ol:not(:first-child) {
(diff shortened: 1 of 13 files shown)


tutor/resources/styles/components/reference-book/page.less
```diff
@@ -5,15 +5,15 @@
     margin-right: -40px;
 
     .center-panel {
-      width: @reference-book-page-width;
+      // width: @reference-book-page-width;
       margin: 0 auto;
```
nathanstitt: // could be removed 😁

tutor/resources/styles/book-content/questions.less
```diff
@@ -23,7 +23,7 @@ section {
     }
 
     // make fully imported questions imitate embedded exercises
-    ol {
+    > ol:not(:first-child) {
       counter-reset: lowerAlpha;
```
nathanstitt: What's special about a ol that's first-child ?  Could we add a comment to indicate that so future us can remember the reason?

tutor/resources/styles/components/reference-book/page.less
pandafulmanda: oh ooops

tutor/resources/styles/book-content/questions.less
pandafulmanda: yes, very good point

PR #1351, opened 2016-09-14
Fix reading scrolling on MPQ

nathanstitt:
https://trello.com/c/28Do07p1/110-mpq-in-reading-scrolls-back-and-forth-between-parts-when-typing-in-free-response-box

While readings shouldn't have MPQ, occasionally one will sneak through.  For an example view the link on the card above.

### DIFF
--- a/tutor/src/components/task-step/exercise.cjsx
+++ b/tutor/src/components/task-step/exercise.cjsx
@@ -123,27 +123,21 @@ module.exports = React.createClass
     @setCurrentStep(current)
 
   setCurrentStep: (currentStep) ->
-    return unless currentStep isnt @state.currentStep
-
+    return if currentStep is @state.currentStep
     @setState({currentStep})
     @props.goToStep(currentStep)
 
-  setCurrentStepByStepId: (id) ->
-    {taskId} = @props
-    stepNavIndex = TaskStore.getStepNavIndex(taskId, id)
-    @setCurrentStep(stepNavIndex)
-
-  onFreeResponseChange: (id, tempFreeResponse) ->
+  onFreeResponseChange: (id, tempFreeResponse, stepIndex) ->
     TaskStepActions.updateTempFreeResponse(id, tempFreeResponse)
 
     # set part to be active if part of multipart
-    @setCurrentStepByStepId(id) unless @isSinglePart(@state.parts)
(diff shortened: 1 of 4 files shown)


tutor/src/components/task-step/exercise.cjsx
```diff
@@ -123,27 +123,21 @@ module.exports = React.createClass
...
-    stepNavIndex = TaskStore.getStepNavIndex(taskId, id)
-    @setCurrentStep(stepNavIndex)
-
-  onFreeResponseChange: (id, tempFreeResponse) ->
+  onFreeResponseChange: (id, tempFreeResponse, stepIndex) ->
     TaskStepActions.updateTempFreeResponse(id, tempFreeResponse)
 
     # set part to be active if part of multipart
-    @setCurrentStepByStepId(id) unless @isSinglePart(@state.parts)
+    @setCurrentStep(stepIndex) unless @isSinglePart(@state.parts)
```
pandafulmanda: Sorry, I didn't make this very clear with comments, but this `stepIndex` getting passed in is the step index based on the actual existing steps.  However, since we've gotten more complicated with our interstitial cards stuffs, this step index does not match the index of the steps on the front end when there are interstitial introduction cards, hence the ugly `TaskStore.getStepNavIndex` stuffs, because that's what `setCurrentStep` wants -- the index that accounts for all crumbs/interstitial stuffs.  I believe these changes will cause weird behavior when the MPQ is after any interstitial cards.  Potentially the current bug is due to unsynced "stepIndices" as well, which the current steps cleanup stuffs in #1354 should help with.

nathanstitt: Hm.  You may be correct but I hope not.

The current bug is because what `Exercise` was getting from `TaskStore.getStepNavIndex` was incorrect when it was for readings.  That would cause it to set the wrong step as active every time a key was pressed and the `onFreeResponseChange` callback occurred, and then as part of the of setting the step active it would scroll to it, causing the page to jump around.

What I'm attempting to do is to have `Exercise tell each of the exercises what their index is, and then they pass that back to the `Exercise` callbacks.  

I think `stepIndex` is the thing to use even with the interstitial cards, since it's derived from the `step.index` from the task, so they don't know anything about the cards.

At any rate, let's hold off for #1354 and see if it breaks.

nathanstitt:
closing in favor of #1368

PR #1356, opened 2016-09-19
Display student identifier in student scores

nathanstitt:
https://trello.com/c/55GFyvdF/125-make-it-possible-to-distinguish-students-with-the-same-name-from-the-roster-page-and-from-the-scores-page

screen shot 2016-09-19 at 2 10 42 pm


kathi-fletcher:
I noticed that Progress and Score are reversed from the mockup (left, right) and also way spread out. @rvnewell, can you take a quick look?

nathanstitt:
@kathi-fletcher that's because the course I took the screenshot on only had two assignments so the columns expanded to fill the table.

If that's not the desired behavior we could change the table to not be full width.

nathanstitt:
Also not sure on the reversed score/progress - I just took a look at QA and it's the same (with score first).  To the best of my recollection they've always been in that order.

rvnewell:
screen shot 2016-09-21 at 2 48 38 pm
Score and Progress does appear to be in the correct order-- Score, left; Progress, right.  The asymmetry between the reading assignment (with no score) and the HW assignment may make it look that way.  But currently it is as designed.

rvnewell:
Checking with Fred about "expanding to fit frame" layout on 1-2 assignments.

rvnewell:
Totally nitpicky, but @nathanstitt, if they're not already this color, could you use #99999 for both the due date and student ID text color?

PR #1359, opened 2016-09-22
Update vendor dependencies

nathanstitt:
This is a WIP so I could test out react joyride.  

Only the student dashboard is covered, but most of the effort was to upgrade the router.  I think the rest should fall into place fairly quickly


pandafulmanda:
hello

**This is supposed to be bolded** and this is now

nathanstitt:
That's ok.   We can just call off doing the update.  It's not like we've worked hard on it or anything 😀

PR #1363, opened 2016-09-22
Update UiSetting's PREVIOUS_SETTINGS only after save

nathanstitt:
Otherwise if `set` is called in quick succession before the `_.debounce` has fired `PREVIOUS_SETTINGS` will be updated before save occurs and the BE will reject the save.

### DIFF
--- a/shared/test/model/ui-settings.spec.coffee
+++ b/shared/test/model/ui-settings.spec.coffee
@@ -6,7 +6,10 @@ Networking = require 'model/networking'
 describe 'UiSettings', ->
 
   beforeEach ->
-    sinon.stub(Networking, 'perform')
+    sinon.stub(Networking, 'perform').returns(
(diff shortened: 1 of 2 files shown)


shared/test/model/ui-settings.spec.coffee
```diff
@@ -28,4 +31,18 @@ describe 'UiSettings', ->
         one: 'five', two: 'II'
       )
       done()
-    , 2)
+    , 20)
+
```
pandafulmanda: c:sunglasses:l

shared/test/model/ui-settings.spec.coffee
```diff
@@ -28,4 +31,18 @@ describe 'UiSettings', ->
...
+  it 'groups saves together', (done) ->
```
pandafulmanda: c:sunglasses:l

PR #1368, opened 2016-09-23
should fix mpq scrolling

pandafulmanda:
step indices were mismatched and confused, and so which step to scroll to was confused as well.  this more carefully matches the visual step indices with real steps on tasks, especially around things like whether steps are part of same multipart

TODO clean up.

hopefully addresses
* https://trello.com/c/28Do07p1
* https://trello.com/c/DBTgj8Ba
* https://trello.com/c/247ytbaE


nathanstitt:
Hm, I like we're you're going with the stop index calculation, but it seems like it's got some issues.  I'm still testing locally and will make you a test case on dev

PR #1376, opened 2016-09-26
change for intro two step message

pandafulmanda:
As defined here: https://trello.com/c/7VIdD1Y3/166-update-the-wording-in-the-first-time-two-step-card-in-tutor-readings-and-homeworks and [here for coach](https://trello.com/c/tCSxdOJh/168-update-the-wording-in-the-first-time-two-step-card-in-concept-coach-readings)

## Tutor
screen shot 2016-09-29 at 1 49 37 am

## Coach
screen shot 2016-09-29 at 1 50 23 am

### DIFF
--- a/shared/src/helpers/step-helps.cjsx
+++ b/shared/src/helpers/step-helps.cjsx
@@ -1,5 +1,6 @@
 React = require 'react'
 _ = require 'underscore'
+capitalize = require 'lodash/capitalize'
 
 PERSONALIZED_GROUP = 'personalized'
 SPACED_PRACTICE_GROUP = 'spaced practice'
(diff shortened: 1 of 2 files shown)


shared/src/helpers/step-helps.cjsx
```diff
@@ -1,5 +1,6 @@
 React = require 'react'
 _ = require 'underscore'
+capitalize = require 'lodash/capitalize'
```
nathanstitt: I like requiring individual stuff this way.  I think it'd be a good thing to do when/if we drop underscore

pandafulmanda: also, i like that `lodash` has those nice convenient string functions

PR #1378, opened 2016-09-29
Fix duplicated choice letter displayed on exercise preview

nathanstitt:
from: 

screen shot 2016-09-29 at 4 05 01 pm

To:

screen shot 2016-09-29 at 3 59 01 pm


nathanstitt:
The extra commits are because this is based off the hot fix branch #1375 .  they should go away once it's merged

PR #1383, opened 2016-10-04
Scroll to current exercise when returning from details view

nathanstitt:
Applies to both QL and HW builder

The selected exercise is focused after it's returned to so it can be identified:
screen shot 2016-10-04 at 4 12 45 pm


pandafulmanda:
nice 👓  :)

PR #1386, opened 2016-10-06
Fix reading task details

pandafulmanda:
screen shot 2016-10-06 at 4 14 09 pm
screen shot 2016-10-06 at 4 13 38 pm


nathanstitt:
Looks pretty good, I think it might be a bit better to put the "viewing as ..." and back button on the navbar, but then it might not fit well on small screens.

PR #1387, opened 2016-10-20
fix missing section titles

pandafulmanda:
https://trello.com/c/qd29BT5E/140-bug-section-title-does-not-show-anywhere-in-the-reading-content

# In Readings

screen shot 2016-10-20 at 3 59 48 am
screen shot 2016-10-20 at 3 59 55 am

# In reference view

screen shot 2016-10-20 at 4 00 25 am

# In other books

screen shot 2016-10-20 at 2 19 43 pm

screen shot 2016-10-20 at 4 02 33 am

### DIFF
--- a/tutor/resources/styles/book-content/note.less
+++ b/tutor/resources/styles/book-content/note.less
@@ -106,16 +106,16 @@
   @except-teacher-rules: @tutor-empty-rules;
   @except-teacher-title-rules: @tutor-empty-rules
 ) {
-  > .note,
+  .book-content > .note,
   section > section > .note,
-  > section > .note:not(.learning-objectives),
+  .book-content > section > .note:not(.learning-objectives),
 
   .grasp-check,
 
-  > [data-element-type="check-understanding"],
+  .book-content > [data-element-type="check-understanding"],
   section > [data-element-type="check-understanding"],
 
-  > .example,
+  .book-content > .example,
   section > .example {
     .tutor-style-note(@all-note-rules; @title-rules);
 
@@ -162,7 +162,7 @@
   }
 
   // .example, check-understanding content currently lacks a data-label attribute
-  > .example,
+  .book-content > .example,
   section > .example {
     .tutor-book-before-manual-label("Worked Example");
   }
--- a/tutor/src/components/related-content.cjsx
+++ b/tutor/src/components/related-content.cjsx
@@ -1,0 +1,40 @@
+React    = require 'react'
+_ = require 'underscore'
+
+{ChapterSectionMixin} = require 'shared'
+
+{StepTitleStore} = require '../flux/step-title'
+
+RelatedContent = React.createClass
+
+  propTypes:
+    contentId: React.PropTypes.string.isRequired
+    title: React.PropTypes.string.isRequired
+    chapter_section: React.PropTypes.array.isRequired
+
+  mixins: [ChapterSectionMixin]
+
+  isIntro: ->
+    (_.isArray(@props.chapter_section) and
+      (@props.chapter_section.length is 1 or
+      @props.chapter_section[1] is 0)) or
+      (_.isString(@props.chapter_section) and 
+      @props.chapter_section.indexOf('.') is -1)
+
+  render: ->
+    {title, contentId, chapter_section} = @props
+
+    return null if _.isEmpty(title) or @isIntro()
+
+    section = @sectionFormat(chapter_section)
+
+    <h4
+      className='related-content'
+      data-has-learning-objectives={StepTitleStore.hasLearningObjectives(contentId)}>
+      <span className="part">
+        <span className="section">{section} </span>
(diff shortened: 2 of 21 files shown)


tutor/resources/styles/book-content/note.less
```diff
@@ -162,7 +162,7 @@
   }
 
   // .example, check-understanding content currently lacks a data-label attribute
-  > .example,
+  .book-content > .example,
   section > .example {
     .tutor-book-before-manual-label("Worked Example");
   }
```
nathanstitt: hm, would it make more sense to import all the book content styles inside a `.book-content` scope?  Or would that be too large a change do you think?

tutor/src/components/related-content.cjsx
```diff
@@ -0,0 +1,40 @@
...
+
+{StepTitleStore} = require '../flux/step-title'
+
+RelatedContent = React.createClass
+
+  propTypes:
+    contentId: React.PropTypes.string.isRequired
+    title: React.PropTypes.string.isRequired
+    chapter_section: React.PropTypes.array.isRequired
+
```
nathanstitt: hm, line 21 below indicates that `chapter_section` might be a string. In that case perhaps the validation should be:
```
  React.PropTypes.oneOfType([
    React.PropTypes.array React.PropTypes.string
  ]).isRequired
``

File: tutor/src/components/task-step/step-with-reading-content.cjsx [hunk 3]
```diff
@@ -27,16 +27,20 @@ ReadingStepContent = React.createClass
...
-        html={content_html} />
-      <RelatedContentLink courseId={courseId} content={related_content} />
+        {...courseDataProps}>
+        <RelatedContent contentId={id} {...related_content?[0]} />
+        <ArbitraryHtmlAndMath
+          className='book-content'
+          shouldExcludeFrame={@shouldExcludeFrame}
+          html={content_html} />
+      </div>
     </div>
```
nathanstitt: Throwing this out for debate:

I've been thinking that when a element has lengthy params, and flows onto multiple lines, and has children it's clearer to leave the trailing `>` indented the same as the start of the element like so:
```
<div
   className="foo"
   param1=1
   param2=2
>
  <a child />
</div>
```

That way you can tell where the open div tag starts and where the child is at.  For instance the above code is kinda hard to tell that `ReatedContent` is a child of the div since the `>` is nestled right after `{...courseDataProps}`.

Thoughts?

tutor/src/flux/step-title.coffee
```diff
@@ -7,9 +7,14 @@ TEXT_LENGTH_CHECK = 110
...
 
+isLearningObjectives = (element) ->
+  (element?.attribs?['class']? and
+    element.attribs['class'].search(/learning-objectives/) > -1) or
+    element.attribs['data-type'] is 'abstract'
+
 isTypedClass = (element) ->
   element?.attribs?['class']? and
-    element.attribs['class'].search(/learning-objectives|references|ap-connection/) > -1
+    element.attribs['class'].search(/learning-objectives|references|ap-connection|solution/) > -1
```
nathanstitt: can we use classList yet?  http://caniuse.com/#search=classlist seems to say it's ok, but I think maybe we had bugs with it once?

tutor/test/components/related-content-link.spec.coffee
```diff
@@ -2,7 +2,7 @@
 React = require 'react/addons'
 RelatedContentLink = require '../../src/components/related-content-link'
 
-describe 'Relalated Content Link Component', ->
+describe 'Related Content Link Component', ->
```
nathanstitt: haha, speellling is hard

nathanstitt:
Looks nice @pandafulmanda - something I didn't see in the code, but I'm assuming you thought of is:  Does the new `RelatedContent` change the color to match the learning objectives that it appears above?  It'd look kinda silly being blue on book content that has the orange objectives.

nathanstitt:
Yep, I see you did.  I deployed it to dev server to test.  Good jorb!

tutor/src/components/related-content.cjsx
pandafulmanda: oh good point

tutor/src/components/task-step/step-with-reading-content.cjsx
pandafulmanda: okay yah makes sense

tutor/src/flux/step-title.coffee
pandafulmanda: this is using a parser, not the actual DOM

tutor/resources/styles/book-content/note.less
pandafulmanda: i think too large for now...should be safe, but i think it'd be better to do it with the upgrades

PR #1390, opened 2016-10-25
Routes combine attempt

pandafulmanda:
Follow up to #1359


nathanstitt:
Looks good @pandafulmanda - specs are failing though, it looks like `./tutor/test/components/helpers/utilities.cjsx` is requiring  `../../../src/router` which doesn't exist.

PR #1391, opened 2016-10-27
Share API handler for flexible and consistent api handling

pandafulmanda:
customizable to each app

In preparation for making exercises have non-default error handling for api endpoints -- https://trello.com/c/35PJj6P6/195-reviewer-views-in-exercises-as-openstax-we-need-a-content-reviewer-role-for-exercises-that-has-a-complete-view-of-everything-abo -- we are making a common api handler
- [x] api handler
- [ ] tests & debug
- [ ] implement handler
  - [x] coach
  - [x] exercises
  - [x] tutor
  - [ ] shared pollers

screen shot 2016-11-07 at 4 36 34 pm
screen shot 2016-11-07 at 3 47 14 pm


shared/src/api/interceptors.coffee
```diff
@@ -0,0 +1,134 @@
...
+
+  requestConfig.url = _.compact(url_parts).join('?')
+
+makeLocalResponse = (response) ->
+  payload = if response.config.data then JSON.parse(response.config.data) else {}
+  response.data = _.extend({}, response.data, payload)
+
+doesErrorMatch = (handledErrors, errorName) ->
+  _.indexOf(handledErrors, errorName) > -1 or
+    _.some(handledErrors, _.partial(minimatch, errorName))
```
nathanstitt: seems like `isnt -1` would be better than `> -1` here.  Or `_.includes`

shared/src/api/interceptors.coffee
```diff
@@ -0,0 +1,134 @@
...
+
+areAllErrorsHandled = (handledErrors, errors, errorNameProperty) ->
+  isErrorHandled = _.partial(doesErrorMatch, handledErrors)
+  _.every errors, _.flow(_.property(errorNameProperty), isErrorHandled)
+
```
nathanstitt: Maybe GitHub isn't displaying this line correctly, but it looks like you're assigning `isErrorHandled` to y the function returned b`_.partial` and then calling `_.every` on the next line, so it doesn't have any effect.  I suspect the partial should be immediately evaluated?  i.e.:
```
 isErrorHandled = _.partial(doesErrorMatch, handledErrors)(
      _.every errors, _.flow(_.property(errorNameProperty), isErrorHandled)
)
```

tutor/src/api/index.coffee
```diff
@@ -0,0 +1,330 @@
...
+      topic: courseId
+      action: 'create'
+    else
+      subject: subject
+      topic: id
+      action: 'update'
+
+    {requestInfo, routeData, requestData}
+  )
+
```
nathanstitt: Some of these seem to get really verbose.  I think we could probably come up with helpers to slim them down a bit,  however I think we should merge this and ponder the issue a bit while we get used to the new loading api.

tutor/src/components/navbar/server-error-monitoring.cjsx
```diff
@@ -132,16 +131,15 @@ module.exports = React.createClass
...
-        handler = message.errors[0].code
-
-    attrs = ERROR_HANDLERS[handler](error, message, @context)
+    if _.isObject(error.data)
+      {data} = error
+      if data.errors?.length is 1 and ERROR_HANDLERS[data.errors[0].code]
+        handler = data.errors[0].code
 
+    attrs = ERROR_HANDLERS[handler](error, data, @context)
     Dialog.show( attrs.dialog ).then(attrs.onOk, attrs.onCancel)
```
nathanstitt: I just noticed that the error dialog wasn't working - glad you fixed it :)

coach/src/concept-coach/error-notification.cjsx
```diff
@@ -37,17 +37,20 @@ ErrorNotification = React.createClass
   componentWillUnmount: ->
     api.channel.off 'error', @onError
 
-  onError: ({response, failedData, exception}) ->
-    return if failedData?.stopErrorDisplay # someone else is handling displaying the error
+  onError: (exception) ->
+    {config, response} = exception
+
+    # return if failedData?.stopErrorDisplay # someone else is handling displaying the error
     if exception?
```
pandafulmanda: This type of thing is no longer needed with the new api handler! :)  this error only get triggered if the endpoint specific error handler hasn't handled it.  for coach, these are defined in the routes as `handledErrors`

shared/index.cjsx
```diff
@@ -9,6 +10,9 @@ module.exports = {
   OXRouter:               require './src/helpers/router'
   OXLink:                 require './src/factories/link'
   OXMatchByRouter:        require './src/helpers/match-by-router'
+  OXMatchByRouter:        require './src/helpers/match-by-router'
+  APIHandler,
+  APIActionAdapter:       require './src/api/action-adapter'
 
   ArbitraryHtmlAndMath:   require './src/components/html'
   Markdown:               require './src/components/markdown'
```
pandafulmanda: hmmm i agree it's getting ridiculous, but i want to ponder this decision a little more.  i dont exactly like the idea of having to look up the location of all the files either and the long paths in all the places we're using `shared` stuffs.  i kinda like that accessing the things we want from shared is very easy, but this is getting a little ridiculous.

shared/src/api/interceptors.coffee
pandafulmanda: oh thanks! i always have a hard time finding/remember `_.includes` cause i try to look for like `_.contain` or `_.has` or something and then i give up and do the `_.indexOf` thing.

coach/src/concept-coach/error-notification.cjsx
```diff
@@ -37,17 +37,20 @@ ErrorNotification = React.createClass
...
+    if response.status is 0 # either no response, or the response lacked CORS headers and the browser rejected it
       errors = ["Unknown response received from server"]
     else
       errors = ["#{response.status}: #{response.statusText}"]
-      if _.isArray(failedData.data?.errors) # we have something from server to display
+      if _.isArray(response.data?.errors) # we have something from server to display
         errors = errors.concat(
-          _.flatten _.map failedData.data.errors, (error) ->
+          _.flatten _.map response.data.errors, (error) ->
             # All 422 errors from BE *should* have a "code" property.  If not, show whatever it is
```
pandafulmanda: okay, will check :)

shared/src/api/interceptors.coffee
pandafulmanda: ah, this is meant to be a non-nested version of this

```
areAllErrorsHandled = (handledErrors, errors, errorNameProperty) ->
  _.every errors, (error) ->
    doesErrorMatch(handledErrors, _.property(errorNameProperty)(error))
```

but perhaps the nested version is easier to understand in this case.  actually, i think i would prefer this:

```
areAllErrorsHandled = (handledErrors, errors, errorNameProperty) ->
  _.every errors, (error) ->
    errorName = _.property(errorNameProperty)(error)
    doesErrorMatch(handledErrors, errorName)
```

shared/index.cjsx
nathanstitt: Yeah, I shoulda said I wasn't  actually suggesting we do it on this PR ;)  Just something we could think about for laters.

PR #1393, opened 2016-10-31
Treat true-false format questions as multi-choice

nathanstitt:
Based of production sha for potential hot-fix release.

Without this, a "panel" type isn't returned and the exercise is treated as "review" way down in the shared/exercise code.

Simular logic is used in Tutor


nathanstitt:
I merged master in, this can now be merged.

PR #1399, opened 2016-11-07
Allow clicks to pass through navbar-overlay

nathanstitt:
Since it's 100% width it was intercepting clicks for the logo and course title.

### DIFF
--- a/tutor/src/components/navbar/index.cjsx
+++ b/tutor/src/components/navbar/index.cjsx
@@ -58,15 +58,13 @@ module.exports = React.createClass
     params = Router.currentParams()
     {courseId} = params
 
-    brand = <TutorLink to='listing' className='navbar-brand'>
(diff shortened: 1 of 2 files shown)


tutor/src/components/navbar/index.cjsx
```diff
@@ -58,15 +58,13 @@ module.exports = React.createClass
...
       <BS.Navbar.Brand>
-        {brand}
+        <TutorLink to='listing' className='navbar-brand'>
+          <i className='ui-brand-logo'></i>
+        </TutorLink>
       </BS.Navbar.Brand>
+      <CenterControls params={params} />
       <BS.Navbar.Collapse>
         <BS.Nav>
           <CourseName course={course}/>
```
nathanstitt: These are just cleanups while i was attempting to figure this out, they aren't really part of the fix.

PR #1401, opened 2016-11-07
fix CC assignment links screen

nathanstitt:
Another link that we missed :(

### DIFF
--- a/tutor/src/components/assignment-links.cjsx
+++ b/tutor/src/components/assignment-links.cjsx
@@ -13,9 +13,10 @@ AssignmentLinks = React.createClass
   bindStore: ReferenceBookStore
 
   componentWillMount: ->
-    {courseId} = Router.currentParams()
+    {courseId} = _.defaults(@props, Router.currentParams() )
+
     course = CourseStore.get(courseId)
-    @ecosystem_id = course.ecosystem_id
+    @setState(ecosystem_id: course.ecosystem_id)
     ReferenceBookActions.load(course.ecosystem_id)
 
   renderRow: (item, baseUrl, bookId) ->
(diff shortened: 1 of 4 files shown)


tutor/src/routes.coffee
```diff
@@ -1,8 +1,7 @@
 async = require './helpers/webpack-async-loader'
 
 getCourseListing = ->
-  {CourseListing} = require './components/course-listing'
-  CourseListing
+  require './components/course-listing'
```
pandafulmanda: is [this](https://github.com/openstax/tutor-js/blob/master/tutor/src/components/course-listing.cjsx#L57) meant to be pulled out of an object then?  and also, since we have a [course-listing directory](https://github.com/openstax/tutor-js/blob/master/tutor/src/components/course-listing), should we move [course-listing.cjsx](https://github.com/openstax/tutor-js/blob/master/tutor/src/components/course-listing) into that folder as `index.cjsx`?

tutor/src/components/assignment-links.cjsx
```diff
@@ -13,9 +13,10 @@ AssignmentLinks = React.createClass
   bindStore: ReferenceBookStore
 
   componentWillMount: ->
-    {courseId} = Router.currentParams()
+    {courseId} = _.defaults(@props, Router.currentParams() )
```
pandafulmanda: For some reason, on my local, `@props` seems immutable, and the result of `_.defaults(@props, Router.currentParams() )` does not pull in the `courseId` from `Router.currentParams()` when i go directly to http://localhost:3001/course/1/assignment-links.

```
{courseId} = _.defaults(Router.currentParams(), @props)
```
appears to work.  Also, `Router.currentParams()` returns a deep clone each time, so using `default` on that should have no unexpected side-effects.

tutor/src/routes.coffee
pandafulmanda: hmmm i see now, this is kinda from #1400

nathanstitt: and yah, this doesn't belong in here.  it confused me also :(  I'm going to rebase it out and make the change on #1400

nathanstitt: Sigh actually I guess I can't rebase out since it's on the same commit.

PR #1405, opened 2016-11-09
adapter was doing weird things with options

pandafulmanda:


shared/src/api/action-adapter.coffee
```diff
@@ -42,15 +42,17 @@ makeRequestHandlers = (Actions, options) ->
...
-resolveOption = (option, key, args...) ->
-  if _.isFunction(option) and key isnt 'handleError'
-    option(args...)
-  else
-    option
+resolveOptions = (args...) ->
+  (option, key) ->
+    if _.isFunction(option) and key isnt 'handleError'
+      console.info(args...)
+      option(args...)
```
nathanstitt: hehe, `console` right back at ya

PR #1415, opened 2016-11-15
Update course cloner

pandafulmanda:
screen shot 2016-11-17 at 4 39 25 pm
screen shot 2016-11-17 at 4 39 30 pm
screen shot 2016-11-17 at 4 39 35 pm
screen shot 2016-11-17 at 4 39 41 pm
screen shot 2016-11-17 at 4 39 51 pm
screen shot 2016-11-17 at 4 39 56 pm
screen shot 2016-11-17 at 4 40 02 pm
screen shot 2016-11-17 at 4 40 06 pm


tutor/specs/components/new-course/copy-ql.spec.cjsx
```diff
@@ -8,14 +8,14 @@ describe 'CreateCourse: choosing to copy QL', ->
...
 
   it 'sets flux values', ->
     wrapper = shallow(<CopyQL />)
-    wrapper.find('tr.true').simulate('click')
+    wrapper.find('[data-copy-or-not="copy"]').simulate('click')
     expect(NewCourseStore.get('copy_question_library')).to.equal(true)
-    wrapper.find('tr.false').simulate('click')
+    wrapper.find('[data-copy-or-not="dont-copy"]').simulate('click')
     expect(NewCourseStore.get('copy_question_library')).to.equal(false)
     undefined
```
nathanstitt: hehe, so what, "true" & "false" can't be class names? 😁

tutor/resources/styles/components/new-course.less
```diff
@@ -1,140 +1,337 @@
...
+      }
+    }
 
-    display: flex;
-    border: 1px solid @tutor-neutral-light;
+    &[data-brand=tutor-beta] {
+      .course-choice-content-wrapper::before {
+        content: url('../../images/openstax-logo-tutor-beta-horizontal.svg');
+      }
+    }
```
nathanstitt: Just curious but why the "-beta" ?  Seems like something that shouldn't be embedded in the code since now we should go through the codebase and remove it once it's "production".  It's not a big deal though since there's also user-visible labels that'll need updating as well.

tutor/src/components/new-course/select-dates.cjsx
```diff
@@ -4,9 +4,10 @@ BS = require 'react-bootstrap'
...
 partial = require 'lodash/partial'
 isEqual = require 'lodash/isEqual'
-S = require '../../helpers/string'
+
 {NewCourseActions, NewCourseStore} = require '../../flux/new-course'
 {OfferingsStore} = require '../../flux/offerings'
+{CourseChoiceItem} = require './choice'
 
 QUARTERS =
   '2017Q4': 'Fall 2017'
```
nathanstitt: Hm, I just noticed that I left the old stub data`QUARTERS` in the file.  If we do further updates that can be removed.

PR #1418, opened 2016-11-18
Course cloner details

pandafulmanda:
screen shot 2016-11-18 at 12 11 56 am

Loading and building animation for course builder widget

### DIFF
--- a/tutor/specs/components/new-course/build-course.spec.cjsx
+++ b/tutor/specs/components/new-course/build-course.spec.cjsx
@@ -13,13 +13,13 @@ describe 'CreateCourse: saving new course', ->
     NewCourseActions.save.restore()
 
   it 'it transitions to ready after save', ->
-    wrapper = mount(<BuildCourse />)
+    wrapper = shallow(<BuildCourse />)
     expect(NewCourseActions.save.calledOnce).to.be.true
     expect(wrapper.text()).to.include('building')
-    expect(wrapper.find('.ox-loader')).to.have.length(1)
+    expect(wrapper.find('OXFancyLoader[isLoading=true]')).to.have.length(1)
     NewCourseActions.created({id: '1'})
     wrapper.setState({})
-    expect(wrapper.find('.ox-loader')).to.have.length(0)
+    expect(wrapper.find('OXFancyLoader')).to.have.length(0)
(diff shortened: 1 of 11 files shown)


tutor/specs/components/new-course/build-course.spec.cjsx
```diff
@@ -12,13 +13,13 @@ describe 'CreateCourse: saving new course', ->
...
     expect(wrapper.text()).to.include('building')
-    expect(wrapper.find('Icon[type="refresh"]')).to.have.length(1)
-    NewCourseActions.created({id: 1})
+    expect(wrapper.find('.ox-loader')).to.have.length(1)
+    NewCourseActions.created({id: '1'})
     wrapper.setState({})
-    expect(wrapper.find('Icon[type="refresh"]')).to.have.length(0)
+    expect(wrapper.find('.ox-loader')).to.have.length(0)
     expect(wrapper.text()).to.include('course is ready')
     undefined
```
nathanstitt: No need to require enzyme, it's `mount` and `shallow` is exported globally in config/test/jest.setup.coffee

tutor/specs/components/new-course/build-course.spec.cjsx
```diff
@@ -12,13 +13,13 @@ describe 'CreateCourse: saving new course', ->
...
     expect(NewCourseActions.save.calledOnce).to.be.true
     expect(wrapper.text()).to.include('building')
-    expect(wrapper.find('Icon[type="refresh"]')).to.have.length(1)
-    NewCourseActions.created({id: 1})
+    expect(wrapper.find('.ox-loader')).to.have.length(1)
+    NewCourseActions.created({id: '1'})
     wrapper.setState({})
-    expect(wrapper.find('Icon[type="refresh"]')).to.have.length(0)
+    expect(wrapper.find('.ox-loader')).to.have.length(0)
     expect(wrapper.text()).to.include('course is ready')
```
nathanstitt: FYI, it's much faster to use `shallow` vs `mount`, the only problem with it is that you can't actually test the dom for nested components since they don't render.  You can test the component itself using enzyme's finders though.  You could have tested something like:

`expect(wrapper.find('OXFancyLoader[isLoading=true]')).to.have.length(1)`

PR #1420, opened 2016-11-18
Upgrade dependencies

nathanstitt:
Updates all dependencies on project.

I removed the inactive react-scroll-components (no updates in 2 years) by taking the file we really need from it and updating to work with latest React.

Currently the only failure seems to be due to react-maskedinput using some removed React internals. That's been corrected by the project but is awaiting a release:  https://github.com/insin/react-maskedinput/issues/76


tutor/index.coffee
```diff
@@ -31,6 +31,9 @@ window._STORES =
   NOTIFICATIONS: require './src/flux/notifications'
   TOC: require './src/flux/toc'
 
+# In dev builds this enables hot-reloading,
+# in production it simply renders the root app
+{AppContainer} = require 'react-hot-loader'
```
pandafulmanda: hmm...interesting choice from `react-hot-loader to require `react` code for what used to be a setting...should we add this to `exercises` and `coach` and/or make a `shared` helper function that will handle this?

pandafulmanda: hmm...also... [comment about not needing `AppContainer` anymore maybe?](https://github.com/gaearon/redux-devtools/commit/64f58b7010a1b2a71ad16716eb37ac1031f93915#commitcomment-19376847)

PR #1423, opened 2016-11-19
Course cloner followup

pandafulmanda:
screen shot 2016-11-20 at 2 59 41 am


shared/resources/styles/variables/openstax-bootstrap.less
```diff
@@ -191,13 +190,13 @@
...
-// @input-border-radius-large:      @border-radius-large;
-// //** Small `.form-control` border radius
-// @input-border-radius-small:      @border-radius-small;
+//** Default `.form-control` border radius
+// This has no effect on `<select>`s in some browsers, due to the limited stylability of `<select>`s in CSS.
+@input-border-radius:            0;
+//** Large `.form-control` border radius
+@input-border-radius-large:      0;
+//** Small `.form-control` border radius
+@input-border-radius-small:      0;
```
nathanstitt: I'm guessing UX is ok with removing the border-radius on all buttons?   If so we should we go through styles and find all the spots where we've already set it to zero and update them?

tutor/resources/styles/global/buttons.less
```diff
@@ -1,28 +1,28 @@
...
-  border-radius: @btn-border-radius;
+  // border-radius: @btn-border-radius;
   transition: box-shadow 0.2s, background 0.2s;
 
   &:hover,
   &:focus {
-    .btn-hover();
+    .tutor-shadow(button);
   }
 }
```
nathanstitt: we should cleanup this file at some point 😁

tutor/resources/styles/tutor.less
```diff
@@ -4,6 +4,7 @@
 @import (inline) '~recordo/recordo.css';
 
 // Variables and Mixins
+// TODO to many renamed variables running around.
 @import './variables';
```
nathanstitt: 👍 agreed

tutor/src/components/course-calendar/add-assignment-sidebar.cjsx
```diff
@@ -1,43 +1,90 @@
...
 PastAssignments = require './past-assignments'
 
 CourseAddMenuMixin = require './add-menu-mixin'
+BindStoreMixin = require '../bind-store-mixin'
+
 {CourseStore} = require '../../flux/course'
+{PastTaskPlansStore} = require '../../flux/past-task-plans'
+
+IS_INTRO_VIEWED = 'viewed-plan-dnd-intro'
+USE_SETTINGS = false
```
nathanstitt: is USE_SETTINGS false so that it'll always be shown for qa or something?

tutor/resources/styles/global/externals.less
```diff
@@ -1,4 +1,5 @@
 @import (less) '~react-datepicker/dist/react-datepicker.css';
 @import (less) '~fixed-data-table-2/dist/fixed-data-table.css';
+@import (css) '~react-joyride/lib/styles/react-joyride-compiled.css';
```
nathanstitt: This is causing errors since joyride isn't in package.json.  Was it added here by mistake?  I don't see it being used anywhere.

tutor/src/components/course-calendar/add-assignment-sidebar.cjsx
```diff
@@ -1,43 +1,90 @@
...
     courseId: React.PropTypes.string.isRequired
     hasPeriods: React.PropTypes.bool.isRequired
-
+    isOpen: React.PropTypes.bool.isRequired
 
   getInitialState: ->
-    {isOpen: false}
+    showIntro: @props.isOpen
+    needsIntro: false
+
```
nathanstitt: The intro never seems to show for me, because the calendar loads with `isOpen` set to false causing `showIntro` to be set to false as well here.  Then when the `shouldShowIntro` method is called below the last clause that checks `showIntro` is always false and fails.

nathanstitt:
@pandafulmanda I'm going to merge this so we can get the styling fixes in for @gregfitch on Monday morning.  Not sure what we should do about https://github.com/openstax/tutor-js/pull/1423/files#r88809405 though, maybe that's by design and i'm mis-understanding it's purpose.

PR #1424, opened 2016-11-20
set needsIntro to true so the sidebar will display it

nathanstitt:


nathanstitt:
closing since the intent was to open the sidebar when `?showIntro=true` is appended to URL.

PR #1436, opened 2016-11-22
try to separate the sections of the teacher menu a little

pandafulmanda:
## Before
screen shot 2016-11-22 at 11 53 14 am

## After
screen shot 2016-11-22 at 11 57 51 am


Fredasaurus:
@pandafulmanda Should "another" be capitalized?

nathanstitt:
Yes, it should be. I think I got that wrong initially though.  @pandafulmanda can you update?

nathanstitt:
Also the spec failures are legit, there's a few specs that deal with testing menu options 😁

PR #1444, opened 2016-11-27
Fix bug with displaying dialog more than once

nathanstitt:
Believe this is the root cause of the various "task builder can't exit" bugs.  It was trying to display the "unsaved data" warning but was failing.


nathanstitt:
yeah, we could/should switch out TutorDialog since we only used it to work around React context bugs with 0.13.x

PR #1446, opened 2016-11-28
Task plan builder scroll to top on mount

nathanstitt:
Scroll task plan builder to top when it mounts

### DIFF
--- a/tutor/src/components/task-plan/index.cjsx
+++ b/tutor/src/components/task-plan/index.cjsx
@@ -8,6 +8,8 @@ LoadableItem = require '../loadable-item'
 {TaskPlanStore, TaskPlanActions} = require '../../flux/task-plan'
 Router = require '../../helpers/router'
 
+{ScrollToMixin} = require 'shared'
+
 PLAN_TYPES =
   reading: ReadingPlan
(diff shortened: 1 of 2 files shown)


tutor/src/components/task-plan/index.cjsx
```diff
@@ -56,9 +58,13 @@ EventShell = React.createClass
 PlanShell = React.createClass
   displayName: 'PlanShell'
```
pandafulmanda: what do you think about this type of thing going into [match-by-router](https://github.com/openstax/tutor-js/blob/master/shared/src/helpers/match-by-router.cjsx) or something?  it seems like scrolling to top should be default behavior, and not scrolling should be optionally set.  or maybe that's just me missing the old router interface.

nathanstitt: hm, yeah that'd be good. I think the old Router used to do that, right?  We'll have to make it so that it can be overridden though, since some routes won't want that, like the metrics review used to

pandafulmanda: will merge for now to take more time for a more global fix, which will require much more testing

PR #1447, opened 2016-11-28
Only bind to 'step.completed' once

nathanstitt:
While debugging a different bug I noticed we were binding step.completed
to different functions.  This binds once and then calls checkPlaceholder
from update.

I think this is a good idea so that we'll have fewer event listeners bound and lessen the maxListeners warning.

### DIFF
--- a/tutor/src/components/breadcrumb/index.cjsx
+++ b/tutor/src/components/breadcrumb/index.cjsx
@@ -78,8 +78,9 @@ BreadcrumbTaskDynamic = React.createClass
     stepId = crumb.id
     taskId = crumb.task.id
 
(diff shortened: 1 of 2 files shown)


pandafulmanda:
what do you think about something like [this](https://github.com/openstax/tutor-js/compare/fewer-listeners...ensure-loading-personalized-only-once)? although probs not needed -- the crumb changing appearance from personalized to regular breadcrumbs is no longer what the design is ever since we added the intro panel logics and the reading task milestones

nathanstitt:
I think that's probably good - seems like we should check if loading before we load.  Feel free to push to this, or make it a PR i guess?

pandafulmanda:
okay, #1452

PR #1455, opened 2016-11-29
Update course banner title

pandafulmanda:
https://trello.com/c/EVoQDgxF/268-t3-09-22-allow-tutor-instructors-to-start-setting-up-assignments-before-the-course-opens

- [x] update course settings with dates and term
- [x] update course banners with term
- [x] open dashboard to course start date if start date is in the future by default
- [x] set min/max for due date for assignment creation
- [x] only allow drop add on in term dates
- [x] update click add as well
- [x] check if existing courses have `starts_at` and `ends_at` from BE
- [x] fix and add tests

screen shot 2016-11-29 at 3 14 55 pm

screen shot 2016-11-29 at 12 40 38 pm
screen shot 2016-11-29 at 12 40 05 pm

### DIFF
--- a/tutor/specs/components/task-plan/mini-editor/editor.spec.cjsx
+++ b/tutor/specs/components/task-plan/mini-editor/editor.spec.cjsx
@@ -1,8 +1,11 @@
 {React, _} = require '../../helpers/component-testing'
 
+moment = require 'moment'
+
 MiniEditor = require '../../../../src/components/task-plan/mini-editor/editor'
 {CourseActions} = require '../../../../src/flux/course'
 {TaskPlanActions, TaskPlanStore} = require '../../../../src/flux/task-plan'
+{TimeStore} = require '../../../../src/flux/time'
 
 COURSE  = require '../../../../api/courses/1.json'
 COURSE_ID = '1'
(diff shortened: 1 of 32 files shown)


tutor/specs/components/task-plan/mini-editor/editor.spec.cjsx
```diff
@@ -92,3 +109,23 @@ describe 'TaskPlan MiniEditor wrapper', ->
...
+    expect(wrapper.find('ServerErrorMessage')).length.to.be(1)
+    undefined
+
+  it 'limits opens date and due date to term dates', ->
+    wrapper = mount(<MiniEditor {...@props} />)
+    datePickers = wrapper.find("DatePicker")
+
+    expect(datePickers.at(0).props().minDate.isSame(@props.termStart, 'day')).to.equal(true)
+    expect(datePickers.at(1).props().maxDate.isSame(@props.termEnd, 'day')).to.equal(true)
+    undefined
```
nathanstitt: cool 👁 🍷

tutor/src/components/course-calendar/add.cjsx
```diff
@@ -38,6 +41,19 @@ CourseAdd = React.createClass
...
+
+    if addDate?.isBefore(termStart, 'day')
+      'day before term starts'
+    else if addDate?.isAfter(termEnd, 'day')
+      'day after term ends'
+    else if addDate?.isBefore(referenceDate, 'day')
+      'past day'
+    else if addDate?.isSame(referenceDate, 'day')
+      'today'
+
```
nathanstitt: if addDate is null because state hasn't set it, then there's no message returned.  Is that OK?

If so, we could do an early `return '' unless addDate?` and then wouldn't need the `?` on each line.   Could use a `switch` statement as well, but that's up to you, I'm kinda ambivalent about wether they're better than `if`

PR #1465, opened 2016-12-01
WIP Unifying CC and Tutor enrollment views and adding a CC conflict view

Dantemss:
https://trello.com/c/zmvY38qo

Requires https://github.com/openstax/tutor-server/pull/1319

- Fixed the ui-settings route which I slightly changed in the BE while reorganizing and grouping routes
- Moved CC enrollment HTML to shared folder
- Added Tutor enrollment view
- Handle errors
- Redirect to dashboard at the end
- New view for when there is a CC registration conflict
- Fixed styles
- Specs

# TODO

- Handle CC courses that have ended (case 3 in Trello card)
- Test this on CC

## New Conflict view (FE):
screen shot 2016-12-07 at 5 08 48 pm

## New Confirm view (FE):
screen shot 2016-12-07 at 5 08 51 pm

## Old (BE view):
screen shot 2016-12-02 at 6 04 38 pm


shared/src/components/enroll/join.cjsx
```diff
@@ -0,0 +1,76 @@
...
+ENTER = 'Enter'
+
+AsyncButton = require '../buttons/async-button'
+
+Join = React.createClass
+
+  propTypes:
+    course: React.PropTypes.object.isRequired
+    errorList: React.PropTypes.instanceOf(React.Component).isRequired
+    optionalStudentId: React.PropTypes.bool
```
nathanstitt: You could spec the errorList as: `React.PropTypes.element.isRequired`

coach/src/course/confirm-join.cjsx
```diff
@@ -1,76 +1,21 @@
...
-        <div className="help-text">
-          <sup>*</sup> You can enter you student ID later by clicking on your name in the top right
-          of your dashboard and selecting "Change Student ID" from the menu.
-        </div>
-      </div>
-    </BS.Row>
+    <Join
+      course={@props.course}
+      errorList={React.createElement(ErrorList, course: @props.course)}
+      optionalStudentId={@props.optionalStudentId}
```
nathanstitt: You can use JSX syntax here:

```
<Join course={@props.course}
  errorList={<ErrorList course={@props.course} />}
  optionalStudentId={@props.optionalStudentId}
/>
```

shared/index.cjsx
```diff
@@ -41,6 +41,7 @@ module.exports = {
 
   FreeResponse:           require './src/components/exercise/free-response'
   GetPositionMixin:       require './src/components/get-position-mixin'
+  Join:                   require './src/components/enroll/join'
   ScrollToMixin:          require './src/components/scroll-to-mixin'
```
nathanstitt: Exporting the component as just `Join` seems too generic to me.  Maybe it could be called `NewEnrollment` or `JoinCourse`?  There's probably something better than either of those, but you get the idea.  The thing that it's exported as here doesn't actually have to be the same as the component's `displayName`

shared/src/components/enroll/join.cjsx
Dantemss: 👍

shared/index.cjsx
Dantemss: How about `ConfirmJoinHtml`?

Dantemss: I guess Html is maybe redundant so... `AbstractConfirmJoin`?

nathanstitt: Hm adding `Abstract` doesn't seem to add anything to me that would help figure out what the thing does..  What I was thinking was that `Join` doesn't really say what is being joined.  When I think `join` I think SQL table, but it could be joining course, combining grades, etc.  `UserJoinCourse` might work, or just `JoinCourse`.  

/meh it's not really that big a deal and as long as you pick something we can easily grep for to rename it's probably good 😁

tutor/src/flux/course-enrollment.coffee
```diff
@@ -0,0 +1,102 @@
...
+
+    # complete and ready for use
+    isRegistered: -> @status is "approved"
+
+    getStudentIdentifier: -> ""
+
+    courseId: -> @to?.course?.id
+
+    confirm: (studentId) -> CourseEnrollmentActions.approve(@id, studentId)
+
```
Dantemss: I see... my issue here is that the shared code does not use flux (because the CC code doesn't) and so it doesn't distinguish between stores and actions...

shared/src/components/enroll/confirm-join-course.cjsx
```diff
@@ -0,0 +1,81 @@
...
+AsyncButton = require '../buttons/async-button'
+CcConflictMessage = require './cc-conflict-message'
+MessageList = require './message-list'
+
+ConfirmJoinCourse = React.createClass
+
+  propTypes:
+    courseEnrollmentActions: React.PropTypes.object.isRequired
+    courseEnrollmentStore: React.PropTypes.object.isRequired
+    optionalStudentId: React.PropTypes.bool
```
nathanstitt: Something you can do here is use `React.PropTypes.shape` to ensure that the object that's passed in has the properties you're using, such as:

```
courseEnrollmentStore: React.PropTypes.shape(
   getEnrollmentChangeId: React.PropTypes.func.isRequred
  hasConflict: React.PropTypes.func.isRequred
  description: React.PropTypes.func.isRequred
  teacherNames: React.PropTypes.func.isRequred
).isRequired,
courseEnrollmentActions: React.PropTypes.shape(
  confirm: React.PropTypes.func.isRequred
).isRequired
```

That way React will perform validation on the object.  Using `propTypes` in this way is nice since it's a dev-only check that's removed when compiled for production

shared/src/components/enroll/message-list.cjsx
```diff
@@ -0,0 +1,34 @@
...
+    alertType: "danger"
+    messagesType: "errors"
+    forceBullets: false
+
+  getUlClassName: ->
+    if @props.messages.length == 1 && !@props.forceBullets
+      "#{@props.messagesType} list-unstyled"
+    else
+      @props.messagesType
+
```
nathanstitt: This could be replaced by the [classnames utility](https://github.com/JedWatson/classnames) (after adding `classnames = require 'classnames'`) and moved to render as:
```coffee
classnames = classnames('message-list', @props.messagesType, 
  {'list-unstyled': @props.messages.length == 1 && !@props.forceBullets}
)
```

Note that the above also adds a "message-list" class, which is a nice thing to do for components so they can be targeted for custom styling.

tutor/src/components/enroll/index.cjsx
```diff
@@ -0,0 +1,90 @@
...
+    router: React.PropTypes.object
+
+  componentWillMount: ->
+    {enrollmentCode} = Router.currentParams()
+    CourseEnrollmentStore.on('changed', @onCourseEnrollmentChange)
+    CourseEnrollmentActions.create(enrollmentCode)
+
+  componentWillUnmount: ->
+    CourseEnrollmentStore.off('changed', @onCourseEnrollmentChange)
+
```
nathanstitt: Rather than doing the manual `on`/`off` store dance here, you can use the `bind-store-mixin`, as https://github.com/openstax/tutor-js/blob/f7af0fd8cd0eb17916cb8152aacd28d699c2df20/tutor/src/components/exercises/preview.cjsx#L19-L23 (or search for BindStoreMixin for others).

tutor/src/components/enroll/index.cjsx
```diff
@@ -0,0 +1,90 @@
...
+  # Wait for the course to load and wait at least 1.5 secs
+  # so our success message is briefly displayed, then redirect to dashboard
+  onComplete: ->
+    loadCourse = new Promise((resolve, reject) ->
+      CourseStore.once('course.loaded', resolve)
+      CourseActions.load(CourseEnrollmentStore.getCourseId())
+    )
+    successTimer = new Promise((resolve, reject) -> _.delay(resolve, 1500))
+    Promise.all([loadCourse, successTimer]).then(@redirectToDashboard)
+
```
nathanstitt: Nice use of `Promise.all` - we should use that more, it's really powerful.

tutor/src/components/enroll/index.cjsx
```diff
@@ -0,0 +1,90 @@
...
+      <h3>Loading...</h3>
+    else if CourseEnrollmentStore.isConflicting()
+      <JoinConflict />
+    else if CourseEnrollmentStore.isPending() or CourseEnrollmentStore.isApproveError()
+      <ConfirmJoin />
+    else if CourseEnrollmentStore.isCreateError()
+      <MessageList messages={CourseEnrollmentStore.errorMessages()} />
+    else
+      @renderComplete()
+
```
nathanstitt: Since `renderComplete` isn't being used anywhere else and is pretty short, it doesn't need to be a function.  You could put the `<h3> block right in the `else` clause

tutor/src/components/enroll/index.cjsx
```diff
@@ -0,0 +1,90 @@
...
+        separate your Tutor responses from those of your students.
+      </p>
+    </div>
+
+  render: ->
+    <div className="tutor-registration">
+      <div className="row">
+        <div className="new-registration">
+          {@teacherMessage() if @isTeacher()}
+          {@renderCurrentStep()}
```
nathanstitt: Since the teacher message doesn't need any references to props, you could extract it to a standalone function component in the same file above this one's `createClass`, and also have it use a `visible` prop.  Something like:
```coffee
TeacherMessage = (props) ->
   return null unless props.visible
   <div className="teacher-message">
          ...
    </div>
```

Then in the render function here you could use:

`<TeacherMessage visible={@isTeacher()} />`

tutor/src/flux/course-enrollment-handled-errors.coffee
```diff
@@ -0,0 +1,13 @@
...
+                    'Please verify the enrollment code and try again.'
+  multiple_roles: 'We currently do not support teacher accounts with multiple associated student enrollments.'
+  dropped_student: 'You have been dropped from this course. Please speak to your instructor to rejoin.'
+  already_processed: 'Your enrollment in this course has been processed. Please reload the page.'
+  already_approved: 'Your enrollment in this course has been approved. Please reload the page.'
+  already_rejected: 'Your request to enroll in this course has been rejected for an unknown reason. Please contact OpenStax support.'
+  taken: 'The provided student ID has already been used in this course. Please try again or contact your instructor.'
+  blank_student_identifier: 'The student ID field cannot be left blank. Please enter your student ID.'
+
+module.exports = ERROR_MAP
```
nathanstitt: Not saying you should do that here, but if the strings got much longer you could use coffee script heredoc.

```coffee
already_enrolled: '''
  You are already enrolled in this course. 
  Please verify the enrollment code and try again.
'''
```

Just wanted to make sure you knew about 'em.

nathanstitt:
Looks really nice @Dantemss - you've really managed to investigate how React stuff works and make it happen here.  Feel free to join FE anytime 😀

shared/src/components/enroll/confirm-join-course.cjsx
Dantemss: I see... makes sense

tutor/src/components/enroll/index.cjsx
Dantemss: Thanks yeah took me a while to figure out how to load the course while the timer was going, but seems to work great.

tutor/src/flux/course-enrollment-handled-errors.coffee
Dantemss: Oh I might do this in some other place actually...

PR #1472, opened 2016-12-06
Ask students for their id via notices bar

nathanstitt:
Makes the "notices" bar be course / role aware and adds a "MissingStudentId" component for it to display when needed.  Updates it's appearance to slide down under the course banner.

Needs BE PR https://github.com/openstax/tutor-server/pull/1322 so it can decide how soon to display the notice.

A "normal" notice:

screen shot 2016-12-06 at 1 34 40 pm

The new student id notice:

screen shot 2016-12-06 at 1 35 53 pm

screen shot 2016-12-06 at 5 09 23 pm
screen shot 2016-12-06 at 5 09 15 pm


nathanstitt:
Thanks @pandafulmanda !   Yeah, after I did it I kinda regretted changing the return value of `getCourseRole`.  It might have been better to add an alternative function that returned the complete role. By then I thought I'd already fixed up all it's uses. :)

`conditional-rendering` is an artifact from the old old router stuff. I've removed the file

shared/resources/styles/components/notifications-bar.less
```diff
@@ -69,5 +85,11 @@
         text-align: center;
       }
     }
+
+
+    &.missing-student-id {
```
nathanstitt: nope 😀

PR #1478, opened 2016-12-07
course settings styling got a little bit mixed up

pandafulmanda:
https://trello.com/c/4plbMd7C/306-bug-start-and-end-dates-merge-together

## Before

screen shot 2016-12-07 at 12 38 43 pm

## After

screen shot 2016-12-07 at 12 38 19 pm

### DIFF
--- a/tutor/resources/styles/components/course-settings.less
+++ b/tutor/resources/styles/components/course-settings.less
@@ -26,7 +26,9 @@
   .settings-section {
     margin-top: 50px;
   }
+}
 
+.course-settings {
   .course-settings-title {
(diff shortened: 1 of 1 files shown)


tutor/resources/styles/components/course-settings.less
```diff
@@ -26,7 +26,9 @@
   .settings-section {
     margin-top: 50px;
   }
+}
 
+.course-settings {
   .course-settings-title {
```
nathanstitt: yeah that was my bad :(  Glad you noticed

PR #1479, opened 2016-12-07
redirect only when the created event is emitted to make sure we dont …

pandafulmanda:
…redirect to last created

https://trello.com/c/y0maTU0s/312-bug-teacher-lands-in-the-wrong-course-after-cloning

### DIFF
--- a/tutor/src/flux/new-course.coffee
+++ b/tutor/src/flux/new-course.coffee
@@ -62,6 +62,7 @@ StoreDefinition = makeStandardStore('NewCourse', {
 # used by api
 # coffeelint: disable=no_empty_functions
   clone: ->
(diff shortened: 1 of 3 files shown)


tutor/src/flux/new-course.coffee
```diff
@@ -63,7 +63,8 @@ StoreDefinition = makeStandardStore('NewCourse', {
 # coffeelint: disable=no_empty_functions
   clone: ->
   create: ->
-# coffeelint: enable=no_empty_functions
+    @_local['newlyCreatedCourse'] = null
+
   created: (newCourse) ->
```
nathanstitt: oops looks like the `coffeelint` re-enable got lost here :(

PR #1481, opened 2016-12-07
fix for adding with due date as today

pandafulmanda:
- [x] can click on add to today
- [x] when default due date is today, open date should also be today
- [x] disable saving when dates are wrong
- [x] fix timezone draggy thingy
- [ ] ~~fix default times -- this is missing async state at the moment~~ will make a new pr for this


nathanstitt:
I'mma gonna merge but there's maybe a bit more to do.  Not sure if UX agrees tho.

When you start a new assignment, the button's default to enabled, even though the assignment starts in an invalid state because there's no title.

pandafulmanda:
yeah i started to make them disabled on mount, but then i realized we had a lot of code in there making sure it was enabled until the save button was hit so i figured it was intentional.  Also, i tried it both ways, and it did make sense initially enabled because otherwise the button is disabled and the user may not know why, and then if the errors show on load that's also jarring.

PR #1483, opened 2016-12-08
add instructions for nvm

philschatz:
This changes the `README.md` to let developers know to use `nvm` so everyone is using the same version of node 😄


nathanstitt:
Thanks!

PR #1495, opened 2016-12-13
Update cloning wizard 

nathanstitt:
Only ask for course type if the teacher has both CC & Tutor types.

Removes the "Copy Question Library?" panel


tutor/resources/styles/components/new-course.less
```diff
@@ -234,29 +226,11 @@
       background-position: left center;
     }
 
-    [data-copy-or-not],
-    [data-new-or-copy] {
-      &::before {
-        background-position: center;
```
pandafulmanda: this is still needed for `[data-new-or-copy] `
screen shot 2016-12-13 at 11 50 45 am

PR #1504, opened 2016-12-14
flex them

pandafulmanda:
screen shot 2016-12-14 at 9 49 46 am

https://trello.com/c/8FAr7Zz5/307-bug-when-calendar-screen-is-narrow-the-buttons-wrap-in-the-wrong-order

needs to be checked in Safari


nathanstitt:
I tested Safari, looks great!

PR #1505, opened 2016-12-14
quick invalid page improvements

pandafulmanda:
screen shot 2016-12-15 at 2 44 14 pm

https://trello.com/c/GL4ILLdw/309-bug-error-pages-need-to-be-styled

will probably be completely redesigned by ux, but this has fixes related to `transitionTo`, a new `ButtonLink` component that should be useful, and removes a lot of dead code


tutor/src/components/invalid-page.cjsx
```diff
@@ -1,13 +1,18 @@
 React  = require 'react'
 {Link} = require 'react-router'
+BS = require 'react-bootstrap'
```
nathanstitt: Eeek, so this has been wrong for awhile huh?

Why not use TutorLink, then we wouldn't need the `TutorRouter.makePathname` below

pandafulmanda: good point, will make some updates

PR #1507, opened 2016-12-14
Use a titleized version of code instead of "default" for course info

nathanstitt:
When a appearance code isn't found it's usually because admin has set
something that's just like the book's title.  In that case we're better
off to send back a cleaned up code instead of 'default'


pandafulmanda:
clever

PR #1509, opened 2016-12-15
Fix timezone missyncing

pandafulmanda:
https://trello.com/c/GMlX0LA7/353-bug-intermittent-assignments-shift-back-one-day-after-dragging-to-intended-due-date

can be triggered as easily as going to calendar, click on an assignment to edit, and then clicking cancel.  Builder unmount to unsync timezone was getting called after the calendar's sync call.

### DIFF
--- a/tutor/src/components/course-calendar/month.cjsx
+++ b/tutor/src/components/course-calendar/month.cjsx
@@ -224,8 +224,6 @@ CourseMonth = React.createClass
     )
 
   onDragHover: (day) ->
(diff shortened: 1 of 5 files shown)


nathanstitt:
I'm a bit confused on how this fixes the bug, but can't really test since I still haven't been able to replicate it :(

It seems to me like we were still calling `TimeHelper.syncCourseTimezone` onMount in both cases, but maybe it was being skipped somehow?

PR #1511, opened 2016-12-15
fix disappearing navbar

pandafulmanda:
this is the simplest way to reset the classes on the pinned components.  `without` wasnt working because the function expects the excluded values as individual and not the whole array of values to exclude. for that, `difference` is the function to use.  also, pinned-on and pinned-shy, classes added or removed based on scroll were not being considered despite needing to be removed on unmount

https://trello.com/c/suWaGT0S/358-bug-menu-disappearing-on-all-browsers-randomly

a reliable way to reproduce this is to go to homework builder from calendar, scroll down to where the main menu hides, and then click the back button

### DIFF
--- a/shared/src/components/pinned-header-footer-card/index.cjsx
+++ b/shared/src/components/pinned-header-footer-card/index.cjsx
@@ -3,7 +3,6 @@ ReactDOM = require 'react-dom'
 
 union = require 'lodash/union'
 omit  = require 'lodash/omit'
-without = require 'lodash/without'
 
 ScrollListenerMixin = require '../../mixins/ScrollListener'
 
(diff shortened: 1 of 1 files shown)


shared/src/components/pinned-header-footer-card/index.cjsx
```diff
@@ -46,7 +46,7 @@ module.exports = React.createClass
     @setBodyClassList()
 
   componentWillUnmount: ->
-    document.body.className = without(document.body.classList, @desiredBodyClassList()).join(' ')
+    document.body.className = @state.originalClassName
```
nathanstitt: hm, I think what's going on here is that if there's two of these on the screen they:

comp A mounts, remembers that body has "C,D" class, it adds E

comp B mounts, sees body has "C,D,E", remembers that list, and add F.

Now comp A unmount, and removes "C,D, E", so body now has "F".

comp B unmount, and sets it back to what it remembered as the original, which is: "C,D,E"

I think the only way to work around this is for the component to remove only things it adds, and forget about attempting to "remember" the original.

PR #1512, opened 2016-12-15
Add error handlers for accessing course when it's not open

nathanstitt:
Depends on BE PR: https://github.com/openstax/tutor-server/pull/1329

Screenshots from when I manually set a course to not be opened / closed and then attempted to practice with it as a student.

screen shot 2016-12-15 at 3 28 19 pm

screen shot 2016-12-15 at 3 27 40 pm


pandafulmanda:
really glad about the `error-monitoring` move into a directory

PR #1515, opened 2016-12-16
Remove old 'match' segment from route

nathanstitt:
The new router doesn't have that, it has pathname as a top-level property.


nathanstitt:
This is based off productions `3aff153f0fc68abce60790124d592745db5901be` and not intended for merging

nathanstitt:
Closing this since we've decided not to do a hot fix release, this will be included in the next regularly scheduled release.

PR #1522, opened 2016-12-21
Update CC enrollment; include enrollment_code param

nathanstitt:
Note the change request doesn't mention changing the large enrollment code that appears at the top, I'm checking on that

screen shot 2016-12-21 at 4 17 12 pm


pandafulmanda:
hmm i just had an idea, i wonder if we should link to `#coach-wrapper` directly for them instead of keeping step 2?

PR #1541, opened 2017-01-03
display multiple notices at once

nathanstitt:
screen shot 2017-01-03 at 1 05 27 pm

### DIFF
--- a/coach/src/navigation/index.cjsx
+++ b/coach/src/navigation/index.cjsx
@@ -40,7 +40,7 @@ Navigation = React.createClass
 
   onCourseEnded: (notice) ->
     NotificationActions.acknowledge(notice)
-    channel.emit('show.registration', view: 'second-semester-registration')
+    channel.emit('show.second-semester-registration', view: 'second-semester-registration')
 
   render: ->
     {course} = @props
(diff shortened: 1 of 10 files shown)


coach/src/navigation/index.cjsx
```diff
@@ -40,7 +40,7 @@ Navigation = React.createClass
 
   onCourseEnded: (notice) ->
     NotificationActions.acknowledge(notice)
-    channel.emit('show.registration', view: 'second-semester-registration')
+    channel.emit('show.second-semester-registration', view: 'second-semester-registration')
```
nathanstitt: this is an unrelated bug fix for the clicking on the second semester link.  I had it as a separate pr but it snuck in here

shared/src/model/notifications.coffee
```diff
@@ -56,11 +56,10 @@ Notifications = {
       @emit('change')
 
   getActive: ->
-    return NOTICES[0] if NOTICES.length
+    active = clone(NOTICES)
```
Dantemss: Hm I'm a bit confused by the interaction between `NOTICES`, `display` and `getActive`.

So `NOTICES` seems to contain an array of the things that are actually displayed and is set by the `display` method. `display` is called from `setCourseRole`, which means when you pick a dashboard?

`getActive` includes all the `NOTICES` but also adds extra notices from the pollers... (how are these different?)

I think we might want to document what's happening here for the future.

Dantemss: Ah I think I see now... `NOTICES` are set by admins... the poller stuff handles the automated notices... we might want to add a comment explaining this.

PR #1542, opened 2017-01-03
remove last of isMounted checks

nathanstitt:


tutor/src/components/tutor-popover.cjsx
```diff
@@ -51,7 +51,6 @@ TutorPopover = React.createClass
     content.querySelectorAll('img')
 
   imageLoaded: (iter) ->
-    return unless @isMounted()
```
Dantemss: I'm guessing this was added to avoid the warning from the @setState below if the image loads after the component is unmounted... is it ok to remove? Unlike the other files I see nothing that removes the listener on unmount...

shared/src/components/buttons/async-button.cjsx
```diff
@@ -40,7 +40,7 @@ module.exports = React.createClass
 
   checkForTimeout: ->
```
Dantemss: Does this timeout need to be disabled somehow when unmounting?

coach/src/user/menu.cjsx
```diff
@@ -38,7 +38,7 @@ UserMenu = React.createClass
     @context.navigator.emit('show.student_id', view: 'student_id')
 
   update: ->
-    @forceUpdate() if @isMounted()
+    @forceUpdate()
```
Dantemss: I'm guessing this component never unmounts, since it's the menu? Nothing in here calls this `update` function, but not sure if something external could call it.

coach/src/user/login-gateway.cjsx
```diff
@@ -68,7 +68,6 @@ LoginGateway = React.createClass
     _.delay(@windowClosedCheck, SECOND)
 
   parseAndDispatchMessage: (msg) ->
```
Dantemss: This one gets removed on Unmount.

coach/src/user/login-gateway.cjsx
```diff
@@ -83,7 +82,6 @@ LoginGateway = React.createClass
     @props.windowImpl.addEventListener('message', @parseAndDispatchMessage)
 
   windowClosedCheck: ->
```
Dantemss: But this timeout doesn't... Might need to cancel it?

shared/src/components/buttons/async-button.cjsx
```diff
@@ -40,7 +40,7 @@ module.exports = React.createClass
...
     {isWaiting} = @props
-    @setState(isTimedout: true, delayTimeout: null) if isWaiting and @isMounted()
+    @setState(isTimedout: true, delayTimeout: null) if isWaiting
```
philschatz: Could removing this `@isMounted()` check cause exceptions when a network request begins and the student closes the window before the server responds?

coach/src/concept-coach/index.cjsx
```diff
@@ -101,7 +101,7 @@ class ConceptCoachAPI extends EventEmitter2
     @removeAllListeners()
 
   remove: ->
```
Dantemss: Both `initialize()` and `destroy()` call this function, but I'm guessing each one happens only once so it's fine?

coach/src/concept-coach/index.cjsx
```diff
@@ -155,7 +155,7 @@ class ConceptCoachAPI extends EventEmitter2
       @open(props)
 
   updateToView: (view) ->
-    if @component?.isMounted()
+    if @component
```
Dantemss: I'm guessing `removeAllListeners()` in `destroy()` takes care of this?

tutor/src/components/book-content-mixin.cjsx
```diff
@@ -25,12 +25,14 @@ IS_INTRO_SELECTORS = '.splash img, [data-type="cnx.flag.introduction"]'
 
 LinkContentMixin =
   componentDidMount:  ->
+    @_isMounted = true
```
philschatz: _Minor:_ Since this is a mixin, would these fields accidentally trample the same field in `resize-listener-mixin`? Just wondering if the field name should be prefixed or not. 

But I'm just mentioning it as a minor issue; the meaning of `_isMounted` is pretty clear to me

shared/src/components/buttons/async-button.cjsx
nathanstitt: No, calling `setState` on an unmounted component doesn't cause an exception, just a warning.  

However Dante points out that we aren't clearing the timeout, but we do have a method to do so. I've added a call to it in `componentWillUnmount`

tutor/src/components/tutor-popover.cjsx
nathanstitt: Yeah, this component doesn't really have any listeners, it's purpose is to load and display a popover when the mouse moves over an element.

Since the `imagesLoaded` call originates from mouse `onEnter` , it's guaranteed to only be called when mounted, otherwise it wouldn't be receiving events.

And if i've mis-read the way the code somehow, the worst that happens is that we call `setState` on an unmounted component and React logs a warning.

coach/src/user/menu.cjsx
nathanstitt: yes, I don't see anywhere that calls it either.  Am tempted to remove it but maybe I've missed something.  Right now it's just a warning to call `forceUpdate` on an unmounted component, but would be an error to remove a method and then it's called somewhere

coach/src/user/login-gateway.cjsx
nathanstitt: Great catch!

coach/src/concept-coach/index.cjsx
nathanstitt: yeah, I hope so 🙏  I'm not entirely sure where all the coaches stuff gets called from, since this is called into from web view

coach/src/concept-coach/index.cjsx
nathanstitt: I **think** so.  Same comment as above really

PR #1544, opened 2017-01-03
upload jest code coverage to codecov.io

philschatz:
Inspired by [cnx-recipes#136](https://github.com/Connexions/cnx-recipes/pull/136), this generates code coverage for each of `tutor`, `exercises`, `coach`, and `shared` using https://facebook.github.io/jest/docs/configuration.html#collectcoveragefrom-array

You can see the results at https://codecov.io/gh/openstax/tutor-js/commits and then click a commit, and then click the "Files" tab near the top

Thanks @nathanstitt for the how-to! It was surprisingly simple!

# Questions

- [x] should I also add a config for `shared`?
- [x] will codecov.io properly OR all the coverage files from the various builds together?
- [ ] should I change the inclusion/exclusion patterns (do they spend extra time instrumenting files that do not end up being run?)
- [ ] should the karma code get removed?

# Screenshots

image

image

refs openstax/napkin-notes#68

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,6 @@
+comment:
+  layout: "header, diff, changes, uncovered, tree"
+  behavior: default
+
+# To Turn off comments completely:
+# comment: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,4 +11,7 @@ script:
 after_failure:
   - npm list
 after_success:
-  - cat ./coverage/all.lcov.txt | node ./node_modules/.bin/coveralls --verbose
+  # Report coverage to codecov
+  - bash <(curl -s https://codecov.io/bash)
+  # Report to coveralls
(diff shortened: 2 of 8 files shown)


nathanstitt:
Looks great @philschatz - thanks!

I'd say go ahead and remove the karma configs, we've been using Jest long enough without problems that I think there's no reason to keep that around.

I don't think the inclusion/exclusion should need tweaked, but feel free to run some tests to find a speedier setup if you want 😁

philschatz:
What if karma is removed in a separate PR? That way it would be easier to search for it if folks wanted to pick it back out for something like selenium tests or something

nathanstitt:
Sounds good, we'll merge after the next release, thanks @philschatz

nathanstitt:
Thanks @philschatz !

PR #1545, opened 2017-01-04
only unmount if @component is present

nathanstitt:


nathanstitt:
self merged so we can quickly deploy for testing

PR #1546, opened 2017-01-04
CC fixes and enrollment verbiage updates

nathanstitt:
screen shot 2017-01-04 at 2 38 00 pm

screen shot 2017-01-04 at 2 52 09 pm

### DIFF
--- a/coach/src/concept-coach/index.cjsx
+++ b/coach/src/concept-coach/index.cjsx
@@ -158,7 +158,7 @@ class ConceptCoachAPI extends EventEmitter2
   updateToView: (view) ->
     if @component
       if view is 'close'
-        @component.props.close()
+        @component.props?.close?()
       else
         navigation.channel.emit("show.#{view}", {view})
     else if componentModel.mounter? and view isnt 'close'
--- a/coach/src/exercise/index.cjsx
+++ b/coach/src/exercise/index.cjsx
@@ -68,7 +68,7 @@ ExerciseBase = React.createClass
         eventData = change: step, data: step, status: 'saving'
 
         channel.emit("change.#{step.id}", eventData)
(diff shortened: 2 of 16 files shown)


coach/src/concept-coach/index.cjsx
```diff
@@ -158,7 +158,7 @@ class ConceptCoachAPI extends EventEmitter2
   updateToView: (view) ->
     if @component
       if view is 'close'
-        @component.props.close()
+        @component.props?.close?()
       else
```
nathanstitt: a small bug, sometimes it would error when closing if the current component didn't have a handler

coach/src/exercise/index.cjsx
```diff
@@ -68,7 +68,7 @@ ExerciseBase = React.createClass
         eventData = change: step, data: step, status: 'saving'
 
         channel.emit("change.#{step.id}", eventData)
-        api.channel.emit("exercise.#{step.id}.save.send", id: step.id, _.pick(step, 'answer_idr'))
+        api.channel.emit("exercise.#{step.id}.save.send", id: step.id, _.pick(step, 'answer_id'))
```
nathanstitt: big bug 😬 this was causing answers to fail to save

coach/src/navigation/settings.coffee
```diff
@@ -14,5 +14,6 @@ settings =
     logout: DEFAULT_PATTERN
     default: '{prefix}{base}'
     close: '{prefix}'
+    'second-semester-registration': DEFAULT_PATTERN
```
nathanstitt: medium bug, since we didn't a route defined for 'second-semester-registration', it would error when the link on the banner was clicked

PR #1549, opened 2017-01-06
Use ecosystem of task for reading table of contents

nathanstitt:
In some cases such as when a course is cloned, a old task can have a different ecosystem than the course currently has, which causes a bug when the course ecosystem isn't found in the TocStore

### DIFF
--- a/tutor/src/components/task-plan/select-topics.cjsx
+++ b/tutor/src/components/task-plan/select-topics.cjsx
@@ -17,6 +17,8 @@ SelectTopics = React.createClass
     onSectionChange: React.PropTypes.func.isRequired
     hide: React.PropTypes.func.isRequired
     selected: React.PropTypes.array
+    header: React.PropTypes.string.isRequired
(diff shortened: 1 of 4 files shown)


tutor/src/components/task-plan/select-topics.cjsx
```diff
@@ -28,8 +30,7 @@ SelectTopics = React.createClass
     @props.onSectionChange?(sectionIds)
 
   renderDialog: ->
-    {courseId, planId, selected, hide, header, primary, cancel} = @props
-    ecosystemId = CourseStore.get(courseId).ecosystem_id
+    {courseId, planId, selected, hide, header, primary, cancel, ecosystemId} = @props
```
nathanstitt: This is the bug fix.  We're now reading the ecosystem from @props, rather than the course.

@Dantemss note that I didn’t default the ecosystem to the course.  I now think that’s a bad idea all around, since we're using the ecosystem is required on @props, and that's also what we're using to load the TocStore below on line 58.

PR #1554, opened 2017-01-09
set a min height on CC launcher's body

nathanstitt:
Otherwise it'll get squished when page is very narrow and things wrap

Turns this:
screen shot 2017-01-09 at 1 02 10 pm

Into this:

screen shot 2017-01-09 at 1 19 22 pm

### DIFF
--- a/coach/resources/styles/components/concept-coach/launcher.less
+++ b/coach/resources/styles/components/concept-coach/launcher.less
@@ -45,7 +45,6 @@
         h2 {
           font-size: 2.6rem;
           letter-spacing: -1px;
-          color: @dark-blue;
           margin: 0;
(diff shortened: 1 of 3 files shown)


coach/resources/styles/components/concept-coach/launcher.less
```diff
@@ -42,6 +43,9 @@
         .justify-content(center);
         color: @dark-blue;
         h2 {
+          font-size: 2.6rem;
+          letter-spacing: -1px;
+          color: @dark-blue;
```
pandafulmanda: i think this color would be better to fix this way: Connexions/webview#1490, otherwise PR looks good

PR #1556, opened 2017-01-09
Add ecosystem_id param to task plan saving endpoint

nathanstitt:
This way the BE knows what ecosystem should be used

### DIFF
--- a/tutor/specs/helpers/task-plan.spec.coffee
+++ b/tutor/specs/helpers/task-plan.spec.coffee
@@ -64,12 +64,12 @@ describe 'task-plan helper', ->
 
     TaskPlanFlux.TaskPlanStore.isNew.mockReturnValue(true)
     expect(TaskPlan.apiEndpointOptions(SINGLE.id, COURSE_ID))
-      .to.deep.equal(url: "courses/#{COURSE_ID}", method: 'POST')
+      .to.deep.equal(url: "courses/#{COURSE_ID}/plans", method: 'POST')
 
     TaskPlanFlux.TaskPlanStore.get.mockReturnValue(ecosystem_id: 42)
     expect(TaskPlan.apiEndpointOptions(SINGLE.id, COURSE_ID))
       .to.deep.equal(
-        url: "courses/#{COURSE_ID}", method: 'POST', params: {ecosystem_id:42}
+        url: "courses/#{COURSE_ID}/plans", method: 'POST', params: {ecosystem_id:42}
       )
 
     undefined
--- a/tutor/specs/helpers/task-plan.spec.coffee
+++ b/tutor/specs/helpers/task-plan.spec.coffee
@@ -1,5 +1,7 @@
 _ = require 'underscore'
 
+jest.mock('../../src/flux/task-plan')
+TaskPlanFlux = require '../../src/flux/task-plan'
 
 TaskPlan = require '../../src/helpers/task-plan'
 {CourseListingActions} = require '../../src/flux/course-listing'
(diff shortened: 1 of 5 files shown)


nathanstitt:
For https://trello.com/c/WZqtDoHp/277-bug-cloning-assignments-does-not-work-unless-the-cloned-course-has-the-same-ecosystem-as-the-assignment

tutor/src/helpers/task-plan.coffee
```diff
@@ -31,4 +32,16 @@ module.exports = {
...
     _.first(dates.sort()) or ''
+
+
+  apiEndpointOptions: (id, courseId) ->
+    task = TaskPlanStore.get(id)
+    options = if TaskPlanStore.isNew(id)
+      url: "courses/#{courseId}"
+      method: 'POST'
+    else
+      url: "plans/#{id}"
```
pandafulmanda: the query can be sent as

```coffee
params: 
  ecosystem_id: task.ecosystem_id
```

as axios will turn it into a query string: https://github.com/mzabriskie/axios#example

tutor/specs/helpers/task-plan.spec.coffee
```diff
@@ -54,3 +56,18 @@ describe 'task-plan helper', ->
...
+    expect(TaskPlan.apiEndpointOptions(SINGLE.id, COURSE_ID))
+      .to.deep.equal(url: "plans/#{SINGLE.id}")
+
+    TaskPlanFlux.TaskPlanStore.isNew.mockReturnValue(true)
+    expect(TaskPlan.apiEndpointOptions(SINGLE.id, COURSE_ID))
+      .to.deep.equal(url: "courses/#{COURSE_ID}", method: 'POST')
+
+    TaskPlanFlux.TaskPlanStore.get.mockReturnValue(ecosystem_id: 42)
+    expect(TaskPlan.apiEndpointOptions(SINGLE.id, COURSE_ID))
+      .to.deep.equal(url: "courses/#{COURSE_ID}?ecosystem_id=42", method: 'POST')
```
pandafulmanda: `method` doesnt seemed to be defined when not `isNew` from the helper?  which would default to `PATCH` in the api handler.

tutor/src/helpers/task-plan.coffee
nathanstitt: ah, neat.  i'll do that

tutor/src/helpers/task-plan.coffee
```diff
@@ -31,4 +32,16 @@ module.exports = {
   earliestDueDate: (plan) ->
     dates = _.map(plan?.tasking_plans, 'due_at')
     _.first(dates.sort()) or ''
+
+
+  apiEndpointOptions: (id, courseId) ->
+    task = TaskPlanStore.get(id)
```
pandafulmanda: also, don't need to get task unless it's not new, but not big deal

tutor/src/helpers/task-plan.coffee
```diff
@@ -31,4 +32,17 @@ module.exports = {
...
+    options = if TaskPlanStore.isNew(id)
+      url: "courses/#{courseId}"
```
pandafulmanda: this used to be `url: "courses/#{courseId}/plans"`

tutor/src/helpers/task-plan.coffee
nathanstitt: hm, no we need to append the ecosystem_id even for new tasks.  Since if they're cloned they should use the original ecosystem id, for now at least

pandafulmanda: ooh i see, yah i read this wrong

PR #1560, opened 2017-01-11
Use the most recently joined course for an ecosystem

nathanstitt:
When a student is taking a second-semester of a course, there will be multiple matching courses for the given ecosystem uuid.  In that case we must use only the most recently joined

### DIFF
--- a/coach/src/user/model.coffee
+++ b/coach/src/user/model.coffee
@@ -50,7 +50,13 @@ User =
     Boolean(@isLoggedIn() and @getCourse(collectionUUID))
 
   getCourse: (collectionUUID) ->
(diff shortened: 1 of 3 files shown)


coach/src/user/model.coffee
```diff
@@ -50,7 +57,13 @@ User =
...
 
   getCourse: (collectionUUID) ->
-    _.findWhere( @courses, ecosystem_book_uuid: collectionUUID )
+    _.chain(@courses)
+      .where( ecosystem_book_uuid: collectionUUID )
+      .sortBy( (course) ->
+        course.getRole().joined_at
+      )
+      .last()
+      .value()
```
nathanstitt: This is the bug fix - before we were just looking at the first match, now we look at the most recently joined

PR #1562, opened 2017-01-11
Switch FE to using latest_enrollment_at instead of joined_at to determine current CC course

Dantemss:
Requires https://github.com/openstax/tutor-server/pull/1343

### DIFF
--- a/coach/specs/user/model.spec.coffee
+++ b/coach/specs/user/model.spec.coffee
@@ -52,7 +52,6 @@ describe 'User', ->
     # test that it doesn't blowup if role latest_enrollment_at is missing
     delete matches[0].roles[0].latest_enrollment_at
     matches[1].roles[0].latest_enrollment_at = "2017-01-06T20:00:00.000Z"
-    delete matches[0].roles[0].latest_enrollment_at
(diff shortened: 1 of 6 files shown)


coach/specs/user/model.spec.coffee
```diff
@@ -41,18 +41,18 @@ describe 'User', ->
...
 
-    # test that it doesn't blowup if role joined_at is missing
-    delete matches[0].roles[0].joined_at
-    matches[1].roles[0].joined_at = "2017-01-06T20:00:00.000Z"
-    delete matches[0].roles[0].joined_at
+    # test that it doesn't blowup if role latest_enrollment_at is missing
+    delete matches[0].roles[0].latest_enrollment_at
+    matches[1].roles[0].latest_enrollment_at = "2017-01-06T20:00:00.000Z"
+    delete matches[0].roles[0].latest_enrollment_at
     expect(User.getCourse(TWO_COURSE_ECOSYSTEM_ID).id).to.equal(matches[1].id)
```
nathanstitt: Hehe, the fact that `delete matches[0].roles[0].joined_at` was present twice on line 53 & 55 was an oversight.  You can remove one of them.

PR #1564, opened 2017-01-11
notify notices store when course and/or role updates

nathanstitt:
The bar only was firing updates on mount, which didn't work for CC when someone logs in


shared/src/components/notifications/bar.cjsx
```diff
@@ -1,5 +1,6 @@
 React = require 'react'
 isEmpty = require 'lodash/isEmpty'
+isEqual = require 'lodash/isEqual'
```
pandafulmanda: hmmm this was added but not used?

nathanstitt: ah, yeah it wasn't

PR #1565, opened 2017-01-12
Set a view to display after a user logs in

nathanstitt:
And then use it to force re-display of enrollment screen


nathanstitt:
Closing this since #1567 fixes it better

PR #1567, opened 2017-01-12
if enrollment code doesnt match user's course, send user to enrollment workflow

pandafulmanda:
Dependent on openstax/tutor-server#1344

## When enrollment code is for a past course
screen shot 2017-01-13 at 6 16 06 am

## Enrollment code matches existing registered course
screen shot 2017-01-12 at 5 09 22 pm

## Enrollment code matches for active course
screen shot 2017-01-12 at 5 11 26 pm

If student is currently enrolled in a previous semester, clicking enroll will start enrolling the student into the course that matches the enrollment code.

If the student is not previously enrolled in courses relating to the book, clicking enroll will start enrolling, as in the above case.

## If no enrollment code or if code is for past course, or is wrong:
for those enrolled in a past course:
screen shot 2017-01-17 at 2 14 44 pm

screen shot 2017-01-17 at 2 16 10 pm

## Otherwise, normal enrollment flow:

with proper enrollment code:
screen shot 2016-12-27 at 2 45 06 am

without proper enrollment code:
screen shot 2017-01-17 at 2 19 57 pm

both will require confirmation
screen shot 2017-01-17 at 2 20 09 pm

after a message about joining and loading first exercise, exercise should load within 2 seconds.

### DIFF
--- a/coach/src/concept-coach/base.cjsx
+++ b/coach/src/concept-coach/base.cjsx
@@ -26,7 +26,8 @@ navigator = navigation.channel
 # TODO Move this and auth logic to user model
 # These views are used with an authLevel (0, 1, 2, or 3) to determine what views the user is allowed to see.
 VIEWS = [
-  'loading', 'login', 'registration',
+  'loading', 'login',
+  ['registration', 'second-semester-registration'],
   [
     'task', 'progress', 'profile', 'dashboard', 'student_id',
     'registration', 'second-semester-registration'
--- a/coach/src/concept-coach/base.cjsx
+++ b/coach/src/concept-coach/base.cjsx
@@ -159,7 +159,8 @@ ConceptCoach = React.createClass
   showTasks: ->
     @updateView(view: 'task')
 
-  updateUser: ->
+  updateUser: (options) ->
+    @setState(ignoreDefaultView: true) if options?.justConfirmed
     userState = User.status(@props.collectionUUID, @props.enrollmentCode)
     view = @getAllowedView(userState)
 
--- a/coach/src/concept-coach/base.cjsx
+++ b/coach/src/concept-coach/base.cjsx
@@ -50,7 +50,7 @@ ConceptCoach = React.createClass
     defaultView: _.chain(VIEWS).last().first().value()
 
   getInitialState: ->
-    userState = User.status(@props.collectionUUID)
+    userState = User.status(@props.collectionUUID, @props.enrollmentCode)
     view = @getAllowedView(userState)
     userState.view = view
     userState
@@ -109,7 +109,10 @@ ConceptCoach = React.createClass
     navigator.off('show.*', @updateView)
 
   getAllowedView: (userInfo) ->
-    {defaultView} = @props
+    {defaultView, collectionUUID, enrollmentCode} = @props
+    course = User.getCourse(collectionUUID, enrollmentCode)
+
+    # console.info(userInfo, course)
     if not userInfo.isLoaded
       authLevel = 0
     else if userInfo.preValidate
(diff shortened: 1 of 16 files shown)


coach/src/concept-coach/base.cjsx
```diff
@@ -109,7 +110,11 @@ ConceptCoach = React.createClass
     navigator.off('show.*', @updateView)
 
   getAllowedView: (userInfo) ->
-    {defaultView} = @props
+    {defaultView, collectionUUID, enrollmentCode} = @props
+    ignoreDefaultView = @state?.ignoreDefaultView or false
+    course = User.getCourse(collectionUUID, enrollmentCode)
+
+    # console.info(userInfo, course)
     if not userInfo.isLoaded
```
nathanstitt: stray `console.log`; commented out tho, so no biggy but also no reason to leave it in

nathanstitt:
Looks good @pandafulmanda - do you think we should go ahead and merge this and then make additional updates as UX desires, or should we deploy it from branch so they can weigh in?

coach/src/concept-coach/base.cjsx
pandafulmanda: good catch!

pandafulmanda:
just found a bug where after initial enrollment and then the task view loading, this happens:

screen shot 2017-01-17 at 2 23 28 pm

will need to fix before this can go out, but otherwise, the enrollment flow is reflective of intended updates.

coach/src/concept-coach/index.cjsx
```diff
@@ -142,8 +142,8 @@ class ConceptCoachAPI extends EventEmitter2
       )
 
 
-  open: (props) ->
-    openProps = _.extend({}, props, open: true)
+  open: (props...) ->
+    openProps = _.extend({}, props..., open: true)
     openProps.triggeredFrom = _.pick(props, 'moduleUUID', 'collectionUUID')
```
nathanstitt: hm, doesn't the splat `...` operator turn the arguments into an array, which would mean that `props` is an array ?  that seems weird to pass to `_.extend`, but if that's true can we call the var something other than `props` ?  When I see that name i think it's an object

coach/src/user/model.coffee
```diff
@@ -46,29 +46,41 @@ User =
...
 
-  getCourse: (collectionUUID) ->
-    _.chain(@courses)
-      .where( ecosystem_book_uuid: collectionUUID )
-      .sortBy( (course) -> course.getRole()?.latest_enrollment_at or '' )
-      .last()
-      .value()
+    filters.push(sortyByJoinedAt)
+
+    _.last(_.compose(filters...)(@courses))
```
nathanstitt: Nice way to compose the search! I'd suggest calling them `operations` rather than `filters` though since it also sorts.

pandafulmanda: ah yes, good idea.  i wasn't loving the name either

shared/src/model/notifications.coffee
```diff
@@ -56,16 +61,31 @@ Notifications = {
...
   #   based on the user's relationship with the course
   # The `getActive` method is the single point for checking which notifications
   # should be displayed
   getActive: ->
-    active = clone(NOTICES)
+    active = clone(@rejectHidden(NOTICES))
     for type, poller of POLLERS
-      active.push(clone(notice)) for notice in poller.getActiveNotifications()
+      active.push(clone(notice)) for notice in @rejectHidden(poller.getActiveNotifications())
     active
```
nathanstitt: Since reject creates a copy we no longer need to call `clone`  I'd only had that there so that if the caller altered the array it wouldn't alter the private `NOTICES`

PR #1570, opened 2017-01-18
Fix bug with CC dashboard

nathanstitt:
When a section on the dashboard has all either correct or incorrect, a JS error would be thrown and no further sections would render.

Turned out it was because we were passing `undefined` as a child to the `BS.ProgressBar` component which didn't like that

screen shot 2017-01-18 at 5 04 51 pm


tutor/src/components/cc-dashboard/section-performance.cjsx
```diff
@@ -29,9 +39,10 @@ SectionPerformance = React.createClass
 
     <div>
       <BS.ProgressBar className="reading-progress-group">
-        {correctBar}
-        {incorrectBar}
+        {bars}
       </BS.ProgressBar>
```
nathanstitt: This is the actual bug - if either `correctBar` or `incorrectBar` is null, the `BS.ProgressBar` component references something it should.

PR #1571, opened 2017-01-18
Only enable code coverage for Travis, not interactive

nathanstitt:
Makes the interactive specs less verbose and run faster when using it for TDD.

Also removes some old gulp based tasks that don't exist any longer

Also replaces a bunch of `to.have.length(x)` with `toHaveRendered`


nathanstitt:
Merging this so we can write tests easier - need it for #1573

PR #1572, opened 2017-01-19
Attempt to fix safari 9 right arrow on narrow screens

nathanstitt:
While I'm not a fan of browser sniffing, this seems like the only way to get this to work :(


pandafulmanda:
nice!

PR #1573, opened 2017-01-19
Fix qa view loading so it waits for ecosystems to be available

nathanstitt:
Otherwise it may render if the loads before the ecosystems are ready.  This happens if the book is extremely small

### DIFF
--- a/tutor/src/components/qa/index.cjsx
+++ b/tutor/src/components/qa/index.cjsx
@@ -30,9 +30,10 @@ QADashboard = React.createClass
     )
 
   render: ->
+    return <h3 className="loading">Loading …</h3> if EcosystemsStore.isLoading()
+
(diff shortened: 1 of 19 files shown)


tutor/src/components/qa/index.cjsx
```diff
@@ -30,9 +30,10 @@ QADashboard = React.createClass
     )
 
   render: ->
+    return <h3 className="loading">Loading …</h3> if EcosystemsStore.isLoading()
+
     <div className="qa">
       <MatchForTutor {...@props} />
-      {<h3>Loading …</h3> if EcosystemsStore.isLoading()}
     </div>
```
nathanstitt: this is the bug fix, previously we were rendering the content even if the ecosystems were still loading

PR #1575, opened 2017-01-23
Hide due date on scores report header for Concept Coach courses

nathanstitt:
Before:

screen shot 2017-01-23 at 10 10 56 am

After:
screen shot 2017-01-23 at 10 10 38 am


pandafulmanda:
spec is sad :(

PR #1578, opened 2017-01-24
Fix safari layout bugs with scores report

nathanstitt:
Fixes:

* [x] title isn't visible on assignment headers
* [x] name column has weird extra large border at the top
* [x] pie progress is offset from center of cell
* [x] sort external assignment columns


tutor/src/components/scores/student-data-sorter.coffee
```diff
@@ -39,6 +39,8 @@ getSortValue = (student, index, dataType, displayAs) ->
             else
               record.completed_on_time_exercise_count
           percent(progress, record.exercise_count)
+    when 'external' # events are also of type 'external' and require no special handling
```
pandafulmanda: hmmm what if we made this the `else`/default?

PR #1579, opened 2017-01-25
DO NOT MERGE: temporarily replace openstax-logo for demo

nathanstitt:
screen shot 2017-01-25 at 4 50 01 pm 1


nathanstitt:
Closing since the demo is complete and we don't need to keep the change around

PR #1585, opened 2017-01-30
Left align popovers, specifically the task builder help one.

nathanstitt:
before:
screen shot 2017-01-30 at 11 27 38 am

after:

screen shot 2017-01-30 at 11 27 07 am

### DIFF
--- a/tutor/resources/styles/global/navbar.less
+++ b/tutor/resources/styles/global/navbar.less
@@ -217,7 +217,6 @@
       border: 0;
       background-color: transparent;
       width: 100%;
-      text-align: left;
       padding-left: 0;
       text-transform: none;
       text-align: left;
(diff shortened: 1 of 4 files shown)


tutor/resources/styles/global/navbar.less
```diff
@@ -217,7 +217,6 @@
       border: 0;
       background-color: transparent;
       width: 100%;
-      text-align: left;
```
pandafulmanda: hmm why remove this one?  was it just not doing anything obvious?

nathanstitt: because that rule was only there to override the global `justified` one.  Now that that's  removed, it isn't needed any longer.  I did a search for `text-align` `left` and that was the only one I found.  There's probably others but they weren't found by my simple grep

nathanstitt: And I checked the las info and it does render fine 😁

PR #1587, opened 2017-02-01
lateness icon should show for college readings and homeworks

pandafulmanda:
pass in isCollege for event row lateness icon logic -- lateness icon should show for both readings and homeworks, whereas they should not show for high school courses

## High school late readings
screen shot 2017-02-01 at 12 59 53 pm

## College late readings
screen shot 2017-02-01 at 1 00 11 pm

- [x] tests


nathanstitt:
Thanks @pandafulmanda !

PR #1589, opened 2017-02-03
we dont use these anymore.  SAD

pandafulmanda:


nathanstitt:
Agree - it's sad :(  but no reason to leave 'em in if they're non-functional.  We can always restore from history if we need them

philschatz:
Is there a way to pin this PR somewhere? Maybe tag a version of tutor w/ the code before it was removed? @RoyEJohnson was interested in getting https://github.com/openstax/os-webview to run Selenium & I mentioned the JS implementation in here that handled several normally-annoying corner-cases.

pandafulmanda:
okay, will add a tag, although for any new use of selenium in javascript, i highly recommend http://webdriver.io/ @RoyEJohnson, it handles some cases of async/promise mixup we were dealing with ourselves.

nathanstitt:
Yes, good idea - I just tagged the commit previous to the removal as `integration-specs` at https://github.com/openstax/tutor-js/tree/integration-specs

Thanks @philschatz !

nathanstitt:
I've also heard good things about [nightwatch](http://nightwatchjs.org/), but haven't used it personally yet

philschatz:
Thanks! And nightwatch looks interesting; I wonder how well it works with integration tests that need to change based on the state in the browser. Like "Select next week in the calendar" sometimes requires "select the next month if today is during the last week in the calendar month"

PR #1590, opened 2017-02-06
Implement an sample CC authorization flow w/Braintree

nathanstitt:
This is an example of one way we could implement CC charging using Braintree's JS api.

Based on my exploration, most other client-side processors have similar API's, and can be implemented something like this

screen shot 2017-02-06 at 10 02 36 am


Fredasaurus:
For the button placement, let's make sure we stick to this pattern.
reset-password

nathanstitt:
Closing since we're now doing payments and no reason to leave this hanging out

PR #1594, opened 2017-02-13
Update dependencies; webpack  2.2.1

nathanstitt:
Tested and believe it's ready for merge now


configs/webpack/base.coffee
```diff
@@ -121,15 +103,22 @@ makeBuildPlugins = (projectConfig) ->
...
+    new ExtractTextPlugin(
+      styleFilename
+    )
+  ]
+
 
+ path = require 'path'
 
-makePathsBase = (projectConfig) ->
+ makePathsBase = (projectConfig) ->
```
pandafulmanda: I can just make the update

nathanstitt: yah, that's not intentional :)  go for it if you're in there

pandafulmanda: also, why `require 'path'` twice?

nathanstitt:
`mangler was too mangly` gets my vote for best commit message 😁

PR #1595, opened 2017-02-16
Use ecosystemId from task when loading exercises

nathanstitt:
Fixes several areas where we were using the course ecosystem id to load exercises.  That works fine normally, but fails when a cloned task is loaded that refers to a past ecosystem.

### DIFF
--- a/tutor/src/components/task-plan/homework/loading-exercises-mixin.cjsx
+++ b/tutor/src/components/task-plan/homework/loading-exercises-mixin.cjsx
@@ -1,6 +1,7 @@
 _     = require 'underscore'
 React = require 'react'
 
+{TaskPlanStore} = require '../../../flux/task-plan'
 {ExerciseStore, ExerciseActions} = require '../../../flux/exercise'
 Icon = require '../../icon'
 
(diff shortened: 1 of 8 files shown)


tutor/src/components/task-plan/homework/loading-exercises-mixin.cjsx
```diff
@@ -12,7 +13,8 @@ LoadingExercisesMixin =
   componentWillUnmount: -> ExerciseStore.removeChangeListener(@exerciseLoadListenerOnUpdate)
 
   requestExerciseLoad: ->
-    ExerciseActions.loadForCourse(@props.courseId, @props.sectionIds, @props.ecosystemId)
+    ecosystemId = @props.ecosystemId or TaskPlanStore.getEcosystemId(@props.planId, @props.courseId)
```
pandafulmanda: so, we will always want to pass the ecosystem id when loading exercises from now on??

nathanstitt: No if you look at the api action https://github.com/openstax/tutor-js/blob/master/tutor/src/api/index.coffee#L92 it only uses the `ecosytemId` if it's present.

PR #1597, opened 2017-02-17
ignore spec and test directories from coverage

nathanstitt:
I noticed on the reports that these folders were being included, seems like they shouldn't be


philschatz:
Who watches the watchmen? I added coverage on tests so 

1. someone didn't accidentally comment out a test
2. why not get credit for writing tests?

nathanstitt:
just a test really, was curious what would happen if those directors were disallowed.  going to merge this and then make another undoing it

PR #1599, opened 2017-02-20
Hotfix coach no sourcemap

pandafulmanda:
Takes #1594 and applies it on top of [coach-20170119.a](https://github.com/openstax/tutor-js/releases/tag/coach-20170119.a) coach release


bin/serve
```diff
@@ -7,4 +7,4 @@ bin/checkinputs "$@"
 export OX_PROJECT=$1
 echo Serving: $OX_PROJECT
 
-webpack-dev-server --progress --config webpack.config.js
+./node_modules/webpack-dev-server/bin/webpack-dev-server.js --debug --progress --config webpack.config.js
```
philschatz: Would `$(npm bin)/webpack-dev-server --debug --progress ....` work here instead?

configs/webpack/base.coffee
```diff
@@ -153,17 +139,23 @@ makeProductionBase = (projectConfig) ->
...
+      rules: BASE_BUILD_LOADERS
     plugins: makeBuildPlugins({styleFilename}).concat([
       # Minify
-      new webpack.optimize.UglifyJsPlugin(minimize: true)
+      # https://webpack.github.io/docs/list-of-plugins.html#uglifyjsplugin
+      # https://github.com/webpack/webpack/issues/2704
+      new webpack.optimize.UglifyJsPlugin(
+        mangle: false
+        comments: false
+        sourceMap: true
```
philschatz: yay! I didn't realize that needed to be specified as an arg to Uglify, for some reason I thought webpack's sourcemap flag would work

bin/serve
nathanstitt: (Replying since I made this change).  Yes,  that'd be better form - i'll make the change in the update PR #1594

PR #1601, opened 2017-02-21
Fixup es6 babel support and other updates

nathanstitt:
Uses the `invalid-page` component as a test since it's super simple.


philschatz:
ah, that diff reminds me of `xmllint --pretty 2 myXMLFile.xhtml`

PR #1602, opened 2017-02-22
Limit CC course name to 40 characters so it doesn't overflow the image

nathanstitt:
Before:
screen shot 2017-02-22 at 2 59 10 pm

After:
screen shot 2017-02-22 at 3 07 11 pm


tutor/src/helpers/string.coffee
```diff
@@ -4,6 +4,12 @@ SMALL_WORDS = /^(a|an|and|as|at|but|by|en|for|if|in|nor|of|on|or|per|the|to|vs?\
 
 module.exports = {
 
+  ellipsis: (string, maxlen) ->
```
pandafulmanda: could use [`lodash.truncate`](https://lodash.com/docs/4.17.4#truncate)

nathanstitt: Good point, I wasn't aware of `truncate`, learned something new. Thanks!

PR #1603, opened 2017-02-22
Right align exercise numbers so they're spaced correctly

nathanstitt:
Regardless of width, i.e. 33) will be aligned to the right just as 1) will be

Fixes badness like this:
screen shot 2017-02-22 at 3 38 11 pm
screen shot 2017-02-22 at 3 38 02 pm

Now looks like:
screen shot 2017-02-22 at 3 37 29 pm
screen shot 2017-02-22 at 3 37 16 pm
screen shot 2017-02-22 at 3 37 05 pm

### DIFF
--- a/tutor/resources/styles/components/task-step/ends.less
+++ b/tutor/resources/styles/components/task-step/ends.less
@@ -81,7 +81,8 @@
       content: attr(data-question-number) ")";
       position: absolute;
       z-index: 1;
-      left: -25px;
+      right: 100%;
+      margin-right: 0.5rem;
       top: -2px;
       .exercise-typography();
     }


tutor/resources/styles/components/task-step/ends.less
```diff
@@ -81,7 +81,8 @@
       content: attr(data-question-number) ")";
       position: absolute;
       z-index: 1;
-      left: -25px;
+      right: 100%;
```
pandafulmanda: Not bigs, but...just checking, this would be the same thing as `left: 0;` right?

pandafulmanda: or hmm, it's not...coo

nathanstitt: Hehe, yeah that was a bit "clever" of me, it set's the right border of the label to the leftmost border of it's container.

PR #1610, opened 2017-02-24
Use individual requires so it doesn't pull in entire lodash

nathanstitt:
I must have been thinking too much es6 importy there


pandafulmanda:
also helps with weird lodash global webview bug: https://github.com/openstax/tutor-js/pull/1519

PR #1614, opened 2017-02-28
Update to use sentence case 'Log out'

nathanstitt:
Per: https://trello.com/c/Owr3K1BA/473-bug-log-in-out-and-sign-in-up-not-consistently-implemented-on-buttons-or-dropdowns-both-in-accounts-and-enrollment-flows#


jpslav:
Looks good.  Needs to be rebased @nathanstitt

PR #1616, opened 2017-03-02
make all images inside figures expand to fill the figure (100% wide)

nathanstitt:
I think on balance this is better than the tiny images we get sometimes.  

example at https://tutor-qa.openstax.org/books/236/section/1.4

screen shot 2017-03-02 at 12 18 00 pm

screen shot 2017-03-02 at 12 17 33 pm


Fredasaurus:
yay!

PR #1618, opened 2017-03-06
Fix scores report CC "finished" icon

nathanstitt:
Not for merge, is hot fix  see #1620 for fix

Sets a viewbox on the icon so it displays.  Without that it thinks it's 300px high for some reason

Based off production `409deeea8d8262fca2c55cd8de6fdc968439e19f`


nathanstitt:
Closing PR, hot fix code has been released

PR #1627, opened 2017-03-14
Fix trouble url location

pandafulmanda:
https://trello.com/c/yF8QUneX/574-bug-not-everything-is-being-sent-to-the-new-errata-tool

### DIFF
--- a/shared/src/components/exercise/index.cjsx
+++ b/shared/src/components/exercise/index.cjsx
@@ -104,7 +104,7 @@ ExerciseMixin =
     {parts} = @props
     step = _.last(parts)
 
-    related_content = step.related_content if related
+    related_content = step.related_content # if related
 
     if step.content?.uid
       <ExerciseIdentifierLink key='exercise-uid'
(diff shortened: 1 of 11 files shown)


pandafulmanda:
[Bug: Not everything is being sent to the new errata tool](https://trello.com/c/yF8QUneX/574-bug-not-everything-is-being-sent-to-the-new-errata-tool)

shared/src/components/exercise/index.cjsx
```diff
@@ -104,7 +104,7 @@ ExerciseMixin =
     {parts} = @props
     step = _.last(parts)
 
-    related_content = step.related_content if related
+    related_content = step.related_content # if related
```
nathanstitt: should that be uncommented?

pandafulmanda: im pretty sure this is no longer being used...this was with the multi-part questions to prevent the "comes from" from being after each part.

nathanstitt: coo, in that case can we remove the `#` :)  could also be: `{ related_content } = step;` but that's no biggy

pandafulmanda: okay nvm, unsure, will just undo for now

PR #1630, opened 2017-03-16
Remember watched and replay tours icon

nathanstitt:
A future PR will actually save the watched tour id to server once I finish the BE for it.  Also UX will be providing a different icon

screen shot 2017-03-16 at 1 50 11 pm


tutor/src/components/navbar/index.jsx
```diff
@@ -0,0 +1,66 @@
...
+              course={course}
+              onItemClick={this.collapseNav}
+            />
+          </Nav>
+        </Navbar.Collapse>
+        <ServerErrorMonitoring />
+      </Navbar>
+    );
+  }
+}
```
nathanstitt: This was a "got confused" conversion.  I converted it because I thought it'd make sense to put the replay icon inline on the bar itself, but as i went along I felt it'd be better to extract into the `navbar/tours-replay.jsx` below.  Since I'd already converted, I left it as there.

configs/webpack/base.coffee
```diff
@@ -26,19 +26,19 @@ STATICS =
   image: { test: /\.(png|jpg|svg|gif)/,    use: [ LOADERS.file ] }
   font:  { test: /\.(woff|woff2|eot|ttf)/, use: [ LOADERS.url  ] }
 
+extractText = ExtractTextPlugin.extract({
```
pandafulmanda: `extractText` doesnt seem to be in use...do we need this?

nathanstitt: heh, nope.  That was leftover from an attempt to share the plugin instance between the different loaders, which  didn't work

PR #1632, opened 2017-03-21
Training wheel updates

nathanstitt:
* Updates replay icon
* Sets colors to match approved mockup
* makes cards smaller
* fixes bug where replay icon would appear on all screens even if they didn't have any tours
* don't cancel tour when overlay is clicked
* Use 1/X format for buttons and move back to left of card
* add a new step to point to reading
* Fixes bug where tour views were saved multiple times
* Only enables tours when spy mode is engaged.

Also adds https://github.com/lodash/babel-plugin-lodash - i've been using stuff like `import { find, isEmpty } from 'lodash';` thinking that webpack would be smart enough not to include all of lodash but apparently it isn't (yet).  That module will convert those type imports into standard `lodash/find` 

screen shot 2017-03-21 at 1 31 51 pm


pandafulmanda:
The `Last` button is a little confusing when there is only one step.
screen shot 2017-03-23 at 11 39 58 pm

Also, when the main builder is hidden and I am in chapter selection mode, clicking replay brings back the initial tour for the builder:

screen shot 2017-03-23 at 11 40 15 pm
screen shot 2017-03-23 at 11 40 20 pm

pandafulmanda:
Unsure why these steps seem to want to alternate between top or bottom?  
screen shot 2017-03-23 at 11 46 58 pm
screen shot 2017-03-23 at 11 47 05 pm
screen shot 2017-03-23 at 11 47 08 pm

PR #1633, opened 2017-03-24
Trying wheel updates II

nathanstitt:
* Next button should read "Got it" for first 3 and last step
* Don't scroll indicator when pointing to fixed element
* Bugfix for joyride to position itself better
* make tour replay navbar icon smaller
* Per @Fredasaurus our light blue should be even lighter
* Adjust tour font colors and use font-awesome chevron for back

screen shot 2017-03-24 at 5 08 59 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "react-dnd": "2.2.3",
     "react-dnd-html5-backend": "2.2.3",
     "react-dom": "15.4.2",
-    "react-joyride": "github:nathanstitt/react-joyride#fix/unchanged-targets",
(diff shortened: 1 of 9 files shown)


pandafulmanda:
looking great overall, small notes:
* the replay toggle gets confused, seems like if i toggle the replay one page, when i change routes, the replay should reset
* whether the replay button should show needs to be reset on route change
  * as in, once im done watching a tour on one page, the replay button shows
  * then if i change pages to somewhere without any tours, the replay button still shows

PR #1635, opened 2017-03-27
Tour for Quick Look analytics & reading assignments

nathanstitt:
Also hides the LMS info popup for preview courses

screen shot 2017-03-27 at 5 13 04 pm

screen shot 2017-03-27 at 6 15 56 pm


pandafulmanda:
this one seems little odd when i know that this assignment is already published..but i guess it might make more sense in the context of a demo course

screen shot 2017-03-30 at 12 04 05 pm

PR #1647, opened 2017-04-05
BL FE updates:  task code to reload itself after each step

nathanstitt:
This is turning into a much bigger thing than thought it'd be :(


nathanstitt:
Yeah, so.. this is a bit much to bite off.  Closing in favor of the much simpler https://github.com/openstax/tutor-js/pull/1649

PR #1650, opened 2017-04-11
hacked together mathjax via npm

nathanstitt:
Probably not a good route to go down, but worth testing


shared/src/helpers/mathjax.js
```diff
@@ -0,0 +1,153 @@
...
+typesetDocument = debounce( typesetDocument, 100);
+
+// typesetMath is the main exported function.
+// It's called by components like HTML after they're rendered
+const typesetMath = function(root, windowImpl) {
+  // schedule a Mathjax pass if there is at least one [data-math] or <math> element present
+
+  if (windowImpl == null) { windowImpl = window; }
+
+  if ((__guard__(__guard__(windowImpl.MathJax, x1 => x1.Hub), x => x.Queue) != null) && root.querySelector(COMBINED_MATH_SELECTOR)) {
```
philschatz: Somebody used decaffeinate 😄

nathanstitt: hehe, yep the __guard__ is a big giveaway. I usually replace it with lodash's `get` but didn't bother since I think this approach is a dead-end.

philschatz:
Were there any problems you noticed with this method? 

I'm guessing the JS file size increased considerably (but chunking _might_ help if mathjax wasn't necessary for the homepage)

nathanstitt:
Yes, my thought's were that if this worked, i'd split it off into a separate webpack entry point and load it dynamically.

The biggest issue is that it attempts to load additional JS in response to different browsers and the type of math it encounters.    The only way to work around this would be to copy the mathjax directory into a place where it's accessibly from the web.

I think we may be able to figure out a build that works for our limited subset of browser and math types, but is probably better to self-host.

There's also an experimental build but they discourage it's use: https://github.com/mathjax/MathJax/wiki/Single-file-built

nathanstitt:
Closing in favor of using alternative CDN

PR #1651, opened 2017-04-12
Only allow teachers who have courses from adding a new one

nathanstitt:
Fixes loophole where `faculty_verified` instructors from elsewhere could wander into Tutor and create a course even though they aren't part of the beta.


gregfitch:
@nathanstitt Can the create new course for those who don't already have one be moved into spy mode? It'll mess with QA otherwise...

PR #1653, opened 2017-04-17
update license filename so GitHub recognizes it

philschatz:
See https://github.com/blog/1530-choosing-an-open-source-license .

It adds some useful indicators for folks (and ppl searching to contribute to AGPL projects) and is detectable via their API

image

image

and Notice GitHub detected it (link to the API call) https://api.github.com/repos/philschatz/test/license


nathanstitt:
Thanks @philschatz !

PR #1655, opened 2017-04-19
Style preview courses on course picker

nathanstitt:
screen shot 2017-04-19 at 9 59 35 am


pandafulmanda:
im really excited about the decision to split up models into ux vs non-ux based models

PR #1664, opened 2017-05-02
csrf token change on User model needs to be observed for change to update xhr options

pandafulmanda:
# Before

screen shot 2017-05-02 at 12 14 56 pm
screen shot 2017-05-02 at 12 13 19 pm

# After
screen shot 2017-05-02 at 12 16 38 pm

screen shot 2017-05-02 at 12 10 59 pm

https://trello.com/c/qxH2IZ8E/672-bug-actions-fail-with-a-422-in-firefox-but-they-works-on-chrome

Actually fails on both


nathanstitt:
Good catch @pandafulmanda .. sigh `crsf_token`

PR #1669, opened 2017-05-11
fix a few missing default exports on teacher metrics

nathanstitt:
Pretty sure I'd gripped for this before, but must have missed 'em.

I checked and this should be it though


nathanstitt:
and by "gripped" I mean grep 😁

PR #1670, opened 2017-05-11
Fix various

pandafulmanda:
- [x] breadcrumb warnings
- [x] links for readings need previews
- [x] homework breadcrumbs should not be "shy", the main nav should be instead
- [x] readings double tap on submit and next step bug

### DIFF
--- a/tutor/src/flux/task-step.coffee
+++ b/tutor/src/flux/task-step.coffee
@@ -62,6 +62,10 @@ TaskStepConfig =
 
   complete: (id) ->
     @_change(id, {is_completed: true})
+    @_save(id)
+
+  completed: (obj, id) ->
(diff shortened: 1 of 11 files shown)


tutor/src/flux/task-step.coffee
```diff
@@ -62,6 +62,10 @@ TaskStepConfig =
 
   complete: (id) ->
     @_change(id, {is_completed: true})
+    @_save(id)
+
+  completed: (obj, id) ->
+    @_asyncStatus[id] = STATES.LOADED
 
   setAnswerId: (id, answer_id) ->
```
nathanstitt: Nice catch 😁

nathanstitt:
tanks for fixing bugs i left laying around on the steps stuff

PR #1676, opened 2017-05-23
Updates for nag logic

nathanstitt:
Makes nag appearances be per-session not per-course and fixes bugs with the every second assignment one so that it doesn't briefly display and disappear

Also redirects to `myCourses` dashboard on preview create since that's the correct per UX.


tutor/src/components/course-calendar/index.jsx
```diff
@@ -38,11 +38,6 @@ const getDisplayBounds = {
   },
 };
 
-function addPlanToListing(plan) {
-  TaskPlans
-    .forCourseId(this.props.params.courseId)
-    .addPublishingPlan(plan);
-}
```
nathanstitt: I removed this because it's never called and has a bug.  `this.props.params.courseId` will be invalid, since `addPlanToListing` is a standalone function.

PR #1677, opened 2017-05-23
Style "MyCourses" page

nathanstitt:
* Sets the preview belt colors
 * Drop shadow for belt
 * Uniform heading color
 * All sections get bottom border

screen shot 2017-05-23 at 5 08 53 pm

### DIFF
--- a/tutor/resources/styles/components/my-courses.less
+++ b/tutor/resources/styles/components/my-courses.less
@@ -30,6 +30,12 @@
 }
 
 .my-courses {
+  .card-title() {
+    #fonts > .book-cover();
+    font-weight: 800;
+    font-size: 2rem;
+    line-height: 1.25em;
+  }
 
   .my-courses-current,
   .my-courses-preview,
@@ -176,6 +182,7 @@
         .name {
           margin: 0;
           padding: 1.5em 0.75em;
+          .card-title();
         }
         .preview-belt {
           position: absolute;
@@ -184,7 +191,7 @@
           height: 90px;
           padding: 0 0.75em;
           background-color: @tutor-neutral-lightest;
-          background: linear-gradient(140deg, @tutor-neutral-lightest, @tutor-neutral-lighter);
+          background: linear-gradient(140deg, @tutor-white, @tutor-neutral-lighter);
           box-shadow: 0px 4px 4px 0px rgba(0,0,0,0.2);
           display: flex;
           flex-direction: column;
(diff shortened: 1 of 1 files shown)


nathanstitt:
@Fredasaurus see screenshot ^^

Fredasaurus:
A few notes:
"Add a course" should be in bold,
Course name should be bold,
"Preview" should be in all caps,
Can we see the gradient?

nathanstitt:
@Fredasaurus gradient is:
linear-gradient(140deg, rgb(249, 249, 249); , rgb(241, 241, 241));

It is visible, it just appears different on the two different preview cards because they have different background colors, which causes an interesting optical illusion (for me at least).

I've made the other changes and updated screenshot

Fredasaurus:
- Can we make the gradient (140deg, rgb(255, 255, 255); , rgb(241, 241, 241));
- The preview course titles (College Biology and College Physics) should be the same size as the current course title and should be bold.

nathanstitt:
@Fredasaurus - good catch, the preview title's weren't bold, but the other's were.  Fixed.

I changed the start gradient to be white.  I can't tell any real difference, but maybe you can 😁 

A closeup of the new preview cards:
screen shot 2017-05-24 at 12 02 59 pm

nathanstitt:
@Fredasaurus Just realized that we didn't anticipate the space needed for the "Loading" animation.

As it is right now the belt isn't tall enough, and the "LOADING PREVIEW" get's pushed off the top.  If I increase the height from 90 to 110px it fits fine and the other cards also look ok to me.  What do you think?

screen shot 2017-05-24 at 2 56 38 pm

Fredasaurus:
OK. I am working on asking Dani if we can replace "openstax" with loading preview.

Fredasaurus:
no go on removing "openstax".
I'd rather not increase the height of the band to support a state that will only exist for a few seconds. It will crowd the card if the title has to go to two lines, so we have to leave room for that.
Dani suggested we use a different loader.

nathanstitt:
@Fredasaurus ok, we can either:

Shrink the loading animation (maybe text should also be smaller):
screen shot 2017-05-24 at 4 55 36 pm

Or place text and loader side-by side:
screen shot 2017-05-24 at 4 59 16 pm

Fredasaurus:
Cool! Let's reduce the font size of "loading preview" by a couple px and keep the same internal padding, I think 30px.

nathanstitt:
Excellent - 14px font gives us:

screen shot 2017-05-24 at 5 03 53 pm

Fredasaurus:
Looks like the padding on the card should be 20px.
screen shot 2017-05-24 at 5 03 43 pm

Fredasaurus:
Also looks like the font size on "check out...." should be a couple px higher.

pandafulmanda:
@Fredasaurus im going to merge this for now and we can fix it after since we've been bad about sitting down and reviewing

Fredasaurus:
👍

PR #1683, opened 2017-05-30
Display error modal when claiming preview fails

nathanstitt:
screen shot 2017-05-30 at 1 03 13 pm

### DIFF
--- a/shared/src/api/index.coffee
+++ b/shared/src/api/index.coffee
@@ -165,13 +165,16 @@ class APIHandlerBase
 
     onSuccess ?= DEFAULT_SUCCESS
     onFail ?= DEFAULT_FAIL
-    return new Promise (resolve) =>
+    return new Promise (resolve, reject) =>
+
       delay =>
         @_xhr.request(requestConfig)
+          .then(((resp) -> resolve(resp); resp), ((resp) -> reject(resp); Promise.reject(resp)))
+
           # if onFail doesn't Promise.reject anything, then the default fail will not fire.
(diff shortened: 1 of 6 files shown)


shared/src/api/index.coffee
```diff
@@ -165,13 +165,16 @@ class APIHandlerBase
 
     onSuccess ?= DEFAULT_SUCCESS
     onFail ?= DEFAULT_FAIL
-    return new Promise (resolve) =>
+    return new Promise (resolve, reject) =>
+
       delay =>
         @_xhr.request(requestConfig)
+          .then(((resp) -> resolve(resp); resp), ((resp) -> reject(resp); Promise.reject(resp)))
```
pandafulmanda: this is a pretty drastic change, does it still work if moved back as the last `then` with the `reject` added?  also, did you try using the `onFail` api setting ?

nathanstitt: Not really - the issue i was working around is that if it comes last then the functions only receive the results of the `.then` block before them, and if the final `rejection` doesn't bother calling `Promise.reject` then the error handler is never called.  That in turn caused the "loading" animation on the preview cards to spin forever. 

That's why I was careful to return the same thing that was passed to then `.then`  Of course there's still the danger that the handler could modify the `resp` and throw off the followup handlers.

nathanstitt: Oh and if you want to test it, mark all your courses as non-preview, then add a new teacher without any courses.

PR #1685, opened 2017-06-01
multiple GA tracker support

jpslav:
Adds support for multiple GA trackers.  Sends tracking commands to each tracker found under `window.ga`.  

The BE PR that adds multiple trackers is https://github.com/openstax/tutor-server/pull/1410.  This PR can be used with or without that PR.

### DIFF
--- a/tutor/specs/helpers/analytics.spec.js
+++ b/tutor/specs/helpers/analytics.spec.js
@@ -9,7 +9,7 @@ describe('Analytics', function() {
 
   beforeEach(function() {
     this.ga = jest.fn();
-    Analytics.setTracker(this.ga);
+    Analytics.setGa(this.ga);
     Courses.bootstrap([ COURSE ]);
     this.sendPageView = Analytics.sendPageView = jest.fn();
   });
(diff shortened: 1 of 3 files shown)


tutor/specs/helpers/analytics.spec.js
```diff
@@ -23,12 +23,52 @@ describe('Analytics', function() {
...
+    // Set up a mock GA function/object
+    function Tracker(name) {
+      this.name = name;
+      this.get = function(fieldName) {
+        if ('name' == fieldName) { return this.name; }
+      };
+    }
+
+    var mockGa = function() {
+      if (arguments.length == 0) { return; }
```
nathanstitt: not a big deal at all, but `var` usage is frowned on with modern js.   Typically you use `const` for variables that won't be re-assiged and `let` for when they will be.

Awesome work on figuring out the stubs and putting them to good use 😁

philschatz:
Just sayin there's https://github.com/philschatz/fetch-vcr . I can add an option for `XMLHttpRequest` if Google Analytics still uses that instead of `fetch`

jpslav:
Cool @philschatz. Agreed probably not needed here (esp since we're not even really caring here about the response and just want to know that we are making the right calls per the docs).  But in general I love VCR approaches :-D

jpslav:
@nathanstitt thanks re `var` `const` `let`.  I updated the PR.

tutor/specs/helpers/analytics.spec.js
jpslav: @nathanstitt thanks re `var` `const` `let`.  I updated the PR.

jpslav: Also re stubs shout out to @philschatz who gave me late night help on that :-D

PR #1694, opened 2017-06-06
Display terms in a modal before allowing access to course dashboards

nathanstitt:
Goes with https://github.com/openstax/tutor-server/pull/1412

screen shot 2017-06-07 at 1 20 32 pm


tutor/src/api/index.coffee
```diff
@@ -260,11 +266,13 @@ startAPI = ->
       end_at: endAt
   )
 
+
+
 start = (bootstrapData) ->
+  console.log bootstrapData
```
pandafulmanda: hehehe, 👀 !

nathanstitt: Sigh.. Glad you noticed 😊

PR #1709, opened 2017-06-13
quick fix for section chooser height

pandafulmanda:
probably not very nice to all browsers though

https://trello.com/c/POEvzixQ/638-change-request-show-show-problems-button-above-the-fold-for-assignment-builder-and-question-library-teachers-fail-to-find-it-and

screen shot 2017-06-13 at 6 17 20 pm
screen shot 2017-06-13 at 6 17 11 pm
screen shot 2017-06-13 at 6 16 57 pm

### DIFF
--- a/tutor/resources/styles/components/questions-dashboard.less
+++ b/tutor/resources/styles/components/questions-dashboard.less
@@ -58,6 +58,8 @@
   .sections-list {
     padding: @padding;
     background-color: @tutor-white;
+    overflow: auto;
+    height: ~"calc(100vh - 250px)";
   }
 
   .pinned-header {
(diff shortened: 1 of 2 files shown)


tutor/resources/styles/components/questions-dashboard.less
```diff
@@ -58,6 +58,8 @@
   .sections-list {
     padding: @padding;
     background-color: @tutor-white;
+    overflow: auto;
+    height: ~"calc(100vh - 250px)";
   }
```
nathanstitt: should this (and rule below) be max-height?   What happens if we have a very small book, should it still expand to have footer at bottom?

PR #1712, opened 2017-06-14
Fix tours additional

pandafulmanda:
https://trello.com/c/Z3EG4nZk/478-osw2onboarding04-training-wheels-as-openstax-we-want-instructors-and-students-to-experience-training-wheels-so-we-can-highlight

- [x] Welcome to your dashboard training wheel should be vertically centered.
- [x] the Browse the Book (as well as Question Library) training wheel should be on the left side so it does not obscure the profile menu.
- [x] selected status for dropdown links should be gray (per earlier Tutor behavior), not orange.
- [x] make super training wheel close button play next
- [x] make super training wheel Continue button able to be scrolled to when page is too short

screen shot 2017-06-14 at 1 48 06 pm
screen shot 2017-06-14 at 1 47 50 pm

### DIFF
--- a/tutor/src/components/tours/custom/centered-wheel.jsx
+++ b/tutor/src/components/tours/custom/centered-wheel.jsx
@@ -1,0 +1,46 @@
+import React        from 'react';
+import ReactDOM     from 'react-dom';
+import { Tooltip }  from 'react-joyride';
+
+import kebabCase    from 'lodash/kebabCase';
+import first        from 'lodash/first';
+import classnames   from 'classnames';
+
+export default class CenteredWheel extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    this.className = kebabCase(this.constructor.name);
+    this.wrapperClassName = this.props.className? `${first(this.props.className.split(' '))}-wrapper` : 'training-wheel-wrapper';
+  }
+
+  setupWrapperClasses() {
+    this.joyrideEl = ReactDOM.findDOMNode(this.props.step.ride.joyrideRef);
+    if (this.joyrideEl) {
+      this.joyrideEl.classList.add(`${this.className}-wrapper`, this.wrapperClassName);
+    }
+  }
+
+  componentDidMount() {
+    this.setupWrapperClasses();
+  }
+
+  componentDidUpdate() {
+    if (!this.joyrideEl) {
+      this.setupWrapperClasses();
+    }
+  }
+
+  componentWillUnmount() {
+    if (this.joyrideEl) {
+      this.joyrideEl.classList.remove(`${this.className}-wrapper`, this.wrapperClassName);
+    }
+  }
+
+  render () {
+    const className = classnames(this.className,  this.props.className);
+
+    return (
+      <Tooltip {...this.props} className={className}/>
+    );
+  }
+}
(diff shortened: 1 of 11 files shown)


tutor/src/components/tours/custom/centered-wheel.jsx
```diff
@@ -0,0 +1,46 @@
...
+import first        from 'lodash/first';
+import classnames   from 'classnames';
+
+export default class CenteredWheel extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    this.className = kebabCase(this.constructor.name);
+    this.wrapperClassName = this.props.className? `${first(this.props.className.split(' '))}-wrapper` : 'training-wheel-wrapper';
+  }
+
```
nathanstitt: I'm not seeing why className needs to be based off `this.constructor.name`.  That seems fragile since the constructor name may be changed when the code's uglified for production.  Why not just make `className` be a property, i.e.:

```javascript
export default class CenteredWheel extends React.PureComponent {
  className = 'centeredWheel'

  setupWrapperClasses() {

...
```



Rather than making `wrapperClassName` static, you could make a computed getter property; that way it'll update when/if props className changes:

```javascript
@computed get wrapperClassName() {
  return this.props.className? `${first(this.props.className.split(' '))}-wrapper` : 'training-wheel-wrapper';
}
```

tutor/src/components/tours/custom/common.jsx
```diff
@@ -1,6 +1,7 @@
 import React        from 'react';
 
 import classnames   from 'classnames';
+import forEach      from 'lodash/forEach';
```
nathanstitt: FYI we can do:  `import { forEach } from 'lodash'`  `babel-plugin-lodash` will translate that to be the same.  Is a bit cleaner when we have multiple imports.

tutor/src/components/tours/custom/common.jsx
```diff
@@ -20,9 +21,45 @@ function TutorBeta() {
...
+    if (el.className.indexOf('joyride-') === 0) {
+      forEach(handlers, (handler, name) => {
+        if (dataType === name) {
+          handled = true;
+
+          clickEvent.preventDefault();
+          clickEvent.stopPropagation();
+
+          return handler(clickEvent);
+        }
```
nathanstitt: Are we meaning to end the `forEach` loop early if `handler` returns false?   That's bitten me before, but I'm assuming that's intended since there's the explicit return.

tutor/src/tours/teacher-calendar.json
```diff
@@ -27,7 +27,8 @@
   "steps": [
     {
       "title": "Welcome to your dashboard!",
-      "body": "Think of your dashboard as base camp -- you can create and manage assignments, see student progress, and review the book."
+      "body": "Think of your dashboard as base camp -- you can create and manage assignments, see student progress, and review the book.",
+      "customComponent": "CenteredWheel"
     }, {
```
nathanstitt: a general observation - we need to add automated js linting to travis.  That'll catch things like this.  I haven't because there's too many failures atm.

tutor/src/components/tours/custom/common.jsx
pandafulmanda: hmm i had done this intentionally, but i guess it doesn't really need to return out, ill remove

PR #1715, opened 2017-06-14
just need to skip new or copy if there are no nonPreviews

pandafulmanda:
https://trello.com/c/dwR1Gsye/732-bug-copy-course-available-even-if-there-isnt-a-past-course

### DIFF
--- a/tutor/src/models/course/builder-ux.js
+++ b/tutor/src/models/course/builder-ux.js
@@ -178,7 +178,7 @@ export default class CourseBuilderUX extends BaseModel {
 
   // skips
   skip_new_or_copy() {
-    return !!this.source;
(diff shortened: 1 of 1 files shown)


tutor/src/models/course/builder-ux.js
```diff
@@ -178,7 +178,7 @@ export default class CourseBuilderUX extends BaseModel {
 
   // skips
   skip_new_or_copy() {
-    return !!this.source;
+    return !!this.source || !Courses.nonPreview.size;
   }
```
nathanstitt: FYI I also added `any`  and `isEmpty` boolean checks.  Could be `Courses.nonPreview.isEmpty`

pandafulmanda: ah cool!

PR #1716, opened 2017-06-15
Show more task spy info fields

Dantemss:


nathanstitt:
Great!  Thanks or updating the class to match what it actually does now

PR #1738, opened 2017-06-16
also need to skip if no courses altogether to clone from

pandafulmanda:
https://trello.com/c/dwR1Gsye/732-j29-bug-copy-course-available-even-if-there-isnt-a-past-course

### DIFF
--- a/tutor/src/models/course/builder-ux.js
+++ b/tutor/src/models/course/builder-ux.js
@@ -178,7 +178,7 @@ export default class CourseBuilderUX extends BaseModel {
 
   // skips
   skip_new_or_copy() {
-    return !!this.source || Courses.nonPreview.isEmpty;
+    return !!this.source || Courses.isEmpty || Courses.nonPreview.isEmpty;
   }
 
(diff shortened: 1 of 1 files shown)


tutor/src/models/course/builder-ux.js
```diff
@@ -178,7 +178,7 @@ export default class CourseBuilderUX extends BaseModel {
 
   // skips
   skip_new_or_copy() {
-    return !!this.source || Courses.nonPreview.isEmpty;
+    return !!this.source || Courses.isEmpty || Courses.nonPreview.isEmpty;
   }
```
nathanstitt: hm, wouldn't `nonPreview` be empty if courses also empty? But if this fixes the bug?

pandafulmanda: ah, but `Courses.isEmpty` could be `false` which would then check the last thing

nathanstitt: Oh yeah duh 😁

PR #1740, opened 2017-06-19
reset viewed tour state when spyMode is turned on

nathanstitt:


Dantemss:
Every time spy mode is turned on? My only concern is that maybe people who use spy mode will get confused.

nathanstitt:
@Dantemss yeah that's why I asked @kajalp and @gregfitch about it - they thought it'd be a good idea.  Of course they may change their minds but it's easy enough to remove if we want to

PR #1746, opened 2017-06-20
vertically align delete assignment button and move to right

nathanstitt:
Before
screen shot 2017-06-20 at 11 32 20 am

after:
screen shot 2017-06-20 at 11 31 33 am


nathanstitt:
[Bug: Enrollment URL does not work in Chrome and Safari with certain characters](https://trello.com/c/WuRZe2cJ/1459-bug-enrollment-url-does-not-work-in-chrome-and-safari-with-certain-characters)

PR #1764, opened 2017-06-21
Only show page tips menu option if there's manually triggered tours

nathanstitt:
Previously it showed if there were any tours, which would also include auto play ones

### DIFF
--- a/tutor/src/models/tour/context.js
+++ b/tutor/src/models/tour/context.js
@@ -117,12 +117,13 @@ export default class TourContext extends BaseModel {
   }
 
   @computed get needsPageTipsReminders() {
-    return this.hasElgibleTour && !find(this.elgibleTours, 'autoplay');
+    return this.hasTriggeredTour && !find(this.elgibleTours, 'autoplay');
   }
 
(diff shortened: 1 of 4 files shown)


tutor/src/models/user.js
```diff
@@ -58,7 +58,7 @@ export class User extends BaseModel {
   }
 
   resetTours() {
-    this.viewed_tour_ids.clear();
+    this.viewed_tour_stats.clear();
   }
```
nathanstitt: This is a bonus 🐛 fix I noticed

pandafulmanda: oh nice!

tutor/src/models/tour/context.js
```diff
@@ -131,7 +132,7 @@ export default class TourContext extends BaseModel {
 
   @action playTriggeredTours() {
     this.elgibleTours.forEach((tour) => {
-      if (!tour.auto_display){ tour.play(); }
+      if (!tour.autoplay){ tour.play(); }
```
pandafulmanda: hehe i was wondering about this

PR #1769, opened 2017-06-21
Copy tweaks to page tips tour

kjdav:
https://trello.com/c/HbML3S3w/773-j29-bug-copy-changes
Nathan or Amanda -- for this PR, can you check one thing? Did I actually add a break just before "And don't worry ..." ? I am not sure if I did that right 👎


nathanstitt:
Not quite, you have to use the newline escape value `\n`, which there's no reason you would know 

I went ahead and fixed it up, it looks like this now:

screen shot 2017-06-21 at 4 18 55 pm

PR #1781, opened 2017-06-30
Render assignment link popover even if shareable_url is missing

nathanstitt:
Some preview courses don't have this, but that's ok because the'll have a "NO URL IN PREVIEW" message
screen shot 2017-06-30 at 2 49 36 pm


tutor/src/components/task-plan/lms-info.jsx
```diff
@@ -137,7 +137,6 @@ export class LmsInfoLink extends React.PureComponent {
   }
 
   render() {
-    if (!this.getStats().shareable_url) { return null; }
```
nathanstitt: this is the bug - sometimes the BE doesn't generate a shareable_url for preview courses

PR #1787, opened 2017-07-06
Style student id modal; use student model

nathanstitt:
screen shot 2017-07-06 at 10 58 49 pm

screen shot 2017-07-06 at 10 55 54 pm


tutor/src/flux/student-id.coffee
```diff
@@ -1,53 +0,0 @@
...
-ERROR_MAP = {
-  student_identifier_has_already_been_taken:
-    'The provided student ID has already been used in this course. ' +
-    'Please try again or contact your instructor.'
-  blank_student_identifier:
-    'The student ID field cannot be left blank. Please enter your student ID.'
-  no_change:
-    'You have not changed the student ID.  Please enter your new student ID and try again.'
-}
-
```
nathanstitt: Note that these api errors are no longer emitted - the BE now allows students to set ID to anything, so we don't need any special handling.

PR #1790, opened 2017-07-10
always render RootCloseWrapper

nathanstitt:
Otherwise React will re-create the menu items on each render, causing the chat to not be enabled/disabled via the display:none style


nathanstitt:
will be related as part of hot fix #1795

PR #1791, opened 2017-07-11
Pay now or later on boarding modal

nathanstitt:
Needs BE PR https://github.com/openstax/tutor-server/pull/1437 to test payments

screen shot 2017-07-11 at 9 00 02 am

screen shot 2017-07-11 at 10 32 22 am


tutor/src/components/payments/no-or-later.jsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react';
```
pandafulmanda: is this `no-or-later` file being used? i couldn't find it being imported anywhere?

PR #1797, opened 2017-07-12
Disable menu highlighting

RoyEJohnson:


RoyEJohnson:
[Bug: Tutor profile dropdown menu always shows Dashboard as selected](https://trello.com/c/vzvksAMN/787-bug-tutor-profile-dropdown-menu-always-shows-dashboard-as-selected)

nathanstitt:
Bah, I'd really hoped that we had code coverage for the highlighting and this would fail Travis.  Oh well, we should add some specs when we re-enable the highlighting.

PR #1800, opened 2017-07-13
Manage Payments

nathanstitt:
currently using stubbed data

screen shot 2017-07-13 at 9 46 22 am

### DIFF
--- a/tutor/src/routes.coffee
+++ b/tutor/src/routes.coffee
@@ -36,6 +36,10 @@ getExternalShell = ->
   {ExternalShell} = require './components/task-plan'
   ExternalShell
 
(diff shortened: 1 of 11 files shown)


tutor/src/routes.coffee
```diff
@@ -154,6 +157,7 @@ ROUTES = [
   {
     pattern: '/student-preview/:courseId?', name: 'studentPreview',       renderer: getStudentPreview
   }
+  { pattern: 'payments',                name: 'managePayments',           renderer: getPaymentsShell     }
```
pandafulmanda: small thing for later, maybe this should be prefixed with a `/` like the other top level routes?

PR #1801, opened 2017-07-13
fix: unbold "What do students see?" button

RoyEJohnson:


RoyEJohnson:
[J05 Bug: student preview videos button text should be regular not bold](https://trello.com/c/yFGqlSYl/828-j05-bug-student-preview-videos-button-text-should-be-regular-not-bold)

RoyEJohnson:
screen shot 2017-07-13 at 4 52 04 pm

tutor/resources/styles/components/task-plan/controls.less
```diff
@@ -8,6 +8,11 @@
     font-weight: 600;
     border-radius: 3px;
 
+    &.preview-btn {
+        border-radius: 4px;
+        font-weight: normal;
```
pandafulmanda: there are 4 spaces here instead of the 2 we usually use.  otherwise looks good.

although i wonder whether in this case, if we should add classes to the other buttons that the `font-weight: 600;` since those are kinda the weird buttons compared to our normal button styles, should apply to, but that is more code change...oh hmm...are those other builder footer buttons supposed to be bolded in the first place @Fredasaurus ?

pandafulmanda: hmmm the mock up has them bolded.  hmm nvm, dont mind me.  @nathanstitt, does this look good to you?  im ready to merge if you are.

RoyEJohnson: Updated the PR to two-space-ify.

PR #1807, opened 2017-07-17
Fix nav styling when logged in

RoyEJohnson:


RoyEJohnson:
[Bug: Global nav visual design specs not matching mockup, inconsistent across logged in/out states](https://trello.com/c/J7ZRcayx/835-bug-global-nav-visual-design-specs-not-matching-mockup-inconsistent-across-logged-in-out-states)

tutor/resources/styles/components/top-nav-bar.less
```diff
@@ -37,9 +37,18 @@
...
+
   a {
     color: #777;
     text-decoration: none;
+
+    &:hover,
+    &[aria-expanded="true"] {
+      color: #333;
+    }
   }
```
nathanstitt: I think the dropdown rules could be moved to the `dropdown-menu` file that's included at the bottom, that way all menu style rules are consolidated.

tutor/specs/components/navbar/__snapshots__/index.spec.jsx.snap
```diff
@@ -93,7 +93,10 @@ exports[`Main navbar renders and matches snapshot 1`] = `
         />
         <span
           title="Menu and settings"
-        />
+        >
+          Hi 
+          
+        </span>
          
         <span
```
nathanstitt: Glad you figured out how to update the snapshots - I just realized I'd never mentioned them to you 😁

tutor/src/components/navbar/course-name.jsx
```diff
@@ -1,24 +0,0 @@
...
-    const { course } = this.props;
-    if (!course) { return null; }
-
-    return (
-      <TutorLink to="dashboard" params={{ courseId: course.id }} className="course-name">
-        {course.name}
-      </TutorLink>
-    );
-  }
-}
```
nathanstitt: Hm, I wasn't aware that UX wanted us to remove the course name altogether, I thought we were only supposed to hide it on narrow screens.  @Fredasaurus can you clarify?

tutor/src/components/navbar/user-actions-menu.jsx
```diff
@@ -100,6 +100,8 @@ export default class UserActionsMenu extends React.PureComponent {
 
     const course = Courses.get(this.props.courseId);
     const isConceptCoach = get(course, 'is_concept_coach', false);
+    const firstName = User.name ? User.name.replace(/ .*/, '') : '';
+
```
nathanstitt: How about we add a method to User for getting the first name.  Probably something like:

```js
@computed get firstName() {
  return this.first_name || User.name.replace(/ .*/, '') || '';
}
```

Then we can just use `User.firstName` from the jsx below.

nathanstitt: fyi - user model file is in `tutor/src/models/user.js`

tutor/src/components/navbar/course-name.jsx
nathanstitt: also, if we do remove the course name then we can no longer click it to go to the dashboard for a course.

PR #1808, opened 2017-07-17
Manage payments

nathanstitt:
Refund is waiting on BE support - coming in https://github.com/openstax/tutor-server/pull/1437

screen shot 2017-07-17 at 3 28 26 pm

### DIFF
--- a/tutor/src/components/payments/manage.jsx
+++ b/tutor/src/components/payments/manage.jsx
@@ -1,14 +1,47 @@
 import React from 'react';
-import Router from '../../helpers/router';
 import { observer } from 'mobx-react';
-import { Grid, Table } from 'react-bootstrap';
+import { Grid, Table, Button } from 'react-bootstrap';
+import moment from 'moment';
+import Router from '../../helpers/router';
 import BackButton from '../buttons/back-button';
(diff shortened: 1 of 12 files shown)


tutor/src/components/payments/manage.jsx
```diff
@@ -53,6 +123,12 @@ export default class ManagePayments extends React.PureComponent {
           <BackButton fallbackLink={this.backLink} />
         </header>
         {Purchases.hasApiRequestPending ? <OXFancyLoader isLoading /> : this.renderTable()}
+        <div className="refund-policy">
+          <NewTabLink to="http://openstax.force.com/support/articles/FAQ/OpenStax-Tutor-Student-Refund-Policy/?q=refund&l=en_US&c=Products%3ATutor&fs=Search&pn=1">
```
pandafulmanda: i kinda think this, along with some other rando links that we have like this, would be nice to pull out into a constant for easy updates later on, but not a big deal for now

nathanstitt: hm, yeah also would be nice if they came from the BE, that way we could have different one's for -dev vs production

PR #1810, opened 2017-07-18
Fix task-step nav-arrow focus weirdness

RoyEJohnson:


RoyEJohnson:
[Bug: Left/Right Nav Arrows are centered oddly on short task steps.](https://trello.com/c/9WDE2pXU/836-bug-left-right-nav-arrows-are-centered-oddly-on-short-task-steps)

nathanstitt:
@RoyEJohnson Hm this seems to have broken on reference book - I didn't test other areas but I suspect they may also have issues.  What the paging on the side are supposed to completely fill the area to the left/right with the arrows centered vertically but no more than a certain distance from the middle.  To accomplish that I'm pretty sure the wrapper elements are needed.

I believe bug isn't about the position of the arrows though, it's only about the blue border around them when they have `:focus`  What we want is for that blue border to go away and some sort of style to be applied only to the arrow.

If we do need to alter the dom & positioning, then we'll have to test carefully on at least the following screens that use the component:  Readings, Reference book, Teacher assignment build, & question library.

screen shot 2017-07-18 at 3 00 33 pm

PR #1817, opened 2017-07-21
Update heading structure on student dashboard

RoyEJohnson:
Updating the structural elements of the page while keeping the appearance and behavior the same.

### DIFF
--- a/tutor/specs/components/scores/__snapshots__/index.spec.cjsx.snap
+++ b/tutor/specs/components/scores/__snapshots__/index.spec.cjsx.snap
@@ -39,163 +39,168 @@ exports[`Scores Report renders and matches snapshot 1`] = `
     >
       <ul
         className="nav nav-tabs"
-        role="tablist"
       >
         <li
           className="active"
-          role="tab"
-          tabIndex={-1}
         >
-          <h2
-            aria-selected="true"
+          <a
+            aria-selected={true}
+            href="#"
             onClick={[Function]}
           >
-            <div
-              className=""
-            >
-              <span
-                aria-describedby="course-periods-nav-tab-0"
-                className="tab-item-period-name"
-                onBlur={[Function]}
-                onClick={null}
-                onFocus={[Function]}
-                onMouseOut={[Function]}
-                onMouseOver={[Function]}
+            <h2>
+              <div
+                className=""
               >
-                1st
-              </span>
-            </div>
-          </h2>
+                <span
+                  aria-describedby="course-periods-nav-tab-0"
+                  className="tab-item-period-name"
+                  onBlur={[Function]}
+                  onClick={null}
+                  onFocus={[Function]}
+                  onMouseOut={[Function]}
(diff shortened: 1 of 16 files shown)


RoyEJohnson:
[Bug: Update heading list on student dashboard](https://trello.com/c/ltlaNy5b/853-bug-update-heading-list-on-student-dashboard)

tutor/specs/components/scores/__snapshots__/index.spec.cjsx.snap
```diff
@@ -67,18 +65,16 @@ exports[`Scores Report renders and matches snapshot 1`] = `
...
         </li>
         <li
           className=""
-          tabIndex={1}
+          role="tab"
+          tabIndex={0}
         >
-          <a
+          <h2
             aria-selected={undefined}
```
nathanstitt: Hm is it acceptable for accessibility to have a clickable `h2` element?  If it is should we add `aria-role=button` ?

tutor/src/components/tabs.cjsx
```diff
@@ -75,15 +75,12 @@ Tabs = React.createClass
...
+    <li key={index} tabIndex={if isSelected then -1 else 0} className={classnames(active: isSelected)} role="tab">
+      <h2 onClick={_.partial(@onTabClick, index)}
         aria-selected={'true' if isSelected}
       >
         {tab}
-      </a>
+      </h2>
     </li>
 
   render: ->
```
nathanstitt: The `Tabs` component is used on several other areas, `course-settings/roster`,  and `course-periods` which is in turned used from `cc-dashboard`, `performance-forecast` and `plan-stats`

Is UX ok with all of them getting much larger?

philschatz: /cc @Fredasaurus @rvnewell  😄

RoyEJohnson: The styling in `tabs.less` is adjusted so that `.nav-tabs li h2` is styled the same as `.nav-tabs li a`. You should not see anything get bigger.

tutor/specs/components/scores/__snapshots__/index.spec.cjsx.snap
RoyEJohnson: The `h2` is inside a `role="tab"`.

tutor/src/components/tabs.cjsx
nathanstitt: ok, in that case the only sticking point is whether it's a good idea to have `h2` be a clickable element accessibility wise.  What's the rational for changing them?  I know h2 has better SEO, but we're not worried about that here 😁

RoyEJohnson: The `tab` is the clickable element. The `h2` is part of the document structure: the panel is a section of information, and the `h2` is its header, but it is placed inside the tab.

nathanstitt:
Nice! No more double border on tabs!

PR #1818, opened 2017-07-21
display "trial" messaging when payments are disabled

nathanstitt:
This is the fallback plan if testing reveals unfixable bad bugs

screen shot 2017-07-21 at 12 45 17 pm

screen shot 2017-07-21 at 5 02 34 pm

### DIFF
--- a/tutor/src/models/course/create.js
+++ b/tutor/src/models/course/create.js
@@ -2,7 +2,7 @@ import {
   BaseModel, identifiedBy, field, belongsTo, computed,
 } from '../base';
 import { observable } from 'mobx';
-import { extend, omit, filter } from 'lodash';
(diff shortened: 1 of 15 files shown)


tutor/src/models/course/create.js
```diff
@@ -15,7 +15,7 @@ export default class CourseCreate extends BaseModel {
   @field num_sections = 1;
   @field estimated_student_count;
   @field is_preview = false;
-  @field is_college = false;
+  @field is_college = true;
   @field time_zone = 'Central Time (US & Canada)';
```
nathanstitt: This is a bug fix i snuck in - @Dantemss noticed we were setting all courses to not be college which is a despicable lie

PR #1819, opened 2017-07-24
Update global nav

RoyEJohnson:


RoyEJohnson:
[Update global nav](https://trello.com/c/SjaZXImZ/690-update-global-nav)

RoyEJohnson:
[Bug: Firefox "Log out" is cut off a bit](https://trello.com/c/ahreSd9L/868-bug-firefox-log-out-is-cut-off-a-bit)

RoyEJohnson:
[Bug: If course name is too long, user is no longer able to access the header bar with "Help", "Student Preview Videos", and User menu](https://trello.com/c/II6yGSqN/891-bug-if-course-name-is-too-long-user-is-no-longer-able-to-access-the-header-bar-with-help-student-preview-videos-and-user-menu)

RoyEJohnson:
[User menu width doesn't match name width](https://trello.com/c/1gvouJuu/895-user-menu-width-doesnt-match-name-width)

tutor/resources/styles/components/top-nav-bar.less
```diff
@@ -21,49 +52,45 @@
...
+
   .ui-brand-logo {
-    margin-top: 3px; // the text on the logo is a bit vertically mis-aligned
     display: inline-block;
-    height: 41px;
-    width: 265px;
+    margin-top: 12px;
+    height: 34px;
+    width: 200px;
     background-repeat: no-repeat;
```
nathanstitt: The Tutor logo is just a bit high vertically.  I think that's supposed to be vertically centered?
screen shot 2017-07-31 at 3 43 39 pm
Rather than rely on the margin to center vertically, you could remove the `margin-top` from `.ui-brand-logo`, and set `.left-side-controls` to flex, and `.brand`  to ` display: flex; align-items: center;`

tutor/resources/styles/components/top-nav-bar.less
```diff
@@ -1,12 +1,43 @@
...
+  color: @openstax-gray;
+  position: fixed;
+  top: 0;
+  right: 0;
+  left: 0;
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+  height: @tutor-navbar-height;
+  padding-right: @navbar-padding-horizontal;
```
nathanstitt: I think we it'd be better to use `min-height` instead of `height`.  Using `height` won't allow it to grow just in case we end up with too many large items on it and it needs to expand.

tutor/resources/styles/components/top-nav-bar.less
```diff
@@ -1,12 +1,43 @@
...
+  padding-left:  @navbar-padding-horizontal;
+  z-index: @zindex-navbar-fixed;
+
+  .tutor-nav-controls {
+    display: flex;
+    justify-content: space-between;
+    margin: 0 15px;
+    width: 1200px;
+  }
```
nathanstitt: Nice idea to have the navbar not expand infinitely.  I think the `1200px` is just a bit too large though, unless maybe the intent is to have it wider than the content?

screen shot 2017-07-31 at 3 53 11 pm
screen shot 2017-07-31 at 3 51 53 pm

nathanstitt:
Looks pretty good @RoyEJohnson but needs rebased and also there's a visual defect with the "center controls", which is used by readings to show student info.  I poked at it using dev tools and looks like the `navbar-overlay` needs:

```
    display: flex;
    align-items: center;
```

screen shot 2017-07-31 at 3 38 03 pm

The top of menu's also are slightly too high, and eat into the navbar:
screen shot 2017-07-31 at 3 40 35 pm

They have a `    margin-top: -4px;` which doesn't seem needed to me.

tutor/resources/styles/components/top-nav-bar.less
RoyEJohnson: @Fredasaurus' specs for the nav are here:
https://sketchviewer.com/sketches/59766aabb57e8900114c89ce/latest/

He specifies margin-top, etc. Centering makes sense to me, but I can't tell whether that was his intent.

tutor/resources/styles/components/top-nav-bar.less
RoyEJohnson: 1170px content width is also in the spec. 1200px with 15px gutters is a standard format we're using on OSweb for general content. We might be looking at widening the rest of the content.

Fredasaurus: Correct. Menu content should be 1170px.

tutor/resources/styles/components/top-nav-bar.less
Fredasaurus: Don't have a preference on the method used, so long as the logo and menu items are aligned to the same baseline.

nathanstitt: hm, yeah I'd say that's very close to baseline then

tutor/resources/styles/components/top-nav-bar.less
nathanstitt: In that case we probably need to make our content a bit wider to be 1170 as well, but that can be debated later 😁 

Thanks @RoyEJohnson & @Fredasaurus

nathanstitt:
Looks good @RoyEJohnson other than I'm still not sure if the "center controls" that appear on readings should be raised like they are? 

screen shot 2017-07-31 at 5 42 27 pm

RoyEJohnson:
Center controls shouldn't be raised; that should have been fixed along with the rebase. I'll double-check that this morning.

RoyEJohnson:
I don't get the raised center controls. Here are Safari, Firefox, and Chrome.
screen shot 2017-08-01 at 9 53 23 am

nathanstitt:
Durp, looks like my git wasn't clean so it failed to pull in your updated commits.  Looks great now. Sorry!

PR #1836, opened 2017-07-27
Display "Trial Activated" message after "try later" is clicked

nathanstitt:
screen shot 2017-07-27 at 11 01 10 am

### DIFF
--- a/tutor/src/components/onboarding/free-trial-activated.jsx
+++ b/tutor/src/components/onboarding/free-trial-activated.jsx
@@ -1,0 +1,43 @@
+import React from 'react';
+import { Button } from 'react-bootstrap';
+import { observer, PropTypes as MobxPropTypes } from 'mobx-react';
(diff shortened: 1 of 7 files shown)


tutor/src/components/onboarding/free-trial-activated.jsx
```diff
@@ -0,0 +1,43 @@
...
+import CourseUX from '../../models/course/ux';
+
+@observer
+export default class FreeTrialActivated extends React.PureComponent {
+
+  static propTypes = {
+    ux: MobxPropTypes.observableObject.isRequired,
+  }
+
+  static className = 'free-trial-activated';
```
pandafulmanda: not big deal but why have this and not use on line 20?

nathanstitt: yeah, that's because the on boarding components are wrapped inside the on boarding modal.  Some of them need to set the class on the outer modal, so the on boarding reads the static prop here so it can set it.

PR #1845, opened 2017-07-31
display new "Individual Review" card

nathanstitt:
https://trello.com/c/mNpmDvT9/663-t31801b-new-spaced-practice-interstitial-card-rationalize-spaced-practice-as-tutor-we-want-to-make-spaced-practice-work-in-a-way

screen shot 2017-07-31 at 2 55 24 pm

screen shot 2017-08-01 at 4 10 15 pm


shared/src/helpers/task.coffee
```diff
@@ -136,9 +137,19 @@ befores[SPACED_PRACTICE_GROUP] = (task, step, stepIndex, isAvailable) ->
       arguments...
     )
 
+isPersonalized = (task, step, stepIndex) ->
+  _.findWhere(task.steps, {group: PERSONALIZED_GROUP})?.id is step.id
+
+befores[INDIVIDUAL_REVIEW] = (task, step, stepIndex, isAvailable) ->
+  return if isPractice(task)
```
pandafulmanda: based on the card, it sounds like this card should show up once in every homework and reading.  As is, it will only show up once for all assignments.  for it to show up once in every reading for example, it would need to return a `step` object based on the proper conditions, like [here](https://github.com/openstax/tutor-js/pull/1845/files#diff-2fe204e9f263bca1bffc1b0895b5ba14R131) for example

PR #1853, opened 2017-08-01
Fix: focus on free response field

RoyEJohnson:
Code to focus on free-response field on update was being skipped.

### DIFF
--- a/shared/src/components/exercise/mode.cjsx
+++ b/shared/src/components/exercise/mode.cjsx
@@ -31,7 +31,7 @@ ExMode = React.createClass
 
   componentDidUpdate: (prevProps) ->
     {mode, focus} = prevProps
-    @setFreeResponseFocusState() if mode is 'free-response' and focus isnt @props.focus
+    @setFreeResponseFocusState() if mode is 'free-response'
 
   componentWillReceiveProps: (nextProps) ->
     {free_response, answer_id, cachedFreeResponse} = nextProps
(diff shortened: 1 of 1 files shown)


RoyEJohnson:
[Bug: Focus not on free-response field](https://trello.com/c/ALfkdoV7/848-bug-focus-not-on-free-response-field)

nathanstitt:
@pandafulmanda - do you recall why we had that check in the code? I know we fought with the focus on multi-part questions zooming back and forth from top to bottom repeatedly.  Was that part of that fix do you think?

shared/src/components/exercise/mode.cjsx
```diff
@@ -31,7 +31,7 @@ ExMode = React.createClass
 
   componentDidUpdate: (prevProps) ->
     {mode, focus} = prevProps
-    @setFreeResponseFocusState() if mode is 'free-response' and focus isnt @props.focus
+    @setFreeResponseFocusState() if mode is 'free-response'
```
pandafulmanda: the way this is set up right now, any type of update will scroll to the top, even if im typing in the textarea.  is it that on mount, we need to scroll to the top?  I'm not sure i understand what the card is asking for

pandafulmanda: Seems to me, the free-response focuses fine without removing the ` and focus isnt @props.focus` clause?

pandafulmanda: also, when scrolling, should it scroll to the textarea as opposed to the top?

pandafulmanda: ooh i see, it's not changing when the question updates, you could test props for question changes

RoyEJohnson: Does `setFreeResponseFocusState` cause scrolling?
The intent of the card is that the free reponse area get the focus when the question loads. This works in the `componentDidMount`, which is to say the first time a free-response field draws, but if you go through more questions and come to another one, it does not get focus.

shared/src/components/exercise/mode.cjsx
```diff
@@ -47,11 +47,12 @@ ExMode = React.createClass
...
     el = ReactDOM.findDOMNode(@refs.freeResponse)
-    if focus
-      el.focus?()
-      window.scrollTo(0, 0);
-    else
-      el.blur?()
+    if el
+      if focus
+        el.focus?()
+        window.scrollTo(0, 0);
```
RoyEJohnson: Ahh, the scrolling.

PR #1874, opened 2017-08-10
Reload student tasks, update student dashboard

nathanstitt:
Goes with: https://github.com/openstax/tutor-server/pull/1508

Starts a slow polling (every 30 seconds) for assignments if an enrollment indicates that the destination course has assignment present.

Changes the "No upcoming events" to be "No upcoming assignments", which will display "Loading upcoming assignments" while the polling is on-going.

Also converts the student-dashboard flux store to a map/model and thenstudent-dashboard components to jsx which grew and grew 😢 

screen shot 2017-08-10 at 6 05 46 pm


pandafulmanda:
something to consider, we do have some [generic flux stuffs](https://github.com/openstax/tutor-js/blob/master/tutor/src/flux/job.coffee) for polling for performance exports status and [assignment publishing status](https://github.com/openstax/tutor-js/blob/master/tutor/src/flux/plan-publish.coffee).  would be nice to combine the logic in `pollForUpdates` with that in the future, but that would likely require the jobs base flux to be converted to mobx.

PR #1877, opened 2017-08-11
Adjust modal styles so that navbar is usuable

nathanstitt:
Not the best ux but is better than it being visible-but-unclickable. Before it's z-index was placing it over the navbar

screen shot 2017-08-11 at 9 09 15 am


Fredasaurus:
It'll do for now. In the future we should do one of two things:
1) make this a real page and keep all app functionality,
2) make this a real modal and remove all external functionality.

PR #1888, opened 2017-08-15
Convert button to anchor tag for screenreader compatibility

nathanstitt:
No UI changes

### DIFF
--- a/shared/src/components/exercise/intro.cjsx
+++ b/shared/src/components/exercise/intro.cjsx
@@ -35,7 +35,7 @@ ExerciseIntro =  React.createClass
         <span>{TITLES[stepGroup]}</span>
       </h1>
       {getIntroText[stepGroup](project)}
-      <BS.Button onClick={onContinue}>Continue</BS.Button>
+      <a className="btn continue" onClick={onContinue}>Continue</a>
     </CardBody>
 
 module.exports = ExerciseIntro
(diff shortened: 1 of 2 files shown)


shared/src/components/exercise/intro.cjsx
```diff
@@ -35,7 +35,7 @@ ExerciseIntro =  React.createClass
         <span>{TITLES[stepGroup]}</span>
       </h1>
       {getIntroText[stepGroup](project)}
-      <BS.Button onClick={onContinue}>Continue</BS.Button>
+      <a className="btn continue" onClick={onContinue}>Continue</a>
```
pandafulmanda: should we use https://react-bootstrap.github.io/components.html#buttons-tags instead here? i guess maybe it doesnt matter except more consistent and also the role="button" attribute on the resulting html

nathanstitt: hm, i seem to recall there was **reasons** for not, but let me test

nathanstitt: Ah, I recall now - when you use`BS.Button`, the only way to get a `a` tag is to add an `href`, but then the `onCancel` doesn't call `.prefentDefault` so that jumps to the top of page when clicked.

Rather than adding custom handlers and stuff, I figured it was simplest to just use `a` since that's what QA wants.

pandafulmanda: okay

PR #1892, opened 2017-08-17
block teacher from enrolling as student

nathanstitt:
Needs BE PR https://github.com/openstax/tutor-server/pull/1516

screen shot 2017-08-17 at 4 52 44 pm


tutor/specs/components/__snapshots__/enroll.spec.jsx.snap
```diff
@@ -1,5 +1,67 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Student Enrollment blocks teavher enrollment 1`] = `
```
Dantemss: `teacher` but maybe doesn't matter

PR #1899, opened 2017-08-31
Delay re-showing onboarding nag for 1 week after first view

nathanstitt:
This prevents it from popping up every time when "I don't know" option is selected

### DIFF
--- a/tutor/src/models/course/onboarding/full-course.js
+++ b/tutor/src/models/course/onboarding/full-course.js
@@ -1,13 +1,17 @@
 import {
   computed, observable,
 } from 'mobx';
-import { includes } from 'lodash';
+import { includes, isNil } from 'lodash';
 
 import BaseOnboarding from './base';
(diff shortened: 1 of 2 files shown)


tutor/src/models/course/onboarding/full-course.js
```diff
@@ -38,18 +42,28 @@ export default class FullCourseOnboarding extends BaseOnboarding {
   }
 
   @computed get displayInitialPrompt() {
-    return this.response === false && (
-      this.courseIsNaggable && this.isOnboardingUndecided
+    return Boolean(
+      this.response === false && this.courseIsNaggable && isNil(this.lastNaggedAgo)
```
pandafulmanda: seems like there should be something with `isOnboardingUndecided` here?  like
```
this.response === false && this.courseIsNaggable && (isNil(this.lastNaggedAgo) || this.isOnboardingUndecided)
``
?

PR #1905, opened 2017-09-11
Update react-router to released version

nathanstitt:


shared/src/helpers/match-by-router.cjsx
```diff
@@ -16,15 +17,16 @@ matchProps = (Router, props, parentParams) ->
...
-    <span>
+    <Switch>
       {for route, i in props.routes
-        <Match key={i} {...matchProps(Router, route, props.params)} />}
-      <Miss component={InvalidPage} />
-    </span>
+        <Route key={i} {...matchProps(Router, route, props)} />}
+      <Route component={InvalidPage} />
+    </Switch>
   match.displayName = displayName
```
nathanstitt: new router doesn't have `Miss` instead you just use a route without a `path` which will match anything.

PR #1919, opened 2017-10-06
Api access class

nathanstitt:


shared/src/api/action-adapter.coffee
```diff
@@ -139,19 +140,21 @@ connectModelAction = (action, apiHandler, klass, method, options) ->
...
-    this.apiRequestsInProgress?.set(action, requestConfig)
+
+    this.api?.requestsInProgress.set(action, requestConfig)
     apiHandler.send(requestConfig, perRequestOptions, firstArg).then((reply) =>
-      this.apiRequestsInProgress?.delete(action)
+      this.api?.requestCounts[action] += 1
+      this.api?.requestsInProgress.delete(action)
       reply
-    ).catch( => this.apiRequestsInProgress?.delete(action))
+    ).catch( => this.apiRequestsInProgress.delete(action))
```
pandafulmanda: seems like this should also be: `this.api?.requestsInProgress.delete(action)` when the XHR fails?

pandafulmanda: Also, would of these changes need to be adapted for `tutor/src/models/map.js` as well?

nathanstitt: ah, yep.  I also missed some instances of `apiErrors`.  Fixing them now

PR #1938, opened 2017-10-20
Fix install issues

pandafulmanda:
`node-inspector` is not fully compatible with all versions of node

updates node and npm to versions we are using now


nathanstitt:
Sounds good to me!

PR #1944, opened 2017-10-24
Update Class Performance (i) copy

kjdav:
Per mockup: http://4tk3oi.axshare.com/tutor_scores_paired_new_tabs_toolbar_dashboard_sty_2.html#choose_settings_tab=lms&CSUM=1


kjdav:
@nathanstitt as you're working on this page, note that I need you to make another change based on this mockup: http://4tk3oi.axshare.com/tutor_scores_paired_new_tabs_toolbar_dashboard_sty_2.html#choose_settings_tab=lms&CSUM=1 

1. The cell called "Overall" needs to be changed to "Course average."
2. We need to add an (i) next to "Course average" that has the copy in the mockup. 

Since there isn't already an (i) next to Overall, I wil leave this to you if that's okay :-)

PR #1948, opened 2017-10-26
Fix milestones

nathanstitt:
Looks like this was broken by the router upgrade.  It now returns only the current route, not the entire tree


nathanstitt:
[Bug: Reading has milestones button missing (on -qa)](https://trello.com/c/cz9vUf13/1047-bug-reading-has-milestones-button-missing-on-qa)

PR #1951, opened 2017-10-27
Fix no periods panel's add button

nathanstitt:


nathanstitt:
[Bug: "Add a Period" and User Menu buttons broken](https://trello.com/c/8tKBm7er/1049-bug-add-a-period-and-user-menu-buttons-broken)

PR #1953, opened 2017-10-30
Display partial failures when sending scores to LMS

nathanstitt:
screen shot 2017-10-31 at 11 47 08 am


nathanstitt:
[T3.03.09 Average score sync - As an instructor, I want to push the final Tutor score to the LMS gradebook.](https://trello.com/c/N8RaMmHW/944-t30309-average-score-sync-as-an-instructor-i-want-to-push-the-final-tutor-score-to-the-lms-gradebook)

PR #1958, opened 2017-11-06
Exercises fix: only add new format if it's not already sent

nathanstitt:
fixes things like this:

screen shot 2017-11-06 at 12 13 49 pm


Dantemss:
Looks good to me

PR #1967, opened 2017-11-14
Fix task plan stats breaking when they're missing for a period

nathanstitt:


tutor/src/components/plan-stats/index.jsx
```diff
@@ -45,7 +45,7 @@ export default class Stats extends React.PureComponent {
   }
 
   @computed get stats() {
-    return find(this.analytics.stats, { period_id: this.period.id });
+    return find(this.analytics.stats, { period_id: this.period.id }) || {};
   }
```
nathanstitt: this is the actual bug fix - the rest is dealing with NaN's and tests

PR #1978, opened 2017-11-21
Add toggle control for viewing annotation summary to navbar 

nathanstitt:
screen shot 2017-11-21 at 5 16 20 pm
screen shot 2017-11-21 at 5 14 42 pm

### DIFF
--- a/tutor/resources/styles/components/reference-book/navbar.less
+++ b/tutor/resources/styles/components/reference-book/navbar.less
@@ -1,38 +1,26 @@
 .reference-book {
 
   @navbar-spacing: 15px;
-  .navbar {
-    height: @tutor-navbar-height; // set in global/navbar.less
-    // override responsive collapse when screen is small,
-    // we handle hiding items via media rules instead
-    .navbar-nav,
-    .navbar-nav > li {
-      margin: 0;
-      float: left;
-      &.navbar-right { float: right; }
-    }
-    .ui-brand-logo { margin: @navbar-spacing 0 0 @navbar-spacing; }
-    .container-fluid {
-      padding: 0;
-    }
-    ul.nav > li > a {
-      margin: 0 @navbar-spacing 0 0;
-      padding: 0;
-      line-height: @tutor-navbar-height;
-    }
+  .tutor-top-navbar {
+    justify-content: space-between;
     .menu-toggle {
-      padding-left: @navbar-spacing;
+      padding: 0 @navbar-spacing;
       height: @tutor-navbar-height;
       cursor: pointer;
       svg { margin-top: @navbar-spacing; }
     }
 
+    &.menu-open {
+      .menu-toggle {
+        border-right: 1px solid @tutor-neutral-light;
+        background: @tutor-neutral-lightest;
+      }
+    }
     .section-title {
       font-size: 1.8rem;
       margin: 0 0 0 2rem;
-      padding-top: 14px;
       display: flex;
-      max-width: calc(~"100% - 550px");
+      max-width: calc(~"100vw - 550px");
       .chapter-section {
         margin-right: 0.5rem;
         text-align: center;
@@ -45,19 +33,4 @@
       }
     }
   }
-  .teacher-edition {
-    .no-content { color: @tutor-neutral-light; }
-  }
-  // hide non-important items when the screen gets narrow
-  @media screen and ( max-width: @reference-book-page-width ){
-    .full-width-only { display: none; }
-  }
-
-  &.menu-open {
-    .menu-toggle {
-      border-right: 1px solid @tutor-neutral-light;
-      background: @tutor-neutral-lightest;
-    }
-  }
-
 }
--- a/tutor/resources/styles/components/reference-book/navbar.less
+++ b/tutor/resources/styles/components/reference-book/navbar.less
@@ -6,6 +6,9 @@
     .left-side-controls, .right-side-controls {
       flex: 1;
     }
+    .right-side-controls {
+      justify-content: flex-end;
+    }
     .menu-toggle {
       padding: 0 @navbar-spacing;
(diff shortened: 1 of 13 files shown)


tutor/resources/styles/components/reference-book/navbar.less
```diff
@@ -1,38 +1,29 @@
...
+        background: @tutor-neutral-lightest;
+      }
+    }
     .section-title {
       font-size: 1.8rem;
       margin: 0 0 0 2rem;
-      padding-top: 14px;
       display: flex;
-      max-width: calc(~"100% - 550px");
+      max-width: calc(~"100vw - 550px");
```
RoyEJohnson: Is the tilde/quote really needed? I see it's used a lot on the source, but I also see a lot of `calc` expressions without it.

tutor/src/components/annotations/annotation.jsx
```diff
@@ -157,27 +149,27 @@ export default class AnnotationWidget extends React.Component {
...
   }
 
-  componentWillMount() {
-    hypothesisStore.fetchAll().then(action((response) => {
-      this.allAnnotationsForThisBook = response
-      .map(dbEntryToPreload)
-      .filter(e => e.selection.ecosystemId === this.props.ecosystemId)
-      .filter(e => e.selection.chapter); // because I have some old entries that cause problems in dev
-    }));
+  @computed get withAnnotations() {
```
RoyEJohnson: This is no longer used and can be removed.

tutor/src/components/annotations/annotation.jsx
```diff
@@ -157,27 +149,27 @@ export default class AnnotationWidget extends React.Component {
...
+    return User.annotations.ux.isSummaryVisible;
+  }
+
   @observable parentRect = {};
   @observable referenceElements = [];
 
   @computed get annotationsForThisPage() {
-    setTimeout(() => this.forceUpdate(), 1000);
-    return this.allAnnotationsForThisBook.filter((item) =>
+    return filter(this.allAnnotationsForThisBook, (item) =>
```
RoyEJohnson: You could use the built-in Array `filter` method instead of the lodash function.

tutor/src/components/annotations/annotation.jsx
```diff
@@ -347,64 +338,52 @@ export default class AnnotationWidget extends React.Component {
   @action.bound
   updateActiveAnnotation(event) {
     const newValue = event.target.value;
-
-    this.activeHighlight.annotation = newValue;
+    this.activeHighlight.text = newValue;
```
RoyEJohnson: lint on osweb insists on a blank line after declarations. :)

tutor/src/components/annotations/annotation.jsx
```diff
@@ -347,64 +338,52 @@ export default class AnnotationWidget extends React.Component {
...
-      }
-    ).then(action((response) => {
-        // Have to make it observable before pushing it, or it won't be the same
-        // object as what is in the observable array
-        const newEntry = observable(dbEntryToPreload(response));
-
-        this.allAnnotationsForThisBook.push(newEntry);
-        return newEntry;
-      }));
+    return User.annotations.create({
```
RoyEJohnson: This makes the code much nicer to read.

tutor/resources/styles/components/reference-book/navbar.less
nathanstitt: hm, no I don't think it is.  I think the tilde is needed if your using less variables like if the 550px was a `@width` or something.   I'm really looking forward to moving to scss since that seems saner to me.

tutor/src/components/annotations/annotation.jsx
nathanstitt: yeah, welcome to our debate!  Since `lodash` is already used pervasively through the code base and is generally faster, we're pretty fond of using it instead of the "builtin" functions.  

For instance https://jsperf.com/array-prototype-filter-vs-filter shows lodash being %2.2 faster than `.filter` on chrome.

Which isn't really enough speedup to bother with, and I normally wouldn't have changed the line to use lodash.  In fact I'm not sure why I did here.  

I think it might be because I'd originally used lodash's object search and had this as: `filter(this.allAnnotationsForThisBook, { chapter, section })`

Then you added the elementId clause and I rebased it to the original and didn't change the `filter` back.

tutor/src/components/annotations/annotation.jsx
nathanstitt: oh that's a nice rule.  We should probably conduct a "lint rules" summit and share our configs to achieve a standard.

tutor/src/components/annotations/summary-toggle.jsx
```diff
@@ -0,0 +1,57 @@
...
+import User from '../../models/user';
+import Courses from '../../models/courses-map';
+import Router from '../../helpers/router';
+import { TaskPanelStore } from '../../flux/task-panel';
+
+@observer
+export default class AnnotationSummaryToggle extends React.Component {
+
+  static propTypes = {
+    windowImpl: React.PropTypes.shape({
```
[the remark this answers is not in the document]
nathanstitt: nope! Good eye.  I removed it

tutor/src/components/annotations/summary-toggle.jsx
```diff
@@ -0,0 +1,57 @@
...
+    return 'reading' === get(crumbs, `[${this.props.taskStepIndex-1}].type`);
+  }
+
+  render() {
+    if (!this.isViewable) { return null; }
+
+    return (
+      <div className="annotation-summary-toggle">
+        <Icon
+          type="pencil-square-o"
```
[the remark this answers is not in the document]
nathanstitt: yeah, I figured as much.  This icon doesn't quite match the mockup but is fairly close for now

PR #1979, opened 2017-11-23
Speedup webpack build using happypack, hard-source and newer uglifyjs

Dantemss:
You will probably have to adjust the package.json file. It seems to do weird things because I am on Ubuntu.

npm packages to be installed are `happypack`, `uglifyjs-webpack-plugin` and `hard-source-webpack-plugin`

- happypack speeds up the first phase (from 1% to 90%) of the initial build
- the updated uglifyjs speeds up the second phase (from 91% to 100%) of all builds
- hard-source caches files for subsequent builds to speed up phase 1 when running build again

On my computer the (repeated) builds go from 2 minutes and a half to 40 seconds. First builds now take around 1 minute and 15 seconds.

Not entirely sure what to do to fix Travis, so will need some help there.


Dantemss:
Turns out tests are probably failing because the fonts are broken

nathanstitt:
Thanks @Dantemss ! Sorry for the delay on merging

PR #1988, opened 2017-11-28
Scroll to highlight after loading page

nathanstitt:


tutor/src/components/annotations/annotation.jsx
```diff
@@ -147,78 +140,118 @@ export default class AnnotationWidget extends React.Component {
...
+  @observable canRenderSidebarButtons = false;
   @observable parentRect = {};
   @observable referenceElements = [];
 
   @computed get annotationsForThisPage() {
     return this.allAnnotationsForThisBook.filter(item =>
       (item.selection.chapter === this.props.chapter) &&
-      (item.selection.section === this.props.section) &&
-      this.referenceElements.find((el) => el.id === item.selection.elementId)
+                                                       (item.selection.section === this.props.section) &&
```
RoyEJohnson: This is weird spacing. Intentional?

tutor/src/components/annotations/annotation.jsx
```diff
@@ -147,78 +140,118 @@ export default class AnnotationWidget extends React.Component {
...
+
     this.handleSelectionChange = debounce(() => {
       if (!this.activeHighlight) {
         this.setWidgetStyle();
       }
     }, 80);
-    this.props.windowImpl.document.addEventListener('selectionchange', this.handleSelectionChange);
+
+    when(
+      () => User.annotations.any,
```
RoyEJohnson: Does this mean that if the user has no annotations, the user will not be able to create any?

tutor/src/components/annotations/annotation.jsx
```diff
@@ -147,78 +140,118 @@ export default class AnnotationWidget extends React.Component {
...
+    this.widgetStyle = null;
+    this.activeHighlight = null;
   }
 
-  componentDidUpdate() {
-    this.getReferenceElements();
-    if (this.savedSelection) {
-      const selection = this.props.windowImpl.getSelection();
+  // componentDidUpdate() {
+  //   this.getReferenceElements();
```
RoyEJohnson: Is this block ready to be removed?

tutor/src/components/annotations/annotation.jsx
```diff
@@ -147,78 +140,118 @@ export default class AnnotationWidget extends React.Component {
...
   highlightEntry(entry) {
     const selection = this.restoreSelectionWithReferenceId(entry.selection);
-
     if (selection) {
       const rect = getSelectionRect(this.props.windowImpl, selection);
-
       entry.style = {
         top: rect.top - this.parentRect.top,
-        position: 'absolute'
+        position: 'absolute',
```
RoyEJohnson: Do you want the trailing comma?

tutor/src/components/annotations/summary-page.jsx
```diff
@@ -87,8 +88,8 @@ class FilterWidget extends React.Component {
             checked={this.props.selectedChapters.length === this.parsedChapters.length}
           />
           {
-            this.parsedChapters.map((ch) => (
-              <CheckBoxEntry key={ch}
+            this.parsedChapters.map((ch, i) => (
+              <CheckBoxEntry key={i}
```
RoyEJohnson: https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

tutor/src/helpers/images-complete.js
```diff
@@ -0,0 +1,22 @@
...
+    let complete = 0;
+    const markComplete = () => {
+      complete += 1;
+      if (complete === images.length) {
+        return resolve(images);
+      }
+    };
+    Array.from(images).forEach((img) => {
+      if (img.naturalWidth) {
+        markComplete();
```
RoyEJohnson: Aha! Nice.

tutor/src/components/annotations/summary-page.jsx
nathanstitt: yeah, I know but I was getting react warnings about duplicate keys if there were multiple annotations per chapter :( In this case I think this is the best we can do.

tutor/src/helpers/images-complete.js
nathanstitt: Yes, I didn't really know about that either until I googled it a bit - is a nice way to check if an image is loaded

tutor/src/components/annotations/annotation.jsx
nathanstitt: Yeah, our (my) preferred eslint config uses trailing comma on objects.  I like it because now if I add another line the git diff doesn't have to pollute the previous line by adding a comma.

tutor/src/components/annotations/annotation.jsx
nathanstitt: yes I think it's not needed. I'd commented it out so I could test that and seems to work fine now with the imagesLoaded helper.  
Thanks for reminding me I'll remove it,

tutor/src/components/annotations/annotation.jsx
nathanstitt: yep, that's the case.  Glad you spotted this.  I think this can check `User.api.isPending`  instead. I'll test that

tutor/src/components/annotations/annotation.jsx
nathanstitt: nope!   not sure what went on there

PR #1990, opened 2017-11-29
moar highlighting fixes

nathanstitt:
* Remove blue focus outline on reading content
* position highlight widget in center of rect
* display an error message when highlighting can't work
* remove highlights from page after delete

### DIFF
--- a/tutor/src/components/annotations/annotation.jsx
+++ b/tutor/src/components/annotations/annotation.jsx
@@ -5,7 +5,7 @@ import { autobind } from 'core-decorators';
 import serializeSelection from 'serialize-selection';
 import './highlighter';
 import User from '../../models/user';
-import { pick, debounce, filter } from 'lodash';
+import { pick, debounce, filter, defer } from 'lodash';
 import Icon from '../icon';
 import SummaryPage from './summary-page';
 import imagesComplete from '../../helpers/images-complete';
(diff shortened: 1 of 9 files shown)


tutor/src/components/annotations/annotation.jsx
```diff
@@ -384,7 +416,12 @@ export default class AnnotationWidget extends React.Component {
 
   @autobind
   deleteEntry(annotation) {
-    User.annotations.destroy(annotation);
+    User.annotations.destroy(annotation).then(() => {
+      const selection = this.restoreSelectionWithReferenceId(annotation.selection);
+      if (selection) {
+        highlighter.removeHighlights(selection.baseNode.parentElement);
```
RoyEJohnson: What's the net effect of this? Seems like either it doesn't matter because all the highlights will be cleared and recreated anyway, or it could remove too many highlights.

nathanstitt: no the `parentElement` of the selection is the actual `.tutor-highlight`, which causes the `highlighter` to only remove the affected highlight.

nathanstitt: .. and I should also say that that's what I've seen thus far.  It wouldn't surprise me if this fails in some situations.

RoyEJohnson: Something to try: a highlight that gets split into multiple `.tutor-highlight`s. That happens when, for example, you highlight from outside a bold element to inside it (but not beyond). The text before the bold is one `.tutor-highlight` and the highlighted portion of the bold element is another.
Each `.tutor-highlight` will have a data-timestamp attribute. Related highlights will have the same timestamp attribute, which you can use to get them all.

PR #1991, opened 2017-12-01
add scripts

philschatz:
**[Trello Card](https://trello.com/c/Ry22O4aP/610-change-request-as-a-qa-dev-ops-i-want-jenkins-to-create-the-build-files-whenever-a-commit-is-pushed-to-a-deployable-repo-so-i-ca)**

The scripts are:

- `./script/setup` install system dependencies (like `yarn`, and `shellcheck`) and then install repo dependencies like `yarn install` (if there is a `package.json`) or `python setup.py install` (if there is a `setup.py`) or `bundle install` (if there is a `Gemfile`)
- `./script/build` to build the `./tutor/dist/`
- `./script/test ${optional_args}` to run the tests
- `./script/ci` to run the **tutor** tests
    - `OX_PROJECT=exercises ./script/ci` runs the exercises tests
- `./script/clean` to remove non-versioned files
- `source ./script/bootstrap` if you want to start a console with all the system dependencies set up

You can prefix every command with the following (not sure if this should be in the README):

- `DEBUG=true` to output each step rather than only outputting the step if it failed
- `TRACE=true` to output each command that bash executes (for debugging the shell script)

# TODO

- [ ] update `README.md`
- [x] have travis run `kcov` and `shellcheck` to add coverage and linting to shell scripts
- [ ] maybe simplify the bin directory?


philschatz:
It seems that Travis is not liking `"phantomjs-prebuilt-that-works": "3.0.1",` because yarn does not find its `package.json`.

I tried using `"phantomjs-prebuilt": "2.1.16",` but that caused 2 tests to fail. Examples:

```
 ● Plan Details Component › renders w/out lms links for events

    TypeError: (0 , _momentTimezone2.default)(...).startOf(...).twix is not a function
      
      at getDay (src/helpers/dates.js:11:563)
```

bin/test-ci
```diff
@@ -7,4 +7,6 @@ set -ev
 
 $(npm bin)/jest --maxWorkers=2 --verbose --coverage --config ./configs/test/jest.$OX_PROJECT.json
 
+echo -en 'travis_fold:start:build\\r'
```
philschatz: ooh, yeah! I love this little "hidden" feature in Travis 😄

nathanstitt: Yeah, it's nice for stuff like this where it generates a huge amount of text

.travis.yml
```diff
@@ -22,20 +22,7 @@ cache:
   directories:
     - node_modules
 
-before_install:
```
philschatz: @nathanstitt Why did you remove kcov? Was it not building?

nathanstitt: I think that could be re-added.  I'd removed it because I thought it was the cause of the failures, but I think the actual cause was the fix in https://github.com/openstax/tutor-js/pull/1991/commits/05d1fc0de9b1e8375a0c76fa3ae4e7de2c4b17b4

However is it the case that `konv` is for coverage of the shell scripts?  In that case I don't think it adds any value since we don't have any specs for the scripts at all and it takes quite awhile to compile on every Travis run.

PR #1995, opened 2017-12-04
refocus to preview tour stop

pandafulmanda:
https://trello.com/c/198YaYBO/1090-change-request-refocus-onboarding-flow-my-courses-page-to-emphasize-more-heavily-the-preview-courses

screen shot 2017-12-04 at 10 52 48 am

### DIFF
--- a/tutor/src/models/user.js
+++ b/tutor/src/models/user.js
@@ -9,6 +9,7 @@ import { UserTerms } from './user/terms';
 import Annotations from './annotations';
 import lazyGetter from '../helpers/lazy-getter';
 import ViewedTourStat from './user/viewed-tour-stat';
+import PreviewCourseOffering from './course/offerings/previews';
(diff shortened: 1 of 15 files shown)


tutor/src/models/user.js
```diff
@@ -46,6 +46,18 @@ export class User extends BaseModel {
...
+
+  @computed get shouldPreview() {
+    const exploreViewStats = this.viewed_tour_stats.find((stat) => stat.id === 'explore-a-preview');
+    if (exploreViewStats) {
+      return exploreViewStats.view_count < 4;
+    }
+    return true;
+  }
+
   @lazyGetter annotations = new Annotations();
```
nathanstitt: This is fine, but I think if we add anymore preview methods we should add a "Previews" class and attach it to User with lazyGetter like we did for Annotations.

PR #1999, opened 2017-12-04
Highlight fixes for loading and positioning

nathanstitt:
Fixes bugs where:
 * We were only loading public annotations, not the ones from the users's group
 *. Annotations were being saved with an incorrect reference element, causing them to appear randomly placed after save


nathanstitt:
[Bug: Bug finding Round 2](https://trello.com/c/v2tkArUu/88-bug-bug-finding-round-2)

PR #2006, opened 2017-12-06
Fix missing fade() sass function

nathanstitt:
fade() doesn't seem to exist, use fade_out instead

fade_in or opacify would be better suited for this conversion but they seem to
require the source color to have alpha specified


nathanstitt:
[Implement the OpenStax Pattern Library](https://trello.com/c/eldsnmxA/1059-implement-the-openstax-pattern-library)

nathanstitt:
[Bug: lost the colored bars on the dashboard](https://trello.com/c/SWD0BMyI/1157-bug-lost-the-colored-bars-on-the-dashboard)

PR #2008, opened 2017-12-06
Highlighting fixes III

nathanstitt:
* Only allow highlighting if feature flag is turned on
* use mouse up to trigger highlight instead of drag
* Only save highlight text when save button is clicked
* look outside selection for highlight element to remove it, fixes FF/Safari bug with deleting highlight

### DIFF
--- a/tutor/src/components/annotations/annotation.jsx
+++ b/tutor/src/components/annotations/annotation.jsx
@@ -126,7 +126,6 @@ export default class AnnotationWidget extends React.Component {
   }
 
   handleSelectionChange = debounce((ev) => {
-    console.log(ev.target)
     if (this.activeAnnotation) {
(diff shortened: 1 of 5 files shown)


nathanstitt:
[T3.highlighting.10 - Highlighting feature flag](https://trello.com/c/vLN3MQvW/35-t3highlighting10-highlighting-feature-flag)

tutor/src/components/annotations/annotation.jsx
```diff
@@ -127,10 +126,10 @@ export default class AnnotationWidget extends React.Component {
   }
 
   handleSelectionChange = debounce((ev) => {
+    console.log(ev.target)
```
pandafulmanda: `console.log` here, should remove later

PR #2009, opened 2017-12-07
Fix review metrics displaying wrong stats for periods

nathanstitt:
The root bug was that the tabs were sorted and filtered out archived (as they should) but the stats were not. 

There also wasn't a loading or "no students" message on the stats page so it looked broken when viewing empty periods.

### DIFF
--- a/tutor/src/components/plan-stats/index.jsx
+++ b/tutor/src/components/plan-stats/index.jsx
@@ -11,7 +11,7 @@ import { ChaptersPerformance, PracticesPerformance } from './performances';
 import Courses from '../../models/courses-map';
 import TeacherTaskPlan from '../../models/task-plan/teacher';
 import LoadingScreen from '../loading-screen';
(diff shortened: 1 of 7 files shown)


nathanstitt:
[Change Request: Do not auto-display all free-responses on the review metrics page](https://trello.com/c/IOcJ19uY/1053-change-request-do-not-auto-display-all-free-responses-on-the-review-metrics-page)

tutor/src/components/plan-stats/index.jsx
```diff
@@ -41,7 +41,7 @@ export default class Stats extends React.PureComponent {
   }
 
   @computed get period() {
-    return this.course.periods[this.currentPeriodIndex];
+    return this.course.periods.sorted[this.currentPeriodIndex];
   }
```
nathanstitt: this is the bug fix.  before the index would be off if periods where archived

PR #2012, opened 2017-12-07
flexbox bug; use height vs min-height

nathanstitt:
For Firefox and probably IE

Before:
screen shot 2017-12-07 at 5 10 12 pm

After:

screen shot 2017-12-07 at 5 08 45 pm


nathanstitt:
[Bug: help and user menus aren't vertically centered](https://trello.com/c/h0vNIqRC/1132-bug-help-and-user-menus-arent-vertically-centered)

PR #2013, opened 2017-12-08
need to browser check further down payments elements so that all paym…

pandafulmanda:
…ents triggering will have browser cer check, also, check for safari versions


nathanstitt:
[CHANGE REQUEST: User taken to different page when unsupported browser is detected in Payments](https://trello.com/c/gfKbpURS/521-change-request-user-taken-to-different-page-when-unsupported-browser-is-detected-in-payments)

PR #2027, opened 2017-12-28
Style and position reference book when menu is on top

nathanstitt:


nathanstitt:
[Bug: Bug finding Round 2](https://trello.com/c/v2tkArUu/1125-bug-bug-finding-round-2)

PR #2034, opened 2017-12-29
set popovers to be same width as training wheels (370px)

nathanstitt:
before:
screen shot 2017-12-29 at 2 56 36 pm

after:
screen shot 2017-12-29 at 2 47 30 pm


nathanstitt:
[Bug: Some Tutor popups have inconsistent styling and are extremely wide](https://trello.com/c/pnGbNYhP/1172-bug-some-tutor-popups-have-inconsistent-styling-and-are-extremely-wide)

PR #2039, opened 2018-01-02
fix sims width in reading content

pandafulmanda:
screen shot 2018-01-01 at 11 21 23 pm
https://trello.com/c/nHB8voyG/1236-bug-flash-simulation-out-of-frame

### DIFF
--- a/tutor/resources/styles/book-content/index.scss
+++ b/tutor/resources/styles/book-content/index.scss
@@ -37,6 +37,7 @@
   @extend %tutor-book-content-teacher;
   @extend %tutor-book-content-questions;
 
+  @include book-content-interactives();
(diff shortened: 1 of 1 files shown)


tutor/resources/styles/book-content/index.scss
```diff
@@ -37,6 +37,7 @@
   @extend %tutor-book-content-teacher;
   @extend %tutor-book-content-questions;
 
+  @include book-content-interactives();
   @include tutor-book-notes();
```
nathanstitt: heh, I bet this was one of those "takes forever to investigate but only a second to fix".   That must have been missed when I translated the less to scss.  Glad you figured it out

PR #2042, opened 2018-01-02
specify paged content as a live region so it notifies when updated

nathanstitt:
Per https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions
this is how you can tell screenreader to re-check a content area


nathanstitt:
[Bug: Free response submission resets keyboard navigation position.](https://trello.com/c/r1XpxKGU/1183-bug-free-response-submission-resets-keyboard-navigation-position)

PR #2044, opened 2018-01-03
Fix `Cannot set property 'time_zone' of undefined`

nathanstitt:
When the time is changed in the task plan builder, a "Set as default" link appears. However clicking it causes a `Cannot set property 'time_zone' of undefined` error and the time default is not saved.


nathanstitt:
[Bug: Unable to update default open/due time](https://trello.com/c/TSqnMBVf/1184-bug-unable-to-update-default-open-due-time)

PR #2046, opened 2018-01-03
remove package-lock.json

nathanstitt:
it's not needed now that we're using yarn


m1yag1:
[T3.highlighting.09 - Improvements to the Highlight Export](https://trello.com/c/ak0WYSBI/1150-t3highlighting09-improvements-to-the-highlight-export)

PR #2047, opened 2018-01-03
add asnc-loaded "screens"; implement qa as POC

nathanstitt:
This is an idea I've been kicking around

I think we should add "screens", each of which has a single entry point that's dynamically loaded using `System.import`

Each goes into it's own directory and anything that's only related to them lives there as well.  (images, styles, etc)


tutor/src/routes.js
```diff
@@ -0,0 +1,193 @@
...
+  return StudentPreview;
+};
+
+const ROUTES = [
+  { path: '/dashboard', name: 'myCourses', renderer: getMyCourses },
+  { path: '/enroll/start/:enrollmentCode', name: 'createEnrollmentChange', renderer: getCreateEnrollmentChange },
+  { path: '/new-course/:sourceId?', name: 'createNewCourse', renderer: getCreateCourse },
+  {
+    path: '/qa', name: 'QADashboard', renderer: () => loadAsync(() => System.import('./screens/qa/index.jsx')),
+
```
nathanstitt: ^^^ This is the call that kicks it all off

tutor/src/routes.js
```diff
@@ -0,0 +1,193 @@
...
+  return StudentPreview;
+};
+
+const ROUTES = [
+  { path: '/dashboard', name: 'myCourses', renderer: getMyCourses },
+  { path: '/enroll/start/:enrollmentCode', name: 'createEnrollmentChange', renderer: getCreateEnrollmentChange },
+  { path: '/new-course/:sourceId?', name: 'createNewCourse', renderer: getCreateCourse },
+  {
+    path: '/qa', name: 'QADashboard', renderer: loadAsync(() => System.import('./screens/qa/index.jsx')),
+
```
nathanstitt: ^^^ This is the call that kicks it all off

PR #2050, opened 2018-01-04
stop emitting debug info when in spy mode

nathanstitt:


nathanstitt:
[Bug: Remove debugging console dumps for tours](https://trello.com/c/xI95EKAE/1187-bug-remove-debugging-console-dumps-for-tours)

PR #2053, opened 2018-01-05
action active menu

pandafulmanda:
screen shot 2018-01-05 at 2 21 35 pm

screen shot 2018-01-05 at 4 40 23 pm

## Before

screen shot 2018-01-05 at 4 38 41 pm

  

**note** seems like we should be able to use the `match` prop being passed into a `Route`'s children, but that doesn't seem to be updating with each route change

### DIFF
--- a/tutor/resources/styles/components/top-nav-bar/drop-down-menu.scss
+++ b/tutor/resources/styles/components/top-nav-bar/drop-down-menu.scss
@@ -75,8 +75,11 @@
       }
     }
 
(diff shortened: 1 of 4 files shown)


nathanstitt:
[Change Request: Post-user test Tutor global nav changes -- change profile dropdown menu header](https://trello.com/c/yK5BzT5y/1102-change-request-post-user-test-tutor-global-nav-changes-change-profile-dropdown-menu-header)

PR #2055, opened 2018-01-08
initial zoning fix for not barfing on invalid zone

pandafulmanda:


tutor/src/helpers/time.coffee
```diff
@@ -1,6 +1,6 @@
 moment = require 'moment-timezone'
 require 'moment-timezone/moment-timezone-utils'
-_ = require 'underscore'
+_ = require 'lodash'
 # Map http://www.iana.org/time-zones names to timezone names in Rails
```
nathanstitt: We shouldn't `require 'lodash'` since that pulls in the entire lodash lib which is pretty large.  In coffeescript we have to do individual require's, i.e. `map = require 'lodash/map`

For JS, we have a babel plugin so we can do `import { map } from 'lodash'`

PR #2058, opened 2018-01-10
Load components async

nathanstitt:
Is on top of #2047 - merge that first

Moves these components into screens that are loaded async:
* Teacher calendar
* Student dashboard
* Reference book


nathanstitt:
[[T3.research.06] Assigning surveys for research](https://trello.com/c/uxeALhOL/1075-t3research06-assigning-surveys-for-research)

PR #2065, opened 2018-01-11
Fix weird student dashboard centering

pandafulmanda:
See also #2062


pandafulmanda:
darn opened against wrong branch

PR #2073, opened 2018-01-15
Convert scores report to dynamically loaded screen

nathanstitt:


nathanstitt:
[T3.03.23: Give instructors a way get meaningful scores sent to the LMS when they only assign readings or want to assign credit for readings](https://trello.com/c/nzF8ZQp5/1088-t30323-give-instructors-a-way-get-meaningful-scores-sent-to-the-lms-when-they-only-assign-readings-or-want-to-assign-credit-for)

PR #2076, opened 2018-01-17
Fix calendar details

pandafulmanda:
https://trello.com/c/8Qt7jbQh/1197-bug-ui-teacher-calendar-view-not-showing-assignment-open-dates and other small fixes


nathanstitt:
Thanks @pandafulmanda - the removal of `checkPublishingStatus` made me nervous but after publishing a few I see that it is indeed not being used.  Good sleuthing !

PR #2078, opened 2018-01-17
Fix double loading tasks when enrolling

nathanstitt:
Initially the enrollment code would poll for tasks, but then later we also added polling anytime the student dashboard is displayed.  

Which caused double loads and occasionally an error when the action-adapter noticed the same api request and returned undefined instead of a Promise


nathanstitt:
[Bug: LMS students with pre-existing accounts are sent to sign-up](https://trello.com/c/mSCZ4V9B/1202-bug-lms-students-with-pre-existing-accounts-are-sent-to-sign-up)

PR #2080, opened 2018-01-17
remove ExtractTextPlugin

nathanstitt:
Styles are now loaded as part of the JS chunk so they can be dynamic
and load via System.import

Goes along with https://github.com/openstax/tutor-server/pull/1652


nathanstitt:
[[T3.research.06] Assigning surveys for research](https://trello.com/c/uxeALhOL/1075-t3research06-assigning-surveys-for-research)

PR #2084, opened 2018-01-18
fix styles on hw builder select and feedback

nathanstitt:
screen shot 2018-01-18 at 4 19 22 pm
 
screen shot 2018-01-18 at 3 51 03 pm


nathanstitt:
[Bug: Feedback not displaying](https://trello.com/c/IV6yEQzJ/1204-bug-feedback-not-displaying)

PR #2093, opened 2018-01-31
don't show deleted periods for event stats

nathanstitt:


nathanstitt:
[Bug: Event in Teacher view is showing all the former names of a renamed section](https://trello.com/c/qSqwUgJS/1275-bug-event-in-teacher-view-is-showing-all-the-former-names-of-a-renamed-section)

PR #2094, opened 2018-01-31
Fix "orig. due date" popover on plan clone

nathanstitt:
Before:
screen shot 2018-01-30 at 9 02 52 pm

After:
screen shot 2018-01-30 at 9 02 12 pm


nathanstitt:
[Change Request: copied course assignment - mouseover messaging looks like an error](https://trello.com/c/tGcCbQ9v/1276-change-request-copied-course-assignment-mouseover-messaging-looks-like-an-error)

PR #2095, opened 2018-01-31
go to second tab for setting time

nathanstitt:
The've been moved so that the time settings are there now


nathanstitt:
[Bug: clicking time zone for an assignment goes to the Student Access tab instead of Dates and Times](https://trello.com/c/kw8xUVZb/1277-bug-clicking-time-zone-for-an-assignment-goes-to-the-student-access-tab-instead-of-dates-and-times)

PR #2097, opened 2018-01-31
Fix progress bars on stats

nathanstitt:
The after label wasn't being displayed and the margin-bottom pushes the following
label down since it's floated

before:
screen shot 2018-01-31 at 12 56 26 pm

after:
screen shot 2018-01-31 at 12 49 00 pm


nathanstitt:
[Bug: odd wrapping for small-value wrong on quick look analytics](https://trello.com/c/5cyzvoVj/1308-bug-odd-wrapping-for-small-value-wrong-on-quick-look-analytics)

PR #2098, opened 2018-01-31
Fix errors that were logged to papertrail

nathanstitt:
Fixes:
 * TypeError: document.querySelectorAll(...).forEach is not a function
    * IE 11 doesn't have forEach for nodeList
 * Uncaught TypeError: Cannot read property 'preventDefault' of undefined
 * TypeError: Cannot read property 'classList' of null  at Constructor.unhighlightMedia
 * Uncaught TypeError: Cannot read property 'isStudent' of undefined 
 * Uncaught Invariant Violation: attempted to add region student-highlighting-reading, but it already exists!
 * TypeError: undefined is not an object (evaluating 'crumb.id')

### DIFF
--- a/shared/specs/helpers/logging.spec.coffee
+++ b/shared/specs/helpers/logging.spec.coffee
@@ -1,5 +1,11 @@
 jest.mock 'model/networking'
 jest.mock 'loglevel'
+jest.mock 'lodash/debounce', ->
+  jest.fn(
+    (fn) -> -> fn()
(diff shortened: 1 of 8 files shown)


nathanstitt:
[Investigate and fix FE errors being logged to paper trail](https://trello.com/c/gFjlgbsk/1272-investigate-and-fix-fe-errors-being-logged-to-paper-trail)

PR #2100, opened 2018-02-01
Fix keyboard navigation on menus

nathanstitt:
One user visible change is that we can no longer wrap the entire menu in a tour region.  Doing so makes the menu stop passing the keyboard handlers down to the MenuItems.

This moves the tour regional that highlighted the help menu from the menu itself to the menu controls at the top.  Basically from this:
screen shot 2018-02-01 at 12 41 26 pm

to this:
screen shot 2018-02-01 at 12 10 31 pm

### DIFF
--- a/tutor/resources/styles/components/top-nav-bar.scss
+++ b/tutor/resources/styles/components/top-nav-bar.scss
@@ -61,14 +61,6 @@ $tutor-navbar-padding-horizontal: 4rem;
     width: 1200px;
   }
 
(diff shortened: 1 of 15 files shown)


nathanstitt:
[Bug: Keyboard navigation skips items in the student name menu.](https://trello.com/c/PZcx8AJh/1179-bug-keyboard-navigation-skips-items-in-the-student-name-menu)

PR #2101, opened 2018-02-01
read first and last name from server

nathanstitt:
this way the initials will be correct even if the user has a surname or title

### DIFF
--- a/tutor/src/models/user.js
+++ b/tutor/src/models/user.js
@@ -25,6 +25,9 @@ export class User extends BaseModel {
   @field account_uuid;
 
   @field name;
+  @field first_name;
+  @field last_name;
+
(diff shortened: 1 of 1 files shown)


nathanstitt:
[Bug: User initials are wrong if anything is entered in the prefix or suffix](https://trello.com/c/HLY4EAuu/1273-bug-user-initials-are-wrong-if-anything-is-entered-in-the-prefix-or-suffix)

gregfitch:
Will the change take care of the case where the first and/or last name is not capitalized?

nathanstitt:
@gregfitch no, @Fredasaurus said that we should honor the user's chosen capitalization preferences.  So it'll still be lowercase if that's what they do

PR #2104, opened 2018-02-02
fix per course tour for new enrollment link, need to display after fr…

pandafulmanda:
…esh course nag

https://trello.com/c/NXwmul6k/1224-change-request-ensure-returning-tutor-user-sends-out-new-correct-enrollment-links-for-new-course

### DIFF
--- a/tutor/specs/screens/course-calendar/plan-details.spec.jsx
+++ b/tutor/specs/screens/course-calendar/plan-details.spec.jsx
@@ -1,18 +1,23 @@
 import { React, ReactTestUtils } from '../../components/helpers/component-testing';
 import { last } from 'lodash';
 import { ReactWrapper } from 'enzyme';
+import { Provider } from 'mobx-react';
 import PlanDetails from '../../../src/screens/teacher-dashboard/plan-details';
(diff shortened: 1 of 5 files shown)


tutor/specs/screens/course-calendar/plan-details.spec.jsx
```diff
@@ -1,18 +1,23 @@
...
 const COURSE_ID = '1';
 
 const renderModal = props =>
   new Promise( function(resolve) {
-    const wrapper = mount(<PlanDetails {...props} />, EnzymeContext.build());
+    const planDetails = <Provider tourContext={new TourContext()}>
+      <PlanDetails {...props} />
+    </Provider>;
+    const wrapper = mount(planDetails, EnzymeContext.build());
     resolve(last(document.querySelectorAll('.modal')));
```
nathanstitt: Heh, the nice thing about the `inject` method from mobx is that you can also provide the values as a prop and don't need to wrap them in a `Provider` for tests.  Pretty sure this would also have worked:

```
   <PlanDetails {...props} tourContext={new TourContext()} />
```

PR #2106, opened 2018-02-02
Fix exercise multi answer

pandafulmanda:
https://trello.com/c/RFLdPdNG/1287-bug-multiple-answers-can-appear-as-selected

### DIFF
--- a/shared/src/components/question/answer.cjsx
+++ b/shared/src/components/question/answer.cjsx
@@ -1,5 +1,5 @@
 React = require 'react'
-_ = require 'underscore'
+{partial, pick, debounce} = require 'lodash'
 classnames = require 'classnames'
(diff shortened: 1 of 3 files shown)


shared/src/components/question/answer.cjsx
```diff
@@ -1,5 +1,5 @@
 React = require 'react'
-_ = require 'underscore'
+{partial, pick, debounce} = require 'lodash'
 classnames = require 'classnames'
```
nathanstitt: This file is still coffeescript so we have to individually import lodash stuffs, i.e.
```
partial = require 'lodash/partial'
...
```

pandafulmanda: oooh right, okay, will change

PR #2115, opened 2018-02-08
space performance key keys, and reading content notes

pandafulmanda:
https://trello.com/c/Oo4ALQx9/1278-bug-spacing-between-or-around-text-is-missing


tutor/resources/styles/book-content/note.scss
```diff
@@ -3,6 +3,7 @@ $feature-title-is-element: '[data-has-label=true][data-label=""]';
 $tutor-note-title-element: '[data-type=title]:first-child';
 
 $tutor-note-margin-horizontal: 38px;
+$tutor-note-margin-vertical: 71px;
 %tutor-empty-rules { }
```
nathanstitt: These units are really strange to me.  why are they so precise?

pandafulmanda: eh, just picked something that looked good

PR #2118, opened 2018-02-09
made a mistake with previous related contented fix

pandafulmanda:
see #2116

### DIFF
--- a/tutor/src/components/task-step/step-with-reading-content.cjsx
+++ b/tutor/src/components/task-step/step-with-reading-content.cjsx
@@ -43,7 +43,11 @@ ReadingStepContent = React.createClass
 
   getContentChapterSection: ->
     {chapter_section, related_content} = TaskStepStore.get(@props.id)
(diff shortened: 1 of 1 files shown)


tutor/src/components/task-step/step-with-reading-content.cjsx
```diff
@@ -43,7 +43,11 @@ ReadingStepContent = React.createClass
 
   getContentChapterSection: ->
     {chapter_section, related_content} = TaskStepStore.get(@props.id)
-    (related_content && related_content[0]) || chapter_section || []
+    (
+      related_content &&
+      related_content[0] &&
+      related_content[0].chapter_section
+    ) || chapter_section || []
```
nathanstitt: fyi - I like using lodash [get](https://lodash.com/docs/#get) for this since it deals with null values, this could be: 

`get(related_content, '[0].chapter_section, chapter_section || [])`

PR #2120, opened 2018-02-13
Reload browser window when an error occurs

nathanstitt:
When I force a load failure by typing gibberish in a scores import it reloads once then displays:

screen shot 2018-02-13 at 1 35 11 pm


nathanstitt:
[Bug: Users receive error when changing views after a deployment](https://trello.com/c/K5aRqm3R/1304-bug-users-receive-error-when-changing-views-after-a-deployment)

PR #2125, opened 2018-02-13
Redirect already enrolled students to their course when using enroll link

nathanstitt:
render already enrolled redirect before errors; otherwise we'd render the generic error message because the error check came before the redirect


nathanstitt:
[Bug: Students trying to pay/login to their course using their enrollment link are not being redirected to the dashboard but receiving an error message](https://trello.com/c/Dj0jpP4y/1308-bug-students-trying-to-pay-login-to-their-course-using-their-enrollment-link-are-not-being-redirected-to-the-dashboard-but-recei)

PR #2128, opened 2018-02-14
Fix html in step titles

nathanstitt:
Previously if a task step had math in it's `stem_html` that would cause us to parse the title incorrectly and we'd display math in the task step title and it's "Continue to .." text link.


nathanstitt:
[Bug: Reading continue text contains HTML code as text](https://trello.com/c/S30ZW0Eq/1290-bug-reading-continue-text-contains-html-code-as-text)

PR #2129, opened 2018-02-14
Use lowercase for reading x-app-date header value

nathanstitt:


nathanstitt:
[Bug: If the student's computer's clock is wrong and an assignment is past-due according to their clock, they are prevented from changing answers that they would otherwise be able to](https://trello.com/c/IuD2w5VC/1142-bug-if-the-students-computers-clock-is-wrong-and-an-assignment-is-past-due-according-to-their-clock-they-are-prevented-from-chan)

PR #2131, opened 2018-02-14
maintain aspect ratio of scaled images in reference book

nathanstitt:
before:

screen shot 2018-02-14 at 4 01 09 pm

after:  Note that the Isaac Newton image looks too tall but that's it's correct height.

screen shot 2018-02-14 at 3 59 11 pm

### DIFF
--- a/tutor/resources/styles/book-content/base.scss
+++ b/tutor/resources/styles/book-content/base.scss
@@ -73,7 +73,10 @@
       }
     }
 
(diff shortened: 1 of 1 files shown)


nathanstitt:
[Bug: Reference View and Readings are having some layout and image issues](https://trello.com/c/yHiak7jj/900-bug-reference-view-and-readings-are-having-some-layout-and-image-issues)

PR #2138, opened 2018-02-19
UX fixes for student scores

nathanstitt:
* Use pattern library buttons styles on weights modal
* Fix overly wide table for students
* Don't render late icon for student's external assignments

screen shot 2018-02-19 at 1 09 13 pm

screen shot 2018-02-19 at 1 08 59 pm

screen shot 2018-02-19 at 1 12 06 pm

### DIFF
--- a/tutor/resources/styles/components/modals.scss
+++ b/tutor/resources/styles/components/modals.scss
@@ -21,10 +21,10 @@
 
   .modal-footer {
     padding: 0 $modal-padding $modal-padding $modal-padding;
-    text-align: left;
     border-top: 0;
+    display: flex;
     button {
-      height: 50px;
+      @include button(); // openstax pattern-library
     }
   }
 
--- a/tutor/resources/styles/components/modals.scss
+++ b/tutor/resources/styles/components/modals.scss
@@ -1,6 +1,6 @@
 .modal {
   $header-padding: 35px;
-  $modal-padding: 50px;
+  $modal-padding: 30px;
 
   .modal-content {
     border-radius: 0;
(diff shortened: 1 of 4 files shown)


nathanstitt:
[Change Request: Add score report for students in Tutor](https://trello.com/c/04Xk18sV/1089-change-request-add-score-report-for-students-in-tutor)

rvnewell:
Nathan, the table looks good now that the bottom border is aligned.  The modal I was referencing, though, was the student view of the weights modal.  It only has the weights listed, with nothing else.  The student view of that modal has a lot more padding on all sides and around the button than spec'd in the pattern library.

nathanstitt:
@rvnewell Hm, the student modal's not styled any different than any of our other modals.  It's using 50px padding on the modal body which I believe is what was specced at one time for modals.

Should I change that for all, or would you like to special case the student modal?

screen shot 2018-02-19 at 2 53 06 pm

rvnewell:
screen shot 2018-02-19 at 2 59 50 pm

rvnewell:
Above is the spec for the instructor weights modal.  This is consistent with the pattern library, in that it's 30px padding around the edges, and also 30px padding between buttons and text.  If you could change the padding for both instructor and student weights modals, that would be excellent.

nathanstitt:
Ok, I just wanted to make sure you were cool with changing from 50px.  I've changed to 35px as shown below:

screen shot 2018-02-19 at 3 20 26 pm
screen shot 2018-02-19 at 3 19 58 pm

rvnewell:
Our intent is to have the weights columns for both moved inward, closer to the text.  We have a card for that, but, since this is now merged, we can go with this for the next release.

PR #2142, opened 2018-02-20
acceptance tests via screenshot generation

nathanstitt:


tutor/acceptance/helpers/git.js
```diff
@@ -0,0 +1,50 @@
+const gitP = require('simple-git/promise');
+const path = require('path');
+const { imagesDir } = require('./images-dir');
+const chalk = require('chalk');
+
+const repoURL = 'https://github.com/openstax/tutor-images.git';
```
philschatz: The images are in a separate repo? One downside of this is that QA or devs will not be able to see which screenshots changed during a PR review (which I think would be helpful).

If the concern is repo size, would git-lfs help? I'm not sure if those changes would render in the PR diff though.

nathanstitt: Yeah, image size was the concern.  I'll investigate git-lfs, maybe it'll work for us

nathanstitt:
actually it may, it shows a preview at least - https://github.com/openstax/tutor-js/blob/e8b143c34958397a91ef4438aad96125fd85da0e/screenshots/student-my-courses.png

philschatz:
Oh cool!

nathanstitt:
@philschatz no travis already has git-lfs so that works... but it doesn't actually work to run the tests ☹️ 

I tracked down the failures to the fact that travis runs on linux and linux doesn't have the HelveticaNue font.  So the screenshots are all different and fail.  I even made a shot at running Travis on OSX which kinda worked but was super slow.  

I think this still has value but we won't be able to run the detection automatically.  We can compare across releases manually though

philschatz:
@Fredasaurus Any chance there's an open-source alternative font we could use for Tutor?

And if not, could the tests run without the font? Like, either the build would conditionally replace `Helvetica Nue` with `roboto`?

nathanstitt:
yeah, we used to use Lato but switched to Helvetica Nue last fall.  Not sure if we can switch back @Fredasaurus ?

PR #2143, opened 2018-02-22
Move base model to shared/model and adjust path to shared

nathanstitt:
This changes shared include path so it points to "shared/src"

Which will allow consumers of shared modules to easily just include the file
so we don't have to add everything to the shared index

This is the first step to updating exercises so it can easily pull things from shared


nathanstitt:
[Change request: Exercises support a nickname so Tutor can fetch exercises via that nickname or question ID](https://trello.com/c/JeVQGCsD/573-change-request-exercises-support-a-nickname-so-tutor-can-fetch-exercises-via-that-nickname-or-question-id)

nathanstitt:
[T3.21.07 improve exercise editor search](https://trello.com/c/UmoB8S9W/1319-t32107-improve-exercise-editor-search)

PR #2144, opened 2018-02-22
Ensure weights are between 0 to 100

nathanstitt:


nathanstitt:
[Bug: able to set negative values for the score weight](https://trello.com/c/hK2bJlF0/1332-bug-able-to-set-negative-values-for-the-score-weight)

PR #2145, opened 2018-02-23
Fix scores moar

pandafulmanda:
screen shot 2018-02-23 at 5 52 33 pm

### DIFF
--- a/tutor/resources/styles/components/modals.scss
+++ b/tutor/resources/styles/components/modals.scss
@@ -1,5 +1,3 @@
-$modal-header-font-size: 16px;
-
 .modal {
   $header-padding: 35px;
   $modal-padding: 50px;
@@ -11,7 +9,7 @@ $modal-header-font-size: 16px;
 
   .modal-header {
     padding: $header-padding $modal-padding;
-    font-size: $modal-header-font-size;
+    font-size: 16px;
     background-color: $tutor-neutral-bright;
     border-bottom: 1px solid $tutor-neutral-light;
   }
@@ -25,9 +23,6 @@ $modal-header-font-size: 16px;
     padding: 0 $modal-padding $modal-padding $modal-padding;
     border-top: 0;
     display: flex;
-    button {
-      @include button(); // openstax pattern-library
-    }
   }
 
 }
--- a/tutor/resources/styles/components/modals.scss
+++ b/tutor/resources/styles/components/modals.scss
@@ -1,9 +1,8 @@
-$modal-header-height: 75px;
 $modal-header-font-size: 16px;
 
 .modal {
-  $header-padding: ($modal-header-height - $modal-header-font-size) / 2;
-  $modal-padding: 30px;
+  $header-padding: 35px;
+  $modal-padding: 50px;
 
   .modal-content {
     border-radius: 0;
@@ -26,6 +25,9 @@ $modal-header-font-size: 16px;
     padding: 0 $modal-padding $modal-padding $modal-padding;
     border-top: 0;
     display: flex;
+    button {
+      @include button(); // openstax pattern-library
+    }
   }
 
 }
--- a/tutor/resources/styles/components/modals.scss
+++ b/tutor/resources/styles/components/modals.scss
@@ -1,6 +1,9 @@
+$modal-header-height: 75px;
+$modal-header-font-size: 16px;
+
 .modal {
-  $header-padding: 35px;
-  $modal-padding: 50px;
+  $header-padding: ($modal-header-height - $modal-header-font-size) / 2;
+  $modal-padding: 30px;
 
   .modal-content {
     border-radius: 0;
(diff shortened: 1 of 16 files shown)


tutor/resources/styles/components/modals.scss
```diff
@@ -1,6 +1,9 @@
+$modal-header-height: 75px;
+$modal-header-font-size: 16px;
+
 .modal {
-  $header-padding: 35px;
-  $modal-padding: 50px;
+  $header-padding: ($modal-header-height - $modal-header-font-size) / 2;
+  $modal-padding: 30px;
```
nathanstitt: @pandafulmanda have you ran this by @Fredasaurus and @rvnewell ?  I'd done this earlier and they didn't want it changed globally.  In particular they didn't like how the enrollment modal looked after this change

tutor/resources/styles/components/modals.scss
```diff
@@ -24,7 +27,7 @@
     text-align: left;
     border-top: 0;
     button {
-      height: 50px;
+      height: 40px;
     }
```
nathanstitt: On my earlier changes to the modal, I'd imported the pattern library buttons.  If we're changing this should we do that?  https://github.com/openstax/tutor-js/commit/85861e418e63f70c4b2486ed0d53f7e78c3242c9#diff-6b525495a136dda2344b10d715890f04

tutor/src/components/error-monitoring/async-load-error.jsx
```diff
@@ -1,21 +1,15 @@
...
   }
 
-  componentWillMount() {
-    reloadOnce();
-  }
-
   render() {
-    if (!isReloaded()) { return null; }
-
     return (
```
nathanstitt: Hm, I'm pretty sure we still want to reload on error.  Why remove this?

tutor/src/screens/scores-report/weights-ux.jsx
```diff
@@ -10,86 +10,151 @@ const DEFAULTS = {
...
+  @computed get values() {
+    return pick(this, SETTINGS);
+  }
+
+  @computed get total() {
+    return flow(
+      toArray,
+      sum,
+      Math.round,
+    )(this.values);
```
nathanstitt: `toArray` is fine, but you could call `this.values.peek()` to get a reference to the internal array that mobx uses.  And if you wanted to modify the array you can call `.slice()` to get a new array

rvnewell:
@nathanstitt If this is what you're referring to above -- the ultimate goal is that Tutor buttons are the medium buttons in the pattern library (40px height).  I think some buttons are still 50px (like course cloning modals?), but the ultimate goal is to move to 40px.  @Fredasaurus please confirm.

nathanstitt:
On the buttons I was just suggesting we use the styles from the pattern library instead of duplicating the height ourselves. 

 I'm more concerned about changing the global padding on the modals, since we decided against doing that previously.  Maybe something changed though?

tutor/resources/styles/components/modals.scss
pandafulmanda: ah, okay, i can make it local to the enrollment modal

tutor/resources/styles/components/modals.scss
pandafulmanda: yes i will do htat

tutor/src/components/error-monitoring/async-load-error.jsx
pandafulmanda: oh i thought we talked about this?  i guess i can across this because when there's an error with the scores report modal saving, the error modal pops up and then the whole page reloads without the user closing or being able to read the modal and it's very disorienting

rvnewell:
@nathanstitt Okay.  Yes, we still need to decide global padding on modals, but we also need to align on modal types and how padding will be applied to each. This card addresses that:  https://trello.com/c/7CPUlEWQ/1337-change-request-align-modal-styling-and-modal-types

tutor/src/screens/scores-report/weights-ux.jsx
pandafulmanda: okay

tutor/src/components/error-monitoring/async-load-error.jsx
nathanstitt: oh, sorry! This is a different type of error.  This error is caused by a JS chunk failing to load, most likely because we deployed and the file that it needs is no longer in the directory.  If we reload then the browser will get the update build and all will be well.

it's a different modal that reloads in response to server errors, it's in https://github.com/openstax/tutor-js/blob/master/tutor/src/components/error-monitoring/index.cjsx

That's the one that I think we can get rid of.

tutor/src/screens/scores-report/weights-ux.jsx
pandafulmanda: ah actually cant use `peek` here because `values` is a computed object

tutor/src/components/error-monitoring/async-load-error.jsx
pandafulmanda: ah my bad, will fix these

PR #2149, opened 2018-02-28
Don't re-use default value, instead always start with ''

nathanstitt:
Fixes bug where the period name would be re-used when adding more than one section/period in a row


nathanstitt:
[Bug: previous section name is in the input box if more than one section is added from the roster](https://trello.com/c/BHMGvn0k/1339-bug-previous-section-name-is-in-the-input-box-if-more-than-one-section-is-added-from-the-roster)

PR #2150, opened 2018-02-28
Fix fetching tasks on enrollment

nathanstitt:
Before it wouldn't immediately re-poll after the initial fetch until 4 hours had past 😿 

And the task spinner wasn't displaying while it loaded


nathanstitt:
[Bug: Assignments do not appear right away when a student is added to a course](https://trello.com/c/quI5hGrT/1341-bug-assignments-do-not-appear-right-away-when-a-student-is-added-to-a-course)

PR #2153, opened 2018-03-01
Fix scores report message for students with no assignments

nathanstitt:
screen shot 2018-03-01 at 11 35 49 am


nathanstitt:
[Bug: setting a weight for a category without any assignments may not show the expected course score](https://trello.com/c/rmS6rHdx/1301-bug-setting-a-weight-for-a-category-without-any-assignments-may-not-show-the-expected-course-score)

PR #2154, opened 2018-03-01
Fix QA view for APUSH & Bio, add book content

nathanstitt:
screen shot 2018-03-01 at 3 12 49 pm
screen shot 2018-03-01 at 3 12 42 pm


nathanstitt:
[Bug: QA view is broken.](https://trello.com/c/s2Xs9V0M/1321-bug-qa-view-is-broken)

PR #2155, opened 2018-03-01
Replace performance report links with BackButton

nathanstitt:
screen shot 2018-03-01 at 3 43 44 pm


nathanstitt:
[Bug: Performance Forecast "Back to ----" button (as will all pages in a Tutor course), should be dynamic and state "Back to and the previous page.](https://trello.com/c/bpJ2TZK6/1342-bug-performance-forecast-back-to-button-as-will-all-pages-in-a-tutor-course-should-be-dynamic-and-state-back-to-and-the-previous)

PR #2157, opened 2018-03-02
weight inputs were not observing weight updates, so the input  would …

pandafulmanda:
…not update it's value despite weights changing

### DIFF
--- a/tutor/src/screens/scores-report/set-weights-modal.jsx
+++ b/tutor/src/screens/scores-report/set-weights-modal.jsx
@@ -14,23 +14,6 @@ const WEIGHTS = [
   'Reading progress',
 ];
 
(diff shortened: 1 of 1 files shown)


tutor/src/screens/scores-report/set-weights-modal.jsx
```diff
@@ -14,23 +14,6 @@ const WEIGHTS = [
   'Reading progress',
 ];
 
-const WeightControl = ( { weightLabel, weightName, weights } ) => (
-  <label className="weight">
```
nathanstitt: If we wanted this as a component we could probably add `observer` to it, i.e. 
```
const WeightControl = observer(( { weightLabel, weightName, weights } ) => (
  ...
);

PR #2163, opened 2018-03-06
NaN for no assignments of type and -- in headers

nathanstitt:


tutor/src/screens/scores-report/assignment-header.jsx
```diff
@@ -26,15 +27,15 @@ const ReviewLink = (props) => {
...
+        {(heading.average_score * 100).toFixed(0)}%
       </span>
     );
-  } if (props.heading.type === 'external') {
-    const p = props.heading.average_progress;
+  } if (heading.type === 'external') {
+    const p = heading.average_progress;
     let percent;
     if (p < 1 && p > 0.99) {
       percent = 99; // Don't round to 100% when it's not 100%!
```
nathanstitt: this was never working as far as I can tell, since `average_score` was on the `heading` not on props

tutor/src/screens/scores-report/ux.js
```diff
@@ -108,8 +108,7 @@ export default class ScoresReportUX {
     return mapValues(averages, (average, key) => {
       const type = scoreKeyToType(key);
       let nullValue;
-
-      if (average === null) {
+      if (isNil(average)) {
         if (type === 'course_average') {
           nullValue = this.nullAverageForCourse;
         } else {
```
nathanstitt: average was `undefined` not `null` which are different. But `isNil` catches both cases

PR #2170, opened 2018-03-08
update factory and moment

pandafulmanda:
not sure why travis is upset :/


nathanstitt:
I re-started travis and it passed.  I really hope we don't have a flaky spec that's gonna fail randomly

PR #2173, opened 2018-03-12
Fixed clue is_real spy mode

Dantemss:


tutor/src/components/performance-forecast/statistics.cjsx
```diff
@@ -46,7 +46,7 @@ Statistics = React.createClass
         <ul>
           { for key, value of @props.section.clue
             value = value.join(' ') if _.isArray(value)
-            <li key={key}><strong>{key}</strong>: {value}</li>}
+            <li key={key}><strong>{key}</strong>: {value.toString()}</li>}
         </ul>
```
nathanstitt: Hm, this will break if value is `null`, `undefined` or really anything that lacks `toString`.  However the `String` class has good co-coercion so you could do: `String(value)`.

PR #2174, opened 2018-03-12
record and use osweb_base_url

nathanstitt:
Goes with https://github.com/openstax/tutor-server/pull/1706


nathanstitt:
[Change Request: add an endpoint so the front end knows which instance set it is a part of (Dev, QA, Staging, Production, Demo, some other down the road)](https://trello.com/c/WBQ8Bo6h/1354-change-request-add-an-endpoint-so-the-front-end-knows-which-instance-set-it-is-a-part-of-dev-qa-staging-production-demo-some-oth)

pandafulmanda:
awesome, let me know when i can merge with BE

PR #2177, opened 2018-03-14
Fix reference view for APUSH content

nathanstitt:
Fixes the reference book so that it will use the first available section instead of defaulting to `1`


nathanstitt:
[Bug: Browse the book is not appearing on tutor-content for APUSH](https://trello.com/c/VT4f5YgJ/1376-bug-browse-the-book-is-not-appearing-on-tutor-content-for-apush)

PR #2178, opened 2018-03-14
Fix NaN% and percent / progress display toggle on scores

nathanstitt:
Fixes the bug where we returned `null` for the default case when there were assignments due but the BE didn't return a score for them (because no work or maybe a bug).

We now will display whatever score the BE sends, but if it's missing we'll look to see if we have any assignments for that type. If we do have assignments we display `n/a`, otherwise we'll display `---`

screen shot 2018-03-14 at 4 22 19 pm


nathanstitt:
[Bug: Scores shows NaN%](https://trello.com/c/N0ivnmUx/1375-bug-scores-shows-nan)

PR #2179, opened 2018-03-14
Fix weird top margin and scroll to linked annotation

nathanstitt:
When an annotation from the summary is clicked it wasn't scrolled to because it was linked via the hash which was cleared for some reason.

Before:
screen shot 2018-03-14 at 4 47 10 pm

After:
screen shot 2018-03-14 at 4 45 48 pm

### DIFF
--- a/tutor/resources/styles/components/annotations/summary-page.scss
+++ b/tutor/resources/styles/components/annotations/summary-page.scss
@@ -45,10 +45,14 @@
     }
 
     .annotation-card {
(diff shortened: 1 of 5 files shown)


nathanstitt:
[Bug: Page does not scroll to annotation when it is clicked](https://trello.com/c/rYGOZGbb/1377-bug-page-does-not-scroll-to-annotation-when-it-is-clicked)

PR #2182, opened 2018-03-15
close highlight summary window shade when esc key is pressed

nathanstitt:
In order to test the effect, this changes annotations to be read fro props with a default so they can be set in the spec


nathanstitt:
[Change Request: Improvement to the highlighting widget](https://trello.com/c/iu0tmeen/1291-change-request-improvement-to-the-highlighting-widget)

PR #2183, opened 2018-03-15
Removed the `archive` arg requirement for generating rev-manifest.json

Dantemss:
Needed so Jenkins builds the archive properly.


nathanstitt:
@Dantemss when should this be merged? I'm guessing it should go out along with the corresponding changes to the deployment scripts?

PR #2186, opened 2018-03-19
use index for finding student's score not identifier

nathanstitt:
Besides being inefficient, idenifier might be duplicated causing incorrect averages to be displayed

### DIFF
--- a/tutor/src/screens/scores-report/overall-cell.jsx
+++ b/tutor/src/screens/scores-report/overall-cell.jsx
@@ -1,12 +1,10 @@
 import React from 'react';
 import cn from 'classnames';
-import { find } from 'lodash';
 import { Cell } from 'fixed-data-table-2';
-import { observer, PropTypes as MobxPropTypes } from 'mobx-react';
+import { observer } from 'mobx-react';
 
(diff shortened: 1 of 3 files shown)


tutor/src/screens/scores-report/overall-cell.jsx
```diff
@@ -1,12 +1,10 @@
...
 import { Cell } from 'fixed-data-table-2';
-import { observer, PropTypes as MobxPropTypes } from 'mobx-react';
+import { observer } from 'mobx-react';
 
-const OverallCell = observer(({ ux, students, rowIndex }) => {
-  const { student_identifier } = students[rowIndex];
-  const studentAverages = find(ux.periodStudentsAverages, { student_identifier });
+const OverallCell = observer(({ ux, rowIndex }) => {
+  const studentAverages = ux.periodStudentsAverages[rowIndex];
   return (
```
nathanstitt: @Dantemss this is the bug, we were finding by student_identifier instead of using the index

Dantemss: OK I think this will work.

PR #2187, opened 2018-03-19
Fix student dashboard past events to list all events

nathanstitt:
We were using the `w` format which made didn't add a leading 0, so weeks less than 10 didn't sort properly.

Testing also revealed that we need to use the GGGG format otherwise dates that belong to a week that overlaps years such as `2017-01-03`, would become `201601` when formatted, rather than the proper `201701`


nathanstitt:
[Bug: old assignments missing from past work](https://trello.com/c/24BCwUDE/1384-bug-old-assignments-missing-from-past-work)

PR #2191, opened 2018-03-21
Fix ref book exercises

pandafulmanda:
embedded exercises were not rendering in reference book

### DIFF
--- a/tutor/src/components/book-page.cjsx
+++ b/tutor/src/components/book-page.cjsx
@@ -6,7 +6,7 @@ classnames = require 'classnames'
 
 {BookContentMixin} = require './book-content-mixin'
 
-{default: {BookPageExerciseShell}} = require './book-page/exercise'
+{default: {ReferenceBookExerciseShell}} = require './book-page/exercise'
 RelatedContent = require './related-content'
 {default: Loading} = require './loading-screen'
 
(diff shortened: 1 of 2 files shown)


tutor/src/components/book-page.cjsx
```diff
@@ -6,7 +6,7 @@ classnames = require 'classnames'
 
 {BookContentMixin} = require './book-content-mixin'
 
-{default: {BookPageExerciseShell}} = require './book-page/exercise'
+{default: {ReferenceBookExerciseShell}} = require './book-page/exercise'
 RelatedContent = require './related-content'
```
nathanstitt: That's a strange way to require a named export, I've usually seen that as:
```
{ReferenceBookExerciseShell} = require './book-page/exercise'
```

Maybe there's a reason for doing it that way though?

tutor/src/components/book-page/exercise.jsx
```diff
@@ -37,7 +37,7 @@ function ReferenceBookExercise(props) {
   }
 
   return (
-    <Question model={question} />
+    <Question question={question} />
   );
```
nathanstitt: I missed this when I changed the prop to be `question`, glad you caught it!

tutor/src/components/book-page.cjsx
pandafulmanda: oh yeah! 👍

PR #2195, opened 2018-03-26
only cache webpack when cache ENV is set

nathanstitt:
HardSource keeps causing errors for me, so this flips the logic so it's only used if the CACHE env is set


Dantemss:
Feel free to just remove it if it's causing issues tbh

nathanstitt:
Nah, I like it for cases where you startup webpack a lot, like when the screenshot specs are running

PR #2197, opened 2018-03-26
Disable highlighting for physics

nathanstitt:
this is (hopefully) a stop-gap measure since highlighting math is pretty bad


nathanstitt:
[Bug:highlighting math is broken, disable highlighting for physics](https://trello.com/c/CvZMHaCC/1392-bughighlighting-math-is-broken-disable-highlighting-for-physics)

PR #2198, opened 2018-03-26
Fix "Waiting for page to load.." message for pages with no images

nathanstitt:
We wait until all the images are loaded before displaying annotations and hiding the message, but it would never resolve if the page had no images


nathanstitt:
[Bug: Firefox - waiting for page to finish loading doesn't go away on readings and left/right arrows aren't working for reading navigation](https://trello.com/c/ATkR5yDj/1391-bug-firefox-waiting-for-page-to-finish-loading-doesnt-go-away-on-readings-and-left-right-arrows-arent-working-for-reading-naviga)

PR #2194, opened 2018-03-23
Fix ql exclusions

pandafulmanda:
https://trello.com/c/jPClXWqR/1387-bug-question-library-exclusions-are-showing-up-in-homework-creation

- [ ] rebase against `release-20180326`
- [ ] add specs

PR #2199, opened 2018-03-28
Hide excluded exercises on HW builder

nathanstitt:
extracted from https://github.com/openstax/tutor-js/pull/2194 + some specs


nathanstitt:
[Bug: Question Library exclusions are showing up in Homework creation](https://trello.com/c/jPClXWqR/1387-bug-question-library-exclusions-are-showing-up-in-homework-creation)

PR #2200, opened 2018-03-28
style next button and note the location of BE styles

nathanstitt:
Before
screen shot 2018-03-28 at 12 00 22 pm

After
screen shot 2018-03-28 at 11 59 42 am


nathanstitt:
[Make research console surveys appear the same as they will in Tutor](https://trello.com/c/Sc2r2HNz/1394-make-research-console-surveys-appear-the-same-as-they-will-in-tutor)

PR #2202, opened 2018-03-28
Fix styles for menu changing after viewing surveys

nathanstitt:
Otherwise they pollute Tutor's styles and make various things strange, such as the logout button


nathanstitt:
[Bug: Styles are incorrect after loading survey page](https://trello.com/c/koYVrxJ8/1410-bug-styles-are-incorrect-after-loading-survey-page)

PR #2204, opened 2018-03-30
Fix survey range and misc styles

nathanstitt:
* Only apply styles to research-surveys course page
 * Set a max-width
 * Style button-groups
 * Style alerts and move them below question

screen shot 2018-03-30 at 2 37 40 pm


nathanstitt:
[Change Request: Survey changes](https://trello.com/c/dS3tcQJD/1401-change-request-survey-changes)

PR #2208, opened 2018-04-10
set color for completed-but-not due pie progress

nathanstitt:
before:
screen shot 2018-04-10 at 1 05 09 pm

after:
screen shot 2018-04-10 at 1 04 36 pm


nathanstitt:
[Bug: pie charts should be light gray until it’s due, then switch to blue](https://trello.com/c/dwlH8t0T/1407-bug-pie-charts-should-be-light-gray-until-its-due-then-switch-to-blue)

PR #2209, opened 2018-04-10
Fix calendar loading multiple times and deleting assignments

nathanstitt:
The reload was caused by create the dashboard render function each time it was called, causing the
react component tree to be unmounted/mounted back-to-back. This broke the loading since the teacher-dashboard/index.jsx held the loader

The deleted remaining on cal was because the flux store didn't set it to 'deleted' in the task listing


nathanstitt:
[Change Request: reduce the number of instructor dashboard page reloads](https://trello.com/c/aVFBJGa4/1410-change-request-reduce-the-number-of-instructor-dashboard-page-reloads)

nathanstitt:
[Bug: Deleting a draft assignment does not remove the draft from the calendar until it is reloaded (maybe race condition)](https://trello.com/c/WBb5nK7z/1409-bug-deleting-a-draft-assignment-does-not-remove-the-draft-from-the-calendar-until-it-is-reloaded-maybe-race-condition)

PR #2211, opened 2018-04-11
don't use scopes for keybinding, and fix paging loop

nathanstitt:
Since only one scope can be active at once, the scopes would cancel each other out.

Also discovered a bug where the mount/unmount could occur in the same event loop causing the page to flip rapidly through several pages


nathanstitt:
[Bug: arrow navigation isn't working on Tutor readings](https://trello.com/c/37eRieSi/1408-bug-arrow-navigation-isnt-working-on-tutor-readings)

PR #2212, opened 2018-04-11
Misc fixes from review

nathanstitt:
* Moves toast down a bit
* fix pie progress being dark grey even for unfinished portions
* Add events popup menu on calendar wasn't navigating on click


nathanstitt:
[Bug: pie charts should be light gray until it’s due, then switch to blue](https://trello.com/c/dwlH8t0T/1407-bug-pie-charts-should-be-light-gray-until-its-due-then-switch-to-blue)

PR #2214, opened 2018-04-12
load past tasks

pandafulmanda:


tutor/src/screens/teacher-dashboard/month.jsx
```diff
@@ -147,7 +147,8 @@ class CourseMonth extends React.Component {
     const daysOfDuration = calendarDuration.iterate('days');
 
     while (daysOfDuration.hasNext()) {
-      mods.push(makeModForDate(daysOfDuration.next(), this.props.date, this.state));
+      // mods.push(makeModForDate(daysOfDuration.next(), this.props.date, this.state));
+      mods.push(makeModForDate(daysOfDuration.next(), this.props.date));
```
pandafulmanda: Not sure about this change, was going to fix it later, taking out state since it's not defined in this component

tutor/src/screens/teacher-dashboard/month.jsx
```diff
@@ -243,26 +244,26 @@ class CourseMonth extends React.Component {
       this.context.router.history.push(url)
     } else { // is a task plan to clone
       this.cloningPlan = extend({}, item, {
```
pandafulmanda: Modifying the `cloningPlan` observable is causing this:

screen shot 2018-04-12 at 12 44 46 pm

nathanstitt: yeah, I'm still not entirely sure what triggered that.  I "solved" it by returning a new object from the drop, instead of extending the existing observable object

nathanstitt:
[Change Request: reduce the number of instructor dashboard page reloads](https://trello.com/c/aVFBJGa4/1410-change-request-reduce-the-number-of-instructor-dashboard-page-reloads)

PR #2215, opened 2018-04-13
upgrade models to version that deals will null hasMany

nathanstitt:


nathanstitt:
[Bug: scores report fails to render when assignment is null](https://trello.com/c/4lql4QOj/1416-bug-scores-report-fails-to-render-when-assignment-is-null)

PR #2223, opened 2018-04-26
Fix scores recalculation when accepting late work

nathanstitt:
Bugs found/fixed:
 * Some of the components weren't observers so they didn't re-render after update
 * Reading assignments was reading all steps instead of only exercises
 * Lack of precision in calculations when togging from accepted/rejected

### DIFF
--- a/tutor/src/models/course/scores/task-result.js
+++ b/tutor/src/models/course/scores/task-result.js
@@ -3,6 +3,7 @@ import { computed, observable, action } from 'mobx';
 import {
   BaseModel, identifiedBy, belongsTo, hasMany, session, identifier, field,
 } from 'shared/model';
+import Big from 'big.js';
 import { TimeStore } from '../../../flux/time';
 
 @identifiedBy('course/scores/task-result')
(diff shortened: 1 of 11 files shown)


tutor/src/models/course/scores/task-result.js
```diff
@@ -130,49 +131,51 @@ export default class TaskResult extends BaseModel {
...
     //----------------------------------------------------------------+
     // scores                                                         |
     //----------------------------------------------------------------+
 
     // student
-    student[`${type}_score`] += (
-      this.scoredCorrectCount / student.scoredStepCount[type]
-    ) - ( prevCorrect / student.scoredStepCount[type] );
+    student[`${type}_score`] = student[`${type}_score`].plus(
+      (this.scoredCorrectCount / student.scoredExerciseCount[type]) -
```
Dantemss: Do these handle division by 0 properly?

Dantemss: I think a task with 0 steps or 0 exercises could end up with a 0 here though?

PR #2225, opened 2018-04-26
Don't allow non-college instructors to use Tutor

nathanstitt:
screen shot 2018-04-26 at 10 53 37 am


Dantemss:
I have to point out they are not necessarily high school. They might be homeschool teachers or something. Whatever categories can get confirmed faculty status but are not college.

Dantemss:
Travis is unhappy.

PR #2228, opened 2018-04-26
Fixes from ux-review of RC candidate

nathanstitt:


nathanstitt:
[Bio 2e Phase 1 -- Notify instructors that Bio 2e is coming July 20, 2018](https://trello.com/c/wyODmtTk/1423-bio-2e-phase-1-notify-instructors-that-bio-2e-is-coming-july-20-2018)

PR #2229, opened 2018-04-27
Fix how scores are calculated when accepting/rejecting late work

nathanstitt:
Previously we were averaging the steps and using that, but that method didn't match how the BE calculated them, which caused the score to be slightly incorrect.


tutor/src/models/course/scores/heading.js
```diff
@@ -0,0 +1,44 @@
...
+  @action adjustScores() {
+    this.average_score = this.averageForType('score');
+    this.average_progress = this.averageForType('progress');
+  }
+
+  averageForType(attr) {
+    return reduce(this.tasks,
+      (acc, s) => acc.plus(s[attr] || 0),
+      new Big(0)
+    ).div(this.tasks.length);
```
Dantemss: Might need to handle when there are 0 tasks.

tutor/src/models/course/scores/period.js
```diff
@@ -0,0 +1,64 @@
...
+      .plus(this.overall_reading_score.times(course.reading_score_weight / 100))
+      .plus(this.overall_reading_progress.times(course.reading_progress_weight / 100));
+
+  }
+
+  averageForType(attr) {
+    return reduce(this.students,
+      (acc, s) => acc.plus(s[attr] || 0),
+      new Big(0)
+    ).div(this.students.length);
```
Dantemss: Same here, might have to handle 0 students.

tutor/src/models/course/scores/student.js
```diff
@@ -0,0 +1,53 @@
...
+      .plus(this.reading_score.times(course.reading_score_weight / 100))
+      .plus(this.reading_progress.times(course.reading_progress_weight / 100));
+  }
+
+  averageTasksOfType(type, attr) {
+    const tasks = this.data.filter(t => t.is_included_in_averages && t.type === type);
+    return reduce(tasks,
+      (acc, t) => acc.plus(t[attr] || 0),
+      new Big(0)
+    ).div(tasks.length);
```
Dantemss: 0 tasks included in averages.

tutor/src/models/course/scores/task-result.js
```diff
@@ -33,11 +33,23 @@ export default class TaskResult extends BaseModel {
...
 
   @computed get columnIndex() {
     return findIndex(this.student.data, s => s === this);
   }
 
+  @computed get progress() {
+    let count;
+    if ('homework' === this.type) {
+      count = this.completed_on_time_exercise_count + this.completed_accepted_late_exercise_count;
+      return count / this.exercise_count;
```
Dantemss: I think this denominator needs to include placeholder steps. In case they did not do enough exercises to trigger the placeholders.

tutor/src/models/course/scores/task-result.js
```diff
@@ -33,11 +33,23 @@ export default class TaskResult extends BaseModel {
...
+    } else {
+      count = this.completed_on_time_step_count + this.completed_accepted_late_step_count;
+      return count / this.step_count;
```
Dantemss: Might be a good idea to handle tasks with 0 steps.

Dantemss: Actually maybe you can just use this logic for all task types.

tutor/src/models/course/scores/task-result.js
```diff
@@ -124,58 +119,16 @@ export default class TaskResult extends BaseModel {
...
   }
 
-  adjustAverages(prevCompletedStepCount, prevCorrect) {
-    const oldScore = this.score;
-    const period = this.student.period;
-    const { course } = period.coursePeriod;
-    const { type } = this;
-    let student = this.student;
+  @action adjustScore() {
+    this.is_included_in_averages = true;
```
Dantemss: Not sure if this is correct. Only tasks that are past-due, have at least 1 exercise and are reading or homework should be in the averages.

PR #2230, opened 2018-04-30
update copy for no-hs teachers message

nathanstitt:
screen shot 2018-04-30 at 10 36 39 am


nathanstitt:
[Change Request: only allow higher ed teachers to create Tutor courses](https://trello.com/c/nwTcoFWn/1298-change-request-only-allow-higher-ed-teachers-to-create-tutor-courses)

PR #2231, opened 2018-04-30
Course weights are in percents already

nathanstitt:
Don't need to divide by 100, doing so causes the average to be way to low


nathanstitt:

PR #2234, opened 2018-04-30
Fixed the sectionizer

Dantemss:
I couldn't find any references anywhere to getCurrentSection so I removed it. But let me know if that was a mistake.


tutor/src/components/exercises/sectionizer.jsx
```diff
@@ -27,7 +26,7 @@ class Sectionizer extends React.Component {
   windowSize = new WindowSize(this.props.windowImpl);
 
   @action.bound renderCount() {
-    ((Math.floor( this.windwSize.width - this.props.nonAvailableWidth) / 42) - 2);
+    return (Math.floor( this.windowSize.width - this.props.nonAvailableWidth) / 42) - 2;
   }
```
nathanstitt: This is completely my fault since I put it there, but I think we should remove the `@action.bound` decorator as well because 1) it's not an action - actions modify state. and 2) the `.bound` part isn't needed since we don't call it from a event handler.


I agree, `getCurrentSection` doesn't appear to be used, not sure what it's purpose was.

PR #2238, opened 2018-05-04
Update ref-view url to use ecosystemId

nathanstitt:
And redirect old urls to it.  There's still a very few things that need a course id, in those cases we search the loaded courses for one that has the correct ecosystem and use it


nathanstitt:

PR #2242, opened 2018-05-09
hide preview from non-college instructors

nathanstitt:
Prevents a 403 because they can't fetch the offering

### DIFF
--- a/tutor/src/components/my-courses/listings.jsx
+++ b/tutor/src/components/my-courses/listings.jsx
@@ -105,7 +105,7 @@ class MyCoursesTitle extends React.Component {
 }
 
 @observer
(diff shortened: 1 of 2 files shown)


tutor/src/components/my-courses/listings.jsx
```diff
@@ -196,7 +196,7 @@ function ExploreAPreview({ course }) {
 }
 
 @observer
-export class MyCoursesPreview extends React.PureComponent {
+export class MyCoursesPreview extends React.Component {
```
nathanstitt: the change here is that we're no longer using PureComponent because it'll cause an error when/if we ever upgrade react

PR #2243, opened 2018-05-09
Fix items detected during internal review

nathanstitt:
* The menu's view book was using the old url
* Reference view was lacking course styles for it's banners.  That was traced back to the ux missing a return so the data-attributes were missing


nathanstitt:
[Bug: Bio, Physics, and Soc 2e reference view and reading view have broken feature styling](https://trello.com/c/sxQt5MOS/1439-bug-bio-physics-and-soc-2e-reference-view-and-reading-view-have-broken-feature-styling)

PR #2245, opened 2018-05-11
Update book urls to have "section" in between id and chapter/section

nathanstitt:
before they were `/book/<ecosystemId>/<chapterSection>` but that didn't match what the BE used.

And also display a simple "not found" message instead of displaying spinner forever
screen shot 2018-05-11 at 1 33 10 pm


nathanstitt:

PR #2246, opened 2018-05-14
Fix exercise books tags controls and error msg

nathanstitt:
screen shot 2018-05-14 at 1 20 55 pm


nathanstitt:
[Bug: exercise tags book selector doesn't work](https://trello.com/c/B1M3q8b1/1462-bug-exercise-tags-book-selector-doesnt-work)

PR #2247, opened 2018-05-15
Redirect non-college teachers to the locked out message

nathanstitt:
Don't attempt to load offerings, instead redirect them to the "thanks for your interest" message

screen shot 2018-05-15 at 5 36 58 pm


nathanstitt:
[Change Request: only allow higher ed teachers to create Tutor courses](https://trello.com/c/nwTcoFWn/1298-change-request-only-allow-higher-ed-teachers-to-create-tutor-courses)

PR #2249, opened 2018-05-22
Adjust image layout so they're either 1/2 or full width

nathanstitt:
After:
screen shot 2018-05-22 at 4 31 35 pm
screen shot 2018-05-22 at 4 31 07 pm


nathanstitt:
[Change Request: Images display half- or full-width](https://trello.com/c/2sxMuFac/1371-change-request-images-display-half-or-full-width)

PR #2257, opened 2018-05-28
Hide section chooser when exercises are being displayed

nathanstitt:
this changes the HW builder and QL so they hide the sections chooser once the exercises are displayed, which prevents scrolling up.  That way the controls are always locked against the navbar.


nathanstitt:
Closing because we decided not to implement at this time.

When/if we choose to do so, we'll also need to add back buttons on a footer and change the cards view to push to the browser history so the back button goes back

PR #2261, opened 2018-06-11
Fix CNX-module editing on exercises

nathanstitt:
the "value" wasn't being initialized causing the field to always be empty.

### DIFF
--- a/exercises/src/components/tags/multi-input.jsx
+++ b/exercises/src/components/tags/multi-input.jsx
@@ -23,6 +23,10 @@ class Input extends React.Component {
   @observable errorMsg = '';
   @observable value = '';
 
+  componentDidMount() {
+    this.value = this.props.tag.value;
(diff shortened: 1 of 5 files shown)


nathanstitt:
[Bug: UUIDs not populating in tags tab](https://trello.com/c/91nn8EWq/1488-bug-uuids-not-populating-in-tags-tab)

exercises/src/components/tags/multi-input.jsx
```diff
@@ -23,6 +23,10 @@ class Input extends React.Component {
   @observable errorMsg = '';
   @observable value = '';
 
+  componentDidMount() {
+    this.value = this.props.tag.value;
+  }
+
   @action.bound onChange(ev) {
```
nathanstitt: this is the actual bug, we were never setting an initial value here

PR #2262, opened 2018-06-14
include counter-resets in book content instead of book-page

nathanstitt:
This way they're available everywhere such as readings

Broken images are because I have very old content, but the figures are working now 😁 
screen shot 2018-06-14 at 1 04 17 pm


nathanstitt:
[Bug: Bio2e labels all figures with Figure 1](https://trello.com/c/Wl0u66PG/1500-bug-bio2e-labels-all-figures-with-figure-1)

PR #2265, opened 2018-06-21
Don't allow use of bio1 past summer 2018

nathanstitt:
The if clause was malformed and would return false for any
year other than 2018, such as 2019 :-(


nathanstitt:
[Update Tutor Bio to use Tutor Bio 2e - DO NOT RELEASE until we are ready to add Bio 2e content to production](https://trello.com/c/JBPYwxhE/1343-update-tutor-bio-to-use-tutor-bio-2e-do-not-release-until-we-are-ready-to-add-bio-2e-content-to-production)

PR #2271, opened 2018-06-28
Delete attachments on the server

nathanstitt:
Instead of just doing so in memory :(

### DIFF
--- a/exercises/src/api/index.js
+++ b/exercises/src/api/index.js
@@ -45,7 +45,7 @@ const start = function() {
 const CSRF_TOKEN = get(document.head.querySelector('meta[name=csrf-token]'), 'content');
 
 Exercise.prototype.uploadImage = function(image, cb) {
-  const url = `/api/exercises/${this.number}@draft/attachments`;
+  const url = `/api/exercises/${this.uid}/attachments`;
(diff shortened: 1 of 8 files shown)


nathanstitt:
[Bug: Can't add assets to existing assessments](https://trello.com/c/7Gvv2Tp5/1503-bug-cant-add-assets-to-existing-assessments)

exercises/src/api/index.js
```diff
@@ -33,12 +33,19 @@ const start = function() {
...
+    onFail: 'onError',
+  });
+
 };
 
 const CSRF_TOKEN = get(document.head.querySelector('meta[name=csrf-token]'), 'content');
 
 Exercise.prototype.uploadImage = function(image, cb) {
-  const url = `/api/exercises/${this.number}@draft/attachments`;
+  const url = `/api/exercises/${this.uid}/attachments`;
```
Dantemss: I think this might fail if the exercise is published?

nathanstitt: Yeah, I'd changed that in hopes that it'd fix the "needs draft to save" bug, and it did, but then the images don't show up for new drafts when it's loaded.  I didn't really mean to check this in, am glad you noticed

PR #2272, opened 2018-06-28
Fix the continue button on reading steps

nathanstitt:
The continue button for reading content didn't work.  It would scroll to the top of the page but wouldn't display new content.


nathanstitt:
[Bug: The continue button on readings is broken](https://trello.com/c/bXX3Yr5h/1511-bug-the-continue-button-on-readings-is-broken)

PR #2273, opened 2018-06-29
prevent global button style reset by scoping under surveys

nathanstitt:
Otherwise the button styles are reset to the bootstrap defaults


nathanstitt:
[Bug: button on-hover is an odd color pair](https://trello.com/c/0UaeNLT4/1512-bug-button-on-hover-is-an-odd-color-pair)

PR #2275, opened 2018-07-02
Fix controls not pinning in safari

nathanstitt:
Also fixes:
  * the loader displaying prematurely on HW builder
  * The "no exercises found" message wasn't showing if only an intro section was selected


nathanstitt:
[Bug: Nav/toolbar for QL and Homework builder does not pin anymore.](https://trello.com/c/hjjKaGQs/1518-bug-nav-toolbar-for-ql-and-homework-builder-does-not-pin-anymore)

nathanstitt:
yeah, the scroll-pinning stuff would be nice to simplify.  it's one of the first react things we wrote and it kind of shows

PR #2276, opened 2018-07-02
add open-ended format

nathanstitt:
screen shot 2018-07-02 at 4 02 44 pm

### DIFF
--- a/shared/src/model/exercise/answer.js
+++ b/shared/src/model/exercise/answer.js
@@ -14,9 +14,7 @@ export default class ExerciseAnswer extends BaseModel {
   @belongsTo({ model: 'exercise/question' }) question;
 
   @computed get validity() {
-    if (
-      isEmpty(this.content_html) && !get(this, 'question.isOpenEnded')
-    ) {
+    if (isEmpty(this.content_html)) {
       return { valid: false, part: 'Answer Distractor' };
     } else {
       return { valid: true };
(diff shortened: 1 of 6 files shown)


nathanstitt:
[Allow open-ended assessments in Exercises](https://trello.com/c/Iccbaq0h/1283-allow-open-ended-assessments-in-exercises)

shared/src/model/exercise/answer.js
```diff
@@ -14,7 +14,9 @@ export default class ExerciseAnswer extends BaseModel {
   @belongsTo({ model: 'exercise/question' }) question;
 
   @computed get validity() {
-    if (isEmpty(this.content_html)) {
+    if (
+      isEmpty(this.content_html) && !get(this, 'question.isOpenEnded')
```
Dantemss: I don't think an empty answer is valid, instead the answers array should be empty?

shared/src/model/exercise/question.js
```diff
@@ -30,6 +31,10 @@ export default class ExerciseQuestion extends BaseModel {
     return this.hasFormat('multiple-choice');
   }
 
+  @computed get isOpenEnded() {
+    return this.hasFormat('open-ended');
```
Dantemss: I think we were using free-response (which most questions have because most questions CAN be answered free-response) but to use that you might need logic that says the free-response format is the only format.

PR #2278, opened 2018-07-03
only name and other info from source course if it's clonable

nathanstitt:
If we're able to clone the course, we shouldn't set the name from it


nathanstitt:
[Change Request: When a user tries to create a new Bio 2e course by first trying to copy Bio 1e, make default name for the course "Biology 2e"](https://trello.com/c/slYYKaNv/1515-change-request-when-a-user-tries-to-create-a-new-bio-2e-course-by-first-trying-to-copy-bio-1e-make-default-name-for-the-course-b)

PR #2280, opened 2018-07-06
update refund policy link, change contact to email

nathanstitt:


nathanstitt:
[Bug: refund and contact support link on the payments screen is broken](https://trello.com/c/skpCOhN3/1519-bug-refund-and-contact-support-link-on-the-payments-screen-is-broken)

PR #2281, opened 2018-07-06
always update new course name to match the offering

nathanstitt:
This way it will update to match the new bio2 offering instead of being stuck on the bio1 name


nathanstitt:
[Change Request: When a user tries to create a new Bio 2e course by first trying to copy Bio 1e, make default name for the course "Biology 2e"](https://trello.com/c/slYYKaNv/1515-change-request-when-a-user-tries-to-create-a-new-bio-2e-course-by-first-trying-to-copy-bio-1e-make-default-name-for-the-course-b)

PR #2285, opened 2018-07-10
handle long content in error modal

TomWoodward:
also changed overflow from `scroll` to `auto` so it doesn't show unnecessary scroll bars

image

### DIFF
--- a/tutor/resources/styles/components/tutor-dialog.scss
+++ b/tutor/resources/styles/components/tutor-dialog.scss
@@ -1,9 +1,11 @@
 .tutor-dialog {
 
   .server-error {
+    word-break: break-all;
+
     .response {
       max-height: 200px;
-      overflow: scroll;
+      overflow: auto;
     }
   }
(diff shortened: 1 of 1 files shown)


tutor/resources/styles/components/tutor-dialog.scss
```diff
@@ -1,9 +1,11 @@
 .tutor-dialog {
 
   .server-error {
+    word-break: break-all;
+
     .response {
       max-height: 200px;
-      overflow: scroll;
+      overflow: auto;
     }
```
nathanstitt: nice catch, auto's better than `scroll`

TomWoodward:
[Bug: Content spilling out of error box](https://trello.com/c/Y53P1wPI/1551-bug-content-spilling-out-of-error-box)

nathanstitt:
awesome way to repro the bug btw!  Looks like the BE isn't sanitizing the inputs very well 😁

PR #2287, opened 2018-07-10
update styles for popovers on student scores page

TomWoodward:
this is what it used to look like:
image

these are what it looks like now:
image
image
image

### DIFF
--- a/tutor/src/screens/scores-report/late-work.jsx
+++ b/tutor/src/screens/scores-report/late-work.jsx
@@ -177,12 +177,10 @@ export class LateWorkPopover extends React.PureComponent {
     const { content } = this;
 
     const status = this.props.task.type === 'homework' ? content.score() : content.progress();
-    const arrowOffsetTop = TH.hasAdditionalLateWork(this.props.task) ? 128 : 95;
     const popoverProps = omit(this.props, 'hide', 'task', 'show', 'columnIndex');
     return (
       <Popover
         {...popoverProps}
-        arrowOffsetTop={arrowOffsetTop}
         title={content.get('title')}
         id={`late-work-info-popover-${content.task.id}`}
         className={content.className()}>
--- a/tutor/src/screens/scores-report/styles/table.scss
+++ b/tutor/src/screens/scores-report/styles/table.scss
@@ -315,7 +315,6 @@ $pie-color-due: #268FBE;
 }
 
 .late-work-info-popover {
-  height: 98px;
   min-width: 315px;
   border-radius: 0;
   border: 1px solid $tutor-neutral-light;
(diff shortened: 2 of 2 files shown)


TomWoodward:
[Bug: new late score message doesn't fit in the box](https://trello.com/c/rI9ZpMjl/1552-bug-new-late-score-message-doesnt-fit-in-the-box)

tutor/src/screens/scores-report/styles/table.scss
```diff
@@ -346,41 +342,38 @@ $pie-color-due: #268FBE;
...
-        display: table-cell;
-        vertical-align: middle;
-        color: $tutor-primary;
-        text-align: center;
+      .late-button {
+        flex: 1;
+        margin: 3px;
+        white-space: normal;
+        text-transform: none !important;
+        font-size: 1.1em !important;
```
nathanstitt: yay for replacing `table-cell` with `flex`!   

Why is the `!important` needed?  I'm guessing it's to override of some other rule?  If it doesn't make it too convoluted I'd prefer to match the specific rule by adding more specific selectors.

If that does make it too complex can we add a comment indicating what it's overriding in hopes (which yeah, probably won't happen) that we can remove the !important if it's ever no longer needed?

tutor/src/screens/scores-report/late-work.jsx
```diff
@@ -177,12 +177,10 @@ export class LateWorkPopover extends React.PureComponent {
     const { content } = this;
 
     const status = this.props.task.type === 'homework' ? content.score() : content.progress();
-    const arrowOffsetTop = TH.hasAdditionalLateWork(this.props.task) ? 128 : 95;
     const popoverProps = omit(this.props, 'hide', 'task', 'show', 'columnIndex');
     return (
       <Popover
         {...popoverProps}
-        arrowOffsetTop={arrowOffsetTop}
         title={content.get('title')}
```
nathanstitt: nice catch, I have no idea why were were setting that

tutor/src/screens/scores-report/styles/table.scss
TomWoodward: i'm not sure i didn't modify these styles i just moved them, do you want me to try to clean it up some more?

tutor/src/screens/scores-report/late-work.jsx
TomWoodward: i also deleted some styles that were making the popover taller for no reason i could figure out

nathanstitt: yes, I saw that, I also have no idea why we had those.  Maybe we had additional absolutely positioned content in the popover at one time?

tutor/src/screens/scores-report/styles/table.scss
nathanstitt: oh, I see.  Yes they had the `!important` originally.  Ok, you've def made the styles better so I think we can cut off the improvements here otherwise we could go on forever 😁

PR #2288, opened 2018-07-10
use modal for assignment delete instead of popover

TomWoodward:
image

before i updated the modal padding it looked like this:
image

### DIFF
--- a/tutor/src/components/task-plan/footer/delete-link.jsx
+++ b/tutor/src/components/task-plan/footer/delete-link.jsx
@@ -1,7 +1,38 @@
-import React from 'react';
-import { AsyncButton, SuretyGuard } from 'shared';
+import { React, observer, observable, action } from '../../../helpers/react';
+import { AsyncButton } from 'shared';
 import Icon from '../../icon';
+import { Modal, Button } from 'react-bootstrap';
 
+const DeleteModal = ({ message, show, onClose, isBusy, onDelete }) => (
+  <Modal
+    show={show}
+    onHide={onClose}
+    className="settings-delete-assessment-modal">
+    <Modal.Header closeButton={true}>
+      <Modal.Title>
+        Delete
+      </Modal.Title>
+    </Modal.Header>
+    <Modal.Body>
+      <p>{message}</p>
+    </Modal.Body>
+    <Modal.Footer>
+      <AsyncButton
+        className="delete"
+        bsStyle="danger"
+        onClick={onDelete}
+        waitingText="Deleting…"
+        isWaiting={isBusy}>
+        Delete
+      </AsyncButton>
+      <Button disabled={isBusy} onClick={onClose}>
+        Cancel
+      </Button>
+    </Modal.Footer>
+  </Modal>
+);
+
+@observer
 export default class DeleteLink extends React.PureComponent {
 
   static propTypes = {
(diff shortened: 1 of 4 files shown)


TomWoodward:
[Change Request: Change "delete assignment" popup into a modal, align w/current modal styling.](https://trello.com/c/DamavnHU/1549-change-request-change-delete-assignment-popup-into-a-modal-align-w-current-modal-styling)

nathanstitt:
It sounds like Roberts in favor of no longer using the popup based version of `SuretyGuard`. 

I think we only use that a few other places, so if he's ok with it let's rewrite `SuretyGuard` so that it uses a modal like you wrote above instead of the `Popover` it uses now.

It lives in https://github.com/openstax/tutor-js/blob/master/shared/src/components/surety-guard.cjsx 

Note that it's currently coffee script based, but I just wrote a JSX version for a15k at https://github.com/a15k/mothership/pull/16/files#diff-990cd117ea7b874b44aee92992e6f5fa 
and even a spec https://github.com/a15k/mothership/pull/16/files#diff-cd06803a8cd16adb59f210386b50660b

Also there appears to be some extra horizontal padding on the modal title so that it doesn't align horizontally with the body text.  Can you investigate that and also check if any of our other modals are having the same bug?  I'm guessing they might since we're setting some global styles on the modals.

TomWoodward:
@nathanstitt that header padding is being set here

https://github.com/openstax/tutor-js/blob/master/tutor/resources/styles/components/modals.scss#L11

i can definitely change it if you want

nathanstitt:
Hm, so it is.

it looks like the separate header styles were set here:
https://github.com/openstax/tutor-js/commit/721226bb733de68dbf607284069549f7e6f61560. That commit relates to setting the scores reports weights though, and then a later commit set the set weights padding differently: https://github.com/openstax/tutor-js/commit/a0455658f91904f68c736aad0dbeb216ed8e5ab8

Based on that I think it's a mistake that the global styles have a different horizontal padding.  I believe it should match the body.

TomWoodward:
@nathanstitt upon mature reflection the actual problem is that this variable is set twice to different values:

https://github.com/openstax/tutor-js/blob/7627a9d098908031e1dc77ac6864da21cde0896d/tutor/resources/styles/components/modals.scss#L3
https://github.com/openstax/tutor-js/blob/57c2336bd6956ee2414c76bde5823b41b5642a76/tutor/resources/styles/variables.scss#L95

i'm removing the one in modals.scss so that everything will use the 20px padding

tutor/src/components/task-plan/footer/delete-link.jsx
```diff
@@ -1,7 +1,40 @@
 import React from 'react';
-import { AsyncButton, SuretyGuard } from 'shared';
+import { observer } from 'mobx-react';
+import { observable, action } from 'mobx';
+import { AsyncButton } from 'shared';
 import Icon from '../../icon';
```
nathanstitt: Just so you're aware, we've just added https://github.com/openstax/tutor-js/blob/master/tutor/src/helpers/react.js which exports common things.  The above could be: 
`import { React, observer, observable, action } from '../../helpers/react';

nathanstitt:
@TomWoodward actually for this we can go to the OpenStax pattern library which sets modal padding to be `$normal-margin*2`
https://github.com/openstax/pattern-library/blob/810c7ea897c093348203819fd306c3f2da9332c1/core/pattern-library/molecules.scss#L21
and `normal-margin` is 1.5rem https://github.com/openstax/pattern-library/blob/810c7ea897c093348203819fd306c3f2da9332c1/core/pattern-library/layout.scss#L1 which a rem being 10px, so the padding should be:  `30px`

TomWoodward:
@nathanstitt this is ready for more review

nathanstitt:
Looks good @TomWoodward thanks!

PR #2292, opened 2018-07-12
remove annotation drawer from dom when closed, and allow tabbing to elements inside

TomWoodward:
There is a little bit of a hack to ensure that when you tab after the drawer opens, it selects the drawer content first. otherwise you have to tab through all the content behind before getting to the stuff you can see. If anyone knows a cleaner way to accomplish that please let me know.


TomWoodward:
[Bug: Make highlighting summary page accessible](https://trello.com/c/vPnqJP4a/1556-bug-make-highlighting-summary-page-accessible)

PR #2293, opened 2018-07-13
remove annotation drawer from dom when closed, and allow tabbing to elements inside

TomWoodward:
There is a little bit of a hack to ensure that when you tab after the drawer opens, it selects the drawer content first. otherwise you have to tab through all the content behind before getting to the stuff you can see. If anyone knows a cleaner way to accomplish that please let me know.

also the ref mocking in the test is super brittle, would love a better solution to that.

### DIFF
--- a/tutor/src/components/annotations/summary-page.jsx
+++ b/tutor/src/components/annotations/summary-page.jsx
@@ -30,6 +30,10 @@ export default class AnnotationSummaryPage extends React.Component {
     this.resetToSection(this.props.currentChapter, this.props.currentSection);
   }
 
+  componentDidMount() {
+    this.prepareFocus();
+  }
+
   componentWillReceiveProps(nextProps) {
     if (
       nextProps.currentSection !== this.props.currentSection ||
@@ -37,6 +41,16 @@ export default class AnnotationSummaryPage extends React.Component {
     ) {
       this.resetToSection(nextProps.currentChapter, nextProps.currentSection);
     }
+  }
+
+  prepareFocus() {
+    const {containerRef} = this;
+    const focusAnchor = document.createElement("a");
+
+    focusAnchor.setAttribute("href", "#");
+    containerRef.insertBefore(focusAnchor, containerRef.firstChild);
+    focusAnchor.focus();
+    focusAnchor.addEventListener("blur", () => containerRef.removeChild(focusAnchor), false);
   }
 
   @computed get annotationsBySection() {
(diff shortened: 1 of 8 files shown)


tutor/resources/styles/components/top-nav-bar/center-controls.scss
```diff
@@ -58,6 +58,7 @@
   .annotation-summary-toggle {
     background-color: transparent;
     border: none;
+    outline: none;
     .i {
```
nathanstitt: I think we do want the focus outline style on the summary toggle, or did I miss a request to remove it?

tutor/src/components/annotations/summary-page.jsx
```diff
@@ -39,6 +43,16 @@ export default class AnnotationSummaryPage extends React.Component {
...
 
+  prepareFocus() {
+    const {containerRef} = this;
+    const focusAnchor = document.createElement("a");
+
+    focusAnchor.setAttribute("href", "#");
+    containerRef.insertBefore(focusAnchor, containerRef.firstChild);
+    focusAnchor.focus();
+    focusAnchor.addEventListener("blur", () => containerRef.removeChild(focusAnchor), false);
+  }
```
nathanstitt: I'm afraid I don't see why we're why we're manually added the `focusAnchor` node?  Maybe there was a requirement that I missed somewhere?

tutor/src/components/annotations/window-shade.jsx
```diff
@@ -29,11 +30,20 @@ export default class WindowShade extends React.Component {
...
+      >
+        {ux.isSummaryVisible &&
+          <div key="shade" className="highlights-windowshade">
+            <div className='centered-content'>
+              {children}
+            </div>
+          </div>
+        }
+      </ReactCSSTransitionGroup>
     );
```
nathanstitt: The `React.Children.toArray(props.children)[0]` seems kinda weird to me.  I think you're doing that so that the page doesn't have an unneeded div until it's needed?

It seems like we could make `WindowShade` be responsible for conditionally rendering the SummaryPage and then do something like:

```jsx
  <ReactCSSTransitionGroup  transitionName="animate" transitionEnterTimeout={300}  transitionLeaveTimeout={300}  className="highlights-windowshade">
         {ux.isSummaryVisible && <SummaryPage {...propsFromAnnotation} />}
  </ReactCSSTransitionGroup>
```

Thoughts on that?

PR #2295, opened 2018-07-17
Fixes for exercise editing

nathanstitt:
Fixes for:
 * Form not clearing after publishing
 * Publish button was enabled even if the exercise was already published
 * Clicking new button would sometimes not display a new record
 * Answer distractor field would not be cleared when a new record was editing
 * JS error when "Exercise contains multiple parts" was checked

### DIFF
--- a/exercises/src/app.jsx
+++ b/exercises/src/app.jsx
@@ -5,6 +5,7 @@ import { Provider, observer } from 'mobx-react';
 import { action } from 'mobx';
 import UX from './ux';
 import Search from './components/search';
+import ExerciseMap from './models/exercises';
 import Exercise from './components/exercise';
 import Preview from './components/preview';
 import UserActionsMenu from './components/user-actions-menu';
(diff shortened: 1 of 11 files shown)


nathanstitt:
[Bug: multi-part assessment creation is broken](https://trello.com/c/8nfFm3Oh/1532-bug-multi-part-assessment-creation-is-broken)

exercises/src/app.jsx
```diff
@@ -35,7 +42,7 @@ export default class App extends React.Component {
                 <NavItem onClick={this.onNav} href="/search">
                   Search
                 </NavItem>
-                <NavItem onClick={this.onNav} href="/exercise/new">
+                <NavItem onClick={this.onNew} href="/exercise/new">
```
TomWoodward: it looks like onNav is now unused, can we remove it?

nathanstitt: Good thought, but onNav is used by the "Search" link.  However it could be rewritten to be `onSearch` since now it's only function.  I'll do that.

PR #2296, opened 2018-07-17
fix question answer alignment when text is long

TomWoodward:
before:
image

after:
image

### DIFF
--- a/shared/resources/styles/components/question.scss
+++ b/shared/resources/styles/components/question.scss
@@ -75,6 +75,7 @@
     }
 
     .answer-answer {
+      flex: 1;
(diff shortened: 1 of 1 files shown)


TomWoodward:
[Bug: correct answer may shift alignment to the left](https://trello.com/c/Z8WSef51/1565-bug-correct-answer-may-shift-alignment-to-the-left)

shared/resources/styles/components/question.scss
```diff
@@ -75,6 +75,7 @@
     }
 
     .answer-answer {
+      flex: 1;
       margin-left: $openstax-answer-horizontal-spacing;
```
nathanstitt: hah, that was a nice fix

PR #2297, opened 2018-07-18
remove exercise style overrides when header changes

TomWoodward:
@nathanstitt these styles were changing the document height when the navbar pinned, making it possible to be in just the right spot where the nav flashes in and out of it's pinned state.

it didn't seem like these styles were super necessary but if you let me know what they were trying to do i can try to update them to not alter the document height

### DIFF
--- a/tutor/resources/styles/components/exercises.scss
+++ b/tutor/resources/styles/components/exercises.scss
@@ -21,12 +21,6 @@ $exercise-details-body-height: calc(100vh - 150px);
     margin-left: 450px;
     z-index: 1; // above details forward/back nav when scrolling up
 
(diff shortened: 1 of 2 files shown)


TomWoodward:
[Bug: Jumping Scrolling in Firefox](https://trello.com/c/IN4EVeXg/1566-bug-jumping-scrolling-in-firefox)

nathanstitt:
Looks good @TomWoodward but I think it needs some more margin when it's pinned.  It looks like this for me:

image

TomWoodward:
@nathanstitt addressed that hidden header issue

nathanstitt:
Looks great Tom, thanks

PR #2301, opened 2018-07-19
use more time units in the trial countdown nav element

TomWoodward:
￼￼￼
image

image

image

image


TomWoodward:
[Bug: Countdown message reads "0 days"](https://trello.com/c/rnzirflA/1576-bug-countdown-message-reads-0-days)

PR #2302, opened 2018-07-19
fix icons in exercise intro cards

TomWoodward:
image


TomWoodward:
[Bug: Personalized info card in student reading has old asset](https://trello.com/c/4w9yboDi/1575-bug-personalized-info-card-in-student-reading-has-old-asset)

PR #2303, opened 2018-07-20
restyle note titles to handle multiline titles

TomWoodward:
image
image

### DIFF
--- a/tutor/resources/styles/book-content/note.scss
+++ b/tutor/resources/styles/book-content/note.scss
@@ -181,14 +181,22 @@ $tutor-book-note-selector: '
     }
 
     @include tutor-book-label-style() {
-      position: absolute;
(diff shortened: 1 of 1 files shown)


tutor/resources/styles/book-content/note.scss
```diff
@@ -181,14 +181,26 @@ $tutor-book-note-selector: '
...
+        flex: 1;
+        background-color: $tutor-white;
+        display: block;
+        margin-left: 40px;
+      }
+
+      & + * {
+        clear: both;
+      }
     };
```
nathanstitt: am curious, why is the `clear` needed since nothings floated?

TomWoodward: i think this is a remnant from an earlier approach, i'll try removing it

TomWoodward:
@nathanstitt comment addressed

PR #2305, opened 2018-07-23
fix link to annotation in summary page

TomWoodward:


TomWoodward:
[Bug: View in book link on summary page doesn't scroll to correct annotation](https://trello.com/c/PgJSK7xH/1587-bug-view-in-book-link-on-summary-page-doesnt-scroll-to-correct-annotation)

tutor/src/components/annotations/annotation-card.jsx
```diff
@@ -87,7 +87,7 @@ export default class AnnotationCard extends React.Component {
   @autobind
   openPage() {
     const { id, selection: { courseId, chapter, section } } = this.props.annotation;
-    let url = `/books/${courseId}/section/${chapter}`;
+    let url = `/book/${courseId}/section/${chapter}`;
     if (section) {
```
nathanstitt: sigh, yeah we renamed that url awhile ago and thought we'd updated all the links.  @TomWoodward can you search through the code and see if you spot any other `books/` urls?

TomWoodward: @nathanstitt does this count https://github.com/openstax/tutor-js/blob/master/tutor/src/screens/new-course/bio2e-unavailable.jsx#L34 ?

nathanstitt: nah that one's to CNX, and does appear to be correct.  Thanks for checking!

PR #2307, opened 2018-07-25
remove ability to delete attachments from exercise

nathanstitt:
Since attachments are shared between versions, removing an attachment might break links on published versions


nathanstitt:
[Bug: Can't delete assets on exercises](https://trello.com/c/swv7Y86z/1589-bug-cant-delete-assets-on-exercises)

PR #2312, opened 2018-08-01
Fix loading APUSH exercises in reference view

nathanstitt:
This is one of the last users of flux stores.  Should probably be converted but am leaving it since it's been working thus far


nathanstitt:
[Open-ended exercises display in reference view](https://trello.com/c/9JFxl3KB/1557-open-ended-exercises-display-in-reference-view)

PR #2314, opened 2018-08-03
fix annotation summary page re-rendering on delete

TomWoodward:
there will now be a rando div wrapping the windowshade, but apparently the transition group needs it to keep track of the children and not re-mount them all willy nilly 

https://github.com/openstax/tutor-js/pull/2314/files?w=1 is a better diff


TomWoodward:
[Bug: Deleting highlighted text from Summary page in reference view hides/collapses the remaining highlighted text](https://trello.com/c/lT8b5kHt/1593-bug-deleting-highlighted-text-from-summary-page-in-reference-view-hides-collapses-the-remaining-highlighted-text)

PR #2315, opened 2018-08-03
fix link to annotation in summary page

TomWoodward:


TomWoodward:
[Bug: View in book link on summary page doesn't scroll to correct annotation](https://trello.com/c/PgJSK7xH/1587-bug-view-in-book-link-on-summary-page-doesnt-scroll-to-correct-annotation)

PR #2321, opened 2018-08-13
upgrade webpack and unify it's config

nathanstitt:
Greatly simplifies the webpack config into a single file


nathanstitt:
Goes along with https://github.com/openstax/tutor-server/pull/1766

TomWoodward:
@nathanstitt am i following this correctly in that its going to compile every project and just serve one of them? rather than just compiling the requested project?

PR #2324, opened 2018-08-16
Fix reference book teacher content indicator

nathanstitt:
Previously it wouldn't check for content when it was loaded


nathanstitt:
[Bug: APUSH Teacher view not working](https://trello.com/c/YGC4A4BJ/1618-bug-apush-teacher-view-not-working)

PR #2331, opened 2018-08-23
fix highlighting issues

TomWoodward:


TomWoodward:
[Bug: Images can be highlighted multiple times; no check to prevent duplicates](https://trello.com/c/Hxm6AEcU/1612-bug-images-can-be-highlighted-multiple-times-no-check-to-prevent-duplicates)

tutor/src/components/annotations/annotation.jsx
```diff
@@ -36,6 +36,26 @@ function getSelectionRect(win, selection) {
...
+  const state = { content: cloneRange.toString() };
+
+  cloneRange.selectNodeContents(referenceNode);
+  cloneRange.setEnd(startContainer, startOffset);
+
+  state.start = cloneRange.toString().length;
+  state.end = cloneRange.toString().length + state.content.length;
+
+  state.restore = serializeSelection.restore.bind(null, state, referenceNode);
+
```
nathanstitt: was `bind` used before?  I'm not sure if we have a polyfill loaded for that.  If it wasn't it might be better to use a fn, i.e. 

`state.restore = () => serializeSelection.restore(state, referenceNode)`

TomWoodward: @nathanstitt that function is mostly copy/paste out of the `serialize-selection` library with the one change that you can pass the range in instead of it grabbing it from `window.getSelection().getRangeAt(0)`, i was leaning towards changing the code as little as possible but i can clean it up if you want

nathanstitt: yep, I see the `bind` in serialize lib now, I'd missed it before I guess

PR #2333, opened 2018-08-23
better highlight overlap checking

TomWoodward:


TomWoodward:
[Bug: Images can be highlighted multiple times; no check to prevent duplicates](https://trello.com/c/Hxm6AEcU/1612-bug-images-can-be-highlighted-multiple-times-no-check-to-prevent-duplicates)

tutor/src/components/annotations/annotation.jsx
```diff
@@ -227,10 +227,10 @@ export default class AnnotationWidget extends React.Component {
...
       if (selection.referenceElementId === other.referenceElementId) {
         if (selection.start >= 0 &&
-          (other.selection.start > selection.start &&
-            other.selection.start < selection.end) ||
-          (other.selection.end > selection.start &&
-            other.selection.end < selection.end)) {
+          (other.selection.start >= selection.start &&
+            other.selection.start <= selection.end) ||
+          (other.selection.end >= selection.start &&
+            other.selection.end <= selection.end)) {
```
nathanstitt: hah, that completely makes sense and explains why the first/lat character would be unhighlightable at times.,

PR #2334, opened 2018-08-24
fix greedy highlighter

TomWoodward:


TomWoodward:
[Bug: Entire page gets highlighted in certain scenarios](https://trello.com/c/bk98GSiS/1609-bug-entire-page-gets-highlighted-in-certain-scenarios)

PR #2335, opened 2018-08-24
1608 bug view in book link in the highlight summary page has issues sporadically with math formulae

TomWoodward:
previously each `ArbitraryHtmlAndMath` would call `typesetMath` but `typesetMath` would just typeset the whole document, there was a debounce in there, but if math took longer than 100ms to process (because, auno, of hundreds of formulas in physics book sections) it could still potentially be causing problems.

so now each node calls `typesetMath` and only that nodes content will by typeset by the call, and each node gets their own debounce for good measure

also moved the typeset calls into the `MathJax.Hub.Queue` because some documentation said that everything was supposed to go through the queue, i think it helped, not sure.


TomWoodward:
[Bug: View in book link in the Highlight Summary page has issues sporadically with math formulae](https://trello.com/c/4PMbpq6O/1608-bug-view-in-book-link-in-the-highlight-summary-page-has-issues-sporadically-with-math-formulae)

TomWoodward:
@nathanstitt converted to js for you

TomWoodward:
@nathanstitt specs pass now

PR #2336, opened 2018-08-26
User interface for pairing a course to LMS

nathanstitt:
screen shot 2018-08-25 at 6 09 35 pm

### DIFF
--- a/tutor/src/models/course/pair-to-lms.js
+++ b/tutor/src/models/course/pair-to-lms.js
@@ -1,0 +1,24 @@
+import {
+  BaseModel, identifiedBy, observable, field, belongsTo, computed, session,
+} from 'shared/model';
+
+@identifiedBy('course/create')
+export default class CoursePair extends BaseModel {
+
+  @observable course;
+  @observable success;
+
+
+  constructor(course) {
+    super();
+    this.course = course;
+  }
+
+  save() {
+    return this; //{ courseId: this.course.id };
+  }
+
+  onPaired({ data: { success } }) {
+    this.success = success;
+  }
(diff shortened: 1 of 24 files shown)


tutor/src/models/course/pair-to-lms.js
```diff
@@ -0,0 +1,24 @@
...
+  @observable success;
+
+
+  constructor(course) {
+    super();
+    this.course = course;
+  }
+
+  save() {
+    return this; //{ courseId: this.course.id };
```
TomWoodward: commented code

webpack.config.js
```diff
@@ -18,7 +18,7 @@ const defaultEntry = `./${project}/index.js`;
 const ENTRIES = {
   tutor: {
     tutor: defaultEntry,
-
+    lmspair: './tutor/lms-pair.js',
```
TomWoodward: out of curiosity, why is this a new entry?

nathanstitt: This sets up a new build that's loaded by the BE for pairing a course.  it's a specialized screen that doesn't allow navigation.  Because of that we can jettison quite a bit of the existing code so it loads faster.

tutor/src/models/course/pair-to-lms.js
nathanstitt: ah thanks. yeah that was an earlier attempt before I decided to just use `this` for interpolating the url

nathanstitt:
[Willo Labs prototype](https://trello.com/c/IxecuKR8/1586-willo-labs-prototype)

PR #2340, opened 2018-08-28
fix propname to be className

nathanstitt:
we'll go with I was thinking I was using preact 😁


nathanstitt:
[Bug: Exercises - tags tab missing from multipart question when it has 5 or more questions](https://trello.com/c/rErgyfXk/1602-bug-exercises-tags-tab-missing-from-multipart-question-when-it-has-5-or-more-questions)

PR #2341, opened 2018-08-31
Add docker configs

TomWoodward:


Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN apt-get update && apt-get install -y \
+    build-essential \
+    g++ \
+    gcc \
+    git \
+    python2.7 \
+  && rm -rf /var/lib/apt/lists/*
+
+RUN ln -s /usr/bin/python2.7 /usr/bin/python2
+
```
nathanstitt: Does tutor-js need python?

TomWoodward: some of the npm modules build themselves and need it.. i forget which one

PR #2342, opened 2018-08-31
Display errors when hypothesis server malfunctions

nathanstitt:
I'd removed the call to `AppActions.setServerError(err)` in https://github.com/openstax/tutor-js/pull/2224 

Then when I was looking at the error message it had a bunch of `undefined` in it because the vars weren't being passed around correctly.

screen shot 2018-08-31 at 12 33 05 pm

screen shot 2018-08-31 at 11 34 07 am

### DIFF
--- a/tutor/src/models/annotations/hypothesis.js
+++ b/tutor/src/models/annotations/hypothesis.js
@@ -65,6 +65,8 @@ class Hypothesis extends BaseModel {
           return this.fetchAllAnnotations();
         });
       } else { return this; }
+    }).catch(err => {
+      AppActions.setServerError(err);
     });
   }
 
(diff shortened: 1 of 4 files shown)


nathanstitt:
[Bug: Make error message sound better for when a student's internet drops](https://trello.com/c/a5Hw1hcd/1645-bug-make-error-message-sound-better-for-when-a-students-internet-drops)

tutor/src/models/annotations/hypothesis.js
```diff
@@ -65,6 +65,8 @@ class Hypothesis extends BaseModel {
           return this.fetchAllAnnotations();
         });
       } else { return this; }
+    }).catch(err => {
+      AppActions.setServerError(err);
     });
```
nathanstitt: this is the bug fix we want in

PR #2344, opened 2018-09-05
Update "all_tasks_are_ready" flag when loaded

nathanstitt:
Since the class inherits from Map it doesn't auto-update like
a normal model

Also adds a safety valve where we only wait if the flag is false.

### DIFF
--- a/tutor/src/models/student-tasks.js
+++ b/tutor/src/models/student-tasks.js
@@ -1,6 +1,7 @@
 import Map from 'shared/model/map';
 import moment from 'moment-timezone';
 import { readonly } from 'core-decorators';
+
 import { computed, action, observable } from 'mobx';
 import { filter, groupBy, sortBy, pickBy } from 'lodash';
 import { TimeStore } from '../flux/time';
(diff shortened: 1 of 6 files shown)


tutor/src/models/student-tasks.js
```diff
@@ -48,9 +49,10 @@ export class CourseStudentTasks extends Map {
   }
 
   // note: the response also contains limited course and role information but they're currently unused
-  onLoaded({ data: { tasks, research_surveys } }) {
+  onLoaded({ data: { tasks, research_surveys, all_tasks_are_ready } }) {
     this.researchSurveys = research_surveys ? new ResearchSurveys(research_surveys) : null;
     this.mergeModelData(tasks);
+    this.all_tasks_are_ready = all_tasks_are_ready;
   }
```
nathanstitt: This is the bug, we never actually updated the flag when the network request finishes.  On normal models we don't need to do that but this is a Map so it doesn't do that for us.

nathanstitt:
[Change Request: BigLearn CritSit resolution](https://trello.com/c/FkQTAwrM/1633-change-request-biglearn-critsit-resolution)

PR #2347, opened 2018-09-12
set width of sims on narrow screens so they'll overflow if needed

nathanstitt:
before:
screen shot 2018-09-12 at 3 18 56 pm

After:
screen shot 2018-09-12 at 3 19 44 pm


nathanstitt:
[Bug: SIM not working](https://trello.com/c/NlKeFHgL/1663-bug-sim-not-working)

PR #2354, opened 2018-09-17
update scores report to use late work & exercise counts correctly

nathanstitt:
The BE now updates the counts for late work in addition to the counts, so the FE needs
to use the max value for them.

Additionally the scores report wasn't using the exercise count for the
human readable X of X score, it was using step counts which was inaccurate
for readings


tutor/src/models/course/scores/task-result.js
```diff
@@ -127,4 +143,73 @@ export default class TaskResult extends BaseModel {
     this.reportHeading.adjustScores();
   }
 
+  @computed get completedPercent() {
```
Dantemss: This will include the late work if the teacher accepted it

tutor/src/models/course/scores/task-result.js
```diff
@@ -127,4 +143,73 @@ export default class TaskResult extends BaseModel {
...
+    const percent = this.isHomework ?
+      (this.completedExerciseCount / this.exercise_count) :
+      (this.completedStepCount / this.step_count);
```
Dantemss: This will include late work if the instructor accepted it.

tutor/src/models/course/scores/task-result.js
```diff
@@ -127,4 +143,73 @@ export default class TaskResult extends BaseModel {
...
+    return Math.round(percent * 100);
+  }
+
+  @computed get humanProgressWithLateWork() {
+    const percent = this.isHomework ?
+      this.completed_exercise_count / this.exercise_count :
+      this.completed_step_count / this.step_count;
```
Dantemss: This will always include late work, even if not accepted.

tutor/src/models/course/scores/task-result.js
```diff
@@ -127,4 +143,73 @@ export default class TaskResult extends BaseModel {
...
+    }
+    return `${Math.round(percent * 100)}%`;
+  }
+
+  @computed get humanUnacceptedProgress() {
+    if (!this.completed_on_time_exercise_count || !this.exercise_count) {
+      return '0%';
+    }
+    const progress = Math.round((
+      this.completed_on_time_exercise_count / this.exercise_count
```
Dantemss: This will never include late work.

tutor/src/models/course/scores/task-result.js
```diff
@@ -127,4 +143,73 @@ export default class TaskResult extends BaseModel {
...
+    ) * 100 );
+    return `${progress}%`;
+  }
+
+  @computed get humanScoreWithLateWork() {
+    if (!this.correct_exercise_count || !this.exercise_count) {
+      return '0%';
+    }
+    const score = Math.round((
+      this.correct_exercise_count / this.exercise_count
```
Dantemss: This will always include late work, even if not accepted.

tutor/src/models/course/scores/task-result.js
```diff
@@ -127,4 +143,73 @@ export default class TaskResult extends BaseModel {
...
+    ) * 100 );
+    return `${score}%`;
+  }
+
+  @computed get humanUnacceptedScore() {
+    if (!this.correct_on_time_exercise_count || !this.exercise_count) {
+      return '0%';
+    }
+    const score = Math.round((
+      this.correct_on_time_exercise_count / this.exercise_count
```
Dantemss: This will never include late work.

tutor/src/models/course/scores/task-result.js
```diff
@@ -41,19 +42,18 @@ export default class TaskResult extends BaseModel {
...
-      return this.completed_exercise_count -
-             this.completed_on_time_exercise_count -
-             this.completed_accepted_late_exercise_count;
+    if (this.isHomework) {
+      return this.completed_exercise_count - this.completed_on_time_exercise_count;
     }
-    return this.completed_step_count -
-           this.completed_on_time_step_count -
-           this.completed_accepted_late_step_count;
+    return this.completed_step_count - this.completed_on_time_step_count;
```
Dantemss: This works, but changes what the method does (i.e it will no longer take into account whether or not the instructor accepted late work). Is this desired?

tutor/src/screens/scores-report/late-work.jsx
```diff
@@ -9,43 +9,35 @@ import omit from 'lodash/omit';
...
+
+  @computed get isAccepted() {
+    return Boolean(
+      this.task.is_late_work_accepted && !this.task.hasAdditionalLateWork
+    );
+  }
+
+  score() {
+    return this.isAccepted ?
+      this.task.humanUnacceptedScore : this.task.humanScoreWithLateWork;
```
Dantemss: I think this is still not right. It will fail if the instructor only accepted part of the late work. Should use the same formula as adjustScore()

tutor/src/screens/scores-report/late-work.jsx
```diff
@@ -9,43 +9,35 @@ import omit from 'lodash/omit';
...
-      return (
-        TH.getHumanUnacceptedProgress(this.task)
-      );
-    } else {
-      return (
-        TH.getHumanProgressWithLateWork(this.task)
-      );
-    }
+    return this.isAccepted ?
+      this.task.humanUnacceptedProgress : this.task.humanProgressWithLateWork;
```
Dantemss: Same as above, should use the same formula as completedPercent()

nathanstitt:
[Bug: Student Reading Score Errors](https://trello.com/c/hbniAWE9/1673-bug-student-reading-score-errors)

tutor/src/models/course/scores/task-result.js
nathanstitt: This method is used to determine if there is late work that could be accepted.  It should be > 0 if there's new work that has not yet been accepted.

tutor/src/screens/scores-report/late-work.jsx
nathanstitt: Hm, what these are doing is either showing what the score would be if the teacher accepts late work, or if they rejects the previously accepted work.  It's on the late work component and is used for the little popover that shows what it would be if the button's clicked.

tutor/src/models/course/scores/task-result.js
Dantemss: I think you might want `this.completed_step_count - this. completedStepCount` in that case?

tutor/src/screens/scores-report/late-work.jsx
Dantemss: OK that makes sense then.

tutor/src/screens/scores-report/late-work.jsx
Dantemss: Ignore the previous comment, since this is only used when accepting/rejecting late work.

tutor/src/models/course/scores/task-result.js
nathanstitt: Ahah, ok I think I see what you're saying.  That makes sense.  Sorry for the confusion

PR #2355, opened 2018-09-17
Search exercises by nickname and other options

Dantemss:


nathanstitt:
looks good @Dantemss , thanks!

PR #2356, opened 2018-09-19
Test that reference element is parent, not the common node

nathanstitt:


nathanstitt:
[User getting an error while highlighting an incomplete sentence](https://trello.com/c/Ypg26v84/1619-user-getting-an-error-while-highlighting-an-incomplete-sentence)

PR #2357, opened 2018-09-19
Adjust left margin of sims when they're very narrow

nathanstitt:
Otherwise it was still being cut-off


nathanstitt:
[Bug: SIM not working](https://trello.com/c/NlKeFHgL/1663-bug-sim-not-working)

PR #2358, opened 2018-09-20
single column layout on summary page

TomWoodward:


TomWoodward:
[Bug: Highlight summary page is not fitting some text that is highlighted](https://trello.com/c/WR5HkQGO/1653-bug-highlight-summary-page-is-not-fitting-some-text-that-is-highlighted)

PR #2359, opened 2018-09-21
generate audit html files

TomWoodward:
artifacts like this will be generated on master

https://jenkins-dev.cnx.org:8080/blue/organizations/jenkins/OpenStax%20-%20Continuous%20Integration%2Ftutor-js/detail/PR-2359/2/artifacts/

had to run 
```groovy
System.setProperty("hudson.model.DirectoryBrowserSupport.CSP", "default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; img-src 'self' data:; style-src 'self' 'unsafe-inline'; font-src *");
```

on jenkins to get it to allow the html content

### DIFF
--- a/jenkins/continuous-delivery.jenkins
+++ b/jenkins/continuous-delivery.jenkins
@@ -12,8 +12,12 @@ pipeline {
             }
             stage('Run Audit') {
               steps {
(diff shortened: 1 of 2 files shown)


nathanstitt:
[Run automated accessibility audit on each tutor page](https://trello.com/c/RUP5EQsg/1585-run-automated-accessibility-audit-on-each-tutor-page)

PR #2360, opened 2018-09-24
Fix payment_overdue error

Dantemss:
This change: 
 - Prevent error modal from showing for payment_overdue errors
 - Will force the "Terms and Conditions" and  "You Must Pay Now!" modals to show anywhere the user in in a course, even if they bypass the dashboard by using an assignment link

### DIFF
--- a/configs/test/jest.config.json
+++ b/configs/test/jest.config.json
@@ -1,6 +1,6 @@
 {
   "verbose": true,
-  "notify": true,
+  "notify": false,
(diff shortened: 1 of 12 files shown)


nathanstitt:
[Bug: When a student whose trial is up goes to an assignment using a direct link, she gets an error](https://trello.com/c/gL0UFzaE/1677-bug-when-a-student-whose-trial-is-up-goes-to-an-assignment-using-a-direct-link-she-gets-an-error)

PR #2383, opened 2018-10-02
update exercise search to use uid instead of just id

nathanstitt:


Dantemss:
Closing since #2382 implemented the same change.

PR #2384, opened 2018-10-03
set "uid" to be the default search

nathanstitt:
The property is only updated when search options change,
meaning the default will be used for a query that's performed without
modifiying the options


Dantemss:
It displays as `Search by uid`... hopefully that's not a problem?

PR #2385, opened 2018-10-03
reload page after a past due payment is made

nathanstitt:
This is needed since the stores are probably in a bad state because the BE would have been raising 403's.


nathanstitt:
[Bug: When a student whose trial is up goes to an assignment using a direct link, she gets an error](https://trello.com/c/gL0UFzaE/1677-bug-when-a-student-whose-trial-is-up-goes-to-an-assignment-using-a-direct-link-she-gets-an-error)

PR #2386, opened 2018-10-03
make sure formats are unique

nathanstitt:
before clicking "open-ended" multiple times would cause "free-response" to appear multiple times


nathanstitt:
[Bug: Adding Open Ended Tag to an Existing Exercise](https://trello.com/c/bYZYBs1a/1659-bug-adding-open-ended-tag-to-an-existing-exercise)

PR #2389, opened 2018-10-05
fix multiple ids on page issue

TomWoodward:
fixes

image

### DIFF
--- a/jenkins/continuous-integration.jenkins
+++ b/jenkins/continuous-integration.jenkins
@@ -10,8 +10,8 @@ pipeline {
         }
         stage('Run Audit') {
           steps {
-            copyArtifacts projectName: 'openstax/tutor-js/master', filter: 'accessibility-report.json'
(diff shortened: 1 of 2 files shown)


nathanstitt:
Hm, the tour step is supposed to point to the task plan on the calendar not to the entire calendar.

I'm pretty sure it doesn't actually use the element's `id` property to target it though.  Instead it uses the `data-tour-anchor-id`.

I think we're only passing the `id` through to the element by mistake because we're using the kinda whacky `filterProps` helper.  https://github.com/openstax/tutor-js/blob/master/tutor/src/components/tours/anchor.jsx#L47

Why don't you try taking the `id` out before setting it there and test if things still work?

TomWoodward:
i did that initially, but couldn't figure out if the id was actually necessary, also it seemed weird to bind it to something that there are a bunch of. on the default dashboard you don't even have any tasks on it so it seemed more helpful to have it on the calendar anyway.

PR #2397, opened 2018-10-09
reposition tour after content loads

nathanstitt:
Before:
c40dfb94-7acb-4114-bef7-7ce1d4c66956

after:
screen shot 2018-10-09 at 1 14 59 pm

### DIFF
--- a/tutor/src/models/reference-book/ux.js
+++ b/tutor/src/models/reference-book/ux.js
@@ -1,5 +1,5 @@
 import { observable, computed, action, observe } from 'mobx';
-import { extend, first } from 'lodash';
+import { first, invoke } from 'lodash';
 
 import WindowSize from '../window-size';
 import Courses from '../courses-map';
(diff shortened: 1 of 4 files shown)


tutor/src/models/reference-book/ux.js
```diff
@@ -75,6 +75,9 @@ export default class BookUX {
     if (this.book && !cs) {
       cs = first(this.book.pages.byChapterSection.keys());
     }
+    if (this.tours && this.tours.tourRide) {
+      setTimeout(() => invoke(this, 'tours.tourRide.joyrideRef.calcPlacement'), 10);
```
TomWoodward: is this timeout waiting for react to re-render? could you invoke the calcPlacement from the componentDidUpdate?

nathanstitt: yes, it's waiting for react to re-render the titles, but also potentially for mathjaxj to run.  Basically it's just waiting for the page to "settle down" and for everything to be re-positioned.

PR #2405, opened 2018-10-15
displays overlays on top of an "Obscured Page" 

nathanstitt:
This way the page can be set to `display: none` when the overlay is displayed, preventing tab and scrolling


tutor/src/components/obscured-page/overlay-registry.js
```diff
@@ -0,0 +1,78 @@
+import { invoke, defer } from 'lodash';
+import { observable, action, computed } from 'mobx';
+import cn from 'classnames';
+
+
```
TomWoodward: some excessive whitespace

tutor/src/components/task/progress/panel.cjsx
```diff
@@ -71,16 +71,17 @@ ProgressPanel = React.createClass
...
-      isForwardEnabled={@state.shouldShowRight}
-      isBackwardEnabled={@state.shouldShowLeft}
-      onForwardNavigation={@goForward}
-      onBackwardNavigation={@goBackward}
-      titles={titles}
-    >
-      {@props.children}
-    </PagingNavigation>
-
+    <ObscuredPage>
```
TomWoodward: would it be better to put this in `components/app.jsx` so that it can be used on any screen without having to include it?

tutor/src/components/obscured-page/index.js
```diff
@@ -0,0 +1,67 @@
...
+  display: flex;
+  position: fixed;
+  left: 0;
+  right: 0;
+  top: 0;
+  height: 0%;
+  z-index: 20;
+  overflow: auto;
+  transition: height 0.4s;
+${props => props.isHidden && css`
```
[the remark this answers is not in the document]
nathanstitt: yeah, that's the one thing that I don't like about styled-components.  Since they're inside a tagged string, auto-indent doesn't work well

tutor/src/components/task/progress/panel.cjsx
nathanstitt: hm that's an interesting idea.  I could see us needing that in the future, but right now the overlay should only render if there's a page that supports overlaying (reading steps & ref view).

We'd need to generalize it more so that the overlay would know more about the current page if we move to app.

PR #2413, opened 2018-10-19
integrate highlighting library

TomWoodward:


nathanstitt:
The code looks good @TomWoodward but I think I've found a few small bugs regarding annotations in reading assignments.  To test I:

 * Created some annotations on a reading assignment using code from the `master` branch.
 * Verified that they appeared identical in both reference view and reading.
 * Switched to your `issue-11` branch, pulled, `yarn install`, and restarted FE without issues
 * Reloaded browser
 * Annotations where present in both reading and reference view. YAY!
 * However!  The annotations in the reading no longer display the comment icon on the right, nor the edit box when they're clicked.  They are scrolled to the correct position on the page but the edit box doesn't appear.
 * I deleted the original annotations and made some new ones with the updated code.
 * Those saved and appeared fine, except they also had the missing icon and edit box in readings.  if the annotation is made with the "note" icon, the edit box does appear at first but is lost once page reloads.

PR #2414, opened 2018-10-23
Upgrade libraries to react 16

nathanstitt:
Also converts the last of the coffee to js


babel.config.js
```diff
@@ -0,0 +1,54 @@
...
+    ['@babel/plugin-proposal-decorators', { legacy: true } ],
+    ['@babel/plugin-proposal-class-properties', { loose: true } ],
+    '@babel/plugin-proposal-export-default-from',
+    '@babel/plugin-syntax-dynamic-import',
+    'babel-plugin-styled-components',
+    'babel-plugin-lodash',
+  ],
+};
+
+// {
```
TomWoodward: commented stuff

bin/upgrade
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env ruby
+
+#require 'dir'
```
TomWoodward: commented code

configs/upgrade.log
```diff
@@ -0,0 +1,1316 @@
+tutor/specs/flux/step-title.spec.js
```
TomWoodward: why commit this?

exercises/src/components/exercise/answer.jsx
```diff
@@ -7,16 +8,17 @@ import cn from 'classnames';
 //import _ from 'underscore';
 //import { AnswerActions, AnswerStore } from 'stores/answer';
 
+export default
```
TomWoodward: does this work? the whitespace is kind of gross but its probably auto generated so if it works /shrug

exercises/src/components/exercise/preview.jsx
```diff
@@ -53,4 +55,4 @@ export default class ExerciseEditingPreview extends React.Component {
     );
   }
 
-}
+};
```
TomWoodward: is this right? classes dont take semicolons do they?

exercises/test/all-source-files.js
```diff
@@ -0,0 +1,2 @@
+const testsContext = require.context('../src', true, /\.(cjsx|coffee)$/);
```
TomWoodward: this seems like it'll be broken now

exercises/test/config/karma.common.js
```diff
@@ -0,0 +1,78 @@
...
+import path from 'path';
+
+export default {
+  basePath: '../../',
+  frameworks: ['mocha', 'chai', 'chai-sinon', 'phantomjs-shim'],
+  browsers: ['PhantomJS'],
+  reporters: ['mocha'],
+
+  files: [
+    'test/all-tests.coffee',
```
TomWoodward: lots of places looking for .coffee or .cjsx extensions are probably all broken now

shared/demo.js
```diff
@@ -0,0 +1,24 @@
+import ReactHelpers from './src/helpers/react';
+
+//require 'react-hot-loader/patch'
+// React = require 'react'
```
TomWoodward: it would be cool to have your script recognize commented code and remove it, but that seems hard

exercises/src/components/exercise/answer.jsx
nathanstitt: yeah the new decorator pattern is that exports, then decorators, then definition.

There wasn't a codmod for that that I could find hence the regex in the update script

exercises/src/components/exercise/preview.jsx
nathanstitt: they can.. I played with it a bit and it seems like ESLint gets a little confused about the decorator and export being attached to the class. If either of those are removed it stops suggesting a semi colon.

PR #2416, opened 2018-10-25
Fix annotation sorting and overlay page scroll position

nathanstitt:
stores the selection's bounding rect and the sorts by it when annotations load

Also fixes the window scroll position. bug discovered by @TomWoodward


TomWoodward:
@nathanstitt when we spoke last night I had wanted to use rely on the sort order of the highlights based on [this](https://github.com/openstax/highlighter/blob/master/src/Highlighter.ts#L62) logic. but i'm realizing that would only work for the current page where we have the highlights mounted, in order to preserve the order on pages you're not looking at we have to store something, so your approach is 👍

PR #2420, opened 2018-11-07
only save task step if it's not completed

nathanstitt:
Goes with 
https://github.com/openstax/tutor-server/pull/1802

### DIFF
--- a/tutor/src/components/task-step/index.cjsx
+++ b/tutor/src/components/task-step/index.cjsx
@@ -39,7 +39,8 @@ TaskStep = React.createClass
 
   onStepCompleted: (id) ->
     {id} = @props unless id? # need to allow `id` from arguments pass through, especially for multi-part
-    if StepPanel.canWrite(id)
+    step = TaskStepStore.get(id)
+    if !step.is_completed && StepPanel.canWrite(id)
       TaskActions.completeStep(id, @props.taskId)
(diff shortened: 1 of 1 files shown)


tutor/src/components/task-step/index.cjsx
```diff
@@ -39,7 +39,8 @@ TaskStep = React.createClass
 
   onStepCompleted: (id) ->
     {id} = @props unless id? # need to allow `id` from arguments pass through, especially for multi-part
-    if StepPanel.canWrite(id)
+    step = TaskStepStore.get(id)
+    if !step.is_completed && StepPanel.canWrite(id)
```
Dantemss: We do want to mark it as completed again if they change the answer. After the first completed, which is required, this could either be conditioned on the answer changing or on the due date not having passed.

nathanstitt: I researched this and the `onStepCompleted` handler here is only called for non-task steps.  This handles reading, video and sim type steps. 

However, I also tested exercises and it currently only hits the "completed" api endpoint once, when the exercise is first answered.  If the exercise is reloaded before it's closed and the answer is changed, it saves the new answer but the "Submit" button isn't shown.  Instead it only has the "Next Answer" button which doesn't send to the "completed" endpoint.

PR #2432, opened 2018-11-26
Fix student task steps loading

nathanstitt:
Also a few small UX things I noted like missing padding on sections chooser, overflowing buttons on student dashboard

### DIFF
--- a/tutor/src/components/task/index.js
+++ b/tutor/src/components/task/index.js
@@ -437,3 +437,5 @@ class TaskShell extends React.Component {
 }
 
 export { Task, TaskShell };
+
+export default TaskShell;
--- a/tutor/src/flux/task.js
+++ b/tutor/src/flux/task.js
@@ -48,16 +48,6 @@ const getChangedSteps = steps =>
   filter(steps, step => (step != null) && TaskStepStore.isChanged(step.id))
 ;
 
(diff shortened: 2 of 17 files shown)


tutor/src/flux/task.js
```diff
@@ -125,10 +114,14 @@ const TaskConfig = {
...
-    TaskStepActions.completed(obj, taskStepId);
+  onStepSaved(obj) {
     this._loaded(obj, obj.id);
     Object.assign(this._local[obj.id], obj);
+  },
+
+  stepCompleted(obj, taskStepId) {
+    TaskStepActions.completed(obj, taskStepId);
+    this.onStepSaved(obj, taskStepId);
     return this.emit('step.completed', taskStepId, obj.id);
```
nathanstitt: @Dantemss this is the actual bug - before it wasn't calling `onStepSaved`

tutor/src/components/task/index.js
```diff
@@ -437,3 +437,5 @@ class TaskShell extends React.Component {
 }
 
 export { Task, TaskShell };
+
+export default TaskShell;
```
nathanstitt: and also this was causing the tasks to fail to load from routes.js

PR #2450, opened 2018-12-10
Misc upgrade style fixes

nathanstitt:
* [x] tick mark in the question library is far from the options
* [x] Add student Id window during course signup is missing background
* [x] Cancel button in Delete modal - Highlights & notes page is orange by default vs. staging where Cancel button is grey
* [x] Delete section modal in course roster has pink and orange color. Cancel button is grey in staging but orange in QA
* [x] Window that opens on clicking an existing assignment in teacher view has yellow panel at the top in staging but after upgrade its grey
* [x] The "X" button in the Window that opens on clicking an existing assignment in teacher view is displayed close the the window title instead of right corner
* [x] The "X" button in most of the pop-up windows are close to the window title vs. the right corner
* [x] Highlight & Notes tool tip - too much white space and the line spacing appears to be less. This spacing issue applies to all tool tips
* [x] Refund request modal has both Continue & Cancel button in Orange. I think Cancel would be grey before upgrade. Same with the next window "Help us improve"

### DIFF
--- a/tutor/specs/components/__snapshots__/lms-info-card.spec.jsx.snap
+++ b/tutor/specs/components/__snapshots__/lms-info-card.spec.jsx.snap
@@ -8,16 +8,18 @@ exports[`LmsInfo Component matches snapshot 1`] = `
 
 .c0 {
   margin: 10px 0 20px 0;
-  display: block;
+  padding-left: 0 !important;
   font-size: 16px;
 }
 
 <div
   className="lms-info"
 >
-  <a
-    className="c0"
+  <button
+    className="c0 btn btn-link"
+    disabled={false}
     onClick={[MockFunction]}
+    type="button"
   >
     <svg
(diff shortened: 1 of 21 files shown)


tutor/specs/components/__snapshots__/lms-info-card.spec.jsx.snap
```diff
@@ -8,16 +8,18 @@ exports[`LmsInfo Component matches snapshot 1`] = `
 
 .c0 {
   margin: 10px 0 20px 0;
-  display: block;
+  padding-left: 0 !important;
```
TomWoodward: 😲

nathanstitt: yeah.. it's overriding a bootstrap default 😭

PR #2456, opened 2018-12-17
fix media previews so they include entire os-figure

nathanstitt:
image


tutor/resources/styles/components/media-preview.scss
```diff
@@ -1,5 +1,5 @@
 .media-preview {
-  max-width: 50%;
+  max-width: 300px;
```
TomWoodward: is this a problem for mobile?

nathanstitt: good thing to check, it's fine though since mobile is always at least 300px

PR #2472, opened 2018-12-31
Fix splash image positioning 

nathanstitt:
This will identify splash images in JS and then skip setting image width class on them

### DIFF
--- a/tutor/specs/screens/new-course/select-dates.spec.jsx
+++ b/tutor/specs/screens/new-course/select-dates.spec.jsx
@@ -1,4 +1,4 @@
-import { React, Factory } from '../../helpers';
+import { React, Factory, TimeMock } from '../../helpers';
 import SelectDates from '../../../src/screens/new-course/select-dates';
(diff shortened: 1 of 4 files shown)


tutor/specs/screens/new-course/select-dates.spec.jsx
```diff
@@ -8,6 +8,9 @@ jest.mock('../../../src/models/user', () => ({
 
 describe('CreateCourse: Selecting course dates', function() {
   let ux;
+
+  TimeMock.setTo('2018-12-01T12:00:00.000Z');
+
```
nathanstitt: this is unrelated to the PR, but the spec failed once the year rolled over to 2019

nathanstitt: nah, it's important to get into master ASAP, so that specs don't continue to fail

brittany-johnson: Ahhhh I thought you were saying that it was failing with the fix. I get it now. Thanks!

PR #2502, opened 2019-02-01
include 0 in mult-choice range

brittany-johnson:


shared/src/components/question/answers-table.jsx
```diff
@@ -12,7 +12,7 @@ import Instructions  from './instructions';
 
 
 const KEYS =
-  { 'multiple-choice-numbers': range(1, 10) }; // 1 - 9
+  { 'multiple-choice-numbers': range(0, 10) }; // 0 - 9
```
nathanstitt: haha!  So simple a bug 😭

nathanstitt:
thanks @brittany-johnson !

PR #2523, opened 2019-03-01
Consider author and vocabulary when calulating readOnly status

nathanstitt:
Previously there was 2 bugs:
 * The ribbon was always visible even when the user could edit
 * The logic behind the ribbon was intended to only check if the current user was the author
    * This failed for vocabulary terms, which aren't editable regardless of who authored them

image

### DIFF
--- a/exercises/src/components/exercise/preview.jsx
+++ b/exercises/src/components/exercise/preview.jsx
@@ -1,14 +1,10 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import { find } from 'lodash';
 import styled from 'styled-components';
 import { ExercisePreview } from 'shared';
 import { Link } from 'react-router-dom';
-import { computed } from 'mobx';
 import { observer } from 'mobx-react';
 import { CornerRibbon } from 'shared';
-import CurrentUser from '../../models/user';
-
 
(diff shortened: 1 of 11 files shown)


exercises/src/components/exercise/preview.jsx
```diff
@@ -24,13 +20,6 @@ class ExerciseEditingPreview extends React.Component {
     showEdit: PropTypes.bool,
   }
 
-  @computed get isAuthor() {
-    const { exercise } = this.props;
-    return Boolean(
-      exercise.isNew || find(exercise.authors, { user_id: CurrentUser.id })
```
reedstrm: Has the 'isNew' case dropped out? Is it needed? What happens when you create a new exercise, does it fall under the "user is author" part of this test?

nathanstitt: Yep, good q 😁   in the model it only fails if the author doesn't match or if it's `is_vocab`. otherwise it'll return true so the isNew check is no longer needed.

nathanstitt: actually I think I'll re-add the `isNew` check since we may stop setting the author on new exercises which would cause this to fail.  Good catch!

PR #2524, opened 2019-03-04
update-readme

brittany-johnson:
Rm mention of coach and add section for Exercises


nathanstitt:
Awesome, thanks @brittany-johnson !

PR #2540, opened 2019-03-12
move task model files into teacher/student directories

nathanstitt:
will enable us to add the new task step models and avoid confusion


tutor/specs/screens/task/index.spec.js
```diff
@@ -0,0 +1,20 @@
...
+  let props;
+
+  beforeEach(() => {
+    props = {
+      course: Factory.course(),
+    };
+  });
+
+  it('renders', () => {
+    mount(<Task {...props} />);
```
TomWoodward: what is this testing? should it be `expect(() => mount(<Task {...props} />)).not.toThrow()`?

PR #2541, opened 2019-03-12
Max width sim exercises

brittany-johnson:
No issue card was created but this is in reference to https://openstax.slack.com/archives/C0F6Z1JUS/p1551884807250900

Msg from @carlycurran: 

> can you look into this one when you get a chance? It’s a sim in an assessment: The simulation for the pre-class reading problem ID# 11240@5 seem to have problems. About half of the students state that they can’t see the simulation.

The sim is in chapter 25.5 in physics

### DIFF
--- a/tutor/resources/styles/book-content/mixins.scss
+++ b/tutor/resources/styles/book-content/mixins.scss
@@ -72,7 +72,7 @@
 }
 
 @mixin book-content-interactives(
-  $narrow-horizontal-margin: $book-content-narrow-horizontal-padding, $inExercise: false
+  $narrow-horizontal-margin: $book-content-narrow-horizontal-padding
 ) {
   iframe.interactive {
     width: 100%;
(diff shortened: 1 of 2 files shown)


brittany-johnson:
Screen Shot 2019-03-12 at 3 00 11 PM
Screen Shot 2019-03-12 at 3 15 58 PM

nathanstitt:
@brittany-johnson not bad!  While that'd certainly work, I think it'd be a bit better if we allowed the sim to expand to completely fill the card's body.  We do that for the splash images using the scss `book-content-full-width` mixin.  You can see an example of that by viewing a reading that has an intro step, and notice how the big image is wider than the text and completely fills the whitespace.

Can you investigate if that mixin will work for sims as well?  Apologies if you already have 😁

brittany-johnson:
@nathanstitt `min-width` doesn't really work on the `<iframe>` well to display the entire sim. As I increase the value for `min-width` it actually starts to shift the `iframe` to the left/right and doesn't display the sim properly. 
I think the `<iframe>` just needs more padding within itself so that the sim will fully display, at least that's what I found from testing it. 
Is there anything else that you think I should try?

tutor/resources/styles/book-content/mixins.scss
```diff
@@ -83,6 +83,10 @@
 
     width: $tutor-interactive-iframe-width;
     margin-left: calc(50% - #{($tutor-interactive-iframe-width / 2)});
+
+    @if ($inExercise: true) {
+      padding: 0px 136px;
+    }
```
nathanstitt: @brittany-johnson I replaced the entire if block here with:
`    @include book-content-full-width(); `

and it looked pretty good to me:




image

Malar-Natarajan:
Verified in QA

PR #2542, opened 2019-03-13
allow editing of vocab-based exercises - apply a ribbon

reedstrm:


reedstrm:
Hmm, poking around a bit , it seems the Ribbon should instead become it;s own component, rather than conditionally rendering bits into it, I suppose. Even more of a newb to React than I am to Rails.

reedstrm:
BTW, I have no idea what this looks like, I haven't deployed it anywhere.

PR #2544, opened 2019-03-15
Fix recalculating student average when they have a 0 score

nathanstitt:
Before we were using boolean operator which treats 0 as false.  That caused the setter to ignore the student if they had a 0 for any of the types


tutor/src/models/scores/student.js
```diff
@@ -29,6 +29,15 @@ class Student extends BaseModel {
...
 
+  get course() {
+    return this.period.course;
+  }
+
+  isValid(type) {
+    return this.course[`${type}_score_weight`] && !isNil(this[`${type}_score`]) &&
+      this.course[`${type}_progress_weight`] && !isNil(this[`${type}_progress`]);
+  }
+
```
nathanstitt: @brittany-johnson sorry for the large PR - most if it is just moving files around to match our standard layout, this is the actual bug.

Before we were just saying `!this.homework_score` here, however that'd also fail if `homework_score` was 0, when we really wanted to skip it only if it was null or undefined.  The `isNil` checks both of those, but not 0

PR #2555, opened 2019-04-03
Task step fixes round 2

nathanstitt:
* adjust line-spacing/font size for exercises
* restore reading progress bar
* Fix milestones so they go to correct step when clicked
* Fix exercise breadcrumbs to not count value prop cards
* record when value props are viewed so that one-time cards won't be re-displayed


tutor/src/components/breadcrumb.js
```diff
@@ -25,7 +25,9 @@ class Breadcrumb extends React.Component {
     let crumbClasses, status, title;
     let isCorrect = false;
     let isIncorrect = false;
-    const { step, canReview, isCurrent, className, stepIndex } = this.props;
+    const {
+      step, canReview, isCurrent, className, stepIndex, dataStepIndex,
```
brittany-johnson: I think we can remove `stepIndex` right?

nathanstitt: no, the stepIndex is still used by the breadcrumb to notify the steps which crumb was clicked.  it's no longer being used for displaying though since using it caused the placeholder steps to be numbered as well

brittany-johnson: ahh, oops 🙈

PR #2569, opened 2019-04-12
do not record nudge message when UI is off

nathanstitt:
previously it wouldn't display the message but would still record it

### DIFF
--- a/tutor/src/screens/task/step/fr-nudge-help.js
+++ b/tutor/src/screens/task/step/fr-nudge-help.js
@@ -37,8 +37,8 @@ const Actions = styled.div`
 const FrNudgeHelp = observer(({ course, step, ux }) => {
   return (
     <StyledFrNudgeHelp>
-      <Title>Need Help?</Title>
-      <Msg>{ux.nudgeMessage}</Msg>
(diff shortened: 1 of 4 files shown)


tutor/src/screens/task/step/fr-nudge-help.js
```diff
@@ -37,8 +37,8 @@ const Actions = styled.div`
 const FrNudgeHelp = observer(({ course, step, ux }) => {
   return (
     <StyledFrNudgeHelp>
-      <Title>Need Help?</Title>
-      <Msg>{ux.nudgeMessage}</Msg>
+      <Title>{ux.nudge.title}</Title>
```
brittany-johnson: Hey Nathan, can this cause a type error since ux.nudge can be null?

nathanstitt: Good thought, but it shouldn't ever be null.  the ux class assigns the nudge in it's constructor by picking a random one from the choices

brittany-johnson: ahhhhhhh okay, thanks for explaining that!

PR #2587, opened 2019-05-01
Remove underline styles from links

nathanstitt:
And also turn off sentry debugging


Brooke-Fail:
Verified in QA.

Brooke-Fail:
Verified in Staging.

PR #2588, opened 2019-05-02
Fetch entire task when current step is a placeholder

nathanstitt:
This will allow BL to add/remove steps if it needs to

### DIFF
--- a/shared/src/model.js
+++ b/shared/src/model.js
@@ -31,6 +31,7 @@ export class BaseModel {
   }
 
   @action onApiRequestComplete({ data }) {
+    this.api.errors = {};
     this.update(data);
   }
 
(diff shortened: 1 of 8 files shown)


shared/src/model.js
```diff
@@ -41,8 +42,10 @@ export class BaseModel {
       this.api.errors = {};
       errors.forEach(e => this.api.errors[e.code] = e);
       error.isRecorded = true;
+    } else if (error) {
+      this.api.errors = { error: error.toString() };
     } else {
-      this.api.errors = null;
+      this.api.errors = {};
     }
```
nathanstitt: this is the bug that was causing the undefined is not an object error

PR #2589, opened 2019-05-02
Fix loading reference view from user menu

nathanstitt:
The user menu wasn't updated so it was loading by ecosystem id, and the code we had in place to support that also had an error

### DIFF
--- a/tutor/src/models/reference-book/ux.js
+++ b/tutor/src/models/reference-book/ux.js
@@ -83,6 +83,10 @@ export default class BookUX {
       // in that case the id that's present will be the ecosystem
       const course = Courses.get(this.courseId) ||
         Courses.forEcosystemId(this.courseId);
+      if (course.id != this.courseId) {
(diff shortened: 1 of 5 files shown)


tutor/src/models/reference-book/ux.js
```diff
@@ -83,6 +83,10 @@ export default class BookUX {
       // in that case the id that's present will be the ecosystem
       const course = Courses.get(this.courseId) ||
         Courses.forEcosystemId(this.courseId);
+      if (course.id != this.courseId) {
+        this.courseId = course.id;
+        this.router.history.push(`/book/${this.courseId}`);
```
Dantemss: Wouldn't we need the chapter-section here if it's present? And maybe remove the last entry so it does not get duplicated (but I'm not sure about this one).

nathanstitt: yeah, we will need chapter-section.  good catch, I'll add that

PR #2590, opened 2019-05-03
Disallow changing answers after assignment becomes due

nathanstitt:
Fixes tricky bug where updates would be allowed if a student started working an assignment before it was due, and did not reload after it was due.

The BE would still error and refuse the update but this will disable the controls once the due date passes

### DIFF
--- a/tutor/specs/models/student-tasks/step.spec.js
+++ b/tutor/specs/models/student-tasks/step.spec.js
@@ -1,7 +1,9 @@
-import { Factory } from '../../helpers';
+import { Factory, TimeMock } from '../../helpers';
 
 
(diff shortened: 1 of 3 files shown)


nathanstitt:
> Ah just had a thought... this will actually not prevent them from trying to change the answer if they had the tab open, unless they navigate away for the component to re-mount?

Nope, actually it will.  The `Time.now` call is `observable`, meaning that whenever it's output changes the components that have read it will re-render.  The update frequency is 1 minute though so it's not immediate.   I tested by setting on a exercise as the assignment became due and a minute afterward it stopped allowing changes and the "Answer" button changed to "Continue"

PR #2591, opened 2019-05-06
Fix add section modal

nathanstitt:
"find" was undefined because it wasn't being imported

And also set course settings/roster to white background

### DIFF
--- a/tutor/src/screens/course-settings/timezone.jsx
+++ b/tutor/src/screens/course-settings/timezone.jsx
@@ -3,7 +3,7 @@ import React from 'react';
 import { observable, action } from 'mobx';
 import { observer } from 'mobx-react';
 import { map, values } from 'lodash';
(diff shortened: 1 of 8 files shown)


tutor/src/screens/course-settings/timezone.jsx
```diff
@@ -172,7 +171,7 @@ class SetTimezone extends React.Component {
               defaultValue={this.props.course.time_zone}
               onChange={val => this.course_timezone = val}
               validate={this.validate}
-              autofocus={true} />
```
Dantemss: Just making sure removing the focus on SetTimezone was intentional.

nathanstitt: yeah, thanks for checking.  the `autofocus` here was never actually read by the component that was rendered here.  Instead it was just setting it to true itself.  I removed it since there was no reason to set it twice.

PR #2594, opened 2019-05-07
ran "npm audit", update a few libraries in response

nathanstitt:
Mainly node-sass had a vulnerability

🤞 this doesn't hose us when it deploys and it needs to recompile on bastion


nathanstitt:
i did this in #2593

PR #2597, opened 2019-05-07
Rework image alignment

nathanstitt:
Re-enables floating images with breaks after the series of them.

Adds new "independent" images that will be horizontally centered

image

image

image


tutor/resources/styles/book-content/figure.scss
```diff
@@ -0,0 +1,5 @@
+.os-figure {
+
+
+
+}
```
Dantemss: This seems very empty :)

PR #2616, opened 2019-05-16
Teacher as student 

nathanstitt:
Adds teacher as student button and a frame around the student view:
image

image

image


tutor/src/components/course-page.jsx
```diff
@@ -1,35 +1,81 @@
...
+
     return (
-      <div className="title-wrapper">
-        <div className="title">{this.props.title}</div>
-        {subtitle}
-      </div>
+      <TitleWrapper>
+        <TitleInner>
+          <LeftSideWrapper>
+            <Title>{this.props.title}</Title>
```
Dantemss: Just `{title}` should work

tutor/src/models/course/period.js
```diff
@@ -51,10 +51,42 @@ class CoursePeriod extends BaseModel {
...
+    await role.become();
+  }
+
+  createTeacherStudent() {
+    return { courseId: this.course.id, id: this.id };
+  }
+
+  @action onCreateTeacherStudent({ data }) {
+    this.course.roles.push(data);
+    this.course.current_role_id = data.id;
```
Dantemss: Technically this part doesn't happen until you call `become`...
And `onBecomeSuccess` already does this so I imagine you could remove it?

tutor/src/models/user/menu.js
```diff
@@ -9,6 +9,7 @@ const ROUTES = {
 
   myCourses: {
     label: 'My Courses',
+    locked(course) { return get(course, 'currentRole.isTeacherStudent'); },
```
Dantemss: My one comment about this (more for UX really) is that they can escape anyway if they bookmarked the `My Courses` page.

tutor/src/models/user/menu.js
```diff
@@ -163,7 +166,8 @@ const UserMenu = observable({
 
     if (course) {
       courseId = course.id;
-      ({ isTeacher, primaryRole: { type: menuRole } } = course);
+      isTeacher = course.currentRole;
```
Dantemss: This assignment seems... odd?

tutor/specs/screens/task/__snapshots__/breadcrumbs.spec.js.snap
```diff
@@ -1664,6 +1664,7 @@ exports[`Homework Breadcrumbs Component matches snapshot 1`] = `
             "appearance_code": "college_physics",
             "book_pdf_url": "https://archive.cnx.org/exports/185cbf87-c72e-48f5-b51e-f14f21b5eabd.pdf",
             "cloned_from_id": null,
+            "current_role_id": 1,
```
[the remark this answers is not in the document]
nathanstitt: eh, that's really not a big deal any longer.   Now that we're not using flux we an deal with both easily

tutor/specs/screens/task/steps/exercise-free-response.spec.js
```diff
@@ -33,11 +33,12 @@ describe('Exercise Free Response', () => {
     expect(<FreeResponseInput {...props} />).toMatchSnapshot();
   });
 
-  it('reviews text', () => {
+  fit('reviews text', () => {
```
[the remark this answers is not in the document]
nathanstitt: lol. no it means it's being "focused". Only that spec will run in the file.  I use that while I'm trying to figure out why a spec is failing and don't wanna run everything.  In this case it was left in by mistake, glad you noticed!

tutor/src/components/buttons/teacher-become-student.js
```diff
@@ -0,0 +1,148 @@
...
+    router: PropTypes.object,
+  }
+
+  @observable isCreating = false;
+  @observable periodMenuIsOpen = false;
+
+  @action.bound onBecomeStudentPeriodSelect(periodId) {
+    this.becomeStudentInPeriod(
+      this.props.course.periods.find(p => p.id == periodId)
+    );
```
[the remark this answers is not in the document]
nathanstitt: no this is called for the `onSelect` callback for the dropdown.  there's also a `onToggle` that's called when the menu's opened but that's only used to change the icon that's displayed

tutor/src/models/user/menu.js
nathanstitt: yeah, there's multiple ways they can escape, but I think that's really ok.  With the frame it's super clear that they're in a student view

tutor/src/models/user/menu.js
nathanstitt: heh, yeah I dunno why it's there, it's not used at all

tutor/src/components/buttons/teacher-become-student.js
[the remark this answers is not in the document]
Dantemss: Ah I see... I thought the button would always be there but it only renders when there is only 1 period. So it does not even exist in this case.

dkhowala:
Some changes, please 
- [ ] The button appearance discrepancy in diff. courses
- [ ] the bottom bar will not be a button 
- [ ] Increase the border width to 4px, please
- [ ] after increasing the border width: move the text 'close student view' 2px up 
- [ ] and move the text 'you are viewing..' 2px down.

PR #2627, opened 2019-05-25
Add refview route with section 

nathanstitt:
And redirect to page id when it loads

This allows code that doesn't know the page id, like the "related content" link to still work


nathanstitt:
> Are the deleted snap files cleanup from other PRs? Wondering because I don't see how these changes would remove them.

Thanks for pointing those out, I checked and it's because the component no longer gets an entire Page object which the snapshot had dumped the entire thing out, but now doesn't.

PR #2631, opened 2019-05-30
log when tasks timeout from BL being unready

nathanstitt:
this will allow us to get an early notice

### DIFF
--- a/tutor/src/models/task-plans/student.js
+++ b/tutor/src/models/task-plans/student.js
@@ -83,6 +83,13 @@ class StudentTaskPlans extends Map {
     );
   }
 
+  @computed get taskReadinessTimedOut() {
+    return Boolean(
+      (false === this.all_tasks_are_ready) &&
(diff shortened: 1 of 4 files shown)


tutor/src/models/task-plans/student.js
```diff
@@ -83,6 +83,13 @@ class StudentTaskPlans extends Map {
     );
   }
 
+  @computed get taskReadinessTimedOut() {
+    return Boolean(
+      (false === this.all_tasks_are_ready) &&
+        this.course.primaryRole.joinedAgo('minutes') > 30
```
Dantemss: This will log whenever biglearn is unready 30 minutes after they joined, specifically it will log right after a new assignment is created long after they joined.

nathanstitt: yep, good point.  just pushed an update so it'll only log if the network request has been attempted twice

PR #2632, opened 2019-05-31
sort tags with LO's first

nathanstitt:
removes the book from aplo tags

### DIFF
--- a/shared/src/model/exercise/tag.js
+++ b/shared/src/model/exercise/tag.js
@@ -6,31 +6,31 @@ import { isObject, first, last, filter, extend, values, pick, isNil } from 'loda
 const TYPES = {
   IMPORTANT: ['lo', 'aplo', 'blooms', 'dok', 'length', 'time'],
 };
-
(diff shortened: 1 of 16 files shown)


shared/src/model/exercise/tag.js
```diff
@@ -54,12 +54,26 @@ class ExerciseTag extends BaseModel {
     return this.asString;
   }
 
+  @computed get sortValue() {
+    // APLO should always appear first
+    if ('aplo' === this.type) { return 'A'; }
+    // and the LO
+    if ('lo' === this.type) { return 'AA'; }
```
Dantemss: Shouldn't we just prepend the A or AA in case there are more LO's?

nathanstitt: we also sort by the tag value itself, this is just to force the APLO & LO's to appear first.  
https://github.com/openstax/tutor-js/pull/2632/files#diff-7521d64a6f3a3194ea1a6b6c894ccbe5R84

PR #2648, opened 2019-06-26
only show menu for ap_bio & ap_physics

nathanstitt:


tutor/src/models/course/information.js
```diff
@@ -68,14 +69,23 @@ export default {
     title:      'Anatomy & Physiology',
     subject:    SUBJECTS.ANATOMY_PHYSIOLOGY,
   },
+};
+
+each(BOOKS, (properties, code) => properties.code = code);
```
Dantemss: This syntax that is the exact opposite order of ruby's always confused me

tutor/src/models/course/information.js
```diff
@@ -68,14 +69,23 @@ export default {
...
+
+// NOTE:
+//  * The 'biology' and 'physics' codes are deprecated, but are retained for older courses that may have them
```
Dantemss: We could probably mass-update the old courses at some point to remove those.

tutor/src/models/user/menu.js
```diff
@@ -40,7 +40,7 @@ const ROUTES = {
   },
   analytics: {
     label: 'Research analytics',
-    isAllowed(course) { return get(course, 'isAP'); },
+    isAllowed(course) { return course && ['ap_biology', 'ap_physics'].includes(course.subject.code); },
```
Dantemss: Isn't `course.subject` a string, meaning it has no `code` method? Or am I missing something?

tutor/src/models/course/information.js
nathanstitt: yeah I know, same here.   This is where the `code` is set that you thought might be missing in the other comment.  Rather than have it typed out on each, I did this so it'll always match the key.

tutor/src/models/user/menu.js
nathanstitt: hm, I had a big long thing typed here pointing out how subject returns course data.  but now I see that my change didn't get included. I wonder why the spec didn't fail 😡    Great catch!

tutor/src/models/course.js
```diff
@@ -157,7 +157,7 @@ class Course extends BaseModel {
   }
 
   @computed get subject() {
-    return get(CourseInformation.forAppearanceCode(this.appearance_code), 'subject', '');
+    return CourseInformation.forAppearanceCode(this.appearance_code);
```
Dantemss: I would suggest renaming this to `information()` now :)

PR #2649, opened 2019-06-26
Tag importance

nathanstitt:


nathanstitt:
@Dantemss yeah they're in the BE fine as they are, the "IMPORTANT" list is just the ones that we display to teachers in the QL/HW builder

PR #2657, opened 2019-07-16
if present; display the LEARNING OBJECTIVES title

nathanstitt:
Previously it was removed but AP books want it

image

### DIFF
--- a/tutor/src/components/book-page.js
+++ b/tutor/src/components/book-page.js
@@ -245,8 +245,8 @@ class BookPage extends React.Component {
       }
 
       // remove all non-lists children to prevent extra text in preamble
-      if (typeof abstractChild.remove === 'function') {
-        abstractChild.remove();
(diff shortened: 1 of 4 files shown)


tutor/src/components/book-page.js
```diff
@@ -245,8 +245,8 @@ class BookPage extends React.Component {
       }
 
       // remove all non-lists children to prevent extra text in preamble
-      if (typeof abstractChild.remove === 'function') {
-        abstractChild.remove();
+      if (true !== invoke(abstractChild, 'matches', '[data-type="title"]')) {
```
Dantemss: Won't matches return an array of matches or something instead of `true`?

nathanstitt: nah, that calls the `matches` method on an element, it returns a bool: https://developer.mozilla.org/en/docs/Web/API/Element/matches

PR #2658, opened 2019-07-16
Hide teacher controls when teacher-student views scores

nathanstitt:
use currentRole instead of testing isTeacher everywhere


nathanstitt:
yeah those are legacy methods that are used still used in a few spots.  I should hunt them down and replace them

PR #2660, opened 2019-07-16
become the student if task is for a teacherStudent role

nathanstitt:
When a teacher is working a task and reloads it, the course's current_role_id is lost.  Check the task to see if the teacher was just viewing a student's work, or if the task was for a ghost student.

Goes along with https://github.com/openstax/tutor-server/pull/1885


tutor/src/screens/task/ux.js
```diff
@@ -33,6 +33,18 @@ export default class TaskUX {
   @lazyGetter scroller = new ScrollTo({ windowImpl: this.window });
   @lazyGetter pageContentUX = new PageContentUX({ main: this });
 
+  @action becomeStudentIfNeeded() {
+    if (!this.course.roles.teacher) { return; }
+    // if the teacher has reloaded while working student tasks
+    // we need to restore it so the frame appears
+    const teacherAsStudentRole = this.course.roles.find(r => {
+      return this._task.students.find(s => s.role_id == r.id);
```
Dantemss: Potential string vs integer comparison here? The role representer coerces the id to String but I think the new backend PR does not.

nathanstitt: yep! that's why I used `==`, it's ugly but will match stuff like `1 == '1'`

However, I should also make a comment there saying why, otherwise someone might decide to "improve" the code and make it triple equals.

PR #2661, opened 2019-07-17
don't link from logo when teacher viewing as student

nathanstitt:


nathanstitt:
yeah that's react "context". its a way for React to do "semi-global" variables.  The variable is available to any of the child nodes.  Over-use can lead to really hard to maintain code, but sometimes it's cleaner to just access the course this way since it's a pain to pass down through everything

PR #2664, opened 2019-07-18
Fix sims without "interactive" class in task steps

nathanstitt:
Rather than attempting to track down all the content, just treat all iframe that way

image

### DIFF
--- a/tutor/specs/components/exercises/__snapshots__/sectionizer.spec.js.snap
+++ b/tutor/specs/components/exercises/__snapshots__/sectionizer.spec.js.snap
@@ -13,23 +13,33 @@ exports[`Exercise Sectionizer Component renders and matches snapshot 1`] = `
   <div
     className="section"
     onClick={[Function]}
-  />
+  >
+    1.1
+  </div>
   <div
     className="section"
     onClick={[Function]}
-  />
+  >
+    2.1
+  </div>
   <div
     className="section"
     onClick={[Function]}
-  />
+  >
+    3.1
+  </div>
   <div
     className="section"
     onClick={[Function]}
-  />
+  >
+    1.2
+  </div>
   <div
     className="section"
     onClick={[Function]}
-  />
+  >
+    2.2
+  </div>
   <div
     className="section ellipsis"
     onClick={[Function]}
@@ -39,7 +49,9 @@ exports[`Exercise Sectionizer Component renders and matches snapshot 1`] = `
   <div
     className="section"
     onClick={[Function]}
-  />
+  >
+    2
+  </div>
   <div
     className="next"
     onClick={[Function]}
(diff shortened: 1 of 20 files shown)


tutor/src/helpers/analytics.js
```diff
@@ -58,7 +58,8 @@ const Translators = {
 
   // Task steps are viewed by both teacher and student with no difference in params
   viewTaskStep({ courseId }) {
-    const role = Courses.get(courseId).primaryRole.type;
+    const course = Courses.get(courseId);
+    const role = course ? course.primaryRole.type : 'unknown';
```
Dantemss: Hmm not `currentRole` here?

tutor/src/models/task-plans/teacher/plan.js
```diff
@@ -302,7 +302,6 @@ class TeacherTaskPlan extends BaseModel {
     return new TeacherTaskPlan({
       ...this.clonedAttributes,
       tasking_plans: course.periods.active.map(period => ({
-        opens_at: course.starts_at,
```
Dantemss: Need to test to make sure assignments created on the day a course starts still work, even if the course has a weird starts_at time, like 10PM.

tutor/specs/components/exercises/__snapshots__/sectionizer.spec.js.snap
```diff
@@ -13,33 +13,23 @@ exports[`Exercise Sectionizer Component renders and matches snapshot 1`] = `
   <div
     className="section"
     onClick={[Function]}
-  >
-    1
```
Dantemss: This change seems odd.

tutor/src/helpers/analytics.js
nathanstitt: no, primaryRole will give priority to teacher role if present, otherwise student.  for this case we just want to log if a teacher or student is viewing the task step so the analytics make sense

tutor/src/models/task-plans/teacher/plan.js
nathanstitt: yeah, I removed that because the tasking was overriding it so having it here didn't make sense.. https://github.com/openstax/tutor-js/blob/master/tutor/src/models/task-plans/teacher/tasking.js#L29-L33

tutor/specs/components/exercises/__snapshots__/sectionizer.spec.js.snap
nathanstitt: yes, turned out the spec wasn't using the model like it should

tutor/src/models/task-plans/teacher/plan.js
Dantemss: Hmm if a course starts on a day after the default open time and you try to create an assignment on that day, will that blow up?

nathanstitt: yep, I think it would.. I'll update that code. thanks!

PR #2666, opened 2019-07-19
only apply "splash" style to images wider than 400px

nathanstitt:
image

### DIFF
--- a/tutor/src/components/book-page.js
+++ b/tutor/src/components/book-page.js
@@ -202,9 +202,7 @@ class BookPage extends React.Component {
       (splashFigure = splashFigure.nextSibling) &&
         (!splashFigure.matches || splashFigure.matches('.os-figure'))
     ) {
-      // text nodes will not have classList
-      const img = invoke(splashFigure, 'querySelector', 'img');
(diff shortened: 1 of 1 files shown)


tutor/src/components/book-page.js
```diff
@@ -195,7 +203,10 @@ class BookPage extends React.Component {
         (!splashFigure.matches || splashFigure.matches('.os-figure'))
     ) {
       // text nodes will not have classList
-      if (splashFigure.classList) { splashFigure.classList.add('splash'); }
+      const img = invoke(splashFigure, 'querySelector', 'img');
```
Dantemss: I think this check could be inside the function, to avoid running the selector twice.

nathanstitt: well we need to check 2 things here, one is that `splashFigure` is a element and not a text node, and if it contains an img.  If it's a text node it won't have a `querySelector` which we guard against by using `invoke`. and if it doesn't have any images then we don't need to do anything with it anyway so no need to do the somewhat expensive wait for images to load

The comment is misleading though, I'll update it

PR #2670, opened 2019-07-23
more fixes for tasking min max

nathanstitt:
adds/subtracts a minute from the dates

### DIFF
--- a/tutor/specs/models/task-plans/teacher/tasking.spec.js
+++ b/tutor/specs/models/task-plans/teacher/tasking.spec.js
@@ -14,6 +14,16 @@ describe('Teacher tasking plan tasking', () => {
     tasking.opens_at = now.toISOString();
   });
 
+  it('converts to course time for save', () => {
+    expect(course.time_zone).toEqual('Central Time (US & Canada)');
+    expect(tasking.dataForSave).toEqual({
+      target_type: 'period',
+      target_id: course.periods[0].id,
+      due_at: '2015-10-14 07:00',  // 12am - 5 hours
+      opens_at: '2015-10-14 07:00',
+    });
+  });
+
   it('sets open/due but not past the opposing open/due', () => {
     tasking.setOpensDate(moment(now).year(2016));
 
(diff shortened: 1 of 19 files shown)


tutor/specs/models/task-plans/teacher/tasking.spec.js
```diff
@@ -19,20 +19,20 @@ describe('Teacher tasking plan tasking', () => {
 
     tasking.due_at = '2016-10-14T12:00:00.000Z';
     tasking.setOpensDate('2016-10-20T12:00:00.000Z');
-    expect(tasking.opens_at).toEqual('2016-10-14T12:00:00.000Z');
+    expect(tasking.opens_at).toEqual('2016-10-14T11:59:00.000Z');
 
     tasking.setOpensTime('2016-10-20T19:42:00.000Z');
-    expect(tasking.opens_at).toEqual('2016-10-14T12:00:00.000Z');
+    expect(tasking.opens_at).toEqual('2016-10-14T11:42:00.000Z');
```
Dantemss: This seems odd... how does 19:42 become 11:42? Due date does not end in :42 either.

tutor/src/components/buttons/teacher-become-student.js
```diff
@@ -86,11 +86,11 @@ class TeacherBecomesStudent extends React.Component {
...
   }
 
-  @action async becomeStudentInPeriod(period) {
+  async becomeStudentInPeriod(period) {
     const { course } = this.props;
     this.isCreating = true;
-    this.context.router.history.push(`/course/${course.id}`);
-    await period.becomeStudent();
+    const role = await period.findOrCreateTeacherStudentRole();
+    this.context.router.history.push(`/course/${course.id}/become/${role.id}`);
```
Dantemss: Doesn't this need to call `role.become()` now? Or does pushing stuff to history actually call the renderer?

tutor/src/models/course.js
```diff
@@ -215,8 +215,8 @@ class Course extends BaseModel {
     return Boolean(this.does_cost && this.userStudentRecord && !this.userStudentRecord.isUnPaid);
   }
 
-  @computed get defaultTimes() {
-    return pick(this, 'default_due_time', 'default_open_time');
+  momentInZone(date) {
+    return moment(date).tz(this.time_zone);
```
Dantemss: Does this convert the date or just append the timezone to a timezone-less date?
I ask because the BE handles opens_at and due_at as a special case and they are supposed to be treated as time-zone-less in the FE (they are always in whatever time zone the course is, but without any conversion).

The BE will completely ignore any timezone you send for opens_at, due_at.

tutor/src/screens/assignment-builder/builder/tasking.js
```diff
@@ -78,6 +78,9 @@ class Tasking extends React.Component {
 
   @action.bound onDueDateChange(date) {
     this.taskings.forEach(t => t.setDueDate(date));
+    // if (moment(date).isBefore(this.taskings[0].opens_at)) {
+
+    // }
```
Dantemss: Commented out code

tutor/src/models/course.js
nathanstitt: no it converts to the course's timezone.  We have to do that so we can send adjust the dates time in the course timezone and also send to the server in that timezone.

tutor/src/components/buttons/teacher-become-student.js
nathanstitt: no, we now call role.become in the new route that lives in the conditionalRoutes.  Calling it here caused the app to start re-rendering while it was still on the teacher dashboard and then it was a race if the route switch would happen or the dashboard would re-render

tutor/src/models/course.js
Dantemss: From our chat, this needs to be a single call to tz to be correct, something like `moment.tz(date, this.time_zone)`

tutor/src/components/buttons/teacher-become-student.js
Dantemss: OK it was just not clear to me what was actually calling the renderer.

tutor/src/models/course.js
nathanstitt: yep, I made the change

PR #2673, opened 2019-07-25
do not prompt ghost or comped students to pay

nathanstitt:
makes ghost students comped and then hides all payment nags for them

### DIFF
--- a/tutor/src/models/course/period.js
+++ b/tutor/src/models/course/period.js
@@ -79,10 +79,10 @@ class CoursePeriod extends BaseModel {
       student.update({
         role_id: role.id,
         student_identifier: '',
-        first_name: 'Review',
+        first_name: 'Instructor',
         is_comped: true,
-        last_name: 'Student',
-        name: 'Teacher Review Student',
+        last_name: 'Review',
+        name: 'Instructor Review',
         period_id: this.id,
         prompt_student_to_pay: false,
       });
--- a/tutor/src/models/course/period.js
+++ b/tutor/src/models/course/period.js
@@ -1,5 +1,6 @@
 import { find, pick, last } from 'lodash';
 import { computed, action, runInAction } from 'mobx';
+import Student from './student';
 import {
   BaseModel, identifiedBy, field, identifier, belongsTo,
 } from 'shared/model';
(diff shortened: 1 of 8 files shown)


tutor/src/models/course/student.js
```diff
@@ -12,6 +12,8 @@ export default
 class CourseStudent extends BaseModel {
   @identifier id;
 
+  static TEACHER_AS_STUDENT_ID = -9;
```
Dantemss: Hmm I *think* a negative number will never appear and so is OK...
Would be better if this was a const but I googled and seem that static consts don't exist...

tutor/src/models/course/period.js
```diff
@@ -60,14 +61,32 @@ class CoursePeriod extends BaseModel {
...
     if (!role) {
       const { data } = await this.createTeacherStudent();
       role = this.course.roles.find(r => r.id == data.id );
     }
+    runInAction(() => {
+      role.joined_at = new Date(); // adjust the date so it always appears new
+      let student = find(this.course.students, { id: Student.TEACHER_AS_STUDENT_ID });
+      if (!student) {
+        this.course.students.push({ id: Student.TEACHER_AS_STUDENT_ID });
+        student = last(this.course.students);
```
Dantemss: This would be clearer like:
```
student = { id: Student.TEACHER_AS_STUDENT_ID };
this.course.students.push(student);
```

Unless push is doing some conversion magic, of course.

tutor/src/models/course/period.js
```diff
@@ -60,14 +61,32 @@ class CoursePeriod extends BaseModel {
...
+      }
+      student.update({
+        role_id: role.id,
+        student_identifier: '',
+        first_name: 'Review',
+        is_comped: true,
+        last_name: 'Student',
+        name: 'Teacher Review Student',
```
Dantemss: Maybe `Instructor` would be better? If this is visible somewhere (scores?)

tutor/src/models/course/student.js
nathanstitt: hah, good idea.  we actually have a @readonly decorator I can use. that'd be good

tutor/src/models/course/period.js
nathanstitt: yeah, push is indeed doing magic :D it converts the plain object to a instance of Student class

tutor/src/models/course/period.js
nathanstitt: yeah that's the only spot.. Instructor is better

PR #2678, opened 2019-07-30
toggle periods in HW builder

nathanstitt:
and also clear cached student data when becoming a student

### DIFF
--- a/tutor/src/screens/assignment-builder/builder/tasking.js
+++ b/tutor/src/screens/assignment-builder/builder/tasking.js
@@ -11,6 +11,7 @@ import { TutorDateInput, TutorTimeInput } from '../../../components/tutor-input'
 import SetTimeAsDefault from './set-time-as-default';
 
 const StyledTasking = styled(Row)`
+min-height: 7rem;
 .tutor-input.form-control-wrapper.tutor-input input[disabled] {
   border-bottom: 0;
 }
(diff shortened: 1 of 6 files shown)


tutor/src/screens/assignment-builder/builder/tasking.js
```diff
@@ -127,79 +129,84 @@ class Tasking extends React.Component {
   }
 
   render() {
-    if (!this.taskings.length) { return null; }
+    const tasking = this.taskings.length ? this.taskings[0] : null;
```
Dantemss: Seems odd that an array of taskings is passed into here but only the first one is used.

tutor/src/screens/assignment-builder/builder/tasking.js
```diff
@@ -127,79 +129,84 @@ class Tasking extends React.Component {
   }
 
   render() {
-    if (!this.taskings.length) { return null; }
```
Dantemss: I'm guessing this got removed because you need it to render something (the `<Col>`) even when tasking is null?

nathanstitt: yep, if it's null we still need to render the checkbox so it can be re-selected

tutor/src/screens/assignment-builder/builder/tasking.js
nathanstitt: yeah, it's a bit strange.  it's because of when they're combined it sets all the tastings for periods at once

PR #2679, opened 2019-07-30
Don't reset tasks when they're viewed

nathanstitt:
but when they are, also reset the API so they're re-fetched


tutor/src/screens/task/ux.js
```diff
@@ -42,7 +42,7 @@ export default class TaskUX {
       return this._task.students.find(s => s.role_id == r.id);
     });
     if (teacherAsStudentRole) {
-      teacherAsStudentRole.become();
+      teacherAsStudentRole.become({ reset: false });
     }
```
nathanstitt: this was the bug... we were resetting whenever the task loaded for a teacher as student

PR #2681, opened 2019-08-05
add feature flag for pulseinsights

nathanstitt:
and default to false

### DIFF
--- a/tutor/src/models/app/pulse-insights.js
+++ b/tutor/src/models/app/pulse-insights.js
@@ -1,5 +1,6 @@
 import loadjs from 'loadjs';
 import Raven from './raven';
+import Flags from '../feature_flags';
 
 function onLoadFailure() {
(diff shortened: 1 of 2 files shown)


tutor/src/models/app/pulse-insights.js
```diff
@@ -14,6 +15,9 @@ class PI {
 
 export default {
   boot() {
+    if (!Flags.pulse_insights) {
+      return Promise.resolve({});
+    }
     window.pi = new PI();
```
reedstrm: Hmm, were does PI come from?

Dantemss: Seems to be the class above

nathanstitt: yeah, it's from the class, this is what CNX does I pretty much just copied from them

PR #2683, opened 2019-08-06
Fix time controls to use native `time` input

Dantemss:
First commit fixes inputs losing focus
Second commit replaces masked inputs with input[type=time]
Polyfill (maybe https://www.npmjs.com/package/time-input-polyfill) will be needed for Safari

### DIFF
--- a/tutor/src/screens/assignment-builder/builder/tasking.js
+++ b/tutor/src/screens/assignment-builder/builder/tasking.js
@@ -167,11 +167,10 @@ class Tasking extends React.Component {
             </Col>
             <Col md={5} xs={4}>
               <TutorTimeInput
-                key={tasking.opensAtTime || 'opens_at_time'}
+                key={'opens-at-time'}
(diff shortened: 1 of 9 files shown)


tutor/src/screens/assignment-builder/builder/tasking.js
```diff
@@ -167,11 +167,10 @@ class Tasking extends React.Component {
             </Col>
             <Col md={5} xs={4}>
               <TutorTimeInput
-                key={tasking.opensAtTime || 'opens_at_time'}
+                key={'opens-at-time'}
```
Dantemss: This is totally my fault but now that I think about it, might be a good idea to add the `${type}` here (I think you can have multiple of these in the same page if you switch out of "All Sections")

nathanstitt: ah, yep.  good catch

PR #2685, opened 2019-08-08
use `displayedChapterSection` so that baked c/s has priority

nathanstitt:
And also hide "Test Prep for AP® Courses" or "Science Practice Challenge Questions"

### DIFF
--- a/tutor/src/models/reference-book/page.js
+++ b/tutor/src/models/reference-book/page.js
@@ -16,9 +16,9 @@ const NON_ASSIGNABLE_TITLES = [
   'Conceptual Questions',
   'Problems & Exercises',
   'Test Prep for AP® Courses',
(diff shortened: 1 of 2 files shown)


tutor/src/models/reference-book/page.js
```diff
@@ -8,15 +8,17 @@ import ChapterSection from '../chapter-section';
...
+  'Glossary',
   'Key Terms',
   'Chapter Summary',
-  'Glossary',
   'Section Summary',
+  'Review Questions',
+  'Conceptual Questions',
+  'Problems & Exercises',
+  'Test Prep for AP® Courses',
+  'Test Prep for AP® Courses',
```
Dantemss: Dupe. From your description I think one of them is supposed to be `"Science Practice Challenge Questions"`

PR #2690, opened 2019-08-12
copy ecosystem_id when cloning a task plan

nathanstitt:
Previously the ecosystem id wouldn't transfer, causing the plan to load the wrong book

### DIFF
--- a/tutor/specs/components/exercises/details.spec.jsx
+++ b/tutor/specs/components/exercises/details.spec.jsx
@@ -17,6 +17,7 @@ describe('Exercise Details Component', function() {
       book,
       exercises,
       pageIds,
+      selectedSection:        book.chapter_section,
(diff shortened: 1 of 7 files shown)


tutor/specs/components/exercises/details.spec.jsx
```diff
@@ -17,6 +17,7 @@ describe('Exercise Details Component', function() {
       book,
       exercises,
       pageIds,
+      selectedSection:        book.chapter_section,
```
Dantemss: I think I removed the `isRequired` for this in the ModalManager PR btw (because a spec kept complaining).

nathanstitt: ah, I probably hadn't pulled master.    is still good to add it.

PR #2692, opened 2019-08-13
Display the terms of use faster

Dantemss:
Set User.terms during bootstrap instead of making it a computed value.
Previously something was causing it to re-initialize and delay the rendering of the terms modal.
Still flashes the content behind the modal briefly though (same as staging).

### DIFF
--- a/tutor/src/screens/student-dashboard/index.jsx
+++ b/tutor/src/screens/student-dashboard/index.jsx
@@ -4,6 +4,8 @@ import { computed } from 'mobx';
 import { observer, inject } from 'mobx-react';
 import StudentDashboard from './dashboard';
 import Courses from '../../models/courses-map';
+import User from '../../models/user';
+import LoadingScreen from 'shared/components/loading-animation';
 import { CourseNotFoundWarning } from '../../components/course-not-found-warning';
 import './styles.scss';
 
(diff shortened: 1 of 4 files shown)


tutor/src/screens/student-dashboard/index.jsx
```diff
@@ -24,6 +26,12 @@ class StudentDashboardShell extends React.Component {
   }
 
   render() {
+    // keep rendering loading screen if the user needs to agree to terms
+    // this way the screen stays the same without a flash of other content
+    if (User.terms_signatures_needed) {
```
Dantemss: OK my one concern is that with the ModalManager if a tour triggers before the terms for some reason (it shouldn't right now but might eventually...) then you'll get a tour of the loading screen..

nathanstitt: hm, not quite, but close.  Since this is the main student dash, it wouldn't render any tour wrapper, which would prevent any tours from playing.  

So what we'd get is the same, but maybe with the old flash of content before terms if the manager did anything else

Dantemss: Ah you're right... the tour will just return isReady == false and exit immediately and the terms will render... looks good then.

PR #2694, opened 2019-08-14
display figures inline and adjust preview modal styles

nathanstitt:
fixes wrapping like:
image

the paren should stay with the figure when it's not set to `inline-block`
image

### DIFF
--- a/tutor/resources/styles/components/media-preview.scss
+++ b/tutor/resources/styles/components/media-preview.scss
@@ -1,6 +1,9 @@
 .media-preview {
-  max-width: 300px;
-
+  max-width: 400px;
+  width: 400px;
+  .popover-body {
+    width: 100%;
+  }
   &.popover {
     z-index: 3;
   }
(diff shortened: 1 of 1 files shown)


tutor/resources/styles/components/media-preview.scss
```diff
@@ -1,6 +1,9 @@
 .media-preview {
-  max-width: 300px;
-
+  max-width: 400px;
+  width: 400px;
```
Dantemss: Seems redundant to have both width and max-width?

tutor/resources/styles/components/media-preview.scss
```diff
@@ -15,7 +18,20 @@
 
   @include tutor-figure();
   @include tutor-tables($tutor-neutral-light, $tutor-neutral-light);
-
+  .os-figure:not(.splash):not(.full-width) {
+    width: 100%;
```
Dantemss: Not full-width == 100% width?

tutor/resources/styles/components/media-preview.scss
Dantemss: Apparently it's only really redundant if they are exactly the same like in this case.

nathanstitt: heh, they're not redundant when flex box gets involved.  With that the width is what I uses for the normal size, but it'll allow expanding it to the maxWidth if there's space.  So if you don't want it to expand you have to set maxWidth as well

tutor/resources/styles/components/media-preview.scss
nathanstitt: full-width is actually 100% + 280px so it expands to fill the entire panel

PR #2696, opened 2019-08-16
Use regex to match multiple variants of titles

nathanstitt:
Some titles have weird UTF characters 😭

### DIFF
--- a/tutor/src/models/reference-book/page.js
+++ b/tutor/src/models/reference-book/page.js
@@ -1,5 +1,5 @@
 import { readonly } from 'core-decorators';
-import { merge, get, pick, includes } from 'lodash';
+import { merge, get, pick } from 'lodash';
 import { action, computed } from 'mobx';
 import {
   BaseModel, identifiedBy, identifier, field, session,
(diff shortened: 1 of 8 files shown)


tutor/src/models/reference-book/page.js
```diff
@@ -121,12 +112,13 @@ class ReferenceBookPage extends BaseModel {
     if (this.bookIsCollated) {
       return this.hasBakedChapterSection;
     }
-    return !this.isIntro && !includes(SUPPLEMENTARY_CONTENT_TITLES, this.title);
+    return !this.isIntro && !this.isAssignable;
```
Dantemss: I think the logic is backwards?
SUPPLEMENTARY_CONTENT_TITLES became NON_ASSIGNABLE_TITLES
The includes for SUPPLEMENTARY_CONTENT_TITLES is negated here.
But the includes for NON_ASSIGNABLE_TITLES is negated twice (once here and once in isAssignable).

nathanstitt: Hm, no I don't see anything obviously incorrect.  I removed `SUPPLEMENTARY_CONTENT_TITLES` just because it was identical to the `NON_ASSIGNABLE_TITLES`

What we want here is all chapters that are not "regular content".  If the book is baked, that's easy we can use the presence of a chapter/section.  For others we need to ignore intros and all the other's at the end, which is anything that's not assignable.

PR #2697, opened 2019-08-16
Fix student dashboard infinite loading

nathanstitt:
and also add api pending condition to terms modal

### DIFF
--- a/tutor/src/components/terms-modal.jsx
+++ b/tutor/src/components/terms-modal.jsx
@@ -32,8 +32,11 @@ class TermsModal extends React.Component {
 
   // for terms to be displayed the user must be in a course and need them signed
   @computed get isReady() {
-    return !!(
-      this.props.canBeDisplayed && User.terms_signatures_needed && !isEmpty(User.unsignedTerms)
+    return Boolean(
+      !User.terms.api.isPending &&
+        this.props.canBeDisplayed &&
+        User.terms_signatures_needed &&
+        !isEmpty(User.unsignedTerms)
     );
   }
 
--- a/tutor/src/components/terms-modal.jsx
+++ b/tutor/src/components/terms-modal.jsx
@@ -6,7 +6,7 @@ import { Modal, Button } from 'react-bootstrap';
 import classnames from 'classnames';
 import Branding from './branding/course';
 import User from '../models/user';
-import { isEmpty, map } from 'lodash';
+import { map } from 'lodash';
 import String from '../helpers/string';
 import ModalManager from './modal-manager';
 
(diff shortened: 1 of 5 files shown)


tutor/src/components/terms-modal.jsx
```diff
@@ -32,8 +32,11 @@ class TermsModal extends React.Component {
 
   // for terms to be displayed the user must be in a course and need them signed
   @computed get isReady() {
-    return !!(
-      this.props.canBeDisplayed && User.terms_signatures_needed && !isEmpty(User.unsignedTerms)
+    return Boolean(
+      !User.terms.api.isPending &&
+        this.props.canBeDisplayed &&
+        User.terms_signatures_needed &&
+        !isEmpty(User.unsignedTerms)
```
Dantemss: We should remove `isEmpty` (and rely on changing `User.terms_signatures_needed`) or also add the isEmpty to the loading screen check.

tutor/src/models/user/terms.js
```diff
@@ -30,6 +30,7 @@ class UserTerms extends BaseModel {
 
   onLoaded({ data }) {
     this.terms = data;
+    this.user.terms_signatures_needed = this.terms.length > 0;
```
Dantemss: I think this should look at `this.unsigned` instead of `this.terms`.

Dantemss: That way it would match the BE's behavior: https://github.com/openstax/tutor-server/blob/master/app/representers/api/v1/user_representer.rb#L74

tutor/src/components/terms-modal.jsx
nathanstitt: yeah, lets change loading to use that. I think this should still be here because we don't want to display unless there are unsignedTerms

tutor/src/models/user/terms.js
```diff
@@ -30,6 +30,7 @@ class UserTerms extends BaseModel {
 
   onLoaded({ data }) {
     this.terms = data;
+    this.user.terms_signatures_needed = this.terms.unsigned > 0;
```
Dantemss: I think it's just `this.unsigned` here. `this.terms` is an object from the API right? Not sure it has an `unsigned` field.

tutor/src/screens/student-dashboard/index.jsx
```diff
@@ -28,7 +28,7 @@ class StudentDashboardShell extends React.Component {
   render() {
     // keep rendering loading screen if the user needs to agree to terms
     // this way the screen stays the same without a flash of other content
-    if (User.terms_signatures_needed) {
+    if (User.unsignedTerms.length) {
```
Dantemss: I think this should look at both `terms_signatures_needed` and `unsignedTerms`. It REALLY should match the logic in `isReady` (minus the `isPending`) part.

tutor/src/models/user.js
```diff
@@ -93,7 +93,11 @@ class User extends BaseModel {
   }
 
   @computed get unsignedTerms() {
-    return this.terms.unsigned;
+    return (this.terms && this.terms.unsigned) || [];
+  }
+
+  @computed get shouldSignTerms() {
+    return this.terms_signatures_needed && !isEmpty(this.unsignedTerms);
   }
```
nathanstitt: nice!

PR #2698, opened 2019-08-20
Display a loading indicator to teacher-as-student

nathanstitt:
needs https://github.com/openstax/tutor-server/pull/1894

image


tutor/src/models/task-plans/student.js
```diff
@@ -109,6 +109,19 @@ class StudentTaskPlans extends Map {
...
   }
 
+  @computed get isTeacherWaitingForLatest() {
+    return Boolean(
+      this.course.primaryRole.isTeacherStudent && !this.isLatestPresent
+    );
+  }
+
+  @computed get isLatestPresent() {
+    if (this.course.primaryRole.isStudent) { return false; }
```
Dantemss: This line seems odd. Is it just because students can't get the plans?

nathanstitt: hm yeah it does seem odd.  I'd put it there just so it wouldn't do anymore work in the case of a student but really there's not much more to do since for a student the teacher plans will be empty.  I'll remove it

PR #2699, opened 2019-08-20
display "only visible to instructors" message for non-open tasks

nathanstitt:
needs https://github.com/openstax/tutor-server/pull/1895

image

### DIFF
--- a/tutor/src/models/task-plans/student/task.js
+++ b/tutor/src/models/task-plans/student/task.js
@@ -1,6 +1,6 @@
 import { observable } from 'mobx';
 import { computed } from 'mobx';
-import { invoke } from 'lodash';
+import { invoke, get } from 'lodash';
 import moment from 'moment';
 import Time from '../../time';
 import {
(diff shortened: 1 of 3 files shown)


tutor/src/models/task-plans/student/task.js
```diff
@@ -93,12 +98,22 @@ class StudentTask extends BaseModel {
...
+  }
+
   @computed get canWork() {
     //students cannot work or view a task if it has been deleted and they haven't started it
     return Boolean(
-      this.opens_at < Time.now && !(
-        this.is_deleted &&
-          this.complete_exercise_count === 0
+      this.isTeacherStudent || (
+        this.isOpen && !(
```
Dantemss: This negation is confusing, but seems correct.

tutor/src/models/task-plans/student/task.js
```diff
@@ -93,12 +98,22 @@ class StudentTask extends BaseModel {
...
 
+  @computed get isTeacherStudent() {
+    return true === get(this, 'tasks.course.currentRole.isTeacherStudent');
+  }
+
+  @computed get isOpen() {
+    return this.opens_at < Time.now;
+  }
+
   @computed get canWork() {
```
Dantemss: `canWork()` is misleading... I think this is more like... `canView()` or `canViewSteps()`
If the task is deleted they can't work but they can sometimes view the steps.

nathanstitt: yeah,  that's true but I'll leave it alone for now since this is used several different spots

PR #2700, opened 2019-08-20
fix editing cloned homeworks

nathanstitt:
And also save the `is_feedback_immediate` setting


Brooke-Fail:
Verified in Staging and Production.

PR #2705, opened 2019-08-23
Tweak task builder ui

nathanstitt:
Fixes a few small irritants:

* Sometimes the individual taskings for a new assignment would be shown one-per-period
* Sometimes the time would be shown as invalid even though it wasn't
* hides the browser auto-complete on the date inputs
* Shows opens at time if it's the same day as today
image

### DIFF
--- a/tutor/src/components/tutor-input.js
+++ b/tutor/src/components/tutor-input.js
@@ -34,6 +34,8 @@ class TutorInput extends React.Component {
     onUpdated: PropTypes.func,
     autoFocus: PropTypes.bool,
     hasValue: PropTypes.bool,
+    defaultValue: PropTypes.string,
+    default: PropTypes.string,
   };
 
   state = { errors: [] };
--- a/tutor/src/components/tutor-input.js
+++ b/tutor/src/components/tutor-input.js
@@ -39,7 +39,7 @@ class TutorInput extends React.Component {
   state = { errors: [] };
 
   componentDidMount() {
-    const errors = this.props.validate(this.props.default);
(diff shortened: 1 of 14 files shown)


tutor/src/components/tutor-input.js
```diff
@@ -323,7 +324,7 @@ class TutorTimeInput extends React.Component {
   };
 
   render() {
-    const inputProps = omit(this.props, 'defaultValue', 'onChange', 'formatCharacters', 'value');
+    const inputProps = omit(this.props, 'default', 'onChange', 'formatCharacters', 'value');
```
Dantemss: Hmm shouldn't `default` and/or `defaultValue` be in `propTypes`?

tutor/src/models/task-plans/teacher/plan.js
```diff
@@ -22,6 +22,25 @@ const TUTOR_SELECTIONS = {
...
+
+  return moment(
+    findLatest(
+      findEarliest(
+        course.bounds.end,
+        defaultOpensAt,
+      ),
+      course.bounds.start.add(1, 'minute'),
+    ),
+  ).hour(hour).minute(minute).startOf('minute').toISOString();
```
Dantemss: I think the `hour` and `minute` need to be applied directly to `defaultOpensAt` (before the comparison). Otherwise subtract a day from `course.bounds.end` and add a day to `course.bounds.start` (instead of 1 minute).

nathanstitt: Hm, I think course start/end is always midnight, but yah wouldn't hurt to add a day to be safe.  i'll do that

PR #2706, opened 2019-08-23
add AP Bio & Physics to errata form

nathanstitt:
The form has a bug that prevents it from loading unless we pass in a matching book title


shared/src/helpers/exercise.js
```diff
@@ -17,6 +17,8 @@ const BOOK_UID_XREF = {
   '185cbf87-c72e-48f5-b51e-f14f21b5eabd': 'Biology 2e', // non baked
   '8d50a0af-948b-4204-a71d-4826cba765b8': 'Biology 2e', // baked
   '405335a3-7cff-4df2-a9ad-29062a4af261': 'College Physics',
+  '8d04a686-d5e8-4798-a27d-c608e4d0e187': 'College Physics',
```
Dantemss: `College Physics for AP® Courses`?

nathanstitt: nope, that's not in their list so we can't use it.  I noted that on the card

Dantemss: Maybe we need to get them to fix it? So we don't mix up the 2 erratas.

Dantemss: Ah seems like `The AP Physics Collection` might work.

PR #2718, opened 2019-09-05
Fetch terms when needed and wait util they're present

nathanstitt:
Also removes the user.unsignedTerms in favor of just User.terms.unsigned

### DIFF
--- a/tutor/src/models/user/terms.js
+++ b/tutor/src/models/user/terms.js
@@ -1,8 +1,7 @@
 import {
   BaseModel, identifiedBy, identifier, field, belongsTo, hasMany,
 } from 'shared/model';
-
-import { computed } from 'mobx';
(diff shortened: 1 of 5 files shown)


tutor/src/models/user/terms.js
```diff
@@ -25,7 +24,7 @@ class UserTerms extends BaseModel {
 
   constructor(attrs) {
     super(attrs);
-    if (this.user.terms_signatures_needed) { this.fetch(); }
+    observe(this.user, 'terms_signatures_needed', this.fetchIfNeeded);
   }
```
nathanstitt: this is the bug.  the user model is created before it's bootstrapped, which causes terms_signatures_needed to be false initially, then later it'll be set to true when the app is bootstrapped

PR #2723, opened 2019-09-06
Fix steps moving backwards after clicking continue

Dantemss:
- Fix steps moving backwards after clicking continue
- Select the correct multipart breadcrumb when saving free response/selecting an answer/answering multipart questions
- Prevent the next multipart question from jumping due to feedback loading


tutor/src/screens/task/ux.js
```diff
@@ -104,11 +104,16 @@ export default class TaskUX {
...
-    if (!this.task.isFeedbackAvailable) {
-      this.onAnswerContinue(step);
-    } else {
-      this.moveToStep(step);
+
+    if (
+      step.multiPartGroup &&
+      this.task.isFeedbackAvailable &&
+      this._stepIndex > this.steps.indexOf(step)
+    ) {
```
nathanstitt: should this be `this._stepIndex != this.steps.indexOf(step)` ?  If we only check for greater then it won't scroll up if you select a previous step

PR #2724, opened 2019-09-13
replace joyride tours with own; update React/Bootstrap/Router

nathanstitt:
Is a pretty big update.   🤞 nothing has broken but there probably is.


tutor/src/models/tour/ride.js
```diff
@@ -28,110 +12,97 @@ class TourRide extends BaseModel {
...
+  }
+
+  @computed get nextLabel() {
+    if (!this.canGoForward || !this.showStepsProgress) { // last step
+      return 'Close';
     }
+    return `Next ${this._stepIndex + 1}/${this.validSteps.length}`;
+  }
+
+  @computed get showStepsProgress() {
```
Dantemss: I would rename this to `hasMultipleSteps()` or similar to be clearer.

tutor/src/screens/qa-view/index.jsx
```diff
@@ -18,20 +20,17 @@ class QAViewWrapper extends React.Component {
...
     topNavbar: PropTypes.instanceOf(NavbarContext).isRequired,
+    history: PropTypes.object.isRequired,
   }
 
-  static contextTypes = {
-    router: PropTypes.object,
-  }
-
-  ux = new UX({ router: this.context.router });
+  ux = new UX({ router: this.props.history });
```
Dantemss: Shouldn't this be `history:`?

tutor/src/screens/student-dashboard/event-row.jsx
```diff
@@ -45,22 +45,20 @@ EventTime.propTypes = {
 };
 
 export default
+@withRouter
```
Dantemss: Does `withRouter` set `props.history` automatically?

tutor/src/screens/task/step/card.js
```diff
@@ -25,6 +25,7 @@ const LoadingCard = styled(InnerStepCard)`
   min-width: 960px;
   padding: 2rem;
 `;
+LoadingCard.displayName = 'LoadingCard';
```
Dantemss: Maybe displayName is actually not displayed? Hmm odd name choice :)

tutor/specs/screens/assignment-builder/builder/tasking.spec.js
```diff
@@ -15,7 +15,7 @@ describe('Tasking Builder', () => {
     props = { ux };
   });
 
-  it('sets times', function() {
+  fit('sets times', function() {
```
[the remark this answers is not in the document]
nathanstitt: heh, the `f` stands for "focused", meaning it'll only run that test.  I use that when attempting to fix a single thing.  Looks like I forgot to return it to be just `it`

tutor/specs/components/navbar/support-menu.spec.jsx
```diff
@@ -36,28 +36,11 @@ describe('Support Menu', () => {
...
     Chat.isEnabled = true;
-    const menu = mount(<SupportMenu {...props} />);
+    const menu = mount(<C><SupportMenu {...props} /></C>);
     menu.find('button.dropdown-toggle').simulate('click');
     menu.find('.chat.enabled a').simulate('click');
     expect(Chat.start).toHaveBeenCalled();
     menu.unmount();
   });
 
-  it('renders support links when in a course for student', () => {
```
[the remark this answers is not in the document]
nathanstitt: You know I'm not sure why that was removed 🤔   I've restored them

tutor/src/components/buttons/teacher-become-student.js
```diff
@@ -17,6 +18,7 @@ const BecomeButton = styled(Button).attrs({
     border-bottom-width: 1px;
   }
 `;
+BecomeButton.displayName = 'BecomeButton';
```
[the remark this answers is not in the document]
nathanstitt: yeah setting the `displayName` is helpful for debugging.  Sometimes when the component has too many decorators applied the displayName is lost so that it looks like just `Component` or something like that to `jest` instead of a properly named one.

tutor/src/screens/student-dashboard/event-row.jsx
nathanstitt: yes, it provides the `history` and `location` props.  https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/withRouter.md

The updated react-router uses new-style React.Context https://reactjs.org/docs/context.html

tutor/src/screens/task/step/card.js
nathanstitt: it's only used  in the React dev tools and in jest/enzyme debugging

PR #2725, opened 2019-09-13
use baked chapter section for page title on reading steps

nathanstitt:
Otherwise AP physics will display the off-by-one verion

also needs https://github.com/openstax/tutor-server/pull/1908

### DIFF
--- a/tutor/src/models/student-tasks/step.js
+++ b/tutor/src/models/student-tasks/step.js
@@ -24,7 +24,9 @@ class StudentTaskPlaceHolderStep extends TaskStepContent { }
 
 class StudentTaskReadingStep extends TaskStepContent {
   @observable title;
-  @lazyGetter chapterSection = new ChapterSection(this.chapter_section);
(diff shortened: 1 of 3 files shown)


tutor/src/models/student-tasks/step.js
```diff
@@ -24,7 +24,9 @@ class StudentTaskPlaceHolderStep extends TaskStepContent { }
 
 class StudentTaskReadingStep extends TaskStepContent {
   @observable title;
-  @lazyGetter chapterSection = new ChapterSection(this.chapter_section);
+  @lazyGetter chapterSection = new ChapterSection(
+    this.baked_chapter_section || this.chapter_section
```
Dantemss: Seems odd that this file and reference-book/page.js accomplish the same thing in slightly different ways.

nathanstitt: yes it is a bit cumbersome.  the data is quite a bit different for the two though so it's kinda hard to merge them

PR #2726, opened 2019-09-13
don't display deleted tasks that are not worked

nathanstitt:
previously we showed them until the student clicked "hide"

### DIFF
--- a/tutor/specs/factories/student-task.js
+++ b/tutor/specs/factories/student-task.js
@@ -18,7 +18,7 @@ Factory.define('StudentDashboardTask')
   .complete_exercise_count(({ object }) =>
     fake.random.number({ min: 0, max: object.exercise_count })
   )
-  .completed_step_count(({ object }) => object.complete_exercise_count)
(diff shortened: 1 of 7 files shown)


tutor/specs/factories/student-task.js
```diff
@@ -18,7 +18,7 @@ Factory.define('StudentDashboardTask')
   .complete_exercise_count(({ object }) =>
     fake.random.number({ min: 0, max: object.exercise_count })
   )
-  .completed_step_count(({ object }) => object.complete_exercise_count)
+  .completed_steps_count(({ object }) => object.complete_exercise_count)
   .correct_exercise_count(({ object }) =>
```
nathanstitt: this was a bug, the attribute was missing the `s` on `steps` so it's value was never used.  Once I corrected that, it caused the snapshots to change since the tasks now had activity.

nathanstitt:
@yep, just commented why on that https://github.com/openstax/tutor-js/pull/2726/files#r324269837

PR #2729, opened 2019-09-17
misc router fixes

nathanstitt:
Fixes a few issued identified in testing:
 * calendar couldn't be clicked to add assignment
 * new course wizard failed
 * task steps would cause errors when rapidly advancing

### DIFF
--- a/tutor/specs/screens/task/__snapshots__/milestones.spec.js.snap
+++ b/tutor/specs/screens/task/__snapshots__/milestones.spec.js.snap
@@ -730,12 +730,7 @@ exports[`Reading Milestones Component matches snapshot 1`] = `
         "webview_url": "https://qa.cnx.org/contents/185cbf87-c72e-48f5-b51e-f14f21b5eabd",
         "year": 2017,
       },
-      "history": Object {
-        "createHref": [MockFunction],
-        "listen": [MockFunction],
-        "push": [MockFunction],
-        "replace": [MockFunction],
-      },
+      "history": undefined,
       "viewedInfoSteps": Array [],
       "window": [Window],
     }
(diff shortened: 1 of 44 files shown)


tutor/specs/screens/task/__snapshots__/milestones.spec.js.snap
```diff
@@ -730,12 +730,7 @@ exports[`Reading Milestones Component matches snapshot 1`] = `
         "webview_url": "https://qa.cnx.org/contents/185cbf87-c72e-48f5-b51e-f14f21b5eabd",
         "year": 2017,
       },
-      "history": Object {
-        "createHref": [MockFunction],
-        "listen": [MockFunction],
-        "push": [MockFunction],
-        "replace": [MockFunction],
-      },
+      "history": undefined,
```
Dantemss: Is this a problem?

tutor/src/screens/task/ux.js
```diff
@@ -195,10 +193,19 @@ export default class TaskUX {
...
+      CenterControls.currentTaskStep = this.currentStep;
     }
   }
 
+  @computed get isApiPending() {
+    return get(this.currentStep, 'api.isPending', false);
+  }
+
   @computed get canGoForward() {
+    if (this.isApiPending) { return false; }
```
Dantemss: I thought we wanted them to be able to move forward while the answer is pending?

tutor/specs/screens/task/__snapshots__/milestones.spec.js.snap
nathanstitt: nah, thats just the snapshot

tutor/src/screens/task/ux.js
nathanstitt: we did but if the step type changes then it's possible the code would start to render one type and then it'd update afterwards.  I suspect that might be part of the `map on undefined` issue

PR #2739, opened 2019-09-23
remove baked_chapter_section fields

nathanstitt:
The BE will no longer send them

Goes along with https://github.com/openstax/tutor-server/pull/1912


Dantemss:
I'm just going to trust that the snaps are correct because github won't render them 😂

PR #2743, opened 2019-09-25
scroll browser window to top of page when screen renders

nathanstitt:
Fixes cases where something at bottom page is clicked and then screen doesn't scroll to top


nathanstitt:
yeah, I thought about that but the problem with that is that some routes shouldn't scroll to top, for instance when going through MPQ or reviewing tasks

PR #2744, opened 2019-09-25
Catch and display errors on task steps

nathanstitt:
Update wording for saving and suggest going back

image

### DIFF
--- a/shared/src/model.js
+++ b/shared/src/model.js
@@ -40,16 +40,17 @@ export class BaseModel {
 
   @action
   setApiErrors(error) {
+    error.isRecorded = true;
     const errors = get(error, 'response.data.errors');
     if (errors) {
       this.api.errors = {};
       errors.forEach(e => this.api.errors[e.code] = e);
-      error.isRecorded = true;
     } else if (error) {
       this.api.errors = { error: error.toString() };
     } else {
       this.api.errors = {};
     }
+    this.api.errors.last = error;
   }
 
 }
--- a/shared/src/model.js
+++ b/shared/src/model.js
@@ -45,12 +45,15 @@ export class BaseModel {
     if (errors) {
       this.api.errors = {};
       errors.forEach(e => this.api.errors[e.code] = e);
+      this.api.errors.last = error;
     } else if (error) {
-      this.api.errors = { error: error.toString() };
+      this.api.errors = {
+        error: error.toString(),
+        last: error,
+      };
     } else {
       this.api.errors = {};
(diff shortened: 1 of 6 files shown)


shared/src/model.js
```diff
@@ -40,16 +40,17 @@ export class BaseModel {
...
     if (errors) {
       this.api.errors = {};
       errors.forEach(e => this.api.errors[e.code] = e);
-      error.isRecorded = true;
     } else if (error) {
       this.api.errors = { error: error.toString() };
     } else {
       this.api.errors = {};
     }
+    this.api.errors.last = error;
```
Dantemss: Won't this be something weird if error is falsy? Why would error be falsy anyway?

tutor/src/screens/task/step/failure.js
```diff
@@ -17,45 +18,53 @@ const StyledFailure = styled(StepCard)`
...
     }
-
-    const { step } = this.props;
     if (!step) {
-      return Raven.log('Failed to load assignment step');
+      errMsg.push(`Failed to ${this.requestAction} assignment step`);
+    }
+
+    const { last: _, ...errors } = step.api.errors;
+    errMsg.push(`Failed to ${this.requestAction} assignment step id: ${step.id}, error: ${titleize(errors)}`);
```
Dantemss: Ohhh so the errors might have happened when trying to save... interesting.

tutor/src/screens/task/step/failure.js
```diff
@@ -17,45 +18,53 @@ const StyledFailure = styled(StepCard)`
...
+    if (errMsg.length) {
+      Raven.log(errMsg.join('\n'));
```
Dantemss: Can you take the `step.id` out of the message and put it in the extras hash? Have to add that argument to the `log` function in the Raven model. This is so Sentry can actually merge those issues automatically.

shared/src/model.js
nathanstitt: hm yep, we probably should't set `last` if there wasn't an error.  I'll update

tutor/src/screens/task/step/failure.js
nathanstitt: yeah, but hopefully not since that's pretty bad.  But if it does we need to log it so we can pay extra attention.

tutor/src/screens/task/step/failure.js
nathanstitt: yep, good idea

PR #2754, opened 2019-10-02
move table caption to bottom and adjust margins

nathanstitt:
AP request


nathanstitt:
@Dantemss yeah it was approved to move them all, glad you noticed though

PR #2764, opened 2019-10-08
Retain same section selections on QL

nathanstitt:
Leave sections selected when showing exercises and then adding more

### DIFF
--- a/tutor/src/components/exercises/details.jsx
+++ b/tutor/src/components/exercises/details.jsx
@@ -3,8 +3,6 @@ import React from 'react';
 import { first, partial, findIndex } from 'lodash';
 import { observer } from 'mobx-react';
 import { observable, computed, action } from 'mobx';
-import ScrollTo from '../../helpers/scroll-to';
-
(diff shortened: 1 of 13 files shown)


tutor/src/components/exercises/details.jsx
```diff
@@ -28,36 +26,33 @@ class ExerciseDetails extends React.Component {
...
+    }
+  }
+
+  componentDidUpdate(prevProps) {
+    if (!this.props.selectedSection.eq(prevProps.selectedSection)) {
+      const index = this.exercises.findIndex(
+        ex => ex.page.chapter_section.eq(this.props.selectedSection)
+      );
+      if (index != -1) {
+        this.currentIndex = index;
```
Dantemss: Shouldn't `currentIndex` just be a computed property?

nathanstitt: nah, what this component does is page through an array of exercises.  it's pretty old though, what we'd probably do now is create a UX model that'd control the state

PR #2765, opened 2019-10-09
decrease polling interval for new assignments

nathanstitt:
sets to 10 secs for "fast polling" state and 1 hour for regular checks

fast polling is enabled for:
 * freshly enrolled students
 * instructors who've just made an assignment then went into student view


nathanstitt:
Good idea, I bumped that to 30 tries which will equal out to 5 minutes

PR #2768, opened 2019-10-10
reset content and fetched status when steps are reused

nathanstitt:
Otherwise the type can change but it's content will not, causing exercises to not have steps

### DIFF
--- a/tutor/specs/screens/task/reading.spec.js
+++ b/tutor/specs/screens/task/reading.spec.js
@@ -8,6 +8,11 @@ describe('Reading Tasks Screen', () => {
 
   beforeEach(() => {
     const task = Factory.studentTask({ type: 'reading' });
+    history = {
+      push: (url) => {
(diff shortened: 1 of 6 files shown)


tutor/specs/screens/task/reading.spec.js
```diff
@@ -8,6 +8,11 @@ describe('Reading Tasks Screen', () => {
 
   beforeEach(() => {
     const task = Factory.studentTask({ type: 'reading' });
+    history = {
+      push: (url) => {
+        props.ux.goToStep(Number(ld.last(url.split('/'))) - 1, false);
```
Dantemss: I think this fails if the url ends in `/`?

nathanstitt: yeah, it probably would but this is just a hacked together mock for the spec so it's ok.  It won't have to deal with stuff like code would "in the wild"

PR #2771, opened 2019-10-12
use book and page uuids to fetch and save notes

nathanstitt:
Otherwise notes are broken for pages without a chapter.section

goes along with https://github.com/openstax/tutor-server/pull/1918


nathanstitt:
@Dantemss I'm going to go ahead and merge this so we can start testing.  I've tested and it seems to work well with the BE PR

PR #2781, opened 2019-10-18
collapse & expand chapters on book TOC menu

nathanstitt:
Makes the sidebar TOC menu be a tree like view. Care is taken so the menu should be able to handle 3 levels of nesting as well

image


tutor/src/components/book-menu/ux.js
```diff
@@ -0,0 +1,80 @@
...
+
+  eachParent(node, fn) {
+    if (node) fn(node);
+    while ((node = node.parent)) {
+      fn(node);
+    }
+  }
+
+  isExpanded(node) {
+    return Boolean(this.expanded.get(node.id));
```
Dantemss: This might get slow if too many things are expanded, but I'm not sure if it's an issue.

nathanstitt: probably not, expanded is a Map so it should be a relatively fast hash lookup

PR #2785, opened 2019-10-22
fix dates when toggling between all and partial sections

nathanstitt:
* don't allow selecting all periods if any are unchecked
* retain due_at date when switching

### DIFF
--- a/tutor/src/screens/assignment-builder/ux.js
+++ b/tutor/src/screens/assignment-builder/ux.js
@@ -78,10 +78,14 @@ class AssignmentBuilderUX {
   }
 
   @computed get isInitializing() {
-
     return !this.isReady || !this.plan || this.plan.api.isPendingInitialFetch;
   }
 
+  @computed get canSelectAllSections() {
+    return Boolean(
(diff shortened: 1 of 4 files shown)


tutor/src/screens/assignment-builder/ux.js
```diff
@@ -78,10 +78,14 @@ class AssignmentBuilderUX {
   }
 
   @computed get isInitializing() {
-
     return !this.isReady || !this.plan || this.plan.api.isPendingInitialFetch;
   }
 
+  @computed get canSelectAllSections() {
+    return Boolean(
+      this.plan.isEditable && this.plan.tasking_plans.length == this.course.periods.length
```
Dantemss: 🤔 when would the second part be false?

nathanstitt: if they unselect a section in the builder it removes that tasking plan.  in that case the lengths won't match and this will prevent them from re-selecting "all sections".  We could also check that the tasking_plans period_id matches periods, but this is close enough and much easier to compute

PR #2786, opened 2019-10-23
Fix deleting sections

nathanstitt:
if the section being deleted was the last one the tabs would throw an error, preventing deletion


nathanstitt:
yeah, they're only soft deleted so hopefully nothing blows up there.

PR #2790, opened 2019-10-25
add a 1.5 seconds delay before a step can be advanced

nathanstitt:
This stops keyboard arrows or clicking on next arrow from rapidly skipping all the reading steps

### DIFF
--- a/tutor/src/screens/task/ux.js
+++ b/tutor/src/screens/task/ux.js
@@ -14,6 +14,7 @@ export default class TaskUX {
   // privateish
   @observable _stepIndex = 0;
   @observable viewedInfoSteps = [];
+  @observable isLocked = false;
 
   constructor({ task, stepIndex = 0, history, windowImpl, course }) {
     this.history = history;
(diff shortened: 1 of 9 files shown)


tutor/src/screens/task/ux.js
```diff
@@ -278,17 +287,15 @@ export default class TaskUX {
...
   }
 
   @computed get currentStepIndex() {
     return this._stepIndex;
   }
 
   @computed get nextStep() {
-    return this.canGoForward ?
-      this.steps[this._stepIndex + 1] : null;
+    return this.steps[this._stepIndex + 1];
```
Dantemss: No risk of going past the last one? Same thing above.

nathanstitt: yeah, a slight one if the calling code doesn't call `canGoForward`, but this method isn't used anywhere outside of this class so seems safe enough.  Also `canGoForward` now calls this method, after it checks if it's safe to do so.  Without removing this you get a circle

PR #2794, opened 2019-10-29
Check period ids on taskings instead of just length

nathanstitt:
If any periods are deleted the length will not match


Dantemss:
Like... shouldn't we just create the missing tasking plans? Maybe copy some attributes from the first or last tasking plan.

nathanstitt:
Assuming they had selected individual for that assignment, they’d be able to select all  sections after checking the new period


If they had not selected individual the new period should be added automatically

PR #2800, opened 2019-10-30
use sans-serif font for table cells

nathanstitt:
before:
image

after:

image


Brooke-Fail:
Verified in Production.

PR #2808, opened 2019-11-06
remove drop cap style from paragraphs

jivey:
Before:
after

After:
before

### DIFF
--- a/tutor/resources/styles/book-content/base.scss
+++ b/tutor/resources/styles/book-content/base.scss
@@ -63,7 +63,6 @@
   section:not(.references) h1 + p:first-of-type,
   [data-type="document-title"] ~ p:first-of-type {
     margin-top: 2rem;
-    @include tutor-reading-first-letter();
   }
 
   // prevent larger images from pushing out past their container.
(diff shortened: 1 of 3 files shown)


tutor/resources/styles/book-content/base.scss
```diff
@@ -63,7 +63,6 @@
   section:not(.references) h1 + p:first-of-type,
   [data-type="document-title"] ~ p:first-of-type {
     margin-top: 2rem;
-    @include tutor-reading-first-letter();
   }
```
nathanstitt: Awesome!  can you also do a search for `tutor-reading-first-letter` and remove the mixin if it's not in use anywhere else? I suspect this was the only spot it was referenced.

jivey: Good call, this was unreferenced and unlikely to be used in the future. Removed in a follow up commit 👍

nathanstitt:
not a big deal on this one since what's happening is entirely self-explanatory, but typically for anything that has visual changes we like to include a before and after screenshot on the PR.

jivey:
Agree, updated in the PR description!

PR #2809, opened 2019-11-06
fix section selector

jivey:
"All Sections" in the UI would be selected in cases it shouldn't, this fixes the check for what to render by changing `isShowingPeriodTaskings` to consider `canSelectAllSections` as well.

### DIFF
--- a/tutor/specs/screens/assignment-builder/builder.spec.js
+++ b/tutor/specs/screens/assignment-builder/builder.spec.js
@@ -40,14 +40,16 @@ describe('Task Plan Builder', function() {
 
   it('selects All Sections when appropriate', () => {
     const builder = mount(<C><Builder {...props} /></C>);
-    expect(builder.find('#hide-periods-radio').props()).toHaveProperty('checked', true);
-    expect(builder.find('#show-periods-radio').props()).toHaveProperty('checked', false);
+    expect(builder).toHaveRendered('#hide-periods-radio[checked=true]');
+    expect(builder).toHaveRendered('#show-periods-radio[checked=false]');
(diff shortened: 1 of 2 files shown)


tutor/specs/screens/assignment-builder/builder.spec.js
```diff
@@ -37,4 +37,17 @@ describe('Task Plan Builder', function() {
...
+    expect(builder.find('#hide-periods-radio').props()).toHaveProperty('checked', true);
+    expect(builder.find('#show-periods-radio').props()).toHaveProperty('checked', false);
+  });
+
+  it('selects Individual Sections when appropriate', () => {
+    props.ux.isShowingPeriodTaskings = true;
+    const builder = mount(<C><Builder {...props} /></C>);
+    expect(builder.find('#hide-periods-radio').props()).toHaveProperty('checked', false);
+    expect(builder.find('#show-periods-radio').props()).toHaveProperty('checked', true);
+  });
```
nathanstitt: Nice specs overall, they're testing exactly what we'd want to ensure happens.

Typically we prefer to test what's rendered instead of reading props, that way more of the stack is checked.  I think you could do:
```
    expect(builder).toHaveRendered('#hide-periods-radio[checked=true]');
    expect(builder).toHaveRendered('#show-periods-radio[checked=false]');
    builder.unmount();

```

It's also important to call `unmount()` on anything mounted.  Otherwise the component will unmount later whenever jest tears down the spec.  If there is any error while it unmounts, it will be very difficult to track down since it won't correspond to any currently running spec.

PR #2810, opened 2019-11-07
chapters never have content

nathanstitt:
regardless of baked status, chapter is not a page


nathanstitt:
Travis is failing because this PR branched from the release tag.  Specs all pass locally so I'm merging it

PR #2813, opened 2019-11-08
use break-word css rule

nathanstitt:
Fixes links that would break anywhere, not just on spaces:
image

after:
image


Brooke-Fail:
Verified in QA.

PR #2827, opened 2019-11-19
set courseId on book ux

nathanstitt:
Otherwise the rewrite book link can't find it and links are incorrect

### DIFF
--- a/tutor/src/screens/task/page-content-ux.js
+++ b/tutor/src/screens/task/page-content-ux.js
@@ -7,6 +7,9 @@ export default class PageContentUX extends BookUX {
   constructor({ main }) {
     super();
     this.mainUX = main;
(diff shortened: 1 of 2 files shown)


tutor/src/screens/task/page-content-ux.js
```diff
@@ -7,6 +7,9 @@ export default class PageContentUX extends BookUX {
   constructor({ main }) {
     super();
     this.mainUX = main;
+    this.update({
+      courseId: this.mainUX.course.id,
```
Dantemss: Don't you need to use the courseId or check that the course is set?

nathanstitt: no, "mainUX" is the ux on the tasks, which will always have a course

PR #2829, opened 2019-11-21
update highlighter, ignore self-mutations

nathanstitt:
highlighter update includes https://github.com/openstax/highlighter/pull/22


tutor/src/components/notes.js
```diff
@@ -128,6 +130,7 @@ class NotesWidget extends React.Component {
     }
 
     if (this.highlighter) {
+      this.highlighter.eraseAll();
```
Dantemss: This... sounds scary?

nathanstitt: eh, not really.  `eraseAll` just means to remove the highlights from the page, it doesn't actually send anything to server 😁

PR #2835, opened 2019-11-22
update MultiSelect and SectionsFilter

jivey:
- Add support for 2 columns if more than 20 items
- Add helpers for selecting all or none
- Fix closeAfterSelect
- Add scrolling for longer lists
- Update checkbox colors and size
- Remove the bold style coming from ChapterSection

### DIFF
--- a/tutor/src/components/multi-select.jsx
+++ b/tutor/src/components/multi-select.jsx
@@ -5,6 +5,24 @@ import classnames from 'classnames';
 import { action, observable } from 'mobx';
 import { observer } from 'mobx-react';
 import { Icon } from 'shared';
+import styled, { css } from 'styled-components';
+
+const MultiSelectWrapper = styled.div`
+  padding: ${props => props.useColumns ? '10px' : '0' };
+  margin-top: ${props => !props.useColumns && props.showHelperControls ? '10px' : '0' };
+  overflow-y: scroll;
+  max-height: 640px;
+  max-width: 976px;
+`
+
+const MultiSelectItems = styled.div`
+  column-count: ${props => props.useColumns ? '2' : 'inherit' };
+  column-gap: 35px;
+
+  @media only screen and (max-width: 979px) {
+    column-count: inherit;
+  }
+`
 
 @observer
 class MultiSelect extends React.Component {
@@ -28,10 +46,14 @@ class MultiSelect extends React.Component {
     onOnlySelection: PropTypes.func,
     onSelect: PropTypes.func,
     tabIndex: PropTypes.number,
+    showHelperControls: PropTypes.bool,
+    useColumns: PropTypes.bool,
   };
 
   static defaultProps = {
     closeAfterSelect: true,
+    showHelperControls: false,
+    useColumns: false,
     tabIndex: 0,
   };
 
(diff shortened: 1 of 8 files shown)


tutor/src/components/chapter-section.js
```diff
@@ -3,11 +3,11 @@ import PropTypes from 'prop-types';
...
 
 export default
-function ChapterSection({ chapterSection }) {
+function ChapterSection({ chapterSection, hideClassName }) {
   if (!chapterSection || chapterSection.isEmpty) { return null; }
   return (
     <span
-      className="chapter-section"
+      className={hideClassName == true  ? '' : 'chapter-section'}
       data-chapter-section={chapterSection.key}
```
nathanstitt: This seems a bit strange to me.  Do we need to hide the class so it's not styled?  If so, lets add a string `className`  prop something like this, and I'll propose an alternative below

```js
import cn from 'classnames';

export default
function ChapterSection({ chapterSection, className }) {
  ...
  return (
    <span className={cn('chapter-section', className)}> 
    ...
  )
}

ChapterSection.propTypes = {
  className: PropTypes.string,
}

File: tutor/src/components/notes/sections-filter.jsx [hunk 2]
```diff
@@ -26,13 +26,25 @@ class SectionsFilter extends React.Component {
...
   @computed get choices() {
     return this.props.notes.summary.sorted().map((s) => (
       {
         id: s.id,
         title: (
           <span>
-            <ChapterSection chapterSection={s.chapter_section} />
+            <ChapterSection chapterSection={s.chapter_section} hideClassName={true} />
+            &nbsp;
             <span>{s.title}</span>
```
nathanstitt: Rather than inserting a raw space it's better to add a margin.  You can wrap any component that accepts a className prop with styled components like this:
```js
const FilteredChapterSection = styled(ChapterSection)`
  margin-right: 1rem;
  font-weight: bold;
`
```
And then you could render: `< FilteredChapterSection chapterSection={s.chapter_section} />` here

tutor/src/components/multi-select.jsx
```diff
@@ -72,28 +106,51 @@ class MultiSelect extends React.Component {
...
+            showHelperControls={this.props.showHelperControls}
+          >
+            {this.renderHelperControls()}
+            <MultiSelectItems useColumns={this.props.useColumns}>
+              {Array.from(this.props.selections).map((selection) => this.renderMenuSelection(selection))}
+            </MultiSelectItems>
+          </MultiSelectWrapper>
         </Dropdown.Menu>
       </Dropdown>
     );
```
nathanstitt: This is fine, but something that would be a bit cleaner would be to add a `helperControls` prop instead of the `showHelperControls` bool.  Doing so is more flexible since you could then pass alternative controls if needed.

You'd then extract the `onSelectAll` and `onSelectNone` methods and `renderHelperControls` into a MultiSelectHelperControls component  and pass it to multiSelect like:

```js
<Multiselect 
  helperControls={< MultiSelectHelperControls onSelectAll={...} />}
/>
```

I don't think it's worth changing to this right now, but wanted to be aware of the technique.

PR #2838, opened 2019-11-27
Fix highlighter dropdown rendering issues in chrome and firefox

jivey:
The will-change property being added by the dropdown library causes stange rendering behavior. Adding `transform: translateZ(0)` forces hardware acceleration on and somehow fixes the bug.

#### In Chrome, the last element in the left column gets cut off and partially rendered in the second column.

Chrome before:
chrome-01-before

Chrome after:
chrome-02-after

#### In Firefox, typography line height gets affected, causing descenders like `j` and `g` to be cut off.

Firefox before:
firefox-01-before

Firefox after:
firefox-02-after

### DIFF
--- a/tutor/resources/styles/components/multi-select.scss
+++ b/tutor/resources/styles/components/multi-select.scss
@@ -17,6 +17,7 @@
 
     &.dropdown-item {
       padding: 0.75rem;
+      transform: translateZ(0);
     }
 
     .title {


jivey:
There's still a bug in Chrome with rendering these items that this doesn't fix, so I'm going to leave this open for now. Even though Chrome no longer shows the element in the second column, it does affect the list in the second column as if it were still rendering there.

PR #2841, opened 2019-12-03
Highlighter Modal

jivey:
- Changes the slide down page into a modal
- Adds a loading animation while note fetches are pending (thanks @nathanstitt!)
- Adds a TagList component to show the selected sections with an easy remove button
- Adds "blank slate" instructions for when you have no highlights
- Prevents section dropdown from appearing if there are no highlights
- Adds a scroll to top button for the modal

Before:

image

After:

image

Blank slate:

image

Large amount of sections selected:

image


tutor/specs/components/__snapshots__/sections-filter.spec.jsx.snap
```diff
@@ -12,16 +52,135 @@ exports[`SectionsFilter component renders 1`] = `
     <button
       aria-expanded={false}
       aria-haspopup={true}
-      aria-label="Display sections"
-      className="dropdown-toggle btn btn-default"
+      aria-label="Sections"
+      className="dropdown-toggle btn btn-link"
       disabled={false}
```
nathanstitt: Hm, I see why you updated the `aria-label` but I don't see why we had those.  Screenreaders will read the button text already.  the `type="button"` prop also not needed here.  Perhaps this used to be a bootstrap Button component or something

tutor/src/components/notes.js
```diff
@@ -419,12 +423,33 @@ class NotesWidget extends React.Component {
...
+                onDelete={this.onNoteDelete}
+                page={this.props.page}
+              />
+              <a href="#"
+                className="modal-scroll-top"
+                onClick={this.onModalScollTop}
+              >
+                <Icon type="angle-up" color="#fff" />
+              </a>
+            </Modal.Body>
```
nathanstitt: Icon has a click handler, I think this could be:

`<Icon type="angle-up" color="#fff" onClick={this.onModalScollTop} className="modal-scroll-top" />`

tutor/resources/images/highlighter/instructions-panel-icon.svg
```diff
@@ -0,0 +1,31 @@
...
+                                <path d="M3.78838821,7.30523305 C3.86540906,7.37237086 3.86540906,7.48224051 3.78838821,7.54937832 L3.50714239,7.79463685 C3.43012154,7.86177466 3.27737682,7.91666667 3.16763209,7.91666667 L0.19953588,7.91666667 C0.0897911458,7.91666667 0,7.8395952 0,7.74539675 L0,7.10930026 C0,7.0151018 0.06295357,6.88313833 0.13997442,6.81600052 L1.54360956,5.59202002 C1.62063041,5.52488221 1.74653755,5.52488221 1.8235584,5.59202002 L3.78838821,7.30523305 Z" id="Path" fill="#FFFFFF" fill-rule="nonzero"></path>
+                                <path d="M9.44444444,8.91666667 L0.555555556,8.91666667 C0.248777778,8.91666667 0,8.69276667 0,8.41666667 C0,8.14056667 0.248777778,7.91666667 0.555555556,7.91666667 L9.44444444,7.91666667 C9.75122222,7.91666667 10,8.14056667 10,8.41666667 C10,8.69276667 9.75122222,8.91666667 9.44444444,8.91666667 Z" id="Path" fill="#0DC0DC" fill-rule="nonzero"></path>
+                            </g>
+                        </g>
+                    </g>
+                </g>
+            </g>
+        </g>
+    </g>
+</svg>
```
nathanstitt: This is fine, but just so you're aware, there's 2 ways we do icons and we haven't really settled on the best approach yet.  

The other method is to convert the SVG to a react component using https://www.smooth-code.com/open-source/svgr/playground/ or one of the other convertors out there.

That has the advantage of playing nicely with styled-components and lets you alter the SVG properties using react props if needed.

tutor/resources/styles/components/notes/summary-page.scss
```diff
@@ -1,7 +1,59 @@
...
+  }
+
+  .modal-scroll-top {
+    position: fixed;
+    bottom: 44px;
+    right: 44px;
+    padding: 4px;
+    text-align: center;
+    font-size: 16px;
+
```
nathanstitt: We use rems for all font sizes, and 1 rem is set to 10px.  This should be: `font-size: 1.6rem;`

Also it'd be nice to set the `44px` to a variable at the top of the file, perhaps `$notes-modal-page-border-size` or something like that and use that everywhere so that it could be easily tweaked.

tutor/specs/components/tag-list.spec.jsx
```diff
@@ -0,0 +1,29 @@
...
+    expect.snapshot(<TagList {...props} />).toMatchSnapshot();
+  });
+
+  it('fires callback when removed', () => {
+    const list = mount(<TagList {...props} />);
+    list.find('.remove-tag').at(3).simulate('click');
+    expect(props.onRemove).toHaveBeenCalledWith(ITEMS[1]);
+    list.unmount();
+  });
+});
```
nathanstitt: nice spec, it's always good to have at least a minimal one to make sure it renders

jivey:
Scroll overflow in the selector isn't kicking in now that it's in the modal:

image

tutor/src/components/notes.js
```diff
@@ -366,6 +371,15 @@ class NotesWidget extends React.Component {
...
 
+  @action.bound setupModalScrollTop() {
+    let modalBody = window.document.body.querySelector('.modal-body');
+    this.modalScrollHelper = new ScrollTo({ windowImpl: modalBody, root: modalBody });
+  }
+
+  @action.bound onModalScollTop() {
+    this.modalScrollHelper.scrollToSelector('.filter-area', { updateHistory: false });
+  }
+
```
nathanstitt: I'm sorry, I hadn't realized that we were scrolling an inner div and not the entire window when you asked about scrolling earlier.  In this case you can remove the `setupModalScrollTop ` and use the  `scroll-into-view` library that we import from https://github.com/KoryNunn/scroll-into-view

an example of it is at: https://github.com/openstax/tutor-js/blob/master/tutor/src/components/book-menu/ux.js#L72

I think it'd look like:
```js
  @action.bound onModalScollTop() {
      scrollIntoView(document.querySelector('.modal-body'), { time: 300 });
   }
```

If the library scrolls things that shouldn't move then you'll have to use the validTarget option like the example in book-menu does.

tutor/src/components/notes.js
```diff
@@ -34,6 +34,11 @@ class NotesWidgetWrapper extends React.Component {
     page: PropTypes.instanceOf(Page).isRequired,
   };
 
+  constructor(props) {
+    super(props);
+    props.course.notes.ensurePageExists(props.page);
+  }
+
```
nathanstitt: I don't think we can call `props.course.notes.ensurePageExists(props.page);` here because it won't run if the props.page updates.  we need to call it before the highlighter is setup in the `initializePage` method, probably right after the `isMounted` check.

tutor/resources/styles/components/notes/summary-page.scss
jivey: Whoops, missed this ref while trying to make it match the spec. Good catch.

tutor/resources/images/highlighter/instructions-panel-icon.svg
jivey: That's interesting, I wonder about the tradeoffs with asset caching.

tutor/src/components/notes.js
jivey: Good to know for the future, I see it can be passed class names too. It does come with some unwanted style baggage because of the other classes that are automatically added; might be a nice future refactor to allow passing an element.

tutor/src/components/notes.js
jivey: I think you moved this there because https://github.com/openstax/tutor-js/pull/2841/files#diff-16bdd0ebc9ce6c81cf33368a50fc4fbaL45 requires it, and it uses the api pending check to `initializePage` after mount. 

I had trouble seeing it in testing, but the potential issue you pointed out makes sense to me... maybe we change `NotesWidget` to handle it internally? What are your thoughts about how to best change that?

tutor/src/components/notes.js
jivey: This is great and exactly what I needed, thanks!!

tutor/src/components/notes.js
jivey: Thanks for mentioning this, I circled back and realized `buttonProps` was passing to the bootstrap button, so I was able to override exactly what I wanted 👍

Brooke-Fail:
Verified in QA.

PR #2842, opened 2019-12-03
prevent section select from partially going off screen

jivey:
This prevents popper from trying to keep the dropdown onscreen, which could cause it to partially go offscreen.

Before:

image

After:

image


jivey:
I think technically we could have changed the offset to consider the fixed navbar, but I think this interaction style is better anyway, and should help out with the future modal.

PR #2843, opened 2019-12-04
Update navbar per highlighting mockups

nathanstitt:
@jivey to finish, this makes the footer optional and moves it's controls to top navbar for readings


tutor/src/theme.js
```diff
@@ -57,6 +57,11 @@ const TutorTheme = {
       notice:   '#eba927',
     },
 
+    navbars: {
+      border: '#d5d5d5',
+      divider: '#cdcdcd',
+      control: '#5e6062',
+    },
   },
```
nathanstitt: Great idea to add these!

PR #2844, opened 2019-12-05
Render HTML markup in exercise selector and performance forecast

nathanstitt:
use BookPartTitle to display chapter/section titles.  This is needed so they'll display any HTML markup


nathanstitt:
yeah, that's difficult to do in client side since you'd have to either use regex, an html parser, or actually render it off-dom somehow.  And like you said, the source is trusted

PR #2850, opened 2019-12-10
move reading milestones into a modal

jivey:
Before:

image

After:

image

- Observes Milestones so we can pass index changes to Milestone
- Hoists PagingNavigation up to ReadingTask since the Overlay isn't used anymore
- Consolidates the new modal style into a mixin so it can be phased in
  and eventually removed when it becomes the base style.


jivey:
One tiny new commit, renamed the modal title "Overview" to "Assignment Overview" after feedback.

dkhowala:
@jivey 
Looks good to go!

PR #2851, opened 2019-12-10
reset summary and selection when deleting last note in a page

nathanstitt:
When the last note in a section is removed, we need to remove it from the summary and selected sections lists

### DIFF
--- a/tutor/src/components/notes/summary-page.jsx
+++ b/tutor/src/components/notes/summary-page.jsx
@@ -80,6 +80,9 @@ class NoteSummaryPage extends React.Component {
 
   @action.bound onDelete(note) {
     this.props.onDelete(note);
+    this.selectedPages.remove(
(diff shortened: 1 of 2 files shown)


tutor/src/components/notes/summary-page.jsx
```diff
@@ -80,6 +80,9 @@ class NoteSummaryPage extends React.Component {
 
   @action.bound onDelete(note) {
     this.props.onDelete(note);
+    this.selectedPages.remove(
+      this.selectedPages.find(s => s.uuid == note.page.uuid)
+    );
```
jivey: As a side effect this would deselect a section that still contains notes, I think?

nathanstitt: no, the selection is held in the component in the `selectedPages` array.  I had to remove it from there as well otherwise it'd never show the "please select some sections" message

PR #2860, opened 2019-12-16
Template card styles

jivey:


tutor/src/screens/grading-templates/card.js
```diff
@@ -1,19 +1,90 @@
...
+  line-height: 2rem;
 
+  .card {
+    border: 1px solid #d5d5d5;
+  }
+  .card-header {
+    border: 0;
+    border-left: 8px solid ${props => props.theme.border};
+    background: ${props => props.theme.background};
+    font-size: 1.6rem;
```
nathanstitt: as I noted below, styled-components always have a `theme` property which we probably shouldn't override like this.  In fact when I first saw this I'd typed up:

>This is my fault, I think I used props in a function elsewhere.  Doing it this way is only really valuable if we had multiple themes in use.  Since tutor's only ever going to have 1 theme we can just use the imported values directly like you did on line 10 above.

> using the values directly allows testing the component directly, otherwise you have to add a theme context to the rendered version

tutor/src/theme.js
```diff
@@ -63,6 +63,16 @@ export const colors = {
...
 
+  templates: {
+    reading: {
+      background: '#fcf6d8',
+      border: '#f4d018',
+    },
+    homework: {
+      background: '#daf3f8',
+      border: '#34bdd7',
+    },
```
nathanstitt: Hm, lets verify that these RGB values are correct, it's weird that readings & homework are getting new colors.

tutor/src/screens/grading-templates/card.js
```diff
@@ -1,19 +1,90 @@
...
 const TemplateCard = observer(({ template, onEdit }) => {
   return (
-    <StyledCard>
-      <Card.Body>
-        {template.name}
-        <Button onClick={() => onEdit(template)}>edit</Button>
-      </Card.Body>
-    </StyledCard>
+    <CardWrapper theme={Theme.colors.templates[template.type]}>
+      <Card>
```
nathanstitt: Hm, `theme` is a pre-existing prop on styled-components that's typically passed down using React context.  Overriding it like this may cause issues, so how about we rename this to `colors` or some other property.

tutor/src/theme.js
jivey: I had wondered about this, there are some differences in what we'd use for border colors (the background would have to be new definitions because of the intended use.)

`reading` old and new:
image

image

`homework` old and new:

image

image

tutor/src/screens/grading-templates/card.js
jivey: Good to know. Renamed it to `templateColors` which better explains the use as well.

PR #2863, opened 2019-12-18
WRM radio component and modal styles

jivey:
Updates the modal and NumberInput styles, and adds RadioInput and Select.

There are some alignments and other things I'd like to finish up later, but this should get us fairly close to the final look so we can test API things.

image

image

image


jivey:
Hmm Travis is complaining about something, I'll take a look tomorrow.

jivey:
@nathanstitt This is probably good for a feedback cycle. Most fields aren't hooked up yet, I'm looking at that in a separate branch so this doesn't get held up.

PR #2864, opened 2019-12-18
fix and run lint cheks on all files

nathanstitt:
Removes the rule that only checked changed files


bin/lint
```diff
@@ -2,31 +2,21 @@
...
+  exit $?
 fi
 
-if [ $# -eq 0 ]; then
-    $(npm bin)/eslint {tutor,shared,exercises}
-    exit $?
+if [[ ! -z "${OX_PROJECT}" ]]; then
+   $(npm bin)/eslint --fix $OX_PROJECT/{src,specs}
+   exit $?
 fi
```
jivey: Wouldn't `--fix` for Travis cause it to pass when we'd want it to fail so it can be committed?

nathanstitt: yep, good catch.  I think I'd copied the command I was running to generate this PR from.  I'll remove that

PR #2866, opened 2019-12-27
bump react-bootstrap to 1.0.0-beta.16

jivey:
Upgrading to the latest react-bootstrap will resolve an accessibility issue with modals that was fixed in react-overlays 2.x (the fix was not applied to the earlier versions):

https://github.com/react-bootstrap/react-bootstrap/issues/4790
https://github.com/react-bootstrap/react-overlays/issues/292


Brooke-Fail:
Verified in Production.

PR #2868, opened 2019-12-31
Add feature tests using cypress

nathanstitt:
Right now only a simple test


exercises/specs/components/tags/__snapshots__/aplo.spec.js.snap
```diff
@@ -28,7 +28,7 @@ exports[`ApLo tags component renders and matches snapshot 1`] = `
       >
         <svg
           aria-hidden="true"
-          className="svg-inline--fa fa-plus-circle fa-w-16 ox-icon ox-icon-plus-circle c0"
+          className="svg-inline--fa fa-plus-circle fa-w-16 c0 ox-icon ox-icon-plus-circle"
           data-icon="plus-circle"
```
nathanstitt: This is what caused all of the snapshot updates.   I'm really confused on what caused the class to switch it's position in the list here.  I think it should be harmless though.

PR #2869, opened 2020-01-02
Don't allow table columns to be less than 300px

nathanstitt:
fixes cases where the columns would get very narrow.  This was caused by the right column having `<br/>` tags inserted in the content which caused the computed line length to be very short

before:
image

After looks like:
image


nathanstitt:
I'm closing this since it's not going to work for us.  There is content with as much as 8 columns, so if we make them all at least 300px the table will grow very wide

PR #2871, opened 2020-01-07
Fix calendar popup

jivey:
This feels slightly hacky because it I think it could be solved by sending styles to popper config instead, but I wasn't able to get the override working.

### DIFF
--- a/tutor/src/screens/teacher-dashboard/add-assignment-popup.js
+++ b/tutor/src/screens/teacher-dashboard/add-assignment-popup.js
@@ -22,7 +22,7 @@ const CustomMenu = React.forwardRef(
 
 CustomMenu.propTypes = {
   'aria-labelledby': PropTypes.string,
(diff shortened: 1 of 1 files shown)


tutor/src/screens/teacher-dashboard/add-assignment-popup.js
```diff
@@ -6,6 +6,26 @@ import Time from '../../models/time';
...
+        {children}
+      </div>
+    );
+  },
+);
+
+CustomMenu.propTypes = {
+  'aria-labelledby': PropTypes.string,
+  children: PropTypes.arrayOf(PropTypes.element),
+  className: PropTypes.string,
```
nathanstitt: You can also use `PropTypes.node` here, that has the advantage of matching anything that can be rendered, while the above will require an array

jivey: Great tip!

PR #2873, opened 2020-01-09
Light style for course screens

jivey:
Adds a lighter style course page that can be opted into, while preserving the existing style for other screens.

Course screen:
image

Other existing screen:
image


tutor/src/screens/grading-templates/index.js
```diff
@@ -33,6 +34,14 @@ class GradingTemplatesScreen extends React.Component {
     this.store.fetch();
   }
 
+  componentDidMount() {
+    document.body.classList.add('light');
+  }
+
+  componentWillUnmount() {
+    document.body.classList.remove('light');
+  }
```
jivey: @nathanstitt Curious to hear what you think about this approach. At one point I looked into using Hooks instead, but had issues integrating it. Feels like there could be a better way though!

PR #2883, opened 2020-01-17
don't display media preview for footnote that points to itself

nathanstitt:
Goes along with https://github.com/openstax/tutor-server/pull/1954

### DIFF
--- a/tutor/src/components/book-page.js
+++ b/tutor/src/components/book-page.js
@@ -160,10 +160,11 @@ class BookPage extends React.Component {
   }
 
   isMediaLink(link) {
-    // TODO it's likely that this is no longer needed since the links being
-    // passed into this are now much stricter and exclude where `href="#"` and
-    // where `href` contains any `/`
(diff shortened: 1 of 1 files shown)


tutor/src/components/book-page.js
```diff
@@ -160,10 +160,11 @@ class BookPage extends React.Component {
   }
 
   isMediaLink(link) {
-    // TODO it's likely that this is no longer needed since the links being
-    // passed into this are now much stricter and exclude where `href="#"` and
-    // where `href` contains any `/`
-    return ((link.hash.length > 0) || (link.href !== link.getAttribute('href'))) && (link.hash.search('/') === -1);
+    // footnotes shouldn't show a preview of themselves
+    if (link.matches('[data-type="footnote-ref-link"]')) {
```
Dantemss: Hmmm maybe we could avoid hardcoding the data-type here somehow?

nathanstitt: hm, well you kinda need to refer to them somehow, and data attributes are a pretty standard thing to use, so it's fine

PR #2885, opened 2020-01-22
add tours to feature flags

nathanstitt:
And move the `_MODELS` to window earlier and stub them from cypress


tutor/specs/models/course.spec.js
```diff
@@ -44,7 +46,7 @@ describe('Course Model', () => {
     expect(c.currentRole.id).toEqual(newRoleAttrs.id);
   });
 
-  it('calculates audience tags', () => {
+  fit('calculates audience tags', () => {
```
jivey: Should this be `it`? Strangely the test seems to pass even though I can't find a definition.

nathanstitt: lol, yes it should be `it`. `fit` is for "focus" meaning it'll only run that test in the file, there's also `xit` which prevents the test from running.

Those are nice shorthands when you're working on a single test and are in "watch" mode and want to re-run it whenever files are modified

jivey: Ah okay, those are nice to have. I see the aliases now in the docs, just couldn't find them yesterday and got confused.

nathanstitt: yep, I'll re-enable the other tests by switching it back to plain `it` on my next PR.  glad you noticed this!

PR #2889, opened 2020-01-28
Validate each step and conditionally enable continue button

nathanstitt:
Adds a validation class with custom logic for each step

### DIFF
--- a/tutor/src/screens/assignments/validations.js
+++ b/tutor/src/screens/assignments/validations.js
@@ -1,0 +1,36 @@
+import { computed } from 'vendor';
+import { isEmpty } from 'lodash';
+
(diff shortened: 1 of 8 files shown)


tutor/src/screens/assignments/validations.js
```diff
@@ -0,0 +1,36 @@
...
+  @computed get chapters() {
+    return !isEmpty(this.plan.pageIds);
+  }
+
+  @computed get questions() {
+    return !isEmpty(this.plan.exerciseIds);
+  }
+}
+
+export default Validations;
```
jivey: This feels really nice being in a dedicated class!

PR #2890, opened 2020-01-29
Retry loading tasks that start with a placeholder step (practice tasks)

Dantemss:
This moves the wait for practice tasks from the BE to the FE so the unicorn workers don't stay busy while waiting.

Goes with https://github.com/openstax/tutor-server/pull/1970


nathanstitt:
the code looks good @Dantemss - I'll test a bit more and can fix the specs

PR #2891, opened 2020-01-30
WRM Preview popup, Questions toolbar, General layout changes

jivey:
- Small body layout refactor to support the different paddings
- Questions toolbar styles (still work to do on that screen though, and Filter bar isn't functional yet)
- Preview popup (left a code comment, wondering if there's a better way)

image

image


tutor/src/screens/assignments/details.js
```diff
@@ -44,10 +44,16 @@ const Details = ({ ux }) => {
             <HintText>(Apply a pre-set submission and grading policy template)</HintText>
           </Label>
           <div>
-            <Select name="grading_template">
+            <Select
+              name="grading_template"
+              onChange={e => {
+                ux.selectedGradingTemplateId = e.target.value;
+                ux.form.setFieldValue('grading_template', e.target.value);
+              }}
```
jivey: @nathanstitt Wondering if there's a better way than overriding to get the selected value. I originally tried checking `form.values.grading_template` directly in a computed, but it doesn't get updated until the step changes.

tutor/src/screens/assignments/preview-tooltip.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  margin-left: 0.8rem;
+`;
+
+const StyledContent = styled(Popover.Content)`
+  .card {
+    border: 0;
+  }
+`;
```
jivey: This might look a little strange, why not just override TemplateBody, but I ran into styled-components specificity issues.

tutor/src/screens/assignments/details.js
nathanstitt: hm, why do we need to set the `ux.selectedGradingTemplateId` ?  If we change the name here to `grading_template_id`, this should set `ux.fom.values.grading_template_id` and UX can then use as needed.

I think we'd have to set add a `@field grading_template_id` on the plan, but we'll need to do that anyway so it saves

tutor/src/screens/assignments/details.js
```diff
@@ -44,10 +44,10 @@ const Details = ({ ux }) => {
             <HintText>(Apply a pre-set submission and grading policy template)</HintText>
           </Label>
           <div>
-            <Select name="grading_template">
+            <Select name="grading_template_id">
```
jivey: @nathanstitt Glad you pointed this out — looking at it again today, I had made an assumption about `grading_template_id` being wired up. Hooking it up gave me what I was looking for.

tutor/src/screens/assignments/ux.js
```diff
@@ -105,7 +105,8 @@ export default class AssignmentUX {
   }
 
   get formValues() {
-    return this.plan; // .serialize();
+    const templateId = get(this.gradingTemplates, '[0].id');
+    return merge(this.plan, { grading_template_id: templateId });
```
jivey: @nathanstitt Formik doesn't seem to set `grading_template_id` to what the initially selected option is when it loads, otherwise I wouldn't need this to set this explicitly here for new forms. Ideas/thoughts?

nathanstitt: ah, I see.  in that case let's set it in the ux.initialize method after we load the templates.

jivey: Because ux is initialized higher up, form isn't set until the `renderStep`. Should we do something like this: 

```
@action.bound renderStep(form) {
    this.form = form;

    if (this.plan.isNew && this.isShowingDetails) {
      this.form.values.grading_template_id =
        this.form.values.grading_template_id || get(this.gradingTemplates, '[0].id');
    }

    return <Step ux={this} />;
  }
```

jivey: Sorry realized what you meant now, we could just set it on the plan itself during initialize, which is nice.

PR #2893, opened 2020-02-03
WRM Steps and actions

nathanstitt:
Moves the steps and actions into separate classes from the main UX to segregate the logic

### DIFF
--- a/tutor/src/screens/assignments/actions.js
+++ b/tutor/src/screens/assignments/actions.js
@@ -1,0 +1,31 @@
+import { action, observe } from 'mobx';
+import { invoke } from 'lodash';
+
+class Actions {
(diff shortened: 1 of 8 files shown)


tutor/src/screens/assignments/actions.js
```diff
@@ -0,0 +1,31 @@
+import { action, observe } from 'mobx';
+import { invoke } from 'lodash';
+
+class Actions {
+  constructor(ux) {
+    this.ux = ux;
+    observe(ux.steps, 'currentStepId', ({ newValue: stepId, oldValue: prevStepId }) => {
+      invoke(this, stepId, { prevStepId });
```
jivey: Neat!

PR #2895, opened 2020-02-04
Save progress between steps and update URL

nathanstitt:
This way it can be reloaded and hopefully have the same state


nathanstitt:
@jivey still some more to do on the review step, the input box for points doesn't actually save yet

PR #2896, opened 2020-02-05
have loading practice task use exponential backoff to reduce load on …

reedstrm:
…tutor backend
Current 30 tries is about 30 * (1 sec sleep + 1.5 sec for fetch), while this is 2^6-1 + 6*1.5 so total wait time of approx. 70 sec is preserved

### DIFF
--- a/tutor/src/models/student-tasks/task.js
+++ b/tutor/src/models/student-tasks/task.js
@@ -64,7 +64,7 @@ class StudentTask extends BaseModel {
           break;
         }
         // exponential backoff between tries: 1, 2, 4 .. 32 (plus some milliseconds of random)
-        await new Promise(resolve => setTimeout(resolve, 1000*(2**(6 - tries_remaining) + Math.random()));
+        await new Promise(resolve => setTimeout(resolve, 1000*(2**(6 - tries_remaining) + Math.random())));
         tries_remaining--;
       }
       if (tries_remaining == 0) {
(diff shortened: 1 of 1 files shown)


tutor/src/models/student-tasks/task.js
```diff
@@ -57,14 +57,14 @@ class StudentTask extends BaseModel {
...
       while(tries_remaining > 0) {
         await this.fetch();
         if (this.isLoaded) {
           break;
         }
-        // wait 1 second in between load attempts
-        await new Promise(resolve => setTimeout(resolve, 1000));
+        // exponential backoff between tries: 1, 2, 4 .. 32 (plus some milliseconds of random)
+        await new Promise(resolve => setTimeout(resolve, 1000*(2**(6 - tries_remaining) + Math.random()));
         tries_remaining--;
```
nathanstitt: Nice!  But.. `Math.random()` returns a float from 0 - 1 so that will either round to 0 or 1.  Probably need to `(Math.random() * 100)`

reedstrm: that's what I wanted, I do the calc in seconds then multiply the whole thing by 1000.  (but failed to get all the nested parens back in place. grrr)

reedstrm: (guess I need to learn how to run the nodejs spec test suite locally … *blush*

reedstrm: Ahh rounding - I worried about that, but tried it at a node prompt, and got a float result.

reedstrm: > tries_remaining=6
6
> (2**(6 - tries_remaining) + Math.random())*1000
1748.102522071031
> (2**(6 - tries_remaining) + Math.random())*1000
1343.0246008299464
>

PR #2898, opened 2020-02-08
Put exercise questions back into columns

jivey:
I pulled over styles from the sass files into styled components — after the switch, we could eventually apply them directly to the exercise card components, but for now it uses the descendant class names. A little fragile but keeps from cloning the card component classes.

Also copied Bootstrap v4's default breakpoints into Theme to maintain an existing margin rule & set it up for future use.

image

image


nathanstitt:
screenshots look great!

jivey:
@nathanstitt Fixed the diff by catching up with wrm branch just in case you wanted to see what changed.

PR #2900, opened 2020-02-12
Hook up question type filters

jivey:
Adds filtering for All / Multiple Choice / Written Response (Open Ended) question types to the assignment builder.

Due to multipart questions, the only reliable method I could find was to check the type of the first question in the list. As far as I can tell, it's not likely a multipart will ever contain mixed types, but if there's a better way let me know.

Example of narrowing it down to WR only:
image

With All or Multiple Choice selected, you'll always see MC at the top:
image

### DIFF
--- a/shared/specs/factories/exercise.js
+++ b/shared/specs/factories/exercise.js
@@ -114,6 +114,28 @@ Factory.define('ExerciseQuestion')
   ])
   .community_solutions(() => []);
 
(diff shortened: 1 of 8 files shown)


jivey:
The spec failures here seem to be from a random failure with loading exercises in the test backend that I haven't looked into yet.

PR #2903, opened 2020-02-15
Add custom checkbox and tweaks to details page

jivey:
- Adds an orange checkbox (CheckboxInput) that can be used anywhere and should be accessible. 
- Fixes some alignment things with rendering period + date time fields

image

### DIFF
--- a/tutor/src/screens/assignments/details.js
+++ b/tutor/src/screens/assignments/details.js
@@ -1,6 +1,6 @@
 import { React, PropTypes, styled, observer } from 'vendor';
 import { isEmpty } from 'lodash';
-import { AssignmentBuilder, SplitRow, Label, HintText, TextInput, Setting, Body } from './builder';
+import { AssignmentBuilder, SplitRow, Label, HintText, TextInput, Body } from './builder';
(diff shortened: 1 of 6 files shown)


tutor/src/screens/assignments/details.js
```diff
@@ -56,9 +68,9 @@ const Details = observer(({ ux }) => {
             <HintText>(Apply a pre-set submission and grading policy template)</HintText>
           </RowLabel>
           <div>
-            <Select name="grading_template_id">
+            <StyledSelect name="grading_template_id">
               {ux.gradingTemplates.map(t => <option key={t.id} value={t.id}>{t.name}</option>)}
-            </Select>
+            </StyledSelect>
             <PreviewTooltip template={ux.gradingTemplate} />
```
nathanstitt: I think we should ask UX what to do when there's only a single grading template for that type of assignment.  In that case it seems like we should just show it's name as text rather than a select that doesn't do anything

jivey: That's a great thing to point out, I'll bring that up.

PR #2907, opened 2020-02-19
move timezone modal into a reusable component, add to assignment builder

jivey:
Also fix zindex and overlap issues with the tooltips.

image

### DIFF
--- a/tutor/src/components/radio-input.jsx
+++ b/tutor/src/components/radio-input.jsx
@@ -55,7 +55,7 @@ const StyledRadioInput = styled.input.attrs( () => ({
 `;
 
 const RadioInput = observer((props) => {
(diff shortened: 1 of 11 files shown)


tutor/src/components/radio-input.jsx
```diff
@@ -55,7 +55,7 @@ const StyledRadioInput = styled.input.attrs( () => ({
 `;
 
 const RadioInput = observer((props) => {
-  const [field] = useField({ type: 'text', ...props });
+  const [field] = props.standalone ? [] : useField({ type: 'text', ...props });
```
jivey: Let's it be used outside of a formik context, but happy to hear suggestions on a different approach.

nathanstitt: yeah I think that's fine.

PR #2909, opened 2020-02-21
Add external and event types, update validations

jivey:
Still some work to do with the time calculation, but this is a good snapshot to revisit.
 
image

image


tutor/src/screens/assignments/ux.js
```diff
@@ -291,7 +313,14 @@ export default class AssignmentUX {
...
   }
 
+  @action.bound async saveAsDraft() {
+    this.plan.update(omit(this.form.values, 'tasking_plans'));
+    this.plan.is_draft = true;
+    await this.saveAndCopyPlan();
+  }
+
   @action.bound async onPublishClick() {
+    this.plan.update(omit(this.form.values, 'tasking_plans'));
```
jivey: I'm omitting these since the dates are set directly to the plan, and I ran into serialization issues when editing a date... for some reason the values were moment objects and not ISOstrings.

nathanstitt: that's fine, I think they're moment objects because that's what the date time component sets, but not sure

PR #2910, opened 2020-02-21
Hook up add new template within builder

jivey:
Adds the template modal to the assignment builder, allowing newly created templates to be selected at assignment during create and edit.

image

image

image

### DIFF
--- a/tutor/src/screens/assignments/details.js
+++ b/tutor/src/screens/assignments/details.js
@@ -1,7 +1,6 @@
 import { React, PropTypes, styled, observer } from 'vendor';
 import { isEmpty } from 'lodash';
 import { AssignmentBuilder, SplitRow, Label, HintText, TextInput, TextArea, Body } from './builder';
-import Select from '../../components/select';
 import RadioInput from '../../components/radio-input';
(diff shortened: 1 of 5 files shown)


tutor/src/screens/assignments/details.js
```diff
@@ -32,10 +34,73 @@ const StyledTextInput = styled(TextInput)`
...
 
-const StyledSelect = styled(Select)`
-  min-width: 24rem;
+const StyledDropdown = styled(Dropdown)`
+  display: inline-flex;
 `;
 
+const StyledToggle = styled(Dropdown.Toggle)`
+  &&& {
+    border: 1px solid ${colors.forms.borders.light};
```
nathanstitt: I'm not familiar with the `&&&` syntax.  Is that just to make the rule specific enough?

you might also be able to nest under body or something else like:
```
  body & {
    border: 1px solid red;
  }
```

jivey: Yeah it's a specificity thing, components from Bootstrap seem to be especially finicky for overriding.

PR #2923, opened 2020-03-05
Add correct/total points to overview

nathanstitt:
image


tutor/specs/screens/teacher-review-metrics/breadcrumbs.spec.jsx
```diff
@@ -6,7 +6,7 @@ import Router from '../../../src/helpers/router';
 
 jest.mock('../../../src/helpers/router');
 
-describe('Task Teacher Review: Breadcrumbs', function() {
+xdescribe('Task Teacher Review: Breadcrumbs', function() {
```
jivey: Just double checking: skipping this temporarily while we build/rearrange things?

nathanstitt: yeah, once we merge this the old review screens will be going away so it wasn't worth fixing their specs

jivey: I think I see why now, I can re-enable while working on the new breadcrumbs

PR #2926, opened 2020-03-10
update assignment modal, fix closes_at & general datetime fields

jivey:
Update Assignment Modal

- Use radio buttons for period toggle
- Add date time fields to change and save inline for period
- Update button styles

Update assignment DateTime fields

- Use setFieldValue in event to make sure the form visually reflects changes we made while running rules (like if the due date was set to before open date)
- Fix closes_at rule

General

- Added a btn-form-action in sass for those new big buttons we've started using.
- Made a TemplateModal component to DRY up that particular modal style

Side note: with both the above, I need to add a pass later that makes things like grading templates modal use them instead of duplicating styles.

image

As you can see from the below image, the metrics screen may need to be touched up now.

image

### DIFF
--- a/tutor/specs/components/template-modal.spec.js
+++ b/tutor/specs/components/template-modal.spec.js
@@ -1,9 +1,10 @@
+import { React } from '../helpers';
 import TemplateModal from '../../src/components/template-modal';
 
 describe('Template Modal', () => {
 
   it('renders and matches snapshot', () => {
-    expect.snapshot(<TemplateModal templateType="neutral" />).toMatchSnapshot();
+    expect(<TemplateModal templateType="neutral" />).toMatchSnapshot();
   });
 
 });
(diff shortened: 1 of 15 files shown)


tutor/specs/components/template-modal.spec.js
```diff
@@ -0,0 +1,9 @@
+import TemplateModal from '../../src/components/template-modal';
+
+describe('Template Modal', () => {
+
+  it('renders and matches snapshot', () => {
+    expect.snapshot(<TemplateModal templateType="neutral" />).toMatchSnapshot();
+  });
+
+});
```
nathanstitt: this isn't actually rendering anything, see the `template-modal.spec.js.snap` file above.  It must be returning null

jivey: Whooooops, thanks!

PR #2927, opened 2020-03-10
Rename assignment to /edit & /review

nathanstitt:
This unifies them and doesn't use `grading` since that's not their purpose


tutor/src/screens/teacher-dashboard/plan-details.jsx
```diff
@@ -277,6 +276,7 @@ class CoursePlanDetails extends React.Component {
         enforceFocus={false}
         data-assignment-type={type}
         templateType={type}
+        className={className}
```
jivey: Good catch! I wonder why my lint run didn't catch that. Thanks for fixing!

PR #2946, opened 2020-03-31
dashboard modal tweaks

jivey:
- All assignment types use a singular "View assignment" button that goes to the Details tab
- Metrics (question performance bars) aren't needed

Note: does not yet include logic to decide if the Grade button should be shown based on WRM question presence

image

### DIFF
--- a/tutor/specs/integration/course.spec.js
+++ b/tutor/specs/integration/course.spec.js
@@ -1,0 +1,33 @@
+context('Course', () => {
+  beforeEach(() => {
+    cy.visit('/course/1');
+    cy.disableTours();
+  });
+
+  it('only displays Grade Answers button on Homework type', () => {
+    cy.get('.event.type-external label').first().click();
+    cy.getTestElement('gradeAnswers').should('not.exist');
+    cy.get('.modal-header .close').click();
+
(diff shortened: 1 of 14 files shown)


jivey:
- Send periodId to the grader
- Disable button based on due_at, and setup some tests to reliably check that by ensuring a prior and upcoming homework is always visible on the calendar.

The `<span>` around the `TutorLink` might be unnecessary but it's a workaround for the tooltip, since we're disabling pointer-events on the disabled button.

tutor/specs/integration/course.spec.js
```diff
@@ -0,0 +1,33 @@
...
+    // Check upcoming assignment
+    cy.get('.upcoming .event.type-homework label').first().click();
+    cy.getTestElement('gradeAnswers').should('exist').should('have.class', 'disabled');
+    cy.get('.modal-header .close').click();
+
+    // Check past due assignment
+    cy.get('.past .event.type-homework label').first().click();
+    cy.getTestElement('gradeAnswers').should('exist').should('not.have.class', 'disabled');
+  });
+});
```
nathanstitt, line 33: awesome spec!

PR #2949, opened 2020-04-06
setup most of the assignment review details screen

jivey:
image

### DIFF
--- a/tutor/src/theme.js
+++ b/tutor/src/theme.js
@@ -3,7 +3,7 @@ import { css } from 'styled-components';
 // Shared with scss in styles/variables
 
 const neutral = {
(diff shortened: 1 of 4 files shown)


tutor/src/theme.js
```diff
@@ -3,7 +3,7 @@ import { css } from 'styled-components';
 // Shared with scss in styles/variables
 
 const neutral = {
-  lightest: '#f9f9f9', // nearly white
+  lightest: '#fafafa', // nearly white
```
jivey, line 6: `fafafa` is used a lot in new designs and `lightest` is only 1 off and previously unused, so changing it permanently here.

PR #2954, opened 2020-04-14
update select grading template assignment type modal

camliu89:
Style changes to the Grading template: "Select Assignment Type" modal

Screenshot from 2020-04-14 16-03-19

Card: https://app.zenhub.com/workspaces/openstax-tutor-560aef617a763234615d6a0f/issues/openstax/tutor/881

### DIFF
--- a/tutor/src/screens/grading-templates/editors.js
+++ b/tutor/src/screens/grading-templates/editors.js
@@ -20,6 +20,21 @@ const StyledTemplateModal = styled(TemplateModal)`
   }
 `;
 
+const StyledSelectAssignmentModal = styled(StyledTemplateModal)`
+  .modal-dialog {
+    margin-top: 135px;
+  }
+  .modal-body {
+    background-color: #F5F5F5 !important;
+  }
+  .btn-default {
+    background-color: #FFFFFF !important;
+  }
+  .btn + .btn {
+    margin-left: 4rem;
+  }
+`;
+
 const Row = styled.div`
   margin: 2.4rem 0;
 `;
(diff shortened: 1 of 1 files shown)


tutor/src/screens/grading-templates/editors.js
```diff
@@ -20,6 +20,21 @@ const StyledTemplateModal = styled(TemplateModal)`
...
+  .modal-dialog {
+    margin-top: 135px;
+  }
+  .modal-body {
+    background-color: #F5F5F5 !important;
+  }
+  .btn-default {
+    background-color: #FFFFFF !important;
+  }
+  .btn + .btn {
```
nathanstitt, line 33: Looks, good!  Glad you're figuring out styled-components.  a few suggestions:  
  * is `StyledTemplateModal` used anywhere else?  If not you could just set it's styles above instead of creating a derived copy for `StyledSelectAssignmentModal`
  * I think you might be able to use `varant="light"` on the buttons instead of overriding it's color here.  Not sure though. https://getbootstrap.com/docs/4.0/components/buttons/
 * are the `!important` needed?   if so can you add a comment so that others know why it was added and can hopefully remove it once it's not needed

camliu89: Thanks for the suggestions Nathan! Please take a look at the new committed code. 

1. `StyleTemplateModal` is used in the next modals. We didn't want to override some of the styling for the other modals.
2. The `light` option in bootstrap seems like a different color from what Devika wanted.
3. I deleted the `!important` in the code. It was not needed.

PR #2955, opened 2020-04-14
grading template form fixes

jivey:
- fix displaying incorrect instructions on homework type
- change placeholder text by type
- extends ranges and adds stepping by 5
- make disabled number input value hidden

image

### DIFF
--- a/tutor/src/screens/grading-templates/editors.js
+++ b/tutor/src/screens/grading-templates/editors.js
@@ -264,11 +264,6 @@ class TemplateForm extends React.Component {
         {map(form.errors.common, (value, key) =>
           <Error key={key}>{value}</Error>)}
         <Line />
-        <Row>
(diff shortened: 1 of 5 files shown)


tutor/src/screens/grading-templates/editors.js
```diff
@@ -244,6 +246,8 @@ class TemplateForm extends React.Component {
 
   renderForm = (form) => {
     const { body } = this.props;
+    const namePlaceholder = this.props.template.task_plan_type == 'reading' ?
```
camliu89, line 249: It looks good! I think `template` can be added in the line above: `{ body, template }`.

jivey: Good call! Added.

camliu89:
@jivey looks good. I don't why the build failed.

PR #2957, opened 2020-04-15
fix small label issues and edit/delete buttons not unfocusing when cl…

camliu89:
- On the OS HW card, The scores and feedback settings are not syncing with the template properly. The setting for ‘For manually-graded questions’ is displaying the selection made for ‘For auto-graded questions’

- The HW template says “After the due date” but the preview card shows “After assignment is due”. The card should also display “After the due date” if that option is selected.

-  The hover state doesn’t go away. To repeat: Click on delete/edit. Click cancel on the modal that opens. On the card, the button is still in the hover state. You have to click on it or somewhere else to remove the hover state

### DIFF
--- a/tutor/specs/screens/grading-templates/__snapshots__/card.spec.js.snap
+++ b/tutor/specs/screens/grading-templates/__snapshots__/card.spec.js.snap
@@ -40,6 +40,10 @@ exports[`Grading Templates Card matches snapshot for homework 1`] = `
   justify-content: space-between;
 }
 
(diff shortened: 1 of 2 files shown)


camliu89:
@jivey Does `./bin/test tutor -u` actually update the tests when you change styles in a component?

camliu89:
I have been changing them manually! That is good to know.

camliu89:
@jivey tests updated.

PR #2959, opened 2020-04-16
Only accept digits in number input

jivey:
Thought about putting this into NumberInput, but feels better at this level in case we need NumberInput to support commas or other formats in the future.

### DIFF
--- a/tutor/specs/integration/grading-templates.spec.js
+++ b/tutor/specs/integration/grading-templates.spec.js
@@ -58,4 +58,33 @@ context('Grading Templates', () => {
     })
 
   });
+
+  it('only accepts digits in number fields', () => {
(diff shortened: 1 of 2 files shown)


tutor/specs/integration/grading-templates.spec.js
```diff
@@ -58,4 +58,33 @@ context('Grading Templates', () => {
...
+    editor('#late_day_deduction')
+      .should(el => expect(el.val()).toEqual('3'));
+
+    editor('#late_work_penalty_assignment').click({ force: true });
+
+    editor('#late_assignment_deduction').clear().type('3a');
+    editor('#late_assignment_deduction')
+      .should(el => expect(el.val()).toEqual('3'));
+  });
 });
```
nathanstitt, line 90: nice spec!

PR #2961, opened 2020-04-17
Fix/wrm 882 hw grading template mid pri

camliu89:
Screenshot from 2020-04-17 10-17-14

- Auto focus the template name when modal is loaded.
- Validate when template name is empty or there is already a template with the same name (case insensitive). ***Bad data: Make sure to delete any previous duplicate names.***
- Style the field text field when validation throws an error.
- Two caveats:
1. When modal is opened, validation happens once when text field is out focus.
2. After, validation happens as the user types in the text field.

### DIFF
--- a/tutor/src/screens/grading-templates/editors.js
+++ b/tutor/src/screens/grading-templates/editors.js
@@ -90,15 +90,20 @@ const HintText = styled.div`
   ${fonts.faces.light};
 `;
 
-const StyledTextInput = styled(Field).attrs((props) => {
-  console.log(props);
-  return { type: 'text' };
+const StyledTextInput = styled(Field).attrs({
+  type: 'text',
 })`
   padding: 0.8rem 1rem;
   border-radius: 4px;
   border: 1px solid ${colors.forms.borders.light};
   font-size: 1.4rem;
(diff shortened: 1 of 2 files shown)


tutor/src/screens/grading-templates/editors.js
```diff
@@ -97,6 +97,13 @@ const StyledTextInput = styled(Field).attrs({
   border-radius: 4px;
   border: 1px solid ${colors.forms.borders.light};
   font-size: 1.4rem;
+  /** styling errors when template name is invalid */
+  background: ${props => props.hasError ? '#fbe8eA' : '#FFFFFF'};
+  color: ${props => props.hasError ? 'red' : 'black'};
+  font-weight: ${props => props.hasError ? 500 : 400};
```
jivey, line 103: I think the Invision spec has this as the default weight instead of bold?

jivey: Oh we can probably bump the font size down to 1.2rem as well.

camliu89: oh yeah. I don't know why at first it looked a little bolder to me. thanks!

PR #2968, opened 2020-04-22
setting up ISO Date based on local timezone

camliu89:
@jivey @nathanstitt this fix is not related to WRM. Is it still ok to push it to the WRM branch?

### DIFF
--- a/tutor/src/helpers/dates.js
+++ b/tutor/src/helpers/dates.js
@@ -33,9 +33,13 @@ export function findLatest(...dateThings) {
   return max(compact(map(flatten(dateThings), toDate)));
 }
 
-
 export function dateWithUnchangedTime(date, t) {
   const time = moment(t);
   return moment(date)
     .hour(time.hour()).minute(time.minute()).seconds(time.seconds()).millisecond(0);
+}
+
+export function guessISODateWithTimezone(isoDate) {
+  const timezone = moment.tz.guess(true);
+  return moment(isoDate).tz(timezone);
 }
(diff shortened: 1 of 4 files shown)


jivey:
@camliu89 `DateTimeInput` is new for WRM, are you saying this should also be applied to the assignment builder on the main branch?

camliu89:
oh nvm. Assignment builder is part of WRM. So this fix should be related to WRM. @jivey @nathanstitt do we want this then for the 5/6 release.

nathanstitt:
@camliu89 good catch. I don't think we need it for the 5/6 release tho, lets just put it in WRM so it can be tested along with that code

tutor/src/helpers/dates.js
```diff
@@ -33,9 +33,13 @@ export function findLatest(...dateThings) {
...
 
-
 export function dateWithUnchangedTime(date, t) {
   const time = moment(t);
   return moment(date)
     .hour(time.hour()).minute(time.minute()).seconds(time.seconds()).millisecond(0);
 }
+
+export function guessISODateWithTimezone(isoDate) {
+  const timezone = moment.tz.guess(true);
```
Dantemss, line 43: I don't think this is quite right... it should read the timezone from the course instead of using the browser's.

camliu89: I might be missing something @Dantemss . I am guessing the timezone of the user here so it reflects the date and time of the course dates in the user's timezone.

PR #2980, opened 2020-05-04
Wrm 929 hw rd assignment fixes

camliu89:
1. Textfield: On page load, the first text field should be in focus and auto-selected so that the user can just start typing right away.

2. The focus state is browser default for the first two fields and it is stylized for the other fields. Let’s use a standard focus state for everything.

3. Update the error state VD. See PL(pattern library) or https://invis.io/EMWORYSXHRD#/402832345_10-v2_Edit_Grading_Template-_ERROR

4. For the ‘NEW” icon, change text weight to 500 and increase character spacing a little. It is currently difficult to read.

5. Show the ‘NEW’ icon only on the default landing page. Need not show for every selection.

6. When a user clicks on ‘select sections’, load only a checklist with section names.
a. Once a user checks/selects a section, THEN load the date and time fields. (Otherwise, the user has to scroll down for the section they want to select and uncheck all the other sections and this could cause confusion.
b. On load, the first available date and time field should be in focus.


cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
+}
```
nathanstitt, line 5: I think this, and the image of the failure below can be removed, no reason to have them in the repo.

Also we should add them to the `.gitignore` so they don't get accidentally committed again.

tutor/src/screens/assignment-edit/builder.js
```diff
@@ -63,15 +63,31 @@ const HintText = styled.div`
...
   border: 1px solid ${colors.forms.borders.light};
-  font-size: 1.4rem;
-
+  font-size: 1.2rem;
+  /** styling errors when template name is invalid */
+  background: ${props => props.haserror ? '#fbe8eA' : '#FFFFFF'};
+  color: ${props => props.haserror ? 'red' : 'black'};
+  border-color: ${props => props.haserror ? '#f4c0c5' : 'd5d5d5'};
+  border-width: ${props => props.haserror ? '2px' : '1px'};
+
```
nathanstitt, line 88: rather than using hex values here, how about adding them to the `colors.states` in theme.js? seems like something that should be standardized.

tutor/src/screens/assignment-edit/details-body.js
```diff
@@ -172,6 +177,10 @@ const DetailsBody = observer(({ ux }) => {
           id="title"
           validate={isRequired}
           data-test-id="edit-assignment-name"
+          
+          innerRef={nameInputField}
+          // Why use 1 and 0 :https://github.com/styled-components/styled-components/issues/1198
+          haserror={!!ux.form.touched.title && !!ux.form.errors.title ? 1 : 0}
         />
```
nathanstitt, line 184: logic looks sound here, but why `1` or `0` instead of `true/false`?   I prefer using `Boolean` to typecast instead of !! since it's more explicit

Seems like this could be: 
`haserror={Boolean(ux.form.touched.title && ux.form.errors.title)}`

tutor/src/screens/grading-templates/editors.js
```diff
@@ -145,6 +121,30 @@ const SettingLabel = styled.label`
...
+  &:focus {
+    outline: 0;
+    border-color: ${colors.forms.borders.focus};
+    box-shadow: 0 0 4px 0 ${colors.forms.borders.focusShadow};
+  }
+
+  &::placeholder {
+    font-size: inherit;
+  }
+`;
```
nathanstitt, line 146: same comment as above about trying to avoid using bare hex values.

tutor/src/screens/grading-templates/editors.js
```diff
@@ -306,7 +306,8 @@ class TemplateForm extends React.Component {
             // Check if field is on focused ('touched') first and then check for any errors.
             // Validate then happens when the field is out of focus for the first time.
             // After, it validates as the user types.
-            hasError={!!form.touched.name && !!form.errors.name}
+            // Why use 1 and 0 :https://github.com/styled-components/styled-components/issues/1198
+            haserror={!!form.touched.name && !!form.errors.name ? 1 : 0}
           />
```
nathanstitt, line 311: and same here as above about using boolean values

tutor/src/screens/assignment-edit/details-body.js
camliu89: It was throwing some console errors but the `Boolean()` worked!

PR #2981, opened 2020-05-05
New timezones

Dantemss:
Goes with https://github.com/openstax/tutor-server/pull/2043

The backend switched to using IANA timezone names so we can do comparisons in Postgres SQL.

- time_zone => timezone
- Use IANA timezone names


Dantemss:
Huh there's a label that I missed in the course settings page.

Dantemss:
We decided to just display the standard timezone names instead of keeping the Rails timezones for display only.

PR #2984, opened 2020-05-05
Wrm 931 events

camliu89:
Events:
1. Open date: If a semester hasn’t started at the time of assignment creation then the default open date should be first day of the semester at 12:01 AM
2. Set the default due date to 7 days from the open date
3. Hide the ‘close date and time’ field

External (Adding #932 since it is similar to #931):
1. Open date: If a semester hasn’t started at the time of assignment creation then the default open date should be first day of the semester at 12:01 AM
2. Set the default Due date to 7 days from the open date
3. Set the default Close date and time to 1 min after the due date.
4. The PB ‘Publish’ gets enabled even when the due date and close date fields are blank. Shouldn't happen.


camliu89:
Actually, I couldn't reproduce the bug. I will add a comment on the card.

PR #2986, opened 2020-05-06
Wrm 930 date time picker style

camliu89:
1. The vertical separator between date and time is missing.
2. The time preview in top left should show the exact time being set.
3. It should show AM/PM instead of seconds
4. Add 01 and 59 also to the seconds option
5. Show the hours in a reverse order 12, 11, 10, 09 - preferred times since preferred due time are 9, 10, 11, 12 etc.
6. The arrows next to the month should be bigger at least 12 /16 px.
7. The confirmation button ‘Okay’ on the date and time modal is easy to miss and makes me wonder if we even need it.
8. The new D&T picker has a ‘now’ option, Update the helper text below ‘Assign’ to reflect that. New helper text-"To immediately open an assignment, select ‘now’ in the calendar."

Screenshot from 2020-05-06 15-07-46

### DIFF
--- a/tutor/specs/integration/assignment-edit.spec.js
+++ b/tutor/specs/integration/assignment-edit.spec.js
@@ -174,7 +174,7 @@ context('Assignment Edit', () => {
     });
   })
 
-  it.only('updates due and closes date manually, select a grading a template, but due and closes date is not updated', () => {
+  it('updates due and closes date manually, select a grading a template, but due and closes date is not updated', () => {
     const templateName = 'Template to not update date'
     const typedDueDate = 'Jun 10 | 05:00 PM'
     const typedClosesDate = 'Jun 22 | 05:00 PM'
--- a/tutor/specs/integration/assignment-edit.spec.js
+++ b/tutor/specs/integration/assignment-edit.spec.js
@@ -2,6 +2,8 @@ import { range } from 'lodash';
 import moment from 'moment-timezone';
 
 context('Assignment Edit', () => {
+  const format = 'MMM D | hh:mm A';
+
   const fillDetails = () => {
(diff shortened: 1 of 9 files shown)


tutor/specs/integration/assignment-edit.spec.js
```diff
@@ -172,10 +174,10 @@ context('Assignment Edit', () => {
     });
   })
 
-  it('updates due and closes date manually, select a grading a template, but due and closes date is not updated', () => {
+  it.only('updates due and closes date manually, select a grading a template, but due and closes date is not updated', () => {
```
jivey, line 177: Might want to remove that... I noticed because I do it all the time haha

tutor/src/components/date-time-input.js
```diff
@@ -64,6 +64,12 @@ const DateTimeInput = (props) => useObserver(() => {
   const id = props.id || uniqueId(props.name);
   const LabelWrapper = props.labelWrapper || React.Fragment;
 
+  const onUpdateDate = dt => {
+    const ev = { target: { name: field.name, value: dt } };
+    field.onChange(ev);
+    props.onChange && props.onChange(ev);
+  };
+
```
nathanstitt, line 72: FYI, I typically prefer to use `@action.bound` decorator instead of using `=>` @action.bound wraps the entire function's changes in a mobx transaction action so they're all applied at once.

tutor/specs/integration/assignment-edit.spec.js
camliu89: :) haha good catch

tutor/src/components/date-time-input.js
camliu89: good to know @nathanstitt !

PR #2987, opened 2020-05-07
remove research analytics menu option

nathanstitt:
Also some bug fixes from sentry


Dantemss:
Hm not sure it's a good idea to disable the cards spec. Isn't that one pretty important?

nathanstitt:
@Dantemss yeah but it's really not testing anything here, we have individual specs that test it on other screens. I think this one is just loading too many cards at once

PR #2988, opened 2020-05-08
WRM Grading Block

jivey:
Adds the Grading Block in Assignment Details with logic to display 4 possible states based on if the exercises are all MCQs or mixed with WRM/OE, and if the due date is past or not. And hides the block on non-homework types.

There's only a basic test for now due to time with trying to set up test data. But I did include a start for getting open ended exercise factories loading from the test server backend. Can pick it up later and get build some specs based around those states mentioned above.

Screenshot_2020-05-07 OpenStax Tutor
Screenshot_2020-05-07 OpenStax Tutor(1)
Screenshot_2020-05-07 OpenStax Tutor(2)
Screenshot_2020-05-07 OpenStax Tutor(3)
Screenshot_2020-05-07 OpenStax Tutor(4)

### DIFF
--- a/shared/src/model/exercise.js
+++ b/shared/src/model/exercise.js
@@ -87,7 +87,7 @@ class Exercise extends BaseModel {
 
   @computed get isMultiPart() { return this.questions.length > 1; }
   @computed get isSinglePart() { return this.questions.length == 1; }
-  @computed get isMultiChoice() { return every(this.questions, 'isMultiChoice'); }
+  @computed get isMultiChoice() { return every(this.questions, 'isMultipleChoice'); }
 
 
   @computed get isPublishable() {
--- a/tutor/src/screens/assignment-review/grading-block.js
+++ b/tutor/src/screens/assignment-review/grading-block.js
@@ -1,5 +1,4 @@
 import { React, PropTypes, styled, observer, css } from 'vendor';
-import { Button } from 'react-bootstrap';
 import { colors } from 'theme';
 import TutorLink from '../../components/link';
 
@@ -69,7 +68,7 @@ const LightBlue = css`
 const Gray = css`
   border-color: ${colors.neutral.pale};
   background: #eeeded;
-`
+`;
 
 const ChartWrapper = styled.div`
   margin: 0 1rem;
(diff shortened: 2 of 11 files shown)


jivey:
@nathanstitt @camliu89 Fixed the spec issue, FYI I changed the scores response to use specific plan ids for different data, so 1 is now a reading, 2 is a homework, etc.

So if you need to test a homework in another spec that will call the scores API just make sure it uses `2` as the plan id.

Open to hearing thoughts on if this should be different! I could look at sending a string id of "homework", "homework-with-all-mc" etc.  and default ints to a homework.

shared/src/model/exercise.js
```diff
@@ -87,7 +87,7 @@ class Exercise extends BaseModel {
 
   @computed get isMultiPart() { return this.questions.length > 1; }
   @computed get isSinglePart() { return this.questions.length == 1; }
-  @computed get isMultiChoice() { return every(this.questions, 'isMultiChoice'); }
+  @computed get isMultiChoice() { return every(this.questions, 'isMultipleChoice'); }
```
camliu89, line 90: I see that `isMultiChoice` and `isMultipleChoice` are being used, what is the difference?

tutor/src/screens/assignment-review/grading-block.js
```diff
@@ -0,0 +1,210 @@
...
+      <ViewScores />
+    </>
+  );
+};
+
+const GradingBlock = observer(({ ux }) => {
+  const { isEveryExerciseMultiChoice, isPastDue } = ux.planScores.taskPlan;
+  const Blocks = isEveryExerciseMultiChoice ?
+    [BeforeDueMCQ, AfterDueMCQ] : [BeforeDueWRQ, AfterDueWRQ];
+  const Block = Blocks[+ isPastDue];
```
camliu89, line 201: ha! nice. this is smart.

shared/src/model/exercise.js
jivey: No difference, just kept it in sync the isMultiPart naming above. Hmm, maybe we should change them to be consistent. `isMultipleChoice` is the actual flag I need and I think it was a typo to call `isMultiChoice`, so this fixes that.

PR #2991, opened 2020-05-11
WMR: Save grades after the teacher inputs them

nathanstitt:
also adds rudimentary support for saving and returning them in the mock
data.  Once the UI is capable of showing already graded questions,
we should be able to test that they're saved

### DIFF
--- a/tutor/src/models/task-plans/teacher/grade.js
+++ b/tutor/src/models/task-plans/teacher/grade.js
@@ -15,7 +15,6 @@ class TeacherTaskGrade extends BaseModel {
     this.question = question;
     this.grader_points = points;
     this.grader_comments = comment;
-  // }, task_step_id: question.task_step_id, grader_points: points, grader_comments: comment,
   }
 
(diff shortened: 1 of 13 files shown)


tutor/src/models/task-plans/teacher/grade.js
```diff
@@ -0,0 +1,40 @@
...
+class TeacherTaskGrade extends BaseModel {
+
+  @belongsTo({ model: TeacherTaskPlan }) taskPlan;
+
+  constructor({ question, points, comment }) {
+    super();
+    this.question = question;
+    this.grader_points = points;
+    this.grader_comments = comment;
+  // }, task_step_id: question.task_step_id, grader_points: points, grader_comments: comment,
```
camliu89, line 18: is this suppose to be deleted?

nathanstitt: yep it was.  thanks, glad yo caught it!

PR #2995, opened 2020-05-13
Hide tabs for external and event types

jivey:
- Don't render Submission and Scores tabs for External and Event types
- Fix an issue with External type in the modal not going to assignment review 

image

### DIFF
--- a/tutor/src/screens/assignment-review/ux.js
+++ b/tutor/src/screens/assignment-review/ux.js
@@ -254,4 +254,8 @@ export default class AssignmentReviewUX {
     return Boolean(['reading', 'homework'].includes(this.planScores.type));
   }
 
+  @computed get canDisplaySubmissionAndScoresTabs() {
+    return Boolean(['reading', 'homework'].includes(this.planScores.type));
+  }
+
 }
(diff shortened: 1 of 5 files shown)


tutor/src/screens/assignment-review/ux.js
```diff
@@ -254,4 +254,8 @@ export default class AssignmentReviewUX {
     return Boolean(['reading', 'homework'].includes(this.planScores.type));
   }
 
+  @computed get canDisplaySubmissionAndScoresTabs() {
+    return Boolean(['reading', 'homework'].includes(this.planScores.type));
+  }
+
```
camliu89, line 260: @jivey isn't this returning the same thing as the function above?

jivey: @camliu89 I thought about DRYing them into a single method but couldn't decide on a good name to refer to these two types. "Workable"? Let me know if you have any suggestions!

camliu89: mmm. I agree with you. I don't like long names. mmm. "hasReadingOrHomeworkType"? It is what is checking for :thinking:

jivey: That's way better and simpler! 😄

PR #3001, opened 2020-05-18
Wrm 958 cloned assignments

camliu89:
This is the cloned assignment modal.


tutor/src/screens/assignment-edit/clone/editor.js
```diff
@@ -0,0 +1,162 @@
...
+
+  async editDetails() {
+    const { ux: { saveAsDraft, navigateToStep } } = this.props;
+    await saveAsDraft();
+    navigateToStep('details');
+  }
+
+  onCancel() {
+    this.props.onHide();
+  }
```
nathanstitt, line 87: functionally looks good, but these methods should be decorated with `@action` so that mobx only fires updates once.  Plus if you make them `@action.bound` you wouldn't need to use functions in the JSX callbacks below.

tutor/src/screens/assignment-edit/clone/editor.js
```diff
@@ -0,0 +1,162 @@
...
+            <Button
+              className="other-button"
+              size="lg"
+              onClick={() => this.publish()}
+            >
+            Publish
+            </Button>
+            <Button
+              size="lg"
+              onClick={() => this.editDetails()}
```
nathanstitt, line 149: `onClick` callbacks like this could just be: `onClick={this.editDetails}` if it was `@action.bound`

tutor/src/screens/assignment-edit/ux.js
```diff
@@ -25,14 +26,16 @@ export default class AssignmentUX {
...
   @observable form;
   @observable activeFilter = 'all';
   @observable templates;
   @observable plan;
   @observable didUserChangeDatesManually = false;
   @observable dueAt;
 
+  // Check if the assignment is old, before WRM release. (No grading template)
+  isCloneOldAssignment;
+
```
nathanstitt, line 38: You'll need to make this `@observable` otherwise mobx won't re-render if it's changed

tutor/src/screens/assignment-edit/ux.js
```diff
@@ -47,8 +50,8 @@ export default class AssignmentUX {
       if (!course.pastTaskPlans.api.hasBeenFetched) {
         await course.pastTaskPlans.fetch();
       }
-      this.sourcePlanId = id;
-      this.plan = course.pastTaskPlans.get(id).createClone({ course });
+      this.plan = course.pastTaskPlans.get(id).createClone({ course, cloned_from_id: id });
+      this.isCloneOldAssignment = Boolean(course.pastTaskPlans.get(id).grading_template_id);
     } else {
```
nathanstitt, line 55: will pastTaskPlans always have a plan for the id?  If not the `get` will return undefined and it'll cause errors.  Maybe need to use lodash's `get`:
`Boolean(get(course.pastTaskPlans.get(id), 'grading_template_id', false));`

tutor/src/screens/assignment-edit/ux.js
```diff
@@ -217,6 +222,11 @@ export default class AssignmentUX {
     return <Step ux={this} />;
   }
 
+  @action.bound renderMiniCloneEditor(form) {
+    this.form = form;
+    return <DetailsBody ux={this} />;
+  }
+
```
nathanstitt, line 229: This seems dangerous because rendering shouldn't set properties, but I don't see where this is called from.  Beside being bad form, React & Mobx will sometimes complain about circular references if setting `form` causes a re-render.

tutor/src/screens/assignment-edit/ux.js
camliu89: it is only set in the constructor so I thought it does not need the `@observable`.

tutor/src/screens/assignment-edit/ux.js
camliu89: Shouldn't there be always a plan from the previous course? Unless the instructor deletes the plan/assignment from the previous source? Can they do that?

PR #3002, opened 2020-05-18
Delete assignment

nathanstitt:
image


tutor/src/screens/assignment-edit/delete-btn.js
```diff
@@ -0,0 +1,90 @@
...
+
+// eslint-disable-next-line
+const DeleteModal = observer(({ isVisible, show, onClose, isBusy, onDelete }) => {
+
+  return (
+    <Modal
+      show={show}
+      onHide={onClose}
+      className="settings-delete-assessment-modal">
+      <Modal.Header closeButton={true}>
```
camliu89, line 19: I think if you just add `closeButton`, without the `{true}`, the component will treat it as `true`.

PR #3009, opened 2020-05-21
Code for pre-wrm release

nathanstitt:
* Add support for force reloading to dashboard when feature flag is set from BE settings


nathanstitt:
this has been released - will merge into master

PR #3010, opened 2020-05-22
Wrm 969 stuent assignment breadcrumbs

camliu89:
Adding new design for homework progress. Need to add the points the user got for each step.


camliu89:
Screenshot from 2020-05-21 20-42-53

PR #3015, opened 2020-05-26
Wrm scores fixes

jivey:
- Display when a student has been granted an extension
- Add a ContentWrapper to start adding min/max width layout requirements
- Display points or dashes based if the question has been graded, has been completed, was not completed before the due date, etc. Regarding the past due date, the BE will send 0 for `points` after the due date if the question wasn't answered, so we don't have to manually check for due dates, just `points`/`gradedPoints` presence.
- Add name order toggle
- Fix displaying "FR" displaying instead of "WRQ"
- Fix many places real values weren't being displayed yet

image

### DIFF
--- a/tutor/specs/integration/assignment-review.spec.js
+++ b/tutor/specs/integration/assignment-review.spec.js
@@ -64,7 +64,6 @@ context('Assignment Review', () => {
     cy.location('pathname').should('include', '/course/1/t/month');
   });
 
-  
   it('can update details', () => {
     cy.server();
     cy.route('GET', '/api/courses/1/grading_templates*').as('getGradingTemplates');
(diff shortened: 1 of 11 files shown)


tutor/specs/integration/assignment-review.spec.js
```diff
@@ -96,15 +98,19 @@ context('Assignment Review', () => {
...
     cy.getTestElement('questions-block').should('exist');
 
-    cy.visit('/course/1/assignment/review/3');
+    cy.visit('/course/1/assignment/review/4');
     cy.getTestElement('grading-block').should('not.exist');
     cy.getTestElement('questions-block').should('not.exist');
 
-    cy.visit('/course/1/assignment/review/4');
+    cy.visit('/course/1/assignment/review/5');
     cy.getTestElement('grading-block').should('not.exist');
```
nathanstitt, line 114: sorry I might have caused this when I made  plan 3 be a WRM assignment

tutor/src/components/icons/extension.js
```diff
@@ -0,0 +1,27 @@
...
+  position: absolute;
+  right: 1rem;
+  background: #009670;
+  color: #fff;
+  font-size: 0.9rem;
+  line-height: 1.4rem;
+  border-radius: 50%;
+  text-align: center;
+`;
+
```
nathanstitt, line 17: this seems like it would be useful on the other screens?  If so how about extracting the: `<GreenCircle>E</GreenCircle>` and put it in `components/icons/extension.js` ?

tutor/src/models/task-plans/teacher/scores.js
```diff
@@ -82,6 +108,10 @@ class TaskPlanScoreHeading extends BaseModel {
     return 'Tutor' !== this.type;
   }
 
+  @computed get displayType() {
+    return this.type === 'FR' ? 'WRQ' : this.type;
+  }
+
```
nathanstitt, line 114: looks the `FR` is coming from the backend, I just made https://github.com/openstax/tutor-server/pull/2067 to fix it so we can remove this

tutor/src/components/icons/extension.js
jivey: This also made me realize I need to move the position stuff to be specific to the scores table, so thanks :D

jivey:
Hmm the time failure seems like a random one on the Travis end of things.

PR #3016, opened 2020-05-27
Revert 2981 new timezones

Dantemss:
Not needed. Timezone stuff is only in master, not in the wrm branch.


Dantemss:
Not needed. Timezone stuff is only in master, not in the wrm branch.

PR #3018, opened 2020-05-27
Calculate number of WRQ questions from the homework selections

Dantemss:


nathanstitt:
spec failure has been fixed in #3019

PR #3020, opened 2020-05-27
Wrm 994 gradebook fixes

camliu89:
small gradebook fixes:
1. The excel export is still downloading the old format (I checked after deployment)
2.  When changing the course avg weight, the course avg doesn't get updated upon clicking save in the modal. Currently it updates only on refreshing the page. Needs to update the avg on save.
3. Class average and min. and max. scores should show triple dash if data is not available (not NaN, as currently displaying)
4. Clicking on an assignment header should open the Assignment page for that assignment

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1019,13 +1019,6 @@
   resolved "https://registry.yarnpkg.com/@emotion/hash/-/hash-0.7.2.tgz#53211e564604beb9befa7a4400ebf8147473eeef"
   integrity sha512-RMtr1i6E8MXaBWwhXL3yeOU8JXRnz8GNxHvaUfVvwxokvayUY0zoBeWbKw1S9XkufmGEEdQd228pSZXFkAln8Q==
 
-"@emotion/is-prop-valid@^0.8.1":
-  version "0.8.2"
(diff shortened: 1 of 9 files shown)


yarn.lock
```diff
@@ -11461,16 +11409,6 @@ styled-components@^5.1.0:
...
-  resolved "https://registry.yarnpkg.com/stylis-rule-sheet/-/stylis-rule-sheet-0.0.10.tgz#44e64a2b076643f4b52e5ff71efc04d8c3c4a430"
-  integrity sha512-nTbZoaqoBnmK+ptANthb10ZRZOGC+EmTLLUxeYIuHNkEKcmKgXX1XWKkUBT2Ac4es3NybooPe0SmvKdhKJZAuw==
-
-stylis@^3.5.0:
-  version "3.5.4"
-  resolved "https://registry.yarnpkg.com/stylis/-/stylis-3.5.4.tgz#f665f25f5e299cf3d64654ab949a57c768b73fbe"
-  integrity sha512-8/3pSmthWM7lsPBKv7NXkzn2Uc9W7NotcwGNpJaa3k7WMM1XDCA4MgT5k/8BIexd5ydZdboXtU90XH9Ec4Bv/Q==
-
 supports-color@5.5.0, supports-color@^5.3.0, supports-color@^5.5.0:
   version "5.5.0"
```
nathanstitt, line 11413: weird, I wonder why this updated?  Seems ok tho

PR #3032, opened 2020-06-01
Hide publish scores on non-gradable assignments

nathanstitt:
nothing to publish otherwise


tutor/src/components/homework-questions.js
```diff
@@ -206,7 +206,7 @@ HomeworkQuestions.displayName = 'HomeworkQuestions';
 HomeworkQuestions.propTypes = {
   headerContentRenderer: PropTypes.func.isRequired,
   questionInfoRenderer: PropTypes.func,
-  footerContentRenderer: PropTypes.func.isRequired,
+  footerContentRenderer: PropTypes.func,
```
jivey, line 209: Whoop good catch.

PR #3038, opened 2020-06-01
Wrm student fixes

jivey:
- Show step progress for homework
- Show extension icon
- Add some spacing in the columns and fix some alignment issues

image

### DIFF
--- a/tutor/src/screens/student-dashboard/cells.js
+++ b/tutor/src/screens/student-dashboard/cells.js
@@ -9,8 +9,14 @@ const Cell = styled.div`
 const TitleCell = styled(Cell).withConfig({
   displayName: 'TitleCell',
 })`
-  padding-left: 1rem;
+  padding: 0 6rem 0 1rem;
   flex: 1;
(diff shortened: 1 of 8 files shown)


tutor/src/screens/student-dashboard/cells.js
```diff
@@ -39,10 +46,14 @@ const ScoreCell = styled(Cell).withConfig({
...
-  width: 10rem;
+  width: 8rem;
   ${({ theme }) => theme.breakpoint.mobile`
      flex: 1;
   `}
+
+  svg.fa-star {
+    padding: 3px;
+  }
 `;
```
nathanstitt, line 57: rather than using the `fa` class, how about using `ox-icon-star` that way it'd continue to work if we ever use other icons with Icon

jivey: Yeah I was torn about that... this is trying to target bumping down the star size within the circle, so the specificity probably helps unexpected results later on.

PR #3044, opened 2020-06-02
Fix double title render and some small visual fixes

jivey:
Consolidates these pages to use the breadcrumb for the title exclusively, and removes the size option since it's not really used.

Also moves the "This assignment is now open for grading." text to be aligned with the Grade button.

image

image

image

image

image

image


jivey:
Also adds a `flex-shrink: 0` to prevent the selected answer circle getting 🥞 

image

PR #3050, opened 2020-06-04
Wrm more grader fixes

camliu89:


tutor/src/screens/assignment-grade/question.js
```diff
@@ -93,31 +92,78 @@ const ExpandGraded = observer(({ ux }) => {
...
+                  response={response}
+                  key={index}
+                  index={index}
+                  ux={ux} />)}
+            </QuestionBody>
+          </StyledQuestion>
+        )
+    }
+  </>
 ));
```
nathanstitt, line 167: hm, instead of a huge ternary, how about extracting the two sides into separate components and:
```
const AssignmentGradingQuestion = observer(({ ux }) => (
  const Component = ux.showOverview ? Overview : Question;
 return <Component ux={ux} />;
})
```

tutor/src/screens/assignment-grade/ux.js
```diff
@@ -90,29 +100,62 @@ export default class AssignmentGradingUX {
...
-    
+
+    if(doGoToOverview) 
+      this.goToOverview();
     // move to next question if any
-    if (isEmpty(this.unGraded)) {
+    else if (isEmpty(this.unGraded) && doMoveNextQuestion) {
       if (this.selectedHeadingIndex < this.headings.length - 1)
         this.selectedHeadingIndex += 1;
     }
```
nathanstitt, line 122: I usually prefer to always use braces, otherwise it's really easy to add a new line right before `this.goToOverview();` and break it.

tutor/src/screens/assignment-grade/question.js
camliu89: that is a good idea! :+1:

tutor/src/screens/assignment-grade/ux.js
```diff
@@ -90,29 +100,62 @@ export default class AssignmentGradingUX {
...
+    this.selectedStudentIndex = 0;
+    this.showOverview = false;
+  }
+
+  @action.bound goToOverview() {
+    this.selectedHeadingIndex = undefined;
+    this.expandGradedAnswers = false;
+    this.selectedStudentIndex = 0;
+    this.showOverview = true;
+  }
```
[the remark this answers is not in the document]
camliu89: thanks!

PR #3054, opened 2020-06-04
Wrm more more grader fixes

camliu89:
1. From the assignment Modal, submission page, and details page - open the first available response for grading.
2. From the specific question in the submission overview page - open the first available response for grading in that particular grading.
3. When all available answers have been graded, the assignment modal would show a default ‘Re-grade’ button
4. Change score to points
5. All number will be in the 0.0 format
6. Make number input box wider
7. When only ‘save and exit’ button is there it will be a primary (orange) button


tutor/src/screens/teacher-dashboard/plan-details.jsx
```diff
@@ -56,12 +57,17 @@ class CoursePlanDetails extends React.Component {
...
 
-  UNSAFE_componentWillMount() {
+  async UNSAFE_componentWillMount() {
     this.props.tourContext.otherModal = this;
+    // get scores to check if there are any students left to grade
+    await this.props.plan.scores.fetch();
+    await this.props.plan.scores.taskPlan.fetch();
+    await this.props.plan.scores.taskPlan.analytics.fetch();
+    await this.props.plan.scores.ensureExercisesLoaded();
   }
```
camliu89, line 67: @nathanstitt @jivey what you guys think about this? fetching the plan scores since it is not available yet when the instructor opens the plan-details. I am like...🤨

tutor/src/models/task-plans/teacher/scores.js
```diff
@@ -311,6 +312,34 @@ class TaskPlanScoresTasking extends BaseModel {
...
+  @computed get hasFinishedGrading() {
+    const completedWrmQuestions = filter(this.allStudentQuestionStatus, s => s.is_completed && s.questionHeading.type === 'WRQ');
+    return every(completedWrmQuestions, q => q.is_completed && !q.needs_grading);
+  }
+
+  @computed get canGrade() {
+    const wrmQuestions = filter(this.allStudentQuestionStatus, s => s.questionHeading.type === 'WRQ');
+    return some(wrmQuestions, q => q.is_completed);
+  }
 }
```
nathanstitt, line 343: nice! these'll come in handy.

tutor/src/screens/teacher-dashboard/plan-details.jsx
nathanstitt: If it turns out that we do need them, I'd suggest we change most to use `ensure Loaded` here and then on the `assignment-review/ux` so that they both won't fetch them twice

PR #3055, opened 2020-06-04
fix not displaying only active assigned periods

jivey:
- Consolidates some of the period options to only provide active periods that were assigned to.
- Fixes an issue with assignment review getting period data from scores api instead of the task plan api, which now allows the selector to render when a period doesn't have students.
- Fixes several period-selector issues with the plan stats/details modal that shows when you click a plan on the calendar.

image

image

### DIFF
--- a/tutor/src/screens/assignment-review/ux.js
+++ b/tutor/src/screens/assignment-review/ux.js
@@ -41,7 +41,6 @@ export default class AssignmentReviewUX {
     this.scroller = new ScrollTo({ windowImpl });
     this.planScores = scores;
     this.course = course;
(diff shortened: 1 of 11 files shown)


tutor/src/screens/assignment-review/ux.js
```diff
@@ -65,6 +64,7 @@ export default class AssignmentReviewUX {
   }
 
   @computed get scores() {
+    if (!this.selectedPeriod) { return null; }
```
jivey, line 67: I know this is kind of an undesired pattern, but something in the render cycle is trying to compute this before it should. Would like to track it down in the future.

PR #3057, opened 2020-06-05
fix snapshots

jivey:
I'm not sure why some of these are just now failing for me, maybe a previous Travis run didn't test all the snapshot renders.


jivey:
Actually I see https://github.com/openstax/tutor-js/pull/3057/files has them updated as well, can close this if that gets merged soon.

nathanstitt:
just merged 3056.  Yeah I dunno why these started complaining now.  I have noticed that sometimes you make a unrelated change and it changes the random number increments which throws off what factories return.

PR #3059, opened 2020-06-05
make drop questions button async

jivey:
Might fix a bug on tutor-dev and is a good idea anyway.

Second api conditional on disabled to ensure it gets disabled while submitting to prevent double clicks... normally the isWaiting logic inside the component takes care of it, but didn't seem to be working in this case 🤔

### DIFF
--- a/tutor/src/screens/assignment-review/ux.js
+++ b/tutor/src/screens/assignment-review/ux.js
@@ -212,6 +212,10 @@ export default class AssignmentReviewUX {
     );
   }
 
+  @computed get isDroppedQuestionsSaving() {
+    return this.taskPlan.api.isPending;
+  }
+
   @action.bound onDelete() {
     this.isDisplayingConfirmDelete = true;
   }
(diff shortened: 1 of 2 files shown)


tutor/src/screens/assignment-review/ux.js
```diff
@@ -211,6 +211,10 @@ export default class AssignmentReviewUX {
     );
   }
 
+  @computed get isDroppedQuestionsSaving() {
+    return this.taskPlan.api.isPending;
+  }
```
nathanstitt, line 216: FYI `api.isPending` will also be true for any network activity like `get`.  To check for put you might add a isUpdating or something like that to api that checks `requestsInProgress` like `fetchInProgress` does: https://github.com/openstax/tutor-js/blob/wrm/shared/src/model/api.js#L30-L32

nathanstitt:
@jivey sorry! I went to merge this and discovered it's got a conflict

tutor/src/screens/assignment-review/ux.js
jivey: Ah good note! I like that idea, I should add that while checking the other modals for async...

PR #3062, opened 2020-06-05
Submission overview changes

jivey:
- Render graded response block to expanded question responses
- Add border to table cells
- Tweak some grade button styles, and only render them when the assignment is past due

I think there's still an issue to look into where the grade buttons are displaying with 0 remaining.

image


nathanstitt:
I'm going to merge this because the gradeAnswers element not found error was fixed in a different pr

PR #3070, opened 2020-06-09
show message when grading template cannot be updated

jivey:
Message will start showing after BE PR is merged

Note that this is only a fallback for if the PATCH is sent... it won't prevent the modal from loading. A future PR will prevent the modal from loading by checking for open tasks in the FE.

image


nathanstitt:
I also made a PR for this ;/.  I think they're both good to have tho since mine is on the click and yours will be on save

PR #3074, opened 2020-06-11
A few fixes to the cloned assignment builder

nathanstitt:
image


tutor/src/screens/assignment-edit/details-body.js
```diff
@@ -168,6 +168,26 @@ const ExternalUrlField = observer(({ ux }) => {
...
+    </div>
+  );
+})`
+  color: ${colors.neutral.dark};
+  ul {
+    margin: 0;
+    padding-left: 3rem;
+  }
+`;
+
```
nathanstitt, line 190: @camliu89 and @jivey - what do you think of this pattern?  It lets us avoid creating a new styled component

PR #3080, opened 2020-06-12
add a no activity yyet message in grader for sections that does not h…

camliu89:
…ave any students

### DIFF
--- a/tutor/src/screens/assignment-grade/index.js
+++ b/tutor/src/screens/assignment-grade/index.js
@@ -53,6 +53,29 @@ class AssignmentGrading extends React.Component {
     });
   }
 
+  renderEmpty(courseId) {
+    return (<>
+      <h2>No activity yet</h2>
+      <p>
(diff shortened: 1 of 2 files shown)


tutor/src/screens/assignment-grade/index.js
```diff
@@ -77,12 +78,25 @@ class AssignmentGrading extends React.Component {
...
+              : <>
+                <QuestionsBar ux={ux} />
+                <BodyBackground>
+                  <Question
+                    ux={ux}
+                  />
+                </BodyBackground>
+              </>
+          }
+
```
nathanstitt, line 99: hm, rather than the big ternary, how about moving each block to either a component or a method, and then:
`ux.headings.length === 0 ? this.renderEmpty() : this.renderContent()`

camliu89: I was just doing that :) thanks nathan.

PR #3084, opened 2020-06-12
Wrm assignment edit fixes

jivey:
- Set originalDueAt when initializing a tasking plan, and check it when validating an existing record. This allows the assignment edit form to save things like name/description while considering the due date as untouched. If they change the due date during edit, normal validation will still occur.

- Only render active periods to be selected in the assignment builder.

- Deep clone the task plan when opening the assignment edit modal. This is to prevent the UI from immediately reflecting potentially temp changes because of computed values.

### DIFF
--- a/tutor/src/screens/assignment-review/ux.js
+++ b/tutor/src/screens/assignment-review/ux.js
@@ -1,5 +1,5 @@
 import { React, observable, action, computed } from 'vendor';
-import { first, pick, sortBy, filter, sumBy, get, find, cloneDeep } from 'lodash';
+import { first, pick, sortBy, filter, sumBy, get, find } from 'lodash';
 import ScrollTo from '../../helpers/scroll-to';
 import DropQuestion from '../../models/task-plans/teacher/dropped_question';
 import EditUX from '../assignment-edit/ux';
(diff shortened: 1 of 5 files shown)


tutor/src/screens/assignment-review/ux.js
```diff
@@ -239,7 +239,7 @@ export default class AssignmentReviewUX {
 
     await this.editUX.initialize({
       ...this.params,
-      plan: this.taskPlan,
+      plan: cloneDeep(this.taskPlan),
       history,
```
nathanstitt, line 243: This doesn't seem like it should be needed.  assignment-edit/ux already creates a new task plan in order to prevent this sort of thing. Otherwise the changes would be saved to the in-memory plan even if the user hit cancel.  If there is leakage of objects, we should fix it there IMO

https://github.com/openstax/tutor-js/blob/wrm/tutor/src/screens/assignment-edit/ux.js#L59-L63

jivey: Thanks for this, didn't realize we were doing that in the edit UX and that we needed to pass the plan id here instead of the object.

PR #3087, opened 2020-06-15
Don't allow creation or modification of pre-wrm grading templates

nathanstitt:
Otherwise they'll mess with the scores for those courses


nathanstitt:
travis did pass, it didn't notify GitHub successfully

PR #3089, opened 2020-06-16
don't update step when teacher is reviewing

nathanstitt:
And change button to read "Next"


camliu89:
@nathanstitt Are these changes the same in #3091 ?

nathanstitt:
no, this is the case when a teacher views a student task. When that happens we can't update the step at all, it should be completely read-only

PR #3090, opened 2020-06-16
Use published points and score in grade books

nathanstitt:
Fixes several more spots that were missed


nathanstitt:
thanks for reviewing @camliu89!

PR #3098, opened 2020-06-16
do not allow editing templates that are in use

nathanstitt:
image
lex


tutor/src/models/grading/templates.js
```diff
@@ -29,6 +29,7 @@ class GradingTemplate extends BaseModel {
   @field default_due_time = '21:00';
   @field default_due_date_offset_days = 7;
   @field default_close_date_offset_days = 7;
+  @field has_task_plans;
```
jivey, line 32: Do we want to change this here and in the representer to use `has_open_task_plans`? Can't remember if they want to allow editing up until open time.

nathanstitt: yeah, I was going to 🅿️ this to ask

Dantemss: Yeah I think we do want to change these.

PR #3103, opened 2020-06-17
fix edit external url

jivey:
- Let formik handle setting the field (name="settings.external_url") via initialValues to prevent uncontrolled/controlled swap issue
- Block the field from being edited once a tasking plan is open.
- Add a handler for the unlikely case someone tries to update the plan when a tasking plan is open, but the front end didn't block it


jivey:
ZenHub is throwing an error when I try to connect this to the relevant issue: https://app.zenhub.com/workspaces/openstax-tutor-560aef617a763234615d6a0f/issues/openstax/tutor/1088

cc @meeragovindaraju

PR #3109, opened 2020-06-18
only use the published fields for calculations

nathanstitt:
The student task should only use the published points
and doesn't need to do any math when computing them

Fixes this:
image

Into the correct:
image


nathanstitt:
we decided to change the display to show points without lateness, so it now looks like:

image

PR #3117, opened 2020-06-19
preserve existing extensions when granting new

nathanstitt:
the BE will remove any extensions not sent in the array


tutor/src/models/task-plans/teacher/plan.js
```diff
@@ -458,7 +458,7 @@ class TeacherTaskPlan extends BaseModel {
     return {
       id: this.id,
       data: {
-        extensions,
```
jivey, line 461: Simple, I like it! 🎉

nathanstitt:
travis passed, it just failed to notify GitHub

PR #3120, opened 2020-06-19
Fix NaN after granting extension

jivey:
In cases were the BE doesn't send `total_points` and `total_fraction` (due to extension or not due yet) we need to display `---`.

image

### DIFF
--- a/tutor/src/screens/assignment-review/homework-scores.js
+++ b/tutor/src/screens/assignment-review/homework-scores.js
@@ -168,9 +168,7 @@ const StudentCell = observer(({ ux, student, striped }) => (
       </NameWrapper>
 
       <Total>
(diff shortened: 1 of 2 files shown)


tutor/src/screens/assignment-review/homework-scores.js
```diff
@@ -168,9 +168,7 @@ const StudentCell = observer(({ ux, student, striped }) => (
       </NameWrapper>
 
       <Total>
-        {ux.displayTotalInPercent ?
-          `${S.asPercent(student.total_fraction || 0)}%` :
```
jivey, line 172: I think my original fallback here to `0` was incorrect, and it should always defer to if BE sent the value or not (show the value if we have it, otherwise `---`)

PR #3121, opened 2020-06-19
use fractional_points field for average scores

nathanstitt:
Only show in % format and treat HW & Readings identically for display

### DIFF
--- a/tutor/src/models/task-plans/teacher/scores.js
+++ b/tutor/src/models/task-plans/teacher/scores.js
@@ -3,7 +3,7 @@ import {
 } from 'shared/model';
 import Exercises from '../../exercises';
 import {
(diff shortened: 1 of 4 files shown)


tutor/src/models/task-plans/teacher/scores.js
```diff
@@ -244,9 +244,8 @@ class TaskPlanScoresTasking extends BaseModel {
   @identifier id;
   @field period_id;
   @field period_name;
+  @field total_fraction;
 
-  @field({ type: 'object' }) average_score;
-  @field({ type: 'object' }) available_points;
   @belongsTo({ model: 'task-plan/scores' }) plan;
```
nathanstitt, line 249: I dunno why these were here.  the BE doesn't send those fields but caused some confusion since it seemed like they'd do what we wanted

PR #3122, opened 2020-06-19
make builder fields consistent for font size and color

jivey:
Use 1.4rem and #424242 for fields in the assignment builder.

image


jivey:
@dkhowala There's an attached screenshot if you'd like to see the end result. Text fields use 1.4rem (14px) and #424242.

jivey:
Haha I was just trying to find the command for that, you saved me some digging!

PR #3124, opened 2020-06-19
create tasking_plans when creating external assignment and events wit…

camliu89:
…h a due_at


jivey:
You can ignore the snapshot failure, the fix is in `wrm` now.

PR #3127, opened 2020-06-22
Wrm fix assignment builder

camliu89:
Use flex. Column-width not working properly in Safari.

Before: 
Screen Shot 2020-06-22 at 10 50 27 AM

After:
Screen Shot 2020-06-22 at 10 50 54 AM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ archive.tar.gz
 .vscode/
 screenshots/
 cypress/fixtures/
+dump.rdb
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -14,4 +14,5 @@ archive.tar.gz
 */test/some-tests.*
 .vscode/
 screenshots/
-cypress/fixtures/
\ No newline at end of file
+cypress/fixtures/
+dump.rdb
```
nathanstitt, line 18: 😁 yay!

PR #3128, opened 2020-06-22
set heading color for practice, deal with unknowns

nathanstitt:
Fixes the `TypeError: Cannot read property 'background' of undefined`
from https://github.com/openstax/tutor/issues/1131


nathanstitt:
travis timed out here, specs were ok

PR #3139, opened 2020-06-25
use course timezones in builder

jivey:
- Update the tasking to use the course timezone when making date calculations
- Use momentInZone in the DateTimeInput when rendering values (if course was passed)
- Fix a bug that prevented opens at from being changed if due at wasn't changed (changed `!due_at` check to `due_at`, now matches `onDueChange`.)
- ~~Recalculate course bounding when changing the course time_zone. (This prevents clicking on the start day of a course, and by changing the time_zone to an earlier one, causing the opens_at to be before the first start date. For example, `July 1st 12:01` can become `June 30 11:01PM`~~ Removed, as the entire timezone modal is being removed from the builder
- Use course timezone anywhere we display times on the teacher side
- Use browser timezone on the student side (this already happens because a bare moment call will use the browser, but you have to use `tz.guess()` to actually get the timezone to display when calling `format` with `'z'`.
- Make some small design adjustments to make the wider time displays fit

image

image

image

image

image


tutor/src/screens/assignment-edit/tasking.js
```diff
@@ -242,6 +219,7 @@ class Tasking extends React.Component {
               name={`tasking_plans[${index}].closes_at`}
               onChange={this.onClosesChange}
               disabledDate={this.course.isInvalidAssignmentDate}
+              course={this.course}
```
camliu89, line 222: Maybe just pass the function `momentInZone` instead of the whole course?

jivey: That's a great idea, I'm going to update that and fix up a different issue I found.

tutor/specs/screens/student-dashboard/__snapshots__/dashboard.spec.jsx.snap
```diff
@@ -513,7 +513,7 @@ exports[`Student Dashboard displays as loading 1`] = `
                       className="sc-fznyAO c2"
                     >
                       <time>
-                        We Oct 21, 7:00am
```
jivey, line 516: Just realized these snapshots might change depending on where they are run 🤔

jivey:
Snapshot failures are probably from the CI needing to have the time configuration changed? I reran all the snapshots in CT so other devs at least should see them passing.

```
- Wed, Oct 21, 6:00am CST
+ Wed, Oct 21, 6:00pm +06
```

tutor/src/screens/assignment-review/details.js
```diff
@@ -240,16 +236,25 @@ const TaskingDates = observer(({ tasking, title }) => {
...
         </div>
         {!tasking.plan.isEvent && (
           <div>
             <dt>Close date</dt>
-            <dd>{format(tasking.closesAtMoment)}</dd>
+            <dd>
+              <div>{tasking.closesAtMoment.format(dateFormat)}</div>
+              <div>{tasking.closesAtMoment.format(timeFormat)}</div>
+            </dd>
           </div>)}
```
nathanstitt, line 258: I'm probably missing something?  Why are each of the dates listed twice?

jivey: It breaks the date and time into separate divs so the line break is nicer than wrapping

nathanstitt: aha! I completely missed the date/time difference there

tutor/src/screens/student-dashboard/event-row.jsx
```diff
@@ -105,7 +109,7 @@ class EventRow extends React.Component {
             <span>{event.title}</span>
           </TitleCell>
           <DueCell>
-            <EventTime event={event} />
+            <EventTime event={event} course={course} />
             {event.is_extended && <StyledEIcon />}
```
nathanstitt, line 113: course doesn't seem to be used in EventTime

jivey: Oops good catch, forgot I changed course there. Weirdly, lint didn't tell me!

PR #3141, opened 2020-06-30
remove any pastTaskPlans in the FE that are coming back from the BE a…

camliu89:
…fter a re-fetch

### DIFF
--- a/tutor/src/screens/assignment-edit/clone/editor.js
+++ b/tutor/src/screens/assignment-edit/clone/editor.js
@@ -61,21 +61,20 @@ class Editor extends React.Component {
   async saveDraft() {
     const { ux: { saveAsDraft }, onHide } = this.props;
     await saveAsDraft();
-    onHide({ shouldRefreshPastPlans: true });
+    onHide();
   }
 
   @action.bound
   async publish() {
     const { ux: { savePlan }, onHide } = this.props;
     await savePlan();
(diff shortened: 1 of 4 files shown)


nathanstitt:
Code looks good @camliu89 but I'm not sure if we want this, I'll 🅿️ so we can discuss

tutor/src/screens/assignment-edit/clone/editor.js
```diff
@@ -102,7 +103,7 @@ class Editor extends React.Component {
           <div>
             <Button
               variant="default"
-              onClick={onHide}
+              onClick={() => onHide({ shouldRefreshPastPlans: false })}
               size="lg"
             >
             Cancel
```
nathanstitt, line 109: I had an alternative thought here.  How about we hook into the `assignment-edit/ux` `saveAndCopyPlan ` https://github.com/openstax/tutor-js/blob/master/tutor/src/screens/assignment-edit/ux.js#L444 if we check `this.plan.isClone` https://github.com/openstax/tutor-js/blob/master/tutor/src/models/task-plans/teacher/plan.js#L163 and if that's set, we remove it from `this.course.pastTaskPlans`

That would remove the past task plan immediately and save us a fetch. 

What do you think?

camliu89: oh that works. great idea @nathanstitt .

PR #3142, opened 2020-06-30
gradebook: delete sorting arrow icon in assignment heading, expand as…

camliu89:
…signment title width, and show tooltip if assignment title is wrapped with ellipsis

### DIFF
--- a/tutor/src/screens/teacher-gradebook/table.js
+++ b/tutor/src/screens/teacher-gradebook/table.js
@@ -1,4 +1,4 @@
-import { React, PropTypes, styled, observer, css } from 'vendor';
+import { React, PropTypes, styled, observer, css, useEffect, useState, useRef } from 'vendor';
 import { StickyTable, Row } from 'react-sticky-table';
 import moment from 'moment';
 import { isNil } from 'lodash';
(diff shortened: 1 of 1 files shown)


tutor/src/screens/teacher-gradebook/table.js
```diff
@@ -282,22 +282,29 @@ const StudentColumnHeader = observer(({ ux }) => {
...
+  const titleTextRef = useRef();
+  const [showToolTip, setShowToolTip] = useState(true);
+  useEffect(() => {
+    setShowToolTip(titleTextRef.current.offsetWidth < titleTextRef.current.scrollWidth);
+  });
+
+  return (   
     <OverlayTrigger
+      // Overlay has a lot of problems when showing at the top. Putting at the bottom for now
       placement="bottom"
```
nathanstitt, line 297: I usually prefer to omit placement so it's `auto` - that way it'll adjust the position so it's shown regardless of the page scroll.

camliu89: Yes, but I think the first choice is the `top` placement which causes a bug. That is why I set it to `bottom`. The bug seems to be caused by some of the cells using the `column` css property which caused a lot of bugs already in other places.

PR #3145, opened 2020-06-30
Wrm student instructions step

camliu89:


tutor/src/screens/task/external.js
```diff
@@ -1,46 +1,15 @@
...
-    <rect x="0" y="0" rx="3" ry="3" width="250" height="10" />
-    <rect x="0" y="10" rx="3" ry="3" width="250" height="10" />
-  </ContentLoader>
-);
-
 @observer
 class ExternalTaskStep extends React.Component {
 
-  static Loader = Loader;
-
```
nathanstitt, line 20: nice catch that loader isn't being used

tutor/src/screens/task/external.js
```diff
@@ -1,46 +1,15 @@
...
-      external_url = `http://${external_url}`;
-    }
-    return external_url;
-  }
-
-  @action.bound onContinue() {
-    const step = this.props.ux.currentStep;
-    step.is_completed = true;
-    step.save();
-  }
```
nathanstitt, line 42: I see that `onContextMenu` isn't being used but that is a bug and we need to hook it back up.  

Right clicking on the link isn't allowed since the only valid use case is opening in a new tab and we already do that for them.  We need to record the visit and we can't do that if they right-click and open in new tab.

tutor/src/screens/task/step/continue-btn.js
```diff
@@ -30,9 +30,11 @@ export default function ContinueBtn({
 
 ContinueBtn.propTypes = {
   label: PropTypes.string,
+  nextIncompleteStepId: PropTypes.string,
   ux: PropTypes.shape({
```
nathanstitt, line 34: I'm not clear what the reason for the `nextIncompleteStepId` is for?   We already jump to that step in index.js when the task loads https://github.com/openstax/tutor-js/blob/master/tutor/src/screens/task/index.js#L125-L130

camliu89: mmm. that is true. It does not go to the instructions step when the task step screen loads. 

I guess this is for the scenario when the student is in the next incomplete step and clicks on the instructions step. It just felt weird that the button in the instructions step says "Continue" and it just goes to the next step.

Are we worry about this scenario?

tutor/src/screens/task/step/instructions.js
```diff
@@ -156,21 +157,46 @@ const ContinueBtn = observer(({ ux }) => {
...
+    return (
+      <TutorLink className="btn btn-primary" to="dashboard" params={{ courseId: ux.course.id }}>Close</TutorLink>
+    );
+  }
+
+  let nextIncompleteStepId;
+  let buttonLabel = 'Start';
+  if (ux.task.steps.every(s => s.is_completed)) {
+    buttonLabel = 'Review';
+  }
```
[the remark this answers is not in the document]
camliu89: It is returning undefined. Maybe the BE took it out? I will create two methods `completed` and `started` in the model.

tutor/src/screens/task/step/continue-btn.js
nathanstitt: Oh sorry, I wasn't really clear there.  yeah I'm 100% behind the wording change.  I was just saying that the continue/review button should always go to the next step instead of jumping to the 1st incomplete step.  It seems to me that the instructions page will be shown in 3 cases:
 
1) The assignment is just getting started, in which case we'd always want to go to the next step 
2) the student has navigated back to the start manually mid-way through the assignment,  in that case I think we should still go to the new step, but I'm not sure what the button could read.  I think either would be appropriate.
3) the assignment is complete and the student clicks to review it.  In that case we want to go to the next step but the button should read review.

Accordingly, I think we should make the button always advance to the next step, but it should read "Review" if that step is completed.

I should also add I don't feel super strong about any of this and could easily change my mind if you've got other thoughts 😁

tutor/src/screens/task/external.js
camliu89: good to know. I will put it back.

tutor/src/screens/task/step/continue-btn.js
camliu89: I totally agree with you @nathanstitt ! Thinking about it more, the button in the instructions step should always go to the next step. I will take out the logic of `nextIncompleteStepId`.

Regarding the label of the button, I still think that:

1. If the assignment is just getting started, it should be labeled "Start".
2. If the assignment is in progress, it should be labeled "Continue".
3. If the assignment is completed, it should be labeled "Review".

nathanstitt: yep, that sounds perfect

PR #3146, opened 2020-07-01
Display question stimulus

nathanstitt:
WIP, needs to hide it if it's a MPQ

### DIFF
--- a/exercises/src/components/exercise/question.jsx
+++ b/exercises/src/components/exercise/question.jsx
@@ -96,7 +96,7 @@ class Question extends React.Component {
 
         <QuestionFormatType question={question} />
         {!question.exercise.isMultiPart && (
-          <div className="questio-stimulus">
+          <div className="question-stimulus">
             <label>
               Question Stimulus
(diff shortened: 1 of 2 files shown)


exercises/src/components/exercise/question.jsx
```diff
@@ -80,24 +80,32 @@ class Question extends React.Component {
...
-           <Alert variant="warning">
-             To save your work, you must fill out the {validity.part}
-           </Alert>)}
+          <Alert variant="warning">
+            To save your work, you must fill out the {validity.part}
+          </Alert>)}
 
         <QuestionFormatType question={question} />
+        {!question.exercise.isMultiPart &&
+          (<div className="questio-stimulus">
```
jivey, line 99: I can't tell from just the diff, is this classname declared somewhere (typo?)

nathanstitt: pretty sure it's a typo.  I don't see it used anywhere.  I'll update it, thanks for noticing!

PR #3155, opened 2020-07-06
only display publish button for ungraded scores

nathanstitt:
We've decided it should be hidden unless there are scores to publish

### DIFF
--- a/tutor/src/screens/assignment-review/publish-scores.js
+++ b/tutor/src/screens/assignment-review/publish-scores.js
@@ -3,7 +3,7 @@ import { OverlayTrigger, Tooltip } from 'react-bootstrap';
 import { AsyncButton } from 'shared';
 
 const PublishScores = observer(({ ux }) => {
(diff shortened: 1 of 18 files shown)


tutor/src/screens/assignment-review/publish-scores.js
```diff
@@ -3,7 +3,7 @@ import { OverlayTrigger, Tooltip } from 'react-bootstrap';
 import { AsyncButton } from 'shared';
 
 const PublishScores = observer(({ ux }) => {
-  if ((ux.scores.hasFinishedGrading && ux.scores.hasUnPublishedScore) || ux.planScores.isManualGradingGrade) {
```
camliu89, line 6: should we keep the `ux.planScores.isManualGradingGrade`? I think we are hiding the publish button if the scores are shown after the instructor grades an answer.

nathanstitt: hm, yes I think that could also be removed.  Good catch!

PR #3160, opened 2020-07-08
wait for scores to be fetched before checking if tab is available

camliu89:
Review index component was not waiting for the scores to be fetched so it will always render the page with just one tab.

### DIFF
--- a/tutor/src/components/tabs.jsx
+++ b/tutor/src/components/tabs.jsx
@@ -56,7 +56,9 @@ const Tabs = ({
     ) {
       const ev = new FakeEvent;
       onSelect(activeIndex, ev);
-      selectTabIndex(activeIndex);
+      if (ev.isDefaultPrevented()) {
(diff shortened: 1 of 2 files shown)


tutor/src/components/tabs.jsx
```diff
@@ -56,9 +56,7 @@ const Tabs = ({
     ) {
       const ev = new FakeEvent;
       onSelect(activeIndex, ev);
-      if (ev.isDefaultPrevented()) {
-        selectTabIndex(activeIndex);
-      }
+      selectTabIndex(activeIndex);
     }
```
nathanstitt, line 60: we need to leave the defaultPrevented check.  That allows onSelect handlers to veto a tab change.

camliu89: oh yeah, forgot to put it back. I was doing some testing. thanks nathan.

PR #3161, opened 2020-07-08
[Wait for release after old scores are frozen] Remove unused fields & methods from gradebook

nathanstitt:
I'm a bit worried that I might have removed something that's used but it all seems to work fine 🤞 

Goes along with https://github.com/openstax/tutor-server/pull/2133


Dantemss:
I am a little worried about old scores changing so how about I work on freezing the old scores and we can have this go in the release after?

nathanstitt:
yeah that's a good idea.  no big hurry on getting this in

PR #3163, opened 2020-07-09
update extensions modal with search and design changes

jivey:
- Change some layout and checkbox inputs to match the design doc
- Add a search field (need a separate student list here otherwise the scores UI under the modal changes as well while searching.)

Screenshot_2020-07-09 OpenStax Tutor

### DIFF
--- a/tutor/src/components/checkbox-input.js
+++ b/tutor/src/components/checkbox-input.js
@@ -5,7 +5,7 @@ import { useField } from 'formik';
 import { Icon } from 'shared';
 
 const StyledWrapper = styled.span`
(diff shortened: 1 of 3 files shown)


camliu89:
Looks great @jivey !

Just a weird behavior I found: When I am searching for a student, and then I click "Select All", it will select all the other students that were not matched by the searching criteria. Should the "Select All" checkbox be hidden if the instructor is searching for a student?

jivey:
@camliu89 Updated to hide the select all button when the search has filtered out results. Good idea!

PR #3164, opened 2020-07-09
Post gb fixes

camliu89:
A lot of gradebook bugs fixed: https://github.com/openstax/tutor/issues/1107


tutor/src/screens/teacher-gradebook/table.js
```diff
@@ -438,20 +446,22 @@ const Aggregates = observer(({ ux }) => {
...
         {ux.headings.map((h, i) => (<MinMaxResult key={i} data={h} ux={ux} type={MinMaxType.MIN} />))}
       </Row>
-    </>
+    </StyledAggregateData>
   );
 });
 
 const GradebookTable = observer(({ ux }) => {
+  // calculating the size of the table
+  const size = useWindowSize();
```
jivey, line 455: Hmm I wonder if we could use CSS's `vh` instead of JS here. I might try that in my branch.

PR #3165, opened 2020-07-10
Conditions for reusing preview courses

Dantemss:
- Only reuse active preview courses with the correct offering id that the user is teaching and are marked as reusable
- Added specs

https://github.com/openstax/tutor-server/pull/2136

### DIFF
--- a/tutor/specs/acceptance/.eslintrc.js
+++ b/tutor/specs/acceptance/.eslintrc.js
@@ -1,6 +1,8 @@
 module.exports = {
   "globals": {
     "React":      false,
+    "beforeAll":  false,
+    "afterAll":   false,
     "beforeEach": false,
     "afterEach":  false,
     "describe":   false,
--- a/tutor/specs/models/courses/offerings/previews.spec.js
+++ b/tutor/specs/models/courses/offerings/previews.spec.js
@@ -1,0 +1,139 @@
+import OfferingsMap from '../../../../src/models/course/offerings';
+import CoursesMap from '../../../../src/models/courses-map';
+import Previews from '../../../../src/models/course/offerings/previews';
+
+describe('Offering Previews Model', () => {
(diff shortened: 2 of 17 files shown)


tutor/specs/acceptance/.eslintrc.js
```diff
@@ -1,6 +1,8 @@
 module.exports = {
   "globals": {
     "React":      false,
+    "beforeAll":  false,
+    "afterAll":   false,
     "beforeEach": false,
     "afterEach":  false,
     "describe":   false,
```
nathanstitt, line 8: eslint now allows shared configs that can extend one another, we should probably use that for the specs to prevent all this duplication. 😭

tutor/specs/models/courses/offerings/previews.spec.js
```diff
@@ -0,0 +1,139 @@
+import OfferingsMap from '../../../../src/models/course/offerings';
+import CoursesMap from '../../../../src/models/courses-map';
+import Previews from '../../../../src/models/course/offerings/previews';
+
```
nathanstitt, line 4: Great spec! The newish way we make test data is using `Factory.create()` methods that are in `specs/factories/index.js`.  I think you could use:
```js
courses = Factory.coursesMap({ count: 1, should_reuse_preview: true })
offerings = Factory.offeringsMap({ count: 1, is_preview_available: false })
```

That has the advantage of not polluting the global map with test data, since if there's another spec that runs using it you have to be really careful to clear the global map

Dantemss:
Yes, feel free to fix my spec :D

tutor/specs/models/courses/offerings/previews.spec.js
nathanstitt: hah, so this got pretty complicated quick due to how the the preview class uses offerings.  I'll take a shot at it later hopefully

PR #3167, opened 2020-07-14
Assignment dates

camliu89:
- In the calendar, disable the dates before the course start date
 - In the calendar, disable the dates after the course end date
 - Allow users to select any enabled date from the calendar
 - If the date/time combination is not valid then show an error (See expected behavior above) Invision Screens for error messages: https://invis.io/V6XZPQ74ZBM
 - If any of the day and time is not valid then disable the Save button
 - When you edit assignment details for an assignment that is past due (or past close), the error state does not show up on the due date (or close date) unless the user makes a date edit

### DIFF
--- a/tutor/src/models/course.js
+++ b/tutor/src/models/course.js
@@ -208,11 +208,8 @@ class Course extends BaseModel {
 
   // bind to this so it can be used in disabledDate check
   isInvalidAssignmentDate = (date) => {
-    return !this.momentInZone(date).isBetween(
-      this.allowedAssignmentDateRange.start,
-      this.allowedAssignmentDateRange.end,
-      'day', '[]'
-    );
+    return date < moment(this.starts_at).endOf('day') ||
+    date > moment(this.ends_at).endOf('day');
   }
 
   @computed get hasStarted() {
--- a/tutor/src/models/task-plans/teacher/tasking.js
+++ b/tutor/src/models/task-plans/teacher/tasking.js
@@ -103,14 +103,26 @@ class TaskingPlan extends BaseModel {
     return moment(this.due_at).isAfter(Time.now);
   }
 
+  @computed get isDueAfterOpen() {
+    return moment(this.due_at).isAfter(moment(this.opens_at));
+  }
+
+  @computed get isCloseAfterDue() {
+    return moment(this.closes_at).isAfter(moment(this.due_at));
+  }
+
   @computed get isValid() {
-    return Boolean(
-      this.target_id &&
-        this.target_type &&
-        this.opens_at &&
-        this.due_at &&
-        ((this.isNew || this.dueAtChanged) ? this.isBeforeDue : true)
-    );
+    let isValid = this.target_id && this.target_type && this.opens_at && this.due_at && this.isDueAfterOpen;
+
+    if(this.isNew || this.dueAtChanged) {
+      isValid = isValid && this.isBeforeDue;
+    }
+    //event does not have a close date (visually)
+    if(!this.plan.isEvent) {
+      isValid = isValid && this.isCloseAfterDue;
+    }
+
+    return isValid;
   }
 
   @computed get dueAtChanged() {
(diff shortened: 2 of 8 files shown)


tutor/src/models/task-plans/teacher/tasking.js
```diff
@@ -103,14 +100,26 @@ class TaskingPlan extends BaseModel {
...
   }
 
+  @computed get isDueAfterOpen() {
+    return moment(this.due_at).isAfter(moment(this.opens_at));
+  }
+
+  @computed get isCloseAfterDue() {
+    return moment(this.closes_at).isAfter(moment(this.due_at));
+  }
+
```
nathanstitt, line 110: you don't need to make sure the argument to `isAfter` is a momentjs instance, it'll take care of that for you.  Should be able to: `moment(this.closes_at).isAfter(this.due_at)`

tutor/src/models/task-plans/teacher/tasking.js
```diff
@@ -103,14 +100,26 @@ class TaskingPlan extends BaseModel {
...
+    let isValid = this.target_id && this.target_type && this.opens_at && this.due_at && this.isDueAfterOpen;
+
+    if(this.isNew || this.dueAtChanged) {
+      isValid = isValid && this.isBeforeDue;
+    }
+    //event does not have a close date (visually)
+    if(!this.plan.isEvent) {
+      isValid = isValid && this.isCloseAfterDue;
+    }
+
```
nathanstitt, line 121: It's a little confusing to me to have isValid being checked inside the if statement How about something like this:
```js
let isValid = ...

if(isValid && (this.isNew || this.dueAtChanged)) {
  isValid =  this.isBeforeDue;
}
if(isValid && !this.plan.isEvent) {
   isValid = this.isCloseAfterDue;
}

```

tutor/src/screens/assignment-edit/tasking.js
```diff
@@ -222,8 +220,9 @@ class Tasking extends React.Component {
               labelWrapper={ux.isShowingPeriodTaskings ? null : NewTooltip}
               name={`tasking_plans[${index}].closes_at`}
               onChange={this.onClosesChange}
-              disabledDate={this.course.isInvalidAssignmentDate}
+              disabledDate={disabledDate}
               timezone={this.course.timezone}
```
nathanstitt, line 224: Why'd this change?   `this.course.isInvalidAssignmentDate` should be doing the same logic I think?  If it's not correct we should update that method since that method is only used here.

Good catch on the bug where the prop was mistakenly called `disabled` 😁

camliu89: Yeah, I should have updated the logic in `isInvalidAssignmentDate`. The current logic does not disable dates before the course start date and after the course end date. I will update it.

tutor/src/models/task-plans/teacher/tasking.js
camliu89: Yeah that works too. I thought about "concatenate" the booleans as we do for strings.

tutor/src/models/task-plans/teacher/tasking.js
camliu89: good to know! :) thanks.

camliu89:
@nathanstitt updated the code with your comments. thanks!

tutor/src/models/course.js
```diff
@@ -208,11 +208,8 @@ class Course extends BaseModel {
...
   // bind to this so it can be used in disabledDate check
   isInvalidAssignmentDate = (date) => {
-    return !this.momentInZone(date).isBetween(
-      this.allowedAssignmentDateRange.start,
-      this.allowedAssignmentDateRange.end,
-      'day', '[]'
-    );
+    return date < moment(this.starts_at).endOf('day') ||
+    date > moment(this.ends_at).endOf('day');
   }
```
nathanstitt, line 213: @camliu89 one more change request - can you remove `allowedAssignmentDateRange` since it's not used any longer?

PR #3168, opened 2020-07-14
Fix new flag before due

jivey:
- Fixes new flag (ex. "3 NEW") on the dashboard showing up before the plan is past due (checks for the earliest due date for the past due check.)
- Fixes an issue with `dateRanges` returning an incorrect due time of `11:59` regardless of the correct value (locally `3:50`). I suspect this is an issue with the date time initialization on that field, and for some reason computed doesn't see that the value has changed. Removing the computed here fixes the issue, I don't think there's any real side effect here besides a very minor performance hit.

Will need a backend change to start filtering `ungraded_step_count` to only include past due.

### DIFF
--- a/tutor/src/models/task-plans/teacher/plan.js
+++ b/tutor/src/models/task-plans/teacher/plan.js
@@ -287,6 +287,7 @@ class TeacherTaskPlan extends BaseModel {
   }
   @computed get isFailed() { return Boolean(this.failed_at || this.killed_at); }
   @computed get isPastDue() { return this.duration.end.isBefore(Time.now); }
+  @computed get isGradeable() { return this.dateRanges.due.start.isBefore(Time.now); }
   @computed get isVisibleToStudents() { return this.isPublished && this.isOpen; }
 
   @computed get isPollable() {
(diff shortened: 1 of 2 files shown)


Dantemss:
I started trying to remove the gradable/ungraded steps column from the task_plan in the BE and just sum them from the tasking_plans which works and is probably better than what we have right now but doesn't account for assignment extensions (can't grade until the extension expires...)

So I'm trying to think what to do now... Maybe I can make use of the Glicko jobs that run when the task becomes due...

Dantemss:
I am doing some BE cleanup on the cache models first, but I'll probably put the changes here as well: https://github.com/openstax/tutor-server/pull/2144

jivey:
I'm moving this back to draft until I know for sure if we even need a FE change here.

Dantemss:
This will probably have to wait... I can work on it after code freeze...

Dantemss:
https://github.com/openstax/tutor-server/pull/2149

jivey:
In most cases we don't really this, but might be good to merge for an edge case: if an assignment has pending gradeable questions, and the due date is moved, but the background job is slow, going back to the dashboard will show a pending count that it shouldn't until the new due date is passed.

PR #3172, opened 2020-07-15
render no student message when showing a period without enrollments

jivey:
image

### DIFF
--- a/tutor/src/screens/assignment-review/index.js
+++ b/tutor/src/screens/assignment-review/index.js
@@ -1,5 +1,6 @@
 import { React, PropTypes, observer, styled, action, observable } from 'vendor';
 import { ScrollToTop } from 'shared';
+import { colors } from 'theme';
 import Courses from '../../models/courses-map';
 import Router from '../../helpers/router';
 import LoadingScreen from 'shared/components/loading-animation';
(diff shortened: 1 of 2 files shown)


tutor/src/screens/assignment-review/index.js
```diff
@@ -122,6 +121,34 @@ class AssignmentReview extends React.Component {
...
+    } = this.ux;
+
+    let body;
+
+    if (selectedPeriod && !selectedPeriod.hasEnrollments) {
+      body = <MessageWrapper><NoStudentsMessage courseId={course.id} /></MessageWrapper>;
+    } else if (!isScoresReady || !scores) {
+      return <LoadingScreen message="Loading Assignment…" />;
+    } else {
+      body = this.renderTabs({ ux: this.ux });
```
jivey, line 149: scores will be null if the period doesn't have students, so that's why this block might look strange. Also needed to wrap `NoStudentsMessage` in a wrapper since styles wouldn't attach 🤔

nathanstitt: Styles wouldn't attach because the `NoStudentsMessage` doesn't accept and use a `className` prop.  That's how styled-components "extends" the styles of another component, it injects a new class onto it.

jivey: Aha! I always assumed it had a way of adding it. This is super helpful for the future, thanks!

PR #3199, opened 2020-07-30
use single assignment edit button

nathanstitt:
Choose between popup details only editor and redirect to full editor depending on task plan open status.

Also load questions for overview tab from task plan so they'll always show up regardless of enrollment status

### DIFF
--- a/tutor/src/screens/assignment-review/index.js
+++ b/tutor/src/screens/assignment-review/index.js
@@ -1,6 +1,5 @@
 import { React, PropTypes, observer, styled, action, observable } from 'vendor';
 import { ScrollToTop } from 'shared';
-import { colors } from 'theme';
 import Courses from '../../models/courses-map';
 import Router from '../../helpers/router';
 import LoadingScreen from 'shared/components/loading-animation';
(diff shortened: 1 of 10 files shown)


tutor/src/screens/assignment-review/ux.js
```diff
@@ -274,21 +278,28 @@ export default class AssignmentReviewUX {
...
+
+      this.isDisplayingEditAssignment = true;
 
-    this.isDisplayingEditAssignment = true;
+    } else {
+      this.onEditAssignment();
+    }
   }
 
+  // taskPlan.isOpen ? undefined : ux.onEditAssignedQuestions}
```
camliu89, line 301: needs to be deleted?

tutor/src/screens/assignment-review/index.js
```diff
@@ -134,19 +125,26 @@ class AssignmentReview extends React.Component {
...
+        <StyledTabs
+          selectedIndex={0}
+          onSelect={this.onTabSelection}
+          tabs={['Details']}
+        />
+        <Details ux={this.ux} />
+      </>
+    );
+  }
+
```
camliu89, line 140: I don't see this function being used anywhere.

nathanstitt: yep good catch..  I'd added this with an earlier version before we decided to hide the "NoStudents" message.

PR #3201, opened 2020-07-31
support either - or . as LO separator

nathanstitt:
Fixes https://github.com/openstax/tutor-js/pull/3176 which got rebased out somehow


Brooke-Fail:
Verified in QA.

PR #3204, opened 2020-08-03
Student assignment instructions responsiveness

camliu89:
Mobile:
Screenshot from 2020-08-03 14-23-44

Tablet:
Screenshot from 2020-08-03 14-24-11

Disregard the student task step table. The responsiveness will be done in another PR.


tutor/src/theme.js
```diff
@@ -173,7 +173,7 @@ export const breakpoint = {
     return css`@media(max-width: 600px) { ${css(...args)} }`;
   },
   tablet(...args) {
-    return css`@media(max-width: 1199px) { ${css(...args)} }`;
+    return css`@media(max-width: 991px) { ${css(...args)} }`;
```
jivey, line 176: This probably needs to stay 1199px?

tutor/src/theme.js
```diff
@@ -183,13 +183,13 @@ export const breakpoint = {
...
-      return css`@media(max-width: 599px) { ${css(...args)} }`;
+      return css`@media(max-width: 600px) { ${css(...args)} }`;
     },
     tablet(...args) {
-      return css`@media(min-width: 600px) and max-width: 1199px) { ${css(...args)} }`;
+      return css`@media(min-width: 601px) and (max-width: 1199px) { ${css(...args)} }`;
     },
     desktop(...args) {
-      return css`@media(min-width: 1200px) and (max-width: 1599px) { ${css(...args)} }`;
+      return css`@media(min-width: 992px) and (max-width: 1599px) { ${css(...args)} }`;
```
jivey, line 192: Same here, desktop should start at 1200px.

tutor/src/theme.js
camliu89: :+1:

AUTHORS
```diff
@@ -8,3 +8,4 @@ Roy E Johnson
 Thomas Woodward <tom.g.woodward@gmail.com>
 Nathan Stitt <nathan@stitt.org>
 Brittany Johnson
+Alexander Cam Liu <acamliu89@gmail.com>
```
nathanstitt, line 11: Yay!

tutor/src/screens/task/exercise-task-header.js
```diff
@@ -95,9 +113,9 @@ class ExercisesTaskHeader extends React.Component {
           </TutorLink>
         </StyledBackLink>
         <StyledHeadingTitle>
-          <span>{ux.task.title}</span>
-          <span> | </span>
-          <span>Due {TimeHelper.toShortHumanDateTimeTz(ux.task.dueAtMoment)}</span>
+          <div><span>{ux.task.title}</span></div>
+          <div><span>|</span></div>
+          <div><span>Due {TimeHelper.toShortHumanDateTimeTz(ux.task.dueAtMoment)}</span></div>
         </StyledHeadingTitle>
```
nathanstitt, line 119: this seems strange to me.  Why wrap each span with a div?  Is it for the `div + div` css rule above?  Seems like we could either set the rules to target span or just remove the inner spans and use div.

tutor/src/screens/task/step/instructions.js
```diff
@@ -21,6 +21,16 @@ const Header = styled.h2`
   line-height: inherit;
   justify-content: space-between;
   align-items: center;
+
+  ${({ theme }) => theme.breakpoint.tablet`
+    > span > span {
+      display: none;
```
nathanstitt, line 27: since we control the html layout, how about adding a class to the element that can be hidden instead of writing nested rules like this.  That seems less fragile to me

tutor/src/screens/task/exercise-task-header.js
camliu89: true. I don't know why I left the span in there. :+1+

camliu89:
@nathanstitt thanks!

PR #3208, opened 2020-08-04
fix only.mobile and only.tablet breakpoint numbers

jivey:
Updated these to match the mobile and tablet number changes.

### DIFF
--- a/tutor/src/theme.js
+++ b/tutor/src/theme.js
@@ -186,7 +186,7 @@ export const breakpoint = {
       return css`@media(max-width: 600px) { ${css(...args)} }`;
     },
     tablet(...args) {
-      return css`@media(min-width: 601px) and max-width: 1199px) { ${css(...args)} }`;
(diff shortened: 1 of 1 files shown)


tutor/src/theme.js
```diff
@@ -183,10 +183,10 @@ export const breakpoint = {
   },
   only: {
     mobile(...args) {
-      return css`@media(max-width: 599px) { ${css(...args)} }`;
+      return css`@media(max-width: 600px) { ${css(...args)} }`;
     },
     tablet(...args) {
-      return css`@media(min-width: 600px) and max-width: 1199px) { ${css(...args)} }`;
+      return css`@media(min-width: 601px) and max-width: 1199px) { ${css(...args)} }`;
```
camliu89, line 189: @jivey it is missing a `(` right before `max-width`. It should be `and (max-width: 1199px).`

jivey: Ohhh good 👁️

PR #3210, opened 2020-08-05
Gradebook show students before due

camliu89:
Requires BE changes: https://github.com/openstax/tutor-server/pull/2165

Screenshot from 2020-08-05 17-37-10


tutor/src/screens/teacher-gradebook/table.js
```diff
@@ -14,10 +14,15 @@ import { getCell } from './styles';
 import AverageInfoModal from './average-info-modal';
 import SetWeightsModal from './set-weights-modal';
 
+const mainDivCSS = css`
+    max-height: 60vh;
+    min-height: auto;
+    margin: 2.2rem 0 1.4rem;
+}
+`;
```
nathanstitt, line 22: On some displays 40vh might not be enough to fit the extra elements on screen.  How about setting this to a calculated height that's 100% - (heading and controls height), something like `calc(100vh - 400px);`

That might now work on mobile though, might need tweaked there

PR #3215, opened 2020-08-10
fix hour scroll not working in firefox

jivey:
There's a difference in overflow implementation with flexbox between Firefox and Chrome: https://bugzilla.mozilla.org/show_bug.cgi?id=1108514#c2

I can't fix this with the `min-height: 0` trick because we don't easily control the html render coming from the date time component.

I'm targeting Firefox with `supports` instead of using `data-browser` because it's set on the wrapper div not the body (this popup is attached to body.)


camliu89:
it is working for me using Firefox. nice!

PR #3227, opened 2020-08-19
Highlights notes responsiveness

camliu89:
Screenshot from 2020-08-19 17-37-47

Screenshot from 2020-08-19 17-38-18


jivey:
I'm trying to test this but I'm running into an issue. When viewing on a phone, I get this when tapping on an existing highlight:

> Invariant Violation: Responsive(...): Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.

Maybe related, this warning came up right before:

> Warning: Failed prop type: EditBox: prop type `textAreaDisabled` is invalid; it must be a function, usually from the `prop-types` package, but received `undefined`.

camliu89:
@jivey , mmm let me check on that.

camliu89:
@jivey nice find. I was missing the `mobile` prop for the `Responsive` component. It is ready again for review.

tutor/src/components/notes/sidebar-buttons.jsx
```diff
@@ -69,8 +69,17 @@ class SidebarButtons extends React.Component {
...
+  // Add a resize dummy listener to listen again and get the new container rect specs
+  @observable fakeRender = false;
+
+  componentDidMount() {
+    window.addEventListener('resize', () => this.fakeRender = !this.fakeRender);
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('resize', () => this.fakeRender = !this.fakeRender);
+  }
```
nathanstitt, line 83: instead of using manual listeners use window size model, it auto listens and unlistens and is what the responsive component is built on. `https://github.com/openstax/tutor-js/blob/main/tutor/src/models/window-size.js`. If you rewrite `getRangeRect` to accept the window size from that it should re-render whenever the window size changes.  It also debounces so you don't re-render a 1000 times on any little resize

PR #3229, opened 2020-08-24
Fix tour on QL cards

nathanstitt:
Discovered that 'el' getter was broken and exercise card class was incorrect


nathanstitt:
yes, I don't know what happened to actions.  they pass locally tho so will merge

PR #3236, opened 2020-08-27
change tablet breakpoint to 1000px but keep it to 1200px for the read…

camliu89:
…ing arrows pagination

### DIFF
--- a/tutor/src/screens/task/step/exercise.js
+++ b/tutor/src/screens/task/step/exercise.js
@@ -81,16 +81,17 @@ class ExerciseTaskStep extends React.Component {
   render() {
     const { ux, step, isMultiPart, isFollowupMPQ } = this.props;
     const { content } = step;
+
     return (
       <>
         <StyledExercise
           step={step}
           questionNumber={ux.questionNumberForStep(step)}
           numberOfQuestions={ux.exerciseSteps ? ux.exerciseSteps.length : 0}
-          goBackward={ux.goBackward}
-          canGoBackward={ux.canGoBackward}
-          goForward={ux.goForward}
-          canGoForward={ux.canGoForward}
+          goBackward={() => ux.goToStep(ux.previousStep)}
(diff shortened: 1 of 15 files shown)


tutor/src/screens/task/step/exercise.js
```diff
@@ -81,16 +81,17 @@ class ExerciseTaskStep extends React.Component {
...
           questionNumber={ux.questionNumberForStep(step)}
           numberOfQuestions={ux.exerciseSteps ? ux.exerciseSteps.length : 0}
-          goBackward={ux.goBackward}
-          canGoBackward={ux.canGoBackward}
-          goForward={ux.goForward}
-          canGoForward={ux.canGoForward}
+          goBackward={() => ux.goToStep(ux.previousStep)}
+          canGoBackward={Boolean(ux.previousStep)}
+          goForward={() => ux.goToStep(ux.nextStep)}
+          canGoForward={Boolean(ux.nextStep)}
```
nathanstitt, line 95: why'd this change?  I don't see any corresponding changes to the ux class 🤔

camliu89: I found a bug that was not showing the forward arrow if you haven't completed the step. The `canGoForward` checks if the current step is completed. But in this case, we don't really want to check if the step is completed or not, we just want to move to the next question. So here I am just checking if there is any `previousStep` and `nextStep` and just call the `goToStep`, same as we do in the task-progress table.

nathanstitt: Hm.  I don't think that's a bug.  I seem to recall that we don't allow navigating forward until steps are completed. I'll 🅿️ so we can discuss

camliu89: mmm. I think you are right. Then I was confused because I was able to navigate to a step in the task-progress table.

PR #3254, opened 2020-08-31
minor fixes in add details step

camliu89:
- In events and external assignment builder, the 'Add Details' label should be left-aligned with the text below.
- When adding an event, in the 'Add details' step, the label for the first text box should be Event name (and not Assignment name)
- When adding an External assignment, the page header in the builder should be Add external assignment
- When reviewing a draft assignment, the Delete button should be a white default button (like the 'Cancel' button next to it)


jivey (approved):
@camliu89 Looks good over here!

PR #3260, opened 2020-09-01
Only display bars that have "real" clues

nathanstitt:
Don't use the placeholder value on student dashboard


camliu89:
In the case where an assignment has all MCQs, do we show the PF with data if the student answered at least one question?

nathanstitt:
Yes I believe the rule is that after a student has 3 questions in a section graded the clue becomes "real".  MCQ are graded immediately so those bars show up right away but WRM won't appear until teacher grades them

PR #3261, opened 2020-09-01
creating a new exercise has id as undefined, use isCorrect to check w…

camliu89:
- new distractors default to incorrect - no check mark in the right-side preview
- "Save Draft" fails if none of the distractors has been marked "correct" by clicking the circled checkmark
- once a distractor has been marked correct, only that distractor has a green check mark next to it in the preview
- that distractor, and only that distractor continues to be marked "correct" upon save and publish


jivey (approved):
Wasn't able to test locally, but seems straightforward from the diff.

PR #3268, opened 2020-09-03
Fix highlight modal responsiveness

camliu89:


tutor/resources/styles/components/notes/summary-page.scss
```diff
@@ -2,11 +2,34 @@
   // Force fixed modal-scroll-top button to attach to modal
   // by using a no-effect transform.
   transform: rotate(0);
+  // make the modal biggers
+  @media screen and ( max-width: $mobile-collapse-breakpoint ){
+    margin: 1% auto;
+    width: 95%;
```
jivey, line 8: Could do something like `calc(100vw - #{$mobile-margin} * 2);` if you wanted to keep the modal aligned and covering the content container below

tutor/resources/styles/components/notes/summary-page.scss
```diff
@@ -220,7 +252,13 @@
             width: 100%;
           }
         }
+
+        img {
+          @media screen and ( max-width: $mobile-collapse-breakpoint ){
+            width: 85%;
```
jivey, line 258: Looks good in the content I tested, but should we make this `max-width` instead so it doesn't enlarge smaller images?

jivey (approved):
Looks good 👍  left a couple of thoughts.

tutor/resources/styles/components/notes/summary-page.scss
nathanstitt: yes, using % is probably fine since this is inside the mobile breakpoint and 1% won't ever grow to large.

nathanstitt (approved):

tutor/resources/styles/components/notes/summary-page.scss
camliu89: thanks nathan 👍

tutor/resources/styles/components/notes/summary-page.scss
camliu89: you are right @jivey

PR #3272, opened 2020-09-08
External event instructions responsiveness

camliu89:
Mobile:
Annotation 2020-09-08 104132

Tablet:
Annotation 2020-09-08 104147

### DIFF
--- a/tutor/src/screens/task/step/instructions.js
+++ b/tutor/src/screens/task/step/instructions.js
@@ -1,4 +1,4 @@
-import { React, PropTypes, observer, styled } from 'vendor';
+import { React, PropTypes, observer, styled, css } from 'vendor';
 import TutorLink from '../../../components/link';
 import { colors } from '../../../theme';
 import { OuterStepCard, InnerStepCard } from './card';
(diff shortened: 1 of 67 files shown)


tutor/src/screens/task/step/instructions.js
```diff
@@ -9,6 +9,12 @@ import Time from '../../../components/time';
 
 const CardBody = styled(InnerStepCard)`
   padding-bottom: 5rem;
+  /* reading assignment has a sticky navigation bar at the bottom */
+  /* all other assignments, make this div to fill up the screen (the header is 20vh) */
+  ${({ theme, taskType }) => taskType !== 'reading' && theme.breakpoint.only.mobile`
+    height: 80vh;
```
jivey, line 15: I might be thinking of the wrong screen here, but you could use `calc` here to subtract 100vh from `navbars.top` + `navbars.readingSubnav` from the theme vars just in case those change in the future.

jivey (approved):

tutor/src/screens/task/step/instructions.js
camliu89: it works! thanks for the advice :)

PR #3274, opened 2020-09-08
Fix request refund button

nathanstitt:
"refunding" must be observable Otherwise the component won't re-render when its set


tutor/src/components/payments/manage.jsx
```diff
@@ -147,7 +147,7 @@ class ManagePayments extends React.Component {
     }),
   };
 
-  @observable refundi
+  @observable refunding;
```
camliu89, line 150: 🦅 👁️

camliu89 (approved):

PR #3279, opened 2020-09-10
add metrics to pulse insight data

nathanstitt:
reports data:
```
{
  "role": "instructor",   // uses "isProbablyTeacher" which combines can_create_courses and self reported role.  It's what tours use
  "is_new_user": true, // true if no past courses
  "is_test_user": true, // from admin settings
  "has_viewed_preview": false, // true if they have preview courses associated 
  "courses": [
    {
      "subject": "biology_2e",  // this is the appearance code
      "type": "real",  // or preview
      "role": "teacher", // or student
      "active": true,  // or preview
      "enrollment": 17, // how many students are a member. if user is a student it will be -1 since we don't know
      "enrollment_type": "links", // or las
      "cost": false. // or true 
    },
    {
      "subject": "college_physics",
      "type": "real",
      "role": "student",
      "active": true,
      "enrollment": -1,
      "enrollment_type": "links",
      "cost": false
    }
  ]
}
```

### DIFF
--- a/tutor/src/models/app.js
+++ b/tutor/src/models/app.js
@@ -47,7 +47,7 @@ export default class TutorApp {
     await documentReady();
 
     const app = new TutorApp();
-    [Raven, PulseInsights, Api].forEach(lib => lib.boot());
+    [Raven, Api].forEach(lib => lib.boot({ app }));
 
     app.data = readBootstrapData();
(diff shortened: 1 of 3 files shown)


camliu89 (approved):
looks straightforward to me. 👍

PR #3280, opened 2020-09-10
add google tag manager

nathanstitt:
Copies what BIT did.

### DIFF
--- a/tutor/src/helpers/analytics.js
+++ b/tutor/src/helpers/analytics.js
@@ -64,12 +64,30 @@ const Translators = {
   },
 };
 
+const TAG_MANAGER_ID = 'GTM-W6N7PB';
 
 let GA = undefined;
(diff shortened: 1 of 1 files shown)


tutor/src/helpers/analytics.js
```diff
@@ -64,12 +64,30 @@ const Translators = {
   },
 };
 
+const TAG_MANAGER_ID = 'GTM-W6N7PB';
```
camliu89, line 67: is it ok to expose this tag? do we have any environment variables?

camliu89 (approved):

tutor/src/helpers/analytics.js
nathanstitt: yah it's ok to expose it since it's included in every post to google.  I thought about making a config for it but doubt it'll ever change so decided to just do this

PR #3281, opened 2020-09-10
Responsive welcome screen

jivey:
This also skips the new student tour after the initial "Welcome" modal. Steps can be skipped by breakpoint by adding "mobile", "tablet" or "desktop" to a disabledBreakpoints array in the step, which will cause isViewable to be false.

Mobile | Tablet | Desktop
----- | ----- | -----
image| image | image


camliu89 (approved):
nice!

PR #3290, opened 2020-09-15
hide payment bar when user is taking an assignment

camliu89:
Screen Shot 2020-09-15 at 8 59 58 AM

Screen Shot 2020-09-15 at 9 00 35 AM

Payment error fix:

Screen Shot 2020-09-15 at 1 15 48 PM

### DIFF
--- a/tutor/src/components/tutor-layout.js
+++ b/tutor/src/components/tutor-layout.js
@@ -117,13 +117,15 @@ class TutorLayout extends React.Component {
     this.courseContext.course = this.props.course;
   }
 
-  /**
-   * Hide the navbar if user is in the 'viewTaskStep' screen.
-   * Use styled-components to check if the screen is mobile width.
-   */
-  isUserInViewTaskStep() {
+  isViewingTaskStep() {
     const routerName = get(Router.currentMatch(), 'entry.name', '');
     if(routerName === 'viewTaskStep') return true;
+    return false;
+  }
+
+  isViewingCourseDashboard() {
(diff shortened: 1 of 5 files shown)


nathanstitt (commented):
This will hide the payments on the task steps, but what about in grade book, performance forecast or other non-dashboard screens?  I think we want to hide it everywhere except the course dashboard, right?

If that's the case, how about changing the logic to be isViewingCourseDashboard or something like that?

camliu89:
Yes, I think we did decide to hide that bar everywhere except on the dashboard. I forgot about that. thanks, Nathan.

tutor/src/components/tutor-layout.js
```diff
@@ -117,16 +117,18 @@ class TutorLayout extends React.Component {
...
     const routerName = get(Router.currentMatch(), 'entry.name', '');
     if(routerName === 'viewTaskStep') return true;
     return false;
   }
 
+  isViewingCourseDashboard() {
+    const routerName = get(Router.currentMatch(), 'entry.name', '');
+    if(routerName === 'dashboard') return true;
+    return false;
+  }
```
nathanstitt, line 130: no big thing but I usually prefer to just return the results of the condition since it'll be a boolean:

```return routerName === 'dashboard';```

nathanstitt (approved):

tutor/src/components/tutor-layout.js
camliu89: yep, it is more readable. thanks.

PR #3299, opened 2020-09-17
Assignment closed message

camliu89:
Add banner to show to the students that the assignment is closed. 
Also, added logic to enable the review and next button so users can navigate to the next step.

Screen Shot 2020-09-17 at 1 57 31 PM

Screen Shot 2020-09-17 at 1 57 42 PM


.idea/.name
```diff
@@ -0,0 +1 @@
+ms-SG.js
```
nathanstitt, line 1: I see you're an Jetbrains user?

nathanstitt (approved):
looks great @camliu89

.idea/.name
camliu89: oh haha. I was just testing Jetbrains. Actually, I am VSCode user :)

PR #3300, opened 2020-09-17
Fix mobile verify text

jivey:
Fixes not displaying the full verify email text on desktop.

### DIFF
--- a/shared/resources/styles/components/notifications-bar.scss
+++ b/shared/resources/styles/components/notifications-bar.scss
@@ -32,6 +32,10 @@
       flex: 1;
     }
 
+    .mobile-text {
+      display: none;
+    }
+
     & + .notification:not(.acknowledged) {
       border-top: 1px solid darken($openstax-primary, 5%);
     }
--- a/shared/src/model/notifications/pollers.js
+++ b/shared/src/model/notifications/pollers.js
@@ -130,7 +130,11 @@ class TutorUpdates extends Poller {
 
 
 class AccountsNagger extends Poller {
-  constructor(type, notices) { super(type, notices, moment.duration(1, 'day')); }
+  constructor(type, notices) {
+    super(type, notices, moment.duration(1, 'day'));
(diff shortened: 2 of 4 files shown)


camliu89 (approved):

shared/resources/styles/components/notifications-bar.scss
```diff
@@ -32,6 +32,10 @@
       flex: 1;
     }
 
+    .mobile-text {
+      display: none;
+    }
+
```
camliu89, line 38: looks good. I was wondering if we can put this in the `global/notifications-bar.scss` file?

jivey: That's there in the shared one so it won't show up on the exercises screens. I thought about putting it all into shared, but I don't think it gets the responsive vars

shared/src/model/notifications/pollers.js
```diff
@@ -130,7 +130,11 @@ class TutorUpdates extends Poller {
 
 
 class AccountsNagger extends Poller {
-  constructor(type, notices) { super(type, notices, moment.duration(1, 'day')); }
+  constructor(type, notices) {
+    super(type, notices, moment.duration(1, 'day'));
+    // uncomment FOR DEBUGGING to get notification bar to show up
+    // this.onReply({ data: { contact_infos: [ { id: -1, is_verified: false } ] } });
+  }
```
jivey, line 137: thanks @nathanstitt for this :D

tutor/resources/styles/global/notifications-bar.scss
```diff
@@ -32,6 +32,13 @@
...
         padding: 4px 8px;
       }
+
+      .desktop-text {
+        display: none;
+      }
+      .mobile-text {
+        display: inline;
+      }
     }
```
nathanstitt, line 42: hm, where's the actually css media rule that switches the text display?

nathanstitt (approved):

PR #3302, opened 2020-09-21
Show sections assigned in the plan details modal

camliu89:
Screen Shot 2020-09-21 at 12 51 43 PM

Screen Shot 2020-09-21 at 12 51 53 PM

### DIFF
--- a/tutor/src/models/task-plans/teacher/plan.js
+++ b/tutor/src/models/task-plans/teacher/plan.js
@@ -365,7 +365,7 @@ class TeacherTaskPlan extends BaseModel {
   }
 
   @computed get assignedSections() {
-    return this.course.referenceBook.pages.all.filter(page => this.pageIds.includes(page.id));
+    return compact(this.pageIds.map(pgId => this.course.referenceBook.pages.byId.get(pgId)));
   }
 
   @action addExercise(ex) {
--- a/tutor/src/models/task-plans/teacher/plan.js
+++ b/tutor/src/models/task-plans/teacher/plan.js
@@ -364,6 +364,10 @@ class TeacherTaskPlan extends BaseModel {
     return get(this, 'settings.page_ids', []);
   }
 
+  @computed get assignedSections() {
+    return this.course.referenceBook.pages.all.filter(page => this.pageIds.includes(page.id));
+  }
+
   @action addExercise(ex) {
     if (!this.exerciseIds.find(exId => exId == ex.id)) {
       this.settings.exercises.push({
--- a/tutor/src/screens/question-library/index.jsx
+++ b/tutor/src/screens/question-library/index.jsx
@@ -19,7 +19,8 @@ class QuestionsDashboardShell extends React.Component {
   }
 
   componentDidMount() {
-    this.course.referenceBook.fetch();
+    if (!this.course.referenceBook.uuid)
+      this.course.referenceBook.fetch();
   }
 
   render() {
@@ -27,4 +28,4 @@ class QuestionsDashboardShell extends React.Component {
     return <Dashboard course={this.course} />;
   }
 
-};
+}
--- a/tutor/src/screens/question-library/index.jsx
+++ b/tutor/src/screens/question-library/index.jsx
@@ -19,8 +19,9 @@ class QuestionsDashboardShell extends React.Component {
   }
 
   componentDidMount() {
-    if (!this.course.referenceBook.uuid)
+    if (!this.course.referenceBook.ensureLoaded()) {
       this.course.referenceBook.fetch();
+    }
   }
 
   render() {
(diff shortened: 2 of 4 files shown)


tutor/src/models/task-plans/teacher/plan.js
```diff
@@ -364,6 +364,10 @@ class TeacherTaskPlan extends BaseModel {
     return get(this, 'settings.page_ids', []);
   }
 
+  @computed get assignedSections() {
+    return this.course.referenceBook.pages.all.filter(page => this.pageIds.includes(page.id));
+  }
```
nathanstitt, line 369: How about `return compact(his.pageIds.map(pgId => this.course.referenceBook.pages.byId.get(pgId))`

That will be sorted correctly since sometimes the teachers do re-order sections on an assignment.

I also think that'll be a bit more efficient since the pages.byId is pre-computed and that won't need to look at every single page id.  Some books are fairly large and have a lot of sections.

I'm not sure if the lodash `compact()` is really needed but there might be some cases where a page isn't found.

tutor/src/screens/question-library/index.jsx
```diff
@@ -19,12 +19,13 @@ class QuestionsDashboardShell extends React.Component {
   }
 
   componentDidMount() {
-    this.course.referenceBook.fetch();
+    if (!this.course.referenceBook.uuid)
+      this.course.referenceBook.fetch();
   }
```
nathanstitt, line 24: You can use `this.course.referenceBook.ensureLoaded()` here.  `ensureLoaded` is part of the base model so it should be on everything

tutor/src/screens/teacher-dashboard/index.jsx
```diff
@@ -89,6 +89,8 @@ class TeacherDashboardWrapper extends React.Component {
     if (course.currentRole.isTeacherStudent && course.roles.teacher) {
       course.current_role_id = null;
     }
+    // getting the reference book data to show chapters and sections in the plan details
+    this.props.course.referenceBook.fetch();
   }
```
nathanstitt, line 94: probably should use `ensureLoaded` here as well since there's no reason to fetch it if it's already loaded.  The book contents never changes.

tutor/src/screens/teacher-dashboard/plan-details.jsx
```diff
@@ -3,8 +3,9 @@ import {
   computed, observable, action, styled,
 } from 'vendor';
 import { first } from 'lodash';
+import cn from 'classnames';
+import { Modal, Row, Col, OverlayTrigger, Tooltip, Button } from 'react-bootstrap';
```
nathanstitt, line 7: `cn` is exported by `vendor` as well ;)

tutor/src/screens/teacher-dashboard/plan-details.jsx
```diff
@@ -245,6 +271,27 @@ class CoursePlanDetails extends React.Component {
...
+  // Show only the first two with a button to show the rest of the sections.
+  renderAssignedSections() {
+    const { plan: { assignedSections } } = this.props;
+
+    if(!assignedSections) return null;
+    return (
+      <SectionsAssignedWrapper>
+        <h6>Sections Assigned</h6>
+        {assignedSections.map((section, i) => 
+          <ul key={i}>
```
nathanstitt, line 284: not super important but its recommended to not use an iterator as `key` unless there's nothing better available.  In this case we could use `section.pathId`

nathanstitt (changes requested):

tutor/src/models/task-plans/teacher/plan.js
camliu89: this is good. it is faster and it is not going through all the pages in a book. 

I think we might need `compact` just in case a weird falsey value gets returned.

tutor/src/screens/question-library/index.jsx
camliu89: @nathanstitt the `ensureLoaded` is working, but it is Promise. I wonder how is that getting evaluated to a Boolean 🤔

nathanstitt: @camliu89 sorry I was unclear here.  the `.ensureLoaded()` method will load the model if it's not already been loaded.  So you don't need an if at all, it could just be:
```
componentDidMount() {
     this.course.referenceBook.ensureLoaded()
}
```

As you noted it does return a promise but that's just so you can wait on it if you need to.  In the already loaded case it returns just `Promise.resolve()`

tutor/src/screens/teacher-dashboard/index.jsx
```diff
@@ -89,6 +89,11 @@ class TeacherDashboardWrapper extends React.Component {
     if (course.currentRole.isTeacherStudent && course.roles.teacher) {
       course.current_role_id = null;
     }
+
+    // getting the reference book data to show chapters and sections in the plan details
+    if (!course.referenceBook.ensureLoaded()) {
+      course.referenceBook.fetch();
+    }
   }
```
nathanstitt, line 97: same here, you can just put a call to `course.referenceBook.ensureLoaded()`,no need for the if

nathanstitt (approved):

PR #3303, opened 2020-09-22
Scroll to top button

jivey:
Adds a scroll to top button on the teacher side once the user has scrolled more than 4 times the browser height.

image

image

image


nathanstitt (changes requested):
@jivey I see we have a super outdated "scroll-tracker" that does pretty much you're doing here in components/scroll-tracker.js

I also don't see anywhere that it's used, so I don't think we need to keep any of it's functionality intact.  How about rewriting ScrollTracker as a hook and using that here?

I see https://dev.to/n8tb1t/tracking-scroll-position-with-react-hooks-3bbj but we could probably use debounce vs the throttle func there.  there's probably also a bunch of alternatives

tutor/src/components/scroll-tracker.js
```diff
@@ -1,127 +1,20 @@
...
-    }
-  },
-
-  scrollToKey(stepKey) {
-    const scrollState = this.getScrollStateByKey(stepKey);
-    return window.scrollTo(0, ((scrollState != null ? scrollState.scrollPoint : undefined) - this.state.scrollTopBuffer));
-  },
+    window.addEventListener('scroll', debounce(_onScroll, 10), { passive: true });
+    return () => window.removeEventListener('scroll', _onScroll);
+  });
```
nathanstitt, line 17: You'll need to use the exact same function on both addEvent and removeEvent, otherwise removeEvent won't work.

To do so you can move the debounce up like so: `const _onScroll => debounce( () => { ... } )`

nathanstitt (approved):

PR #3304, opened 2020-09-22
500 word limit

camliu89:
1. Show red border color in the response text area input.
2. Show word info error message.
3. Disable Submit button.

Screen Shot 2020-09-22 at 1 39 29 PM

### DIFF
--- a/tutor/src/screens/task/step/exercise-free-response.js
+++ b/tutor/src/screens/task/step/exercise-free-response.js
@@ -181,7 +181,7 @@ class FreeResponseInput extends React.Component {
             <AnswerButton
               size="lg"
               data-test-id="submit-answer-btn"
-              disabled={ux.isSubmitDisabled || ux.isOverWordLimit}
+              disabled={ux.isSubmitDisabled}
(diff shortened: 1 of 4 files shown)


tutor/src/screens/task/step/exercise-free-response.js
```diff
@@ -171,7 +181,7 @@ class FreeResponseInput extends React.Component {
             <AnswerButton
               size="lg"
               data-test-id="submit-answer-btn"
-              disabled={ux.isSubmitDisabled}
+              disabled={ux.isSubmitDisabled || ux.isOverWordLimit}
               onClick={this.onSave}>
```
nathanstitt, line 185: rest of pr looks great, but how about changing the `isSubmitDisabled` in ux to check the word count itself that way it's controlling the logic

nathanstitt (approved):

tutor/src/screens/task/step/exercise-free-response.js
camliu89: :+1:

PR #3306, opened 2020-09-23
Disable save draft button

camliu89:
Disable if:

1. No title is given to the assignment
2. Dates are incorrect
3. Assignment is published

### DIFF
--- a/tutor/specs/integration/assignment-edit.spec.js
+++ b/tutor/specs/integration/assignment-edit.spec.js
@@ -249,15 +249,16 @@ context('Assignment Edit', () => {
     })
   });
 
-  it('shows error if due date is before open date', () => {
+  it.only('shows error if due date is before open date', () => {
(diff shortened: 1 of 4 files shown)


tutor/specs/integration/assignment-edit.spec.js
```diff
@@ -249,15 +249,16 @@ context('Assignment Edit', () => {
     })
   });
 
-  it('shows error if due date is before open date', () => {
+  it.only('shows error if due date is before open date', () => {
     typedDueDate = moment().subtract(1, 'weeks').format('MMM D [| 05:00 PM]')
```
nathanstitt, line 253: I think you probably left the `.only` on by mistake?

nathanstitt (approved):
looks good, but take the .only off unless it was intended for some reason

tutor/specs/integration/assignment-edit.spec.js
camliu89: yes :) thanks for catching it.

PR #3308, opened 2020-09-24
Fix grading block buttons

camliu89:
Fixes the grading block button that was not doing anything when they were clicked.


nathanstitt (approved):
nice catch!

PR #3312, opened 2020-09-25
OpenEnded questions must be of type practice

nathanstitt:
Disable changing the controls when doing so would make question invalid

image

### DIFF
--- a/shared/src/model/exercise.js
+++ b/shared/src/model/exercise.js
@@ -100,8 +100,4 @@ class Exercise extends BaseModel {
     this.tags.findOrAddWithType('type').value = 'practice';
   }
 
-  @computed get isTypeLocked() {
-    return Boolean(this.isSinglePart && this.isOpenEnded);
-  }
-
 }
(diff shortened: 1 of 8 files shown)


shared/src/model/exercise.js
```diff
@@ -95,4 +95,15 @@ class Exercise extends BaseModel {
   @computed get isPublishable() {
     return Boolean(!this.isNew && this.validity.valid && !this.published_at);
   }
+
+  @action onQuestionFreeResponseSelected() {
+    if (this.isMultiPart) { return; }
```
Dantemss, line 100: This return seems odd to me. Are multipart WRQs allowed elsewhere?

shared/src/model/exercise/question.js
```diff
@@ -41,6 +41,16 @@ class ExerciseQuestion extends BaseModel {
 
   @belongsTo({ model: 'exercise' }) exercise;
 
+  @computed get allowedFormatTypes() {
+    if (this.exercise.isSinglePart) {
```
Dantemss, line 45: Same

Dantemss (commented):
See comments

shared/src/model/exercise.js
nathanstitt: yeah it also seemed odd to me but that's exactly what the card read and that's why I asked about practice 😁

Dantemss (approved):

PR #3316, opened 2020-09-25
show exercise tags on question, fix course id link

jivey:
image


jivey:
Based on a convo we only want DOK, Blooms, and APLO (for AP courses.)

camliu89 (approved):
nice!

PR #3317, opened 2020-09-25
Display course not found warning

nathanstitt:
Also updates the message to address 2 cases if they're not a member of course, or maybe they are a member but the material is sensitive.

image

image

### DIFF
--- a/tutor/src/screens/task/practice.js
+++ b/tutor/src/screens/task/practice.js
@@ -34,7 +34,7 @@ class TaskPractice extends React.Component {
 
   render() {
     if (!this.course) {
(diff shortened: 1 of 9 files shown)


tutor/src/screens/task/practice.js
```diff
@@ -34,7 +34,7 @@ class TaskPractice extends React.Component {
 
   render() {
     if (!this.course) {
-      return <CourseNotFoundWarning area="assignment" />;
+      return <CourseNotFoundWarning area="assignment" messageType="notAllowed" />;
```
camliu89, line 37: should the prop be `areaName`?

nathanstitt: yep! it sure should be.  thanks!

PR #3318, opened 2020-09-28
add min width to progress bar based on char display length

jivey:
Set a minimum width + increase it based on how many digits are being displayed. Fixes a bug where a large class size to task completion ratio would cause the completion bar to not be wide enough.

image
image
image


tutor/src/screens/assignment-review/grading-block.js
```diff
@@ -102,6 +102,9 @@ const ChartItem = styled.div`
   &:last-child { border-width: 1px 1px 1px 0; }
   &:only-child { border-width: 1px; }
   width: ${props => props.width}%;
+  min-width ${props => css`
+    calc(1rem + 1rem * ${props.value.toString().length});
+  `};
```
camliu89, line 107: nice. 👍

camliu89 (approved):

PR #3322, opened 2020-10-01
fix total scores in RD assignment

camliu89:
Before due date:
Screen Shot 2020-10-01 at 12 11 06 PM

After due date:
Screen Shot 2020-10-01 at 12 11 41 PM


jivey (approved):
Good fix. This makes me wonder if we should change the ScoresHelper to return unworked if passed null values 🤔

camliu89:
@jivey actually, that's a really good idea. But we should do it in another PR as it might affect a lot of places.

PR #3325, opened 2020-10-06
fill screen with payments modal on mobile

nathanstitt:
This way it has enough room to display buttons.

Goes along with https://github.com/openstax/ospayments/pull/221

image


jivey (approved):
Ah good catch, I must have missed copying over some browser styles!

PR #3337, opened 2020-10-21
add flex-flow column for gradebook table container

camliu89:
Screen Shot 2020-10-21 at 10 47 28 AM

### DIFF
--- a/tutor/src/screens/teacher-gradebook/index.js
+++ b/tutor/src/screens/teacher-gradebook/index.js
@@ -23,6 +23,7 @@ const titleControls = (ux) => {
 
 const StyledTourRegion = styled(TourRegion)`
   display: flex;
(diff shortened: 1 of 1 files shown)


tutor/src/screens/teacher-gradebook/index.js
```diff
@@ -23,6 +23,7 @@ const titleControls = (ux) => {
 
 const StyledTourRegion = styled(TourRegion)`
   display: flex;
+  flex-flow: column;
 `;
```
nathanstitt, line 27: Usually I see flex-flow with 2 arguments, like `flex-flow: column wrap`, since you're not specifying the 2nd argument for wrap behavior, this is equivalent to `flex-direction: column`

nathanstitt (approved):

PR #3341, opened 2020-10-26
Fix loading exercises

nathanstitt:
The old method would fail wit the new script tags, due to the doc already being ready when the script executed.


camliu89 (commented):
it is working for me 👍

camliu89 (approved):

PR #3345, opened 2020-10-29
make the div of the assignment event row clickable

camliu89:


tutor/src/screens/student-dashboard/event-row.jsx
```diff
@@ -143,7 +143,9 @@ class EventRow extends React.Component {
     const { event, course } = this.props;
     return (
       <>
-        <MobileCell>
+        <MobileCell
+          href={Router.makePathname('viewTask', { courseId: course.id, id: event.id })}
+          onClick={this.onClick}>
           <div className="assignment-title-icon">
```
nathanstitt, line 149: we need to make `MobileCell` be a `a` tag for accessibility   I see that `MobileCell` is only being used here, so we can rename it to `MobileRow` and have it be a styled.a tag

tutor/src/screens/student-dashboard/cells.js
```diff
@@ -54,9 +55,12 @@ const ScoreCell = styled(Cell).withConfig({
   }
 `;
 
-const MobileCell = styled.div`
+const MobileRow = styled.div.attrs({
+  as: 'a',
+})`
```
nathanstitt, line 60: heh, you can just do: `styled.a` here instead of making it a div and then switching to `a` 😁

nathanstitt (approved):

PR #3346, opened 2020-10-29
Make upper rows go to task, don't wrap scores link

nathanstitt:
Only the upper rows should go to the task, if the entire row is wrapped in a click handler then it block clicking the score link to grade book

image


camliu89:
@nathanstitt 👍  I totally forgot about handling the score link which is probably the reason why this bug was introduced.

tutor/specs/integration/student-dashboard.spec.js
```diff
@@ -1,16 +1,27 @@
...
   });
+
+  it('loads assignments', () => {
+    withScreenSizes(() => {
+      setupDashboard();
+      cy.get('.task.homework').first().click()
+      cy.getTestElement('student-task').should('exist')
+    })
+  })
 });
```
camliu89, line 27: thanks for adding the tests.

camliu89 (approved):

PR #3348, opened 2020-10-29
do not show the save practice button in wrq when assignment is closed

camliu89:


jivey (approved):

tutor/src/screens/task/step/exercise-question.js
```diff
@@ -183,7 +183,7 @@ class ExerciseQuestion extends React.Component {
           {step.canAnswer && this.needsSaved ?
             this.renderSaveButton() : this.renderNextButton()}
           {
-            ux.task && !ux.task.isSavedPractice &&
+            ux.task && !ux.task.isPractice && !step.isWrittenResponseExercise &&
             <SavePracticeButton
```
nathanstitt, line 187: We'd decided that the save practice button should appear on practice worst and practice by section, but should not appear on saved practice.  To continue that we need to test for the ` !ux.task.isSavedPractice`

camliu89: I merged it. I can revert it back.

PR #3351, opened 2020-11-02
Fix save practice

camliu89:
- During practice quiz: When a student has completed all steps (reaches the "you are done" page), the "exit practice" button should not be displayed

- During assignment: "Save to practice" button should have a white background

- During assignment: When a student saves any part of a MPQ: On the other parts, show a tooltip when student hovers over the ‘Remove from practice’ button- “All parts get saved in a multi-part question“

- In MPP: "Exclude question" should have a dark grey background (#424242)
 Selecting in MPP: The white and blue checkmark (when the question has been selected) is placed farther down than it should be (related to note below)

- In MPP: Video and Multi-part tags should not be displayed (will allow checkmark to be moved up per the note above)


shared/src/components/exercise-preview/index.jsx
```diff
@@ -126,7 +126,10 @@ class ExercisePreview extends React.Component {
               actions={this.props.overlayActions}
               onClick={this.props.onOverlayClick} /> )}
           <div className="exercise-body">
-            <ExerciseBadges multiPart={info.isMultiPart} video={info.hasVideo} interactive={info.hasInteractive} writtenResponse={info.isWrittenResponse} />
+            {
+              this.props.questionType !== 'student-mpp' &&
+              <ExerciseBadges multiPart={info.isMultiPart} video={info.hasVideo} interactive={info.hasInteractive} writtenResponse={info.isWrittenResponse} />
+            }
             {!isEmpty(info.context) && !!this.props.isInteractive ? <ArbitraryHtmlAndMath className="context" block={true} html={info.context} /> : undefined}
```
nathanstitt, line 133: It's almost always better to have the component handle the logic instead of using inline logic via `{}`.  How about adding a `questionType` prop to the ExerciseBadges` which will return null if it's a `student-mpp`? https://github.com/openstax/tutor-js/blob/main/shared/src/components/exercise-badges/index.jsx#L117

tutor/resources/images/icons/exercise-exclude-card-view-action-transparent.svg
```diff
@@ -0,0 +1,9 @@
...
+</defs>
+</svg>
```
nathanstitt, line 9: where'd this SVG come from?  It looks like it's a SVG that continues a rasterized image.  It'd be better to use one that has paths.

tutor/src/components/buttons/save-practice.js
```diff
@@ -94,30 +95,30 @@ const SavePracticeButton = observer(({
...
+      <StyledSavePracticeButton
+        onClick={saveOrRemovePracticeQuestion}
+        isSaved={isSaved()}
+        className="save-practice-button"
+        data-test-id="save-practice-button"> 
+        {getIconAndLabel(isSaved(), practiceQuestions.isAnyPending)}
+      </StyledSavePracticeButton>
+    );
+  };
+
```
nathanstitt, line 120: nice cleanup. this reads much easier to me.

nathanstitt (approved):
looks pretty good, I'll go ahead and approve so you can merge after looking at the suggestions

tutor/resources/images/icons/exercise-exclude-card-view-action-transparent.svg
camliu89: I see the image embedded here. I got it from figma, let me take a look.

tutor/src/components/buttons/save-practice.js
camliu89: thanks. I changed it mainly because the `<></>` blocks the render of the tooltip.

PR #3352, opened 2020-11-02
A few fixes for release

nathanstitt:
* Clear exercises store when viewing practice and leaving QL view
* Don't show save practice button when user is a teacher
* Don't mention answering questions on closed bar when user is a teacher


camliu89 (approved):
looks good.

PR #3353, opened 2020-11-02
Add/Edit main branch

camliu89:
Main branch for ADD/EDIT


tutor/src/components/add-edit-question/ux.js
```diff
@@ -0,0 +1,240 @@
...
+  /** form */
+  // chapter/sections
+  @observable selectedChapter;
+  @observable selectedChapterSection;
+  // question
+  @observable questionText;
+  @observable answerKeyText;
+  @observable isTwoStep;
+  @observable options = [];
+  @observable detailedSolution;
```
jivey, line 33: @camliu89 @nathanstitt I think you can send answerKeyText and detailedSolution as the same field

camliu89: @jivey yes. I realized that as I was working on my other branch.

PR #3354, opened 2020-11-02
Ql choose section

camliu89:
Before:
Screen Shot 2020-11-02 at 3 44 52 PM

After:
Screen Shot 2020-11-04 at 1 06 12 PM

### DIFF
--- a/tutor/specs/components/sections-chooser.spec.jsx
+++ b/tutor/specs/components/sections-chooser.spec.jsx
@@ -20,7 +20,7 @@ describe('Sections Chooser', () => {
     expect(props.onSelectionChange).toHaveBeenCalled();
     props.onSelectionChange.mockReset();
     const pageId = book.pages.byId.keys()[2];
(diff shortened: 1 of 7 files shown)


jivey:
@camliu89 I think we have an orange checkbox component for new UI stuff (may require a `standalone` prop): https://github.com/openstax/tutor-js/blob/main/tutor/src/components/checkbox-input.js Would that work for this purpose?

camliu89:
@jivey I might be able to use it, I think. I was using the radio-input in the next screen :). thanks.

camliu89:
@jivey actually, the checkbox needs a third state: orange for when the whole chapter is selected, blank for when it is not selected, and grayed out when some sections are selected. I can modify the new checkbox UI to handle the third state, but I think it works now with the changes I made in the `TriStateCheckbox` (I just realized now why it has the `Tri` at the beginning: 3 states).

nathanstitt:
yes go ahead and modify `TriStateCheckbox`, at one time that was used elsewhere but I think the book section picker is the only spot now

PR #3367, opened 2020-11-17
HTML editor

nathanstitt:
image


nathanstitt:
@camliu89 I pulled in master to get the lint fix PR

tutor/src/components/editor/index.js
```diff
@@ -0,0 +1,49 @@
...
+    if (html) {
+      setHTML(html);
+      setEditing(false);
+    }
+  }, [setEditing, setHTML]);
+  const body = isEditing ?
+    <Editing html={currentHTML} onSave={onSave} /> : <HTML html={currentHTML} onClick={() => setEditing(true)}/>;
+
+  return (
+    <div style={{ margin: '100px', height: '500px' }}>
```
camliu89, line 38: maybe use `styled` here? or are you overriding some parent styles?

camliu89:
@nathanstitt it looks good. I am not sure about the `Save` button in the editor toolbar. Do the users need to save the question text before they can save the question?

camliu89 (approved):
looks good @nathanstitt , just left a couple of comments.

tutor/src/components/editor/index.js
nathanstitt: good idea, that was just being lazy and copying from the demo which didn't have styled-components

nathanstitt:
> @nathanstitt it looks good. I am not sure about the `Save` button in the editor toolbar. Do the users need to save the question text before they can save the question?

No but it needs a way to switch from editing html to displaying it.  Perhaps the "Save" button isn't really the best control, but  I'm not sure what'd be better

camliu89:
> > @nathanstitt it looks good. I am not sure about the `Save` button in the editor toolbar. Do the users need to save the question text before they can save the question?
> 
> No but it needs a way to switch from editing html to displaying it. Perhaps the "Save" button isn't really the best control, but I'm not sure what'd be better

Maybe just an `onBlur` and `onFocus` listener?

PR #3372, opened 2020-11-24
Edit question

camliu89:
Editing a question.


tutor/src/components/add-edit-question/terms-of-use.js
```diff
@@ -43,9 +43,9 @@ const AddEditQuestionTermsOfUse = observer(({ ux }) => {
   const [agree, setAgree] = useState(false);
   return (
     <StyledAddEditQuestionTermsOfUseModal
-      show={ux.showAddEditTermsOfUse}
+      show={true}
       backdrop="static"
```
nathanstitt, line 47: was the change to be `true` here for debugging?

tutor/src/models/reference-book/node.js
```diff
@@ -163,7 +163,8 @@ class ReferenceBookNode extends BaseModel {
     if(!this.chapter_section || !this.chapter_section.chapter) {
       return this.titleText;
     }
-    return `${this.chapter_section.chapter}. ${this.titleText}`;
+    const chapterSection = this.chapter_section;
+    return `${chapterSection.chapter}.${chapterSection.section || ' '} ${this.titleText}`;
   }
```
nathanstitt, line 168: I know this isn't new, but chapter section has a format method that is better to use, it handles the case where section may be missing when it's 0. https://github.com/openstax/tutor-js/blob/main/tutor/src/models/chapter-section.js#L18-L23

nathanstitt (approved):

tutor/src/components/add-edit-question/terms-of-use.js
camliu89: yes! :)

PR #3374, opened 2020-12-02
Create question

camliu89:
Add createExercise api endpoint, and add also feedback tip modal and exit warning. modal


shared/resources/styles/components/toasts.scss
```diff
@@ -1,6 +1,6 @@
 .toast-notification {
   position: fixed;
-  z-index: $zindex-modal;
+  z-index: 1051;
   min-height: 50px;
```
nathanstitt, line 4: can you do a ` z-index: $zindex-modal + 1`;  here?  I'd prefer to continue using variables so they'll update if we ever update bootstrap.  The defaults are listed in `styles/variables/tutor-bootstrap.scss`

tutor/resources/styles/global.scss
```diff
@@ -7,6 +7,16 @@ body {
...
+  // blur out the previous modals, except the last one
+  > div.modal.show:not(:last-child) {
+      position: fixed;
+      top: 0;
+      left: 0;
+      z-index: 1040;
+      width: 100vw;
+      height: 100vh;
+  }
 }
```
nathanstitt, line 20: same as above, how about `$zindex-modal - 10` here?

tutor/src/components/add-edit-question/form/index.js
```diff
@@ -39,7 +39,8 @@ const FormButtons = observer(({ ux }) => {
     <Button
       variant="default"
       className="preview"
-      onClick={() => window.alert('preview')}>
+      onClick={() => window.alert('preview')}
+      disabled={!ux.isReadyToPublish}>
```
nathanstitt, line 43: 😁 nice preview!

tutor/src/components/add-edit-question/ux.js
```diff
@@ -300,6 +381,103 @@ export default class AddEditQuestionUX {
...
+      this.showExitWarningModal = true;
+    }
+    else {
+      this.onDisplayModal(false);
+    }
+  }
+
   /**
    * This method checks if the `fields` (fields in the form) are empty.
    * If so, set this.isEmpty[field] to true.
```
nathanstitt, line 483: I think this'll work fine for now Alex, but typically I find it cleaner to create a model to hold all the data that's being edited, and the UX class would manipulate the model as needed when events occur.

That'd greatly simplify the creation and reset.  Creation could be handled by the model's constructor, and on reset, the UX would throw away that model and create a new one.

nathanstitt (approved):

shared/resources/styles/components/toasts.scss
camliu89: oops, I was testing it and forgot to change it back. will do 👍

tutor/src/components/add-edit-question/ux.js
camliu89: 👍

PR #3375, opened 2020-12-02
Update sidebar action styles

jivey:
image

### DIFF
--- a/tutor/src/screens/question-library/exercises-display.jsx
+++ b/tutor/src/screens/question-library/exercises-display.jsx
@@ -203,6 +203,13 @@ class ExercisesDisplay extends React.Component {
         handler: this.onExerciseToggle,
       };
     }
+    if (!exercise.isMultiPart) {
+      const isUserGeneratedQuestion = exercise.belongsToCurrentUserProfileId(User.profile_id);
+      actions.copyEdit = {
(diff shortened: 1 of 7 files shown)


jivey:
This doesn't match the design in Figma yet, more time is needed to make the side arrows work well with the controls above them, but we can get some of these changes in and add the rest later.

tutor/src/screens/question-library/exercises-display.jsx
```diff
@@ -203,6 +203,13 @@ class ExercisesDisplay extends React.Component {
         handler: this.onExerciseToggle,
       };
     }
+    if (!exercise.isMultiPart) {
+      const isUserGeneratedQuestion = exercise.belongsToCurrentUserProfileId(User.profile_id);
+      actions.copyEdit = {
```
nathanstitt, line 208: @camliu89 @jivewise we should add a `canEdit(user)` method to exercise that would check it it's not multipart and all this

nathanstitt (approved):

tutor/src/screens/question-library/exercises-display.jsx
camliu89: yep, that sounds good.

PR #3378, opened 2020-12-03
More question detail updates

jivey:
localhost_3000_course_14_questions (3)

- Adds author and section link to footer
- Updates padding and text styles to match Figma (besides the location of the side controls)
- Fixes null icon issue

### DIFF
--- a/shared/src/components/exercise-preview/controls-overlay.js
+++ b/shared/src/components/exercise-preview/controls-overlay.js
@@ -49,7 +49,7 @@ class ControlsOverlay extends React.Component {
                   className={`action ${type}`}
                   onClick={partial(this.onActionClick, partial.placeholder, action.handler)}
                 >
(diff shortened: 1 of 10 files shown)


shared/src/components/exercise-preview/controls-overlay.js
```diff
@@ -48,7 +48,7 @@ class ControlsOverlay extends React.Component {
                   className={`action ${type}`}
                   onClick={partial(this.onActionClick, partial.placeholder, action.handler)}
                 >
-                  <Icon type={iconTypes[type]} />
+                  {iconTypes[type] && <Icon type={iconTypes[type]} />}
                   <span>
```
nathanstitt, line 52: hah, this is good but we should also add a `details` icon above as well, I did that though so lets see if they conflict after this one is merged

nathanstitt (approved):

PR #3384, opened 2020-12-07
Reading no copy edit

camliu89:
Do not show the copy and edit action in Reading questions


tutor/src/models/exercises/exercise.js
```diff
@@ -105,6 +105,10 @@ class TutorExercise extends BaseModel {
     return this.has_video;
   }
 
+  @computed get canCopyOrEdit() {
+    return !this.isMultiPart && this.isHomework;
+  }
+
```
nathanstitt, line 111: this also needs to check the `is_copyable` flag.  that will be true even if the user is the owner

nathanstitt (approved):

PR #3389, opened 2020-12-08
Add edit card view fixes

camliu89:
QL card view fixes


tutor/src/components/exercises/no-exercises-found.jsx
```diff
@@ -31,7 +31,7 @@ const noResultsInFiltering = () =>
 const noQuestionsInSection = (isHomework, onSelectSections) =>
   <>
     <h3>
-        No exercises found.
+        No exercises questions found.
     </h3>
```
nathanstitt, line 35: this sounds weird.   I think it should just be "No questions found", right?

nathanstitt (approved):
looks good, just need to fix wording

PR #3391, opened 2020-12-09
Ql terms of use footer

camliu89:
New footer in QL with terms of use for creating/editing a question:

Screen Shot 2020-12-09 at 10 04 57 AM

### DIFF
--- a/tutor/src/components/add-edit-question/index.js
+++ b/tutor/src/components/add-edit-question/index.js
@@ -10,7 +10,9 @@ import {
 
 const AddEditQuestionModals = observer(({ ux }) => {
   if(!ux.didUserAgreeTermsOfUse) {
-    return <AddEditQuestionTermsOfUse ux={ux} />;
+    return <AddEditQuestionTermsOfUse
+      onClose={() => ux.onDisplayModal(false)}
+      show={!ux.didUserAgreeTermsOfUse} />;
   }
   return (
     <>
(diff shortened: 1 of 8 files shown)


tutor/src/components/add-edit-question/index.js
```diff
@@ -10,7 +10,9 @@ import {
 
 const AddEditQuestionModals = observer(({ ux }) => {
   if(!ux.didUserAgreeTermsOfUse) {
-    return <AddEditQuestionTermsOfUse ux={ux} />;
+    return <AddEditQuestionTermsOfUse
+      onClose={() => ux.onDisplayModal(false)}
+      show={!ux.didUserAgreeTermsOfUse} />;
   }
```
nathanstitt, line 16: show seems like it'll always be true here since we're not rendering `AddEditQuestionTermsOfUse` at all unless it's true.

tutor/src/components/add-edit-question/terms-of-use.js
```diff
@@ -43,45 +45,63 @@ const StyledAddEditQuestionTermsOfUseModal = styled(AddEditQuestionTermsOfUseMod
...
+            <Button
+              variant="primary"
+              disabled={!agree}
+              onClick={agreeTermsOfUse}>
+              Accept and continue
+            </Button>
+          </div>
+        </>
+        }
       </Modal.Body>
```
nathanstitt, line 105: instead of a huge if block, how about extracting to a TermAgreement component and render that if needed?

nathanstitt (changes requested):

tutor/src/components/add-edit-question/index.js
camliu89: technically it has also an if statement above so it won't render this component at all if `!ux.didUserAgreeTermsOfUse` is `false`.

camliu89: so it can be `show={true}`

nathanstitt (approved):

PR #3393, opened 2020-12-09
Sort exercises by how closely related to user they are

nathanstitt:
Sort should be:
* My questions
* Co-teachers’ questions
* OST questions
* OST community

### DIFF
--- a/tutor/src/components/exercises/cards.jsx
+++ b/tutor/src/components/exercises/cards.jsx
@@ -1,15 +1,15 @@
 import {
-  React, PropTypes, observer, ArrayOrMobxType, styled,
+  React, PropTypes, observer, ArrayOrMobxType, styled, computed,
 } from 'vendor';
 import { map, isEmpty } from 'lodash';
 import ExercisePreview from './preview';
 import BookPartTitle from '../book-part-title';
 import ScrollTo from '../../helpers/scroll-to';
-import { ExercisesMap } from '../../models/exercises';
+import { ExercisesMap, exerciseSort } from '../../models/exercises';
 import Exercise from '../../models/exercises/exercise';
-import Book from '../../models/reference-book';
+import Course from '../../models/course';
 import NoExercisesFound from './no-exercises-found';
-
+import User from '../../models/user';
 
 const SectionLabel = styled.label`
   font-size: 2.8rem;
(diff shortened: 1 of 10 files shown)


tutor/src/models/exercises/exercise.js
```diff
@@ -112,6 +112,6 @@ class TutorExercise extends BaseModel {
 
   // Openstax exercises returns an id of 0;
   @computed get belongsToOpenStax() { return this.author.id === '0'; }
-  belongsToCurrentUserProfileId(profileId) { return this.author.id == profileId; }
-  belongsToOtherAuthorProfileIds(profileId) { return !this.belongsToOpenStax && this.author.id != profileId; }
+  belongsToUser(user) { return this.author.id == user.profile_id; }
+  belongsToOtherUser(user) { return !this.belongsToOpenStax && !this.belongsToUser(user); }
```
camliu89, line 116: This is much more readable.

tutor/src/components/exercises/cards.jsx
```diff
@@ -30,10 +30,18 @@ class SectionsExercises extends React.Component {
...
+    );
+  }
+
   render() {
-    const { pageId, book, exercises, ...previewProps } = this.props;
-    const page = book.pages.byId.get(pageId);
-    const sectionExercises = exercises.byPageId[pageId];
+    const { pageId, course, ...previewProps } = this.props;
+    const sectionExercises = this.exercises;
+    const page = course.referenceBook.pages.byId.get(pageId);
```
camliu89, line 44: I ran into issues sometimes where `course.referenceBook` is `undefined` because it has not started a fetch yet for the book. Should we ensure it is fetched before it gets into this part of the component?

tutor/src/models/exercises/exercise.js
nathanstitt: thanks!

tutor/src/components/exercises/cards.jsx
nathanstitt: yes good idea, I'll add a guard for that

nathanstitt: after looking at it, referenceBOok will always be set.  when it's not loaded page won't be found though, so I'll check for that

camliu89 (approved):

PR #3408, opened 2020-12-14
only show the feedback tip modal once per question building

camliu89:
CASE:
add/edit modal is opened -> question is filled without feedback -> try to publish -> show feedback tip -> feedback tip is closed -> tries to publish again -> do not show the feedback tip.

*If the user opens the add/edit modal again then the above logic is applied.


jivey (approved):
Works over here 👍

PR #3411, opened 2020-12-15
sort exercises first by chapter-section, then user

nathanstitt:
Otherwise details view gets out of order when the section changes

### DIFF
--- a/tutor/src/components/exercises/details.jsx
+++ b/tutor/src/components/exercises/details.jsx
@@ -53,20 +53,6 @@ class ExerciseDetails extends React.Component {
     }
   }
 
-  componentDidUpdate(prevProps) {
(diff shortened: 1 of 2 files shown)


jivey (approved):
Tested well over here

jivey:
Going to the right seems to be consistent now, sometimes going left a few times jumps to the beginning? I'm not having it occur consistently, so maybe it's just my local data (I've done a lot of manual DB changes to test add/edit things.)

nathanstitt:
the cause of the jumping around is here https://github.com/openstax/tutor-js/blob/main/tutor/src/components/exercises/details.jsx#L60-L66

I'm not sure why that block is needed since the controlling component never sets the current section. I think I'll take that out and test it that way

PR #3412, opened 2020-12-15
Switch to card view after deleting in details

nathanstitt:
Otherwise you get stuck

### DIFF
--- a/tutor/src/models/exercises.js
+++ b/tutor/src/models/exercises.js
@@ -13,7 +13,7 @@ export { COMPLETE, PENDING };
 
 export const exerciseSort = (exercises, course, user) => {
   return sortBy(exercises, (ex) => {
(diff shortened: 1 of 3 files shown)


tutor/src/models/exercises.js
```diff
@@ -13,7 +13,7 @@ export { COMPLETE, PENDING };
 
 export const exerciseSort = (exercises, course, user) => {
   return sortBy(exercises, (ex) => {
-    const cs = ex.page.chapter_section.asString;
+    const cs = ex.page.chapter_section.asNumber;
     if (ex.belongsToUser(user)) {
```
nathanstitt, line 17: need to use the numeric version of c/s so it sorts correctly otherwise stuff like intro will appear at the end

camliu89 (approved):

PR #3419, opened 2020-12-18
add-edit-context-fix

camliu89:
`@observable exercise` was deleted. This PR creates an `observable context` that tracks if the exercise that is being edited has a `context`.

### DIFF
--- a/tutor/src/components/add-edit-question/form/question.js
+++ b/tutor/src/components/add-edit-question/form/question.js
@@ -300,8 +300,7 @@ const Form = observer(({ ux }) => {
   };
   
   const contextInfo = () => {
-    const exercise = ux.exercise;
-    if(!exercise || !exercise.context) {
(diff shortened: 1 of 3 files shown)


nathanstitt:
Oh I'm so sorry Alex, I caused this    I removed setting `exercise` since it was info that could be obtained using the `from_exercise_id` and I was afraid they might get out-of-sync at some point.  I replaced it with a `@computed get fromExercise()`  but missed updating the check in the context info component. 

I think if you update context info to look at `ux.fromExercise.context` it should work.

camliu89:
oh I see it! I am changing it now.

camliu89:
@nathanstitt `fromExercise` looks good. I updated the code with it.

nathanstitt (approved):
Looks great, thanks Alex.  Sorry again for mucking up the exercise before

PR #3421, opened 2020-12-18
fix preview info displaying on exericse

nathanstitt:
image

### DIFF
--- a/shared/src/components/exercise-preview/index.jsx
+++ b/shared/src/components/exercise-preview/index.jsx
@@ -63,22 +63,23 @@ class ExercisePreview extends React.Component {
   };
 
   getCleanPreview = () => {
-    return trimEnd(last(this.exercise.questions).stem_html);
+    return this.props.extractedInfo.preview ? this.props.extractedInfo.preview : trimEnd(last(this.exercise.questions).stem_html);
   };
 
+  @computed get showPreview() {
+    return Boolean(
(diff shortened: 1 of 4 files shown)


shared/src/components/exercise-preview/index.jsx
```diff
@@ -63,22 +63,23 @@ class ExercisePreview extends React.Component {
   };
 
   getCleanPreview = () => {
-    return trimEnd(last(this.exercise.questions).stem_html);
+    return this.props.extractedInfo.preview ? this.props.extractedInfo.preview : trimEnd(last(this.exercise.questions).stem_html);
   };
```
nathanstitt, line 67: this is the really crazy bit.  the `getCleanPreview` method wasn't even reading the preview

camliu89: I was wondering, in a MPQ without video or sim, would it always show the last question stem at the top?

camliu89 (approved):
It looks good. I just left a comment about the `getCleanPreview` logic we had before.

shared/src/components/exercise-preview/index.jsx
nathanstitt: hm, good point.  yes that does seem like it would.  I'd just copied the existing code, but I agree that ti'd make more sense to pick the first part

PR #3422, opened 2020-12-18
Ql scroll fix

camliu89:
Screen Shot 2020-12-18 at 3 22 00 PM
Screen Shot 2020-12-18 at 3 22 24 PM

### DIFF
--- a/tutor/src/screens/question-library/exercise-controls.jsx
+++ b/tutor/src/screens/question-library/exercise-controls.jsx
@@ -1,4 +1,4 @@
-import { React, PropTypes, observer, inject, autobind, styled } from 'vendor';
+import { React, PropTypes, observer, inject, autobind, styled, cn } from 'vendor';
 import { Button, Popover, OverlayTrigger } from 'react-bootstrap';
 import { partial } from 'lodash';
 import Course from '../../models/course';
@@ -15,7 +15,6 @@ const StyledExerciseControls = styled.div`
   .exercise-controls-bar {
     display: flex;
     height: 60px;
-    border-top: 1px solid ${colors.neutral.pale};
     label {
       font-size: 1.4rem;
     }
@@ -23,6 +22,7 @@ const StyledExerciseControls = styled.div`
   .sections-control {
     justify-content: flex-start;
     overflow-x: auto;
+    border-bottom: 1px solid ${colors.neutral.pale};
     .sectionizer {
       margin: 0 1.7rem;
     }
@@ -34,9 +34,11 @@ const StyledExerciseControls = styled.div`
         margin-left: 0;
       }
     }
+    &.is-details {
+      border-bottom: inherit;
+    }
   }
   .filters-control {
-    border-bottom: 1px solid ${colors.neutral.pale};
     justify-content: space-between;
     padding: 0 2.2rem;
     .library-label {
@@ -130,7 +132,7 @@ class ExerciseControls extends React.Component {
       </StyledPopover>;
 
     const sections =
-      <div className="exercise-controls-bar sections-control">
+      <div className={cn('exercise-controls-bar sections-control', { 'is-details': showingDetails })}>
         <Button className="back-to-section" variant="link" onClick={this.props.onSelectSections}>
           <Icon
             size="lg"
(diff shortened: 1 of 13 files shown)


tutor/src/screens/question-library/exercise-controls.jsx
```diff
@@ -86,8 +89,15 @@ class ExerciseControls extends React.Component {
     showingDetails: PropTypes.bool,
     topScrollOffset: PropTypes.number,
     onDisplayAddEditQuestionModal: PropTypes.func.isRequired,
+    setSecondaryTopControls: PropTypes.func.isRequired,
   };
 
+  constructor(props) {
+    super(props);
+    //props.setSecondaryTopControls(this.renderControls);
```
nathanstitt, line 97: if this isn't needed, you can remove the entire `constructor`, however it's not clear to me where `setSecondaryTopControls` is being called from if this is removed.

tutor/src/screens/question-library/exercises-display.jsx
```diff
@@ -106,6 +110,28 @@ class ExercisesDisplay extends React.Component {
...
 
+  componentWillUnmount() {
+    this.props.exercises.clear();
+    document.removeEventListener('scroll', this.checkIsControlsSticky);
+  }
+
+  componentDidMount() {
+    document.addEventListener('scroll', this.checkIsControlsSticky);
+  }
+
```
nathanstitt, line 121: if we add event listeners to `scroll`, its important to debounce the callback so it doesn't block scrolling and lead to jittery behavior. In fact the latest chrome will yell at you for it.  

We have a observable model for this, https://github.com/openstax/tutor-js/blob/main/tutor/src/models/window-scroll.js   Use it instead.

tutor/src/screens/question-library/exercises-display.jsx
```diff
@@ -106,6 +110,28 @@ class ExercisesDisplay extends React.Component {
...
+  // Checking if the controls-bar is sticked to the top.
+  // If so, show box-shadow.
+  @action.bound checkIsControlsSticky() {
+    const windowImpl = this.windowImpl || window;
```
nathanstitt, line 125: windowImpl should be from props, however it won't be needed if you use windowScroll.

Instead how about this:
```js

class ExercisesDisplay extends React.Component {
...

windowScroll = new WindowScroll(this.props.windowImpl)

@computed areControlsSticky() {
  const topOffset = this.showingDetails ? 160 : 250;
   return this.windowScroll.y >=  topOffset;
}
```

tutor/src/screens/question-library/exercises-display.jsx
```diff
@@ -389,7 +415,7 @@ class ExercisesDisplay extends React.Component {
     return (
       <StyledExerciseDisplay>
         <StyledCourseBreadcrumb course={course} currentTitle="Question Library" />
-        <div className="controls-wrapper">
+        <div className={cn('controls-wrapper', { 'isSticky': this.isControlsSticky })}>
           <ExerciseControls
```
nathanstitt, line 419: css classNames should use lowercase & dashes, not camelCase, I also kinda prefer declarative names for them instead of `is-`, so this could be just `sticky-controls` maybe?

nathanstitt (changes requested):

tutor/src/screens/question-library/exercises-display.jsx
camliu89: this works! 👍

tutor/src/screens/question-library/exercises-display.jsx
```diff
@@ -89,9 +96,7 @@ class ExercisesDisplay extends React.Component {
     exercises: sharedExercises,
   };
 
-  componentWillUnmount() {
-    this.props.exercises.clear();
-  }
+  @observable isControlsSticky = false;
```
nathanstitt, line 99: I think this is unused now

nathanstitt (approved):
looks good @camliu89 , found one unused var is all

tutor/src/screens/question-library/exercises-display.jsx
camliu89: good catch :)

PR #3429, opened 2021-01-07
Hide excluded exercises in HW builder

nathanstitt:
Previously we weren't using ux.displayedExercises which checks that and a few other good things


jivey (approved):
Works over here

PR #3431, opened 2021-01-11
add margin between index items title and link

nathanstitt:
Thanks to @lwtchu for listening while I floundered on this for a bit 🤣 

changes:
image

to be:
image


camliu89 (approved):
👍

PR #3435, opened 2021-01-15
read bootstrap data from api endpoint

nathanstitt:
Reading it from the dom means it's cached and hitting backbutton would
use stale data.

Plus we're about to make index.html a static file which means it
couldn't be delivered like that anyway


jivey (approved):
Tested over here, everything seems to load like normal. I did see a prop requirement warning, maybe from the async changes?
```
Warning: Failed prop type: The prop `app` is marked as required in `App`, but its value is `undefined`
```

nathanstitt:
@jivey that's interesting. I actually just added the propType definitions to App, it lacked them before.

PR #3438, opened 2021-01-19
Rename conceptual/recall to reading/homework and make multi-select

nathanstitt:
image

### DIFF
--- a/exercises/src/components/tags/assignment-type.jsx
+++ b/exercises/src/components/tags/assignment-type.jsx
@@ -7,7 +7,6 @@ import MultiSelect from './multi-select';
 const TYPES = [
   { value: 'reading', label: 'Reading' },
   { value: 'homework', label: 'Homework' },
-  { value: 'practice', label: 'Practice' },
 ];
 
 function QuestionTypeTag(props) {
--- a/exercises/src/components/tags/assignment-type.jsx
+++ b/exercises/src/components/tags/assignment-type.jsx
@@ -1,11 +1,9 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import { Icon } from 'shared';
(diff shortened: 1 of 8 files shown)


exercises/src/components/tags/assignment-type.jsx
```diff
@@ -3,22 +3,25 @@ import React from 'react';
...
   const { isOpenEnded } = props.exercise;
   return (
-    <SingleDropdown
+    <MultiSelect
+      {...props}
       readonly={isOpenEnded}
-      {...props} label="Question Type" type={TYPE} choices={TYPES}
+      tagType="assignment-type"
+      label="Assignment Type"
+      type={TYPE}
```
camliu89, line 23: @nathanstitt I don't think `type` is used in the `MultiSelect` component.

exercises/src/components/tags/assignment-type.jsx
```diff
@@ -3,22 +3,25 @@ import React from 'react';
...
+      options={TYPES}
       icon={isOpenEnded && <Icon variant="infoCircle" tooltip='"Open Ended" exercises must be type practice' />}
```
camliu89, line 25: `MultiSelect` also does not take an `icon` prop. Should we add this check outside of the `MultiSelect` component?

nathanstitt: ug good catch.  actually we want to remove this tooltip altogether so that'll be fine once we do

exercises/src/components/tags/assignment-type.jsx
nathanstitt: yep, its tagType for MultiSelect.  thanks!

camliu89 (approved):

PR #3439, opened 2021-01-21
Integration branch for my-courses

nathanstitt:


package.json
```diff
@@ -140,7 +147,10 @@
     "survey-react": "^1.1.7",
     "svg-url-loader": "3.0.0",
     "time-input-polyfill": "^1.0.7",
+    "ts-loader": "^8.0.14",
     "twix": "1.3.0",
+    "typescript": "^4.1.3",
+    "typescript-eslint": "^0.0.1-alpha.0",
```
camliu89, line 153: This is a really old version of `typescript-eslint`. It seems that they are at `v4.14.0`.

.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "trailingComma": "es5",
+  "tabWidth": 2,
+  "semi": false,
+  "singleQuote": true
+}
```
camliu89, line 6: 🙌

tsconfig.json
```diff
@@ -0,0 +1,35 @@
...
+    "strict": true,
+    "forceConsistentCasingInFileNames": true,
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "downlevelIteration": true,
+    "noEmit": true
+  },
+  "exclude": [
+    "node_modules",
+  ]
```
camliu89, line 34: I like to do also `"include": ["**/*.ts", "**/*.tsx"]`.

camliu89:
@nathanstitt I gave a look at your PR and left some comments. 👍

package.json
nathanstitt: Yah I think this one was installed by mistake, this is https://www.npmjs.com/package/typescript-eslint but we want https://www.npmjs.com/package/@typescript-eslint/eslint-plugin.  I'll update

tsconfig.json
nathanstitt: does that help with resolving or speed it up or something?  Don't mind adding, but am curious why

camliu89: I think it helps with resolving the files faster. You can also specifically do `["tutor/**/*.ts", "shared/**/*.ts", "exercises/**/*.ts"]` so it specifically look for only ts files in those folders.

PR #3443, opened 2021-01-27
add redux store with courses data

nathanstitt:
Pretty barebones but gives us an idea of what would be involved


tutor/src/store/api.ts
```diff
@@ -0,0 +1,47 @@
...
+    return createAsyncThunk<RetT, ArgT>(storePath, async (obj: ArgT) => request(method, urlPattern, reducer(obj)))
+}
+
+
+// URLS do not include /api/ prefix.  the request method below will add it
+
+export const updateCourse = r<Course, Course>('PUT', 'courses/${id}', 'courses/updateCourse', (c: Course) => ({
+    params: { id: c.id }, data: { name: c.name } // TODO include other updateable properties
+}))
+
```
nathanstitt, line 45: blocks like the `updateCourse` here would replace the ones in https://github.com/openstax/tutor-js/blob/main/tutor/src/api/index.js

tutor/src/store/courses.ts
```diff
@@ -0,0 +1,46 @@
...
+        }
+    },
+    extraReducers: (builder) => {
+        builder.addCase(updateCourse.fulfilled, (state, { payload: course }) => {
+            state.ids[course.id] = course
+        })
+        builder.addCase(bootstrap, (state, action) => {
+            courseAdapter.setAll(state, action.payload.courses)
+        })
+    }
```
nathanstitt, line 35: `reducers` are how the store is updated.  the plain ones only operate on the `courses` part of the store, while `extraReducers` can act on the entire store if needed

tutor/src/store/courses.ts
```diff
@@ -0,0 +1,46 @@
...
+import { bootstrap } from './bootstrap'
+
+const courseAdapter = createEntityAdapter<Course>({
+    sortComparer: (a, b) => a.name.localeCompare(b.name),
+})
+
+const initialState = courseAdapter.getInitialState()
+interface CourseSlice { courses: typeof initialState }
+
+const selectors = courseAdapter.getSelectors<CourseSlice>(s => s.courses)
```
nathanstitt, line 14: selectors have a bunch of nice functions: https://redux-toolkit.js.org/api/createEntityAdapter#return-value

tutor/src/store/courses.ts
```diff
@@ -0,0 +1,46 @@
...
+    reducers: {
+        rename(state, { payload: { id, name } }: PayloadAction<{ id: string, name: string }>) {
+            console.log(current(state))
+            const course = state.entities[id]
+            if (course) { course.name = name }
+            console.log(current(state))
+        }
+    },
+    extraReducers: (builder) => {
+        builder.addCase(updateCourse.fulfilled, (state, { payload: course }) => {
```
camliu89, line 29: I don't see any action created with `updateCourse.fulfilled`.

nathanstitt: the action is `updateCourse` from the `api` endpoints file: https://github.com/openstax/tutor-js/pull/3443/files/d4d8049b44fdbdc141672a7b4a104ac2b544df99

Maybe they should be put in the same file as the slice, but we've found it's better to keep all the urls together so you can find them by url.  I'd be fine either way really

camliu89: oh I see. I was thinking we could split the apis into different files, so it is not a massive file like the one we have now. I am ok adding it into the same file as the slice.

camliu89: ok. I got the `updateCourse.fulfilled`. It is from the `createAyncThunk`.

nathanstitt:
ok, I'm going to merge this and make a new branch for my-courses I think

PR #3444, opened 2021-01-28
Mycourses course settings UI

camliu89:
1. New course settings UI that has a form to update the name, code, and timezone of the code.
2. Delete course is available on this screen, but it is disabled if there are students in the course.
3. Shows a success toast when changes are saved.

Screen Shot 2021-01-28 at 1 46 53 PM


tutor/src/screens/course-settings/delete-course-button.js
```diff
@@ -0,0 +1,98 @@
...
+  @action.bound open() {
+    this.showModal = true;
+  }
+
+  @action.bound deleteCourse() {
+    const { course, history } = this.props;
+    const courseTeacherMembership = course.currentCourseTeacher;
+    if (courseTeacherMembership.isTeacherOfCourse) {
+      courseTeacherMembership.drop().then(history.push('/dashboard'));
+    }
```
jivey, line 36: I was still seeing a test course on the dashboard after deleting it. Not sure if we need to refetch or if there's a computed property that isn't updating, hmm.

jivey (changes requested):

tutor/src/screens/course-settings/delete-course-button.js
camliu89: I merged the `my-courses` branch and it is working for me. @jivey

PR #3445, opened 2021-01-29
update sentry and filter out a few errors

nathanstitt:
* "cannot add button after initialization" is from SF
* ChunkLoadError is due to client losing internet


tutor/src/models/app/raven.js
```diff
@@ -35,6 +36,13 @@ const RavenErrorLogging = {
         if (isMathjax(breadcrumb)) { return null; }
         return breadcrumb;
       },
+      beforeSend(event, hint){
+        const error = hint.originalException;
+        if (error?.message?.match(IGNORED)) {
```
nathanstitt, line 41: @openstax/tutor-devs heads up we can use the optional chaining syntax now

jivey (approved):

PR #3452, opened 2021-02-05
My courses dashboard

camliu89:
Screen Shot 2021-02-05 at 3 39 21 PM
Screen Shot 2021-02-05 at 3 39 34 PM


tutor/src/screens/my-courses/dashboard/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+                return null
+            }));
+            return { ...o, courses: offeringCourses }
+        });
+    }
+
+    return {
+        offeringsWithCourses,
+    }
+}
```
camliu89, line 192: @nathan, Do you think we should just create a selector to return this data? Basically, I am just adding the courses to their respective offering.

nathanstitt: since offerings and courses will not change once bootstrapped, it seems like should add a "offerings" slice to the store, and then just use the it to get the offering for a given course.offering_id wherever needed.

tutor/src/screens/my-courses/dashboard/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+export default connect(mapStateToProps)(MyCoursesDashboard)
```
nathanstitt, line 194: a few more suggestions.  

Most of this probably belongs inside a new file for the teacher-has-courses case.  Maybe `screens/my-courses/courses` ?  That would leave index.ts only containing the logic for switching between the create course and display course screens.

I think instead of using `connect` here we can use hooks almost all the time, either with redux `useStore` or use the exported selectors.  For instance to get all the courses inside a component you could:
```ts
import { selectors  }  from 'store/selectors'
// ... inside component
const courses  = selectors.selectAll()
```

camliu89: I have all of these files for the "teacher-has-courses" case inside of `screens/my-courses/dashboard`. 

Yes, I am switching to it now. FTW hooks now. :)

tutor/src/screens/my-courses/dashboard/index.tsx
camliu89: 👍

tutor/src/screens/my-courses/dashboard/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+    // no students courses put them at the end of the list
+    if (useNumberOfStudents(a) === 0) { return -1 }
+    if(moment(a.ends_at).isAfter(b.ends_at)) { return 1 }
+    return 0
+})
+const sortPastCourses = (courses: Course[]) => courses.sort((a, b) => {
+    if(moment(a.ends_at).isAfter(b.ends_at)) { return 1 }
+    return 0
+})  
+
```
nathanstitt, line 88: This doesn't handle the a is before b case, sort should return -1 then.  Do we need to sort the courses without students as well?  With this courses will appear unsorted at the end.

We could make the sortPastCourses be `sortByCourseEndsAt` then:
```ts
const sortByCourseEndsAt = (courses: Course[]) => courses.sort((a, b) => {
     if(moment(a.ends_at).isAfter(b.ends_at)) { return 1 }
     if(moment(a.ends_at).isBefore(b.ends_at)) { return -1 }
     return 0
 })

const sortCurrentCourses = (courses: Course[]) => courses.sort((a, b) => {
   if (useNumberOfStudents(a) == 0 && useNumberOfStudents(b) != 0) { return -1 }
   return sortByCourseEndsAt(a, b);
}
```

tutor/src/screens/my-courses/dashboard/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+                return <PastCourses courses={pastCourses} />
+            } 
+            case 2: { 
+                return <ResourcesInfo
+                  appearanceCode={offering.appearance_code}
+                  os_book_id={offering.os_book_id}
+                  isFirstBlock={isFirstBlock} />
+            } 
+            default: { 
+               return <p>How did you get here?!</p>
```
nathanstitt, line 164: we should probably ask @dkhowala if she has a desired error message here, and if not use the standard PageNotFound one

tutor/src/screens/my-courses/dashboard/preview-course.tsx
```diff
@@ -0,0 +1,88 @@
...
+        onClick={onClick}
+      >
+          <h3 className="name">{offering.title}</h3>
+          <div className="preview-belt">
+            {previewMessage()}
+          </div>
+      </a>
+      </div>
+  </StyledPreviewCourse>
+  )
```
nathanstitt, line 85: code isn't indented properly here. I think there must be something wrong with lint check, that should have caught this.

tutor/src/screens/my-courses/dashboard/view-course.tsx
```diff
@@ -0,0 +1,91 @@
...
+        data-is-preview={course.is_preview}
+        data-term={useTermFull(course.id)}
+        data-is-teacher={useCurrentRole(course.id) === 'teacher'}
+        data-course-id={course.id}
+        className={cn('my-courses-item', className, { 'is-past': isPast })}
+      >
+        {/* If we are gonna be using Redux, need to set the current_role_id inside the course dashboard component which is for now a class component.
+        Hooks can only be called inside of a function component. */}
+        <div className="my-courses-item-title" onClick={() => dispatch(setCurrentRole({ roleId: primaryRole.id, id: course.id }))}>
+          <TutorLink to="dashboard" params={{ courseId: course.id }}>
```
nathanstitt, line 78: I don't understand what this is doing.  Clicking on the thumbnail is changing the user role?

tutor/src/store/offering.ts
```diff
@@ -0,0 +1,30 @@
+import { createSlice, createEntityAdapter } from '@reduxjs/toolkit'
+import { useSelector } from 'react-redux'
+import { Offering } from './types'
+import { getOfferings } from './api'
+
+const offeringAdapter = createEntityAdapter<Offering>({
+    sortComparer: (a, b) => a.default_course_name.localeCompare(b.default_course_name),
+})
```
nathanstitt, line 8: nice!  I see you've been reading the docs 😍

nathanstitt (approved):
looks great @camliu89 - mostly just comments around the sorting

tutor/src/components/tabs.jsx
```diff
@@ -37,10 +38,12 @@ const Tabs = ({
...
   const selectTabIndex = (tab) => {
     tab = Number(tab);
+    if(pushToPath) {
     const query = extend(Router.currentQuery(windowImpl), { tab });
     history.push(
       windowImpl.location.pathname + '?' + qs.stringify(query)
-    );
+      );
+    }
     setActiveIndex(tab);
```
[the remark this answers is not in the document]
camliu89: @nathanstitt this is because we are using tabs in each of the offerings. so there is no need to have the `tab=` in the url. Each tab has its own logic.

tutor/src/screens/my-courses/dashboard/index.tsx
camliu89: yeah, somehow I missed it. Thanks.

tutor/src/screens/my-courses/dashboard/index.tsx
camliu89: yeah :) , this was just a placeholder.

tutor/src/screens/my-courses/dashboard/view-course.tsx
camliu89: This is replicating what this is doing: https://github.com/openstax/tutor-js/blob/cb44f9ef3fb2480cfc158d37b01efff7595f144c/tutor/src/models/course/role.js#L42. I realized that `current_role_id` is not from the BE, but we were setting it up when the user goes to a course.

PR #3453, opened 2021-02-09
look at answers to detect multi-choice vs tag

nathanstitt:
This way the FE behaviour matches what the back-end does. In the case where the tag is mis-applied we need to be in sync


camliu89 (approved):

nathanstitt:
specs should be fixed in a different pr, will merge this one

PR #3459, opened 2021-02-23
Update lint rules

nathanstitt:
eslint defaulted to only running on `.js` files, meaning it's been skipping jsx files all this time


.eslintrc.js
```diff
@@ -35,7 +35,7 @@ module.exports = {
     "no-console":  "error",
     "no-debugger": "error",
     "no-unused-vars": [2, {'varsIgnorePattern': '_+'}],
-    "consistent-return": ["error", { "treatUndefinedAsUnspecified": true }],
```
camliu89, line 38: what is the difference between `error` and `0`?

camliu89 (approved):

PR #3462, opened 2021-02-24
update playwright version

jivey:
The locked version of playwright is incompatible with macOS 11.2, this bumps the version to one that is compatible.


nathanstitt (approved):
looks fine, I wonder what's going on with the spec failure, it's not even playwright related 🤔

jivey:
It's one I think you already fixed, I forgot to catch up this branch with the latest.

PR #3463, opened 2021-02-25
Preview course side panel

jivey:
image

### DIFF
--- a/tutor/src/components/side-panel.tsx
+++ b/tutor/src/components/side-panel.tsx
@@ -1,0 +1,111 @@
+import { React, styled, css } from 'vendor'
+import { useEffect, useState, useRef } from 'react'
+import { Icon } from 'shared'
+import TutorTheme, { colors, breakpoints } from 'theme'
+
+const Panel = styled.div`
+    position: absolute;
+    top: 33vh;
+    right: 0;
+    z-index: ${TutorTheme.zIndex.sidePanel};
+    display: flex;
+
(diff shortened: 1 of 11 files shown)


tutor/src/components/side-panel.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+    useEffect(
+        () => {
+            if (mounted.current) {
+                setPanelState(false)
+            } else {
+                mounted.current = true
+            }
+        },
+        [windowImpl.location.pathname],
```
nathanstitt, line 98: I think this will work, It won't be triggered on its own but will pickup the change when `SidePanel` re-renders after react-router switches the route.

jivey:
I'm going to put in a review request for this now despite the failing spec. I may need to fix it in a new PR, the time out is quite strange and hard to debug at the moment.

tutor/src/screens/my-courses/new-teacher.tsx
```diff
@@ -401,7 +401,7 @@ const OfferingList: React.FC<OfferingListProps> = ({ subject, offerings }) => {
 
     return (
         <div>
-            <button aria-expanded="true" onClick={toggleShow}>
+            <button aria-expanded={showList} onClick={toggleShow}>
```
nathanstitt, line 404: nice catch!

nathanstitt (approved):

PR #3464, opened 2021-02-25
Add bootstrap for offerings

nathanstitt:
was stopping tutor from booting


camliu89:
@nathanstitt I think this is what I was trying to fix in my PR.

camliu89:
👍

jivey (approved):

jivey:
The spec failure seems to be from something flaky, the screenshot when the selector failed was:

image

nathanstitt:
I'm going to merge this so the main tutor app boots, will make a followup with the role fix for specs in a bit

PR #3469, opened 2021-02-26
Add redirect for old url /dashboard

nathanstitt:
The reload toast for current code uses /courses

Also a few other minor tweaks

### DIFF
--- a/tutor/src/routes.js
+++ b/tutor/src/routes.js
@@ -9,6 +9,9 @@ const getRoutes = (router) => {
     const ConditionalHandlers = getConditionalHandlers(router);
 
     return [
+        { path: '/dashboard', name: 'oldMyCourses',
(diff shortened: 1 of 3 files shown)


tutor/src/routes.js
```diff
@@ -9,6 +9,9 @@ const getRoutes = (router) => {
     const ConditionalHandlers = getConditionalHandlers(router);
 
     return [
+        { path: '/dashboard', name: 'oldMyCourses',
```
camliu89, line 12: I think the reload sends the user to `/courses`

nathanstitt: it does on the current code, but the code that's already on production uses /courses, and that's what users will be running when this is deployed

camliu89: 👍

camliu89 (approved):

PR #3472, opened 2021-03-01
Don't blow up if course is missing appearance_code

nathanstitt:
QA has several courses without it

### DIFF
--- a/tutor/src/components/tutor-layout.js
+++ b/tutor/src/components/tutor-layout.js
@@ -4,7 +4,7 @@ import {
 } from 'vendor';
 import Theme                from '../theme';
 import Toasts               from 'shared/components/toasts';
-import Course               from '../models/course';
+import Courses, { Course }  from '../models/courses-map';
 import { Logo }             from './navbar/logo';
 import { Menus }            from './navbar/menus';
 import TermsModal           from './terms-modal';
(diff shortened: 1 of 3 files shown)


tutor/src/components/tutor-layout.js
```diff
@@ -94,30 +94,29 @@ const Content = styled.div`
...
+            <TutorLink
+                className="btn btn-primary"
+                to="createNewCourseFromOffering"
+                params={{ appearanceCode: course.appearanceCode }}
+                data-test-id="preview-panel-create-course"
+            >
+            Create a course
+            </TutorLink>
+        </SidePanel>
+    )
```
jivey, line 119: Good refactoring of the multiple returns

tutor/src/store/offering.ts
```diff
@@ -31,4 +31,8 @@ export const useAllOfferings = () => useSelector<OfferingSlice, Offering[]>(stat
     return selectors.selectAll(state)
 })
 
+export const useOfferingForId = (offeringId: ID) => useSelector<OfferingSlice, Offering | undefined>(state => {
+    return selectors.selectById(state, offeringId)
+})
+
```
jivey, line 37: Used somewhere else?

jivey (approved):

tutor/src/store/offering.ts
nathanstitt: oh hah, I was going to use this but then saw the file was JS and using models so I just used that

PR #3474, opened 2021-03-01
don't render if appearance code is missing

nathanstitt:
found one more spot that was using it


jivey (approved):

camliu89:
Maybe don't show the create course component or appearance_code is null?

camliu89:
Nvm. I read it wrong. 😐

camliu89 (approved):

PR #3477, opened 2021-03-04
Only display available offerings

nathanstitt:
and only show the preview/create when the offering has that availability indicated

PR #3478, opened 2021-03-04
reload page after joining a preview

nathanstitt:
the old code did some pretty intricate role/course stuff to make
the FE match the new BE state, but it's probably best to just reload


nathanstitt:
spec is failing for same cause fixed in #3477 - will merge that one first then rebase this

jivey (approved):

PR #3480, opened 2021-03-05
add resetState for e2e specs

nathanstitt:
This will reset all modifications so changes don't bleed over from one spec to another

### DIFF
--- a/tutor/scripts/test-server/backend.js
+++ b/tutor/scripts/test-server/backend.js
@@ -11,78 +11,91 @@ const os = require('os');
 faker.seed(12345);
 
 server.all('/*', function(req, res, next) {
-  res.setHeader('X-App-Date', now);
-  res.setHeader('Access-Control-Expose-Headers', 'X-App-Date');
-  next();  // call next() here to move on to next middleware/router
+    res.setHeader('X-App-Date', now);
+    res.setHeader('Access-Control-Expose-Headers', 'X-App-Date');
+    next();  // call next() here to move on to next middleware/router
 });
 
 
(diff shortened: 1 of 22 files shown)


tutor/scripts/test-server/backend.js
```diff
@@ -11,78 +11,91 @@ const os = require('os');
...
-  }
-  MULTI_HANDLERS.forEach((handler) => { handler.setRole(role); });
+    for (let route in GET_HANDLERS) {
+        GET_HANDLERS[route].setRole(role);
+    }
+    MULTI_HANDLERS.forEach((handler) => { handler.setRole(role); });
+}
+function resetState() {
+    for (let route in GET_HANDLERS) {
+        GET_HANDLERS[route].resetState();
```
camliu89, line 86: should we check if each route has a `resetState`? That way we don't have to always write a `resetState` in each of the future routes.

camliu89 (approved):
@nathanstitt left a comment, but it looks good to me.

tutor/scripts/test-server/backend.js
nathanstitt: no, I don't think we should use this unless we absolutely have to.

Jest runs multiple processes concurrently, so if we happen to call setState while another process is in the middle of building an assignment it will reset the the state that the other spec had carefully setup.

nathanstitt: ug I'm sorry, I completely mis-read your comment @camliu89 - I took it to mean adding it to beforeEach, but you were thinking of checking for the fn on the handler before calling it.  that's not a bad idea, I'll update to do that

PR #3484, opened 2021-03-08
Delelte subject block bug

camliu89:
Drop current instructor from all courses before deleting the block.
Make sure there are no students in any of the courses, past or current.

Waiting for the BE changes to pass the role.id instead of the roster.id to drop an instructor from a course.


tutor/src/screens/my-courses/dashboard/index.tsx
```diff
@@ -147,16 +150,31 @@ export const MyCoursesDashboard = () => {
...
-        if (!offeringHasCourses) {
+        const offeringCourses = filter(courses, c => c.offering_id === deleteOfferingIdModal)
+        const areAllPreviewCourses = every(offeringCourses, oc => oc.is_preview)
+        const areAllNonPreviewCoursesWithoutStudents = every(filter(offeringCourses, c => String(c.term) !== 'preview'), oc => sumBy(oc?.periods, p => {
+            return p.num_enrolled_students
+        }) === 0)
+        if (areAllPreviewCourses || areAllNonPreviewCoursesWithoutStudents) {
             return (
                 <DeleteOfferingModal
                     show={Boolean(deleteOfferingIdModal)}
```
nathanstitt, line 180: This component has gotten really large.  Can we move any of the functions into their own hook?

nathanstitt (commented):
@camliu89 looks pretty good but would like to shrink the component if possible, 200+ lines is way to large of a function.

Also can we add a spec for deleting an offering? I just made a PR #3483 to add some useful ways to set bootstrap data if that helps

tutor/src/screens/my-courses/dashboard/index.tsx
camliu89: I agree, I will break it down. thanks nathan!

camliu89: @nathanstitt moved the displayOfferingIds logic into a custom hook. That should shrink the file a little bit.

tutor/src/screens/my-courses/dashboard/useDisplayedOfferings.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { useState, useEffect, useMemo } from 'react'
```
nathanstitt, line 1: While I also personally prefer camelcase filenames that match the component/hook they contain, everything else is dashes so we should adhere to that standard vs ending up with a mismatch.   accordingly this should probably be: `use-displayed-offerings`

tutor/src/screens/my-courses/dashboard/index.tsx
```diff
@@ -109,54 +111,45 @@ const StyledMyCoursesDashboard = styled.div`
...
+            const offeringCourses = filter(courses, c => c.offering_id === deleteOfferingIdModal && String(c.term) !== 'preview')
+            Promise.all(map(offeringCourses, c => {
+                const currentRole = find(c.roles, r => r.type === 'teacher')
+                const currentTeacher = currentRole && find(c.teachers, t => t.role_id === currentRole.id);
+                if(currentTeacher) {
+                    dispatch(dropCourseTeacher(currentTeacher.id))
+                } else {
+                    Promise.resolve()
+                }
+            }))
```
nathanstitt, line 137: I think you need to return a promise here, either the result of dispatch or Promise.resolve.  

since the function that is iterating through map never returns anything you're ending up with `[undefined]` and promise.all won't wait on anything

nathanstitt (approved):
looks pretty good Alex, other than the returning promise issue.  Feel free to merge once it's fixed.

PR #3486, opened 2021-03-08
fix create course wizard's final step

nathanstitt:
Also do the same thing from the my-courses page where we need to reload after create course

### DIFF
--- a/tutor/specs/e2e/create-course.e2e.ts
+++ b/tutor/specs/e2e/create-course.e2e.ts
@@ -1,0 +1,74 @@
+import { visitPage, setTimeouts, setRole } from './helpers'
+
+describe('Create Course wizard', () => {
+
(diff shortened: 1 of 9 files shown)


tutor/specs/e2e/create-course.e2e.ts
```diff
@@ -0,0 +1,74 @@
...
+            await page.fill('#number-students', '10')
+            await expect(page).toHaveText('testEl=next-btn', 'Finish')
+            await page.click('testEl=next-btn')
+            await page.waitForNavigation()
+            expect(
+                await page.evaluate(() => document.location.pathname)
+            ).toMatch(/course\/\d+/)
+        })
+    })
+})
```
camliu89, line 74: nice tests @nathanstitt

PR #3488, opened 2021-03-09
Manage subjects bug

camliu89:
Disabled course item cards when in edit mode.
Persist the order of the offerings after a refresh.

Screen Shot 2021-03-09 at 11 55 46 AM

### DIFF
--- a/tutor/src/screens/my-courses/dashboard/use-displayed-offerings.tsx
+++ b/tutor/src/screens/my-courses/dashboard/use-displayed-offerings.tsx
@@ -11,9 +11,7 @@ const useDisplayedOfferings = () : ReturnUseDisplayedOfferings => {
     // getting all the data: offerings and courses
     const courses = useAllCourses()
     const offerings = useAvailableOfferings(courses)
-    const [displayedOfferingIds, setDisplayedOfferingIds] = useState<ID[]>(
-        UiSettings.get('displayedOfferingIds') || useDisplayedCourseOfferingIds()
-    )
+    const [displayedOfferingIds, setDisplayedOfferingIds] = useState<ID[]>(useDisplayedCourseOfferingIds())
 
     // update the `displayedOfferingIds` if users adds/delete offerings
     useEffect(() => {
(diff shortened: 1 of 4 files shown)


tutor/src/screens/my-courses/dashboard/use-displayed-offerings.tsx
```diff
@@ -12,7 +12,7 @@ const useDisplayedOfferings = () : ReturnUseDisplayedOfferings => {
     const courses = useAllCourses()
     const offerings = useAvailableOfferings(courses)
     const [displayedOfferingIds, setDisplayedOfferingIds] = useState<ID[]>(
-        useDisplayedCourseOfferingIds()
+        UiSettings.get('displayedOfferingIds') || useDisplayedCourseOfferingIds()
```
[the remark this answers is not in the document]
camliu89: @nathanstitt I didn't check the `useDisplayedCourseOfferingIds`, but I made some changes in there now.

tutor/src/store/courses.ts
```diff
@@ -108,7 +108,7 @@ export const useLatestCoursePreview = (offeringId: string | number) => useSelect
 export const useDisplayedCourseOfferingIds = () => {
     const courseOfferingIds = useAllCourses().map(c => c.offering_id)
     const selectedIds = UiSettings.get('displayedOfferingIds') || []
-    return union(courseOfferingIds, selectedIds)
+    return union(selectedIds, courseOfferingIds)
```
camliu89, line 111: @nathanstitt I swapped these two because the courseOfferingIds was basically overriding the selectedIds from the UISettings.

nathanstitt: hm, it seems like something else must be going on then.  maybe some of the ids are string and others number?

the order that things are passed to `union`  should not matter, for instance this should both be the same
`union([2], [1, 2,3])` and `union([1, 2],[2,3])` should both return `[1,2,3]`

nathanstitt: oh wait, maybe you mean the order that they're returned in is inconsistent?  I can see that happening sine union doesn't make any representations about the order.

nathanstitt (approved):

tutor/src/store/courses.ts
camliu89: yes, the selectedIds is in the order that the user changed. The courseOfferingIds is in the default order.

PR #3489, opened 2021-03-09
clear subject when typing a suggestion

jivey:
We decided to clear the selected subject when typing a suggestion, and clear the suggestion when selecting a subject, so that only 1 primary button will be visible on screen at any time.

### DIFF
--- a/tutor/specs/e2e/my-courses.e2e.ts
+++ b/tutor/specs/e2e/my-courses.e2e.ts
@@ -9,7 +9,8 @@ describe('My Courses', () => {
 
     it('allows a new teacher select and suggest subjects', async () => {
         await visitPage(page, '/courses')
-        await expect(page).toHaveSelector('testEl=existing-teacher-screen', { timeout: 1000 })
+        // The 500ms timeouts may be too low when using DEBUG, temp set to 1000 if you see issues
+        await expect(page).toHaveSelector('testEl=existing-teacher-screen', { timeout: 500 })
         await modifyBootstrapData(page, (data) => ({ ...data, courses: [] }))
         await visitPage(page, '/courses')
         await expect(page).toHaveSelector('testEl=new-teacher-screen', { timeout: 500 })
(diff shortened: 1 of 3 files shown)


tutor/specs/e2e/my-courses.e2e.ts
```diff
@@ -31,6 +31,15 @@ describe('My Courses', () => {
...
         ).toContain('onboarding=0')
+        await expect(page).not.toHaveSelector('testEl=show-detail', { timeout: 1000 })
+        await page.type('testEl=input-suggested-subject', 'test')
+        await page.click('testEl=offering-0', { force: true })
+        await expect(page).toHaveSelector('testEl=show-detail', { timeout: 1000 })
+        await expect(page).not.toHaveSelector('testEl=submit-suggested-subject', { timeout: 100 })
+        await page.type('testEl=input-suggested-subject', 'test')
+        await expect(page).not.toHaveSelector('testEl=show-detail', { timeout: 100 })
+        await page.click('testEl=offering-0', { force: true })
+        await page.click('testEl=show-detail')
```
nathanstitt, line 42: why the need to `force` the click?  also the `timeout` of 1000  (1 second) is a bit long, I normally use only `100`. Keep in mind that on a `.not.` will always wait that long then test for it to appear at the end of the timeout

nathanstitt (approved):

tutor/src/screens/my-courses/new-teacher.tsx
```diff
@@ -443,6 +443,10 @@ const SubjectSelect: React.FC<SubjectSelectProps> = ({
     }
 
     const onChangeSuggestion = (value) => {
+        if (value.length > 0) {
```
camliu89, line 446: I think you can do `if(value)` since `''` is a falsey value.

camliu89 (approved):

tutor/src/screens/my-courses/new-teacher.tsx
jivey: Ah good tip!

PR #3498, opened 2021-03-11
sort subjects by a specific order, sort offerings within subject alphabetically

jivey:
Fixes a bug where the subjects in the onboarding screen weren't displaying in the specific order we'd like, and also ensures the offerings within the subject are alphabetical.

### DIFF
--- a/tutor/src/models/subject-order.ts
+++ b/tutor/src/models/subject-order.ts
@@ -1,0 +1,8 @@
+export const SubjectOrder = [
+    'Math',
+    'Science',
+    'Social Sciences',
+    'Humanities',
+    'Business',
+    'AP® Courses',
+]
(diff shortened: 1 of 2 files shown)


jivey:
Example (ignore the missing covers)
image

tutor/src/models/subject-order.ts
```diff
@@ -0,0 +1,8 @@
+export const SubjectOrder = [
+    'Math',
+    'Science',
+    'Social Sciences',
+    'Humanities',
+    'Business',
+    'AP® Courses',
+]
```
camliu89, line 8: I wonder if there is a way to persist the order from the database, instead of in-memory.

camliu89 (approved):

tutor/src/models/subject-order.ts
jivey: Yeah that'd be nice, maybe something we can add to admin sometime (might be nice to create a Subject model then too.)

PR #3500, opened 2021-03-12
No preview bug

camliu89:
catching the errors in the redux api calls using the flux error handlers.


jivey (approved):
Works for me locally, nice fix!

PR #3502, opened 2021-03-12
Create course wizard bug

camliu89:
Moves to the term step if there is an `offeringId` preselected.

### DIFF
--- a/tutor/specs/screens/new-course/ux.spec.js
+++ b/tutor/specs/screens/new-course/ux.spec.js
@@ -46,6 +46,13 @@ describe('Course Builder UX Model', () => {
         expect(ux.newCourse.save).toHaveBeenCalled();
     }
 
+    it('goes to the next term step if there is a preselected offeringId in the params', () => {
+        const course = courses.array[0];
+        testRouter.match.params = { offeringId: course.offering_id };
+        ux = createTestUX();
+        expect(ux.stage).toEqual('term');
+    });
+
     it('sets cloned course when sourceId is present', () => {
         const course = courses.array[0];
         testRouter.match.params = { sourceId: course.id };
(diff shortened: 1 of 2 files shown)


tutor/specs/screens/new-course/ux.spec.js
```diff
@@ -46,6 +46,13 @@ describe('Course Builder UX Model', () => {
...
     }
 
+    it('goes to the next term step if there is a preselected offeringId in the params', () => {
+        const course = courses.array[0];
+        testRouter.match.params = { offeringId: course.offering_id };
+        ux = createTestUX();
+        expect(ux.stage).toEqual('term');
+    });
+
     it('sets cloned course when sourceId is present', () => {
```
nathanstitt, line 56: nice spec!

nathanstitt (approved):

tutor/src/screens/new-course/ux.js
```diff
@@ -91,9 +91,9 @@ export default class CourseBuilderUX extends BaseModel {
   }
 
   @action.bound onOfferingsAvailable() {
-      if (this.preselectedAppearanceCode) {
+      if (this.preselectedOfferingId) {
           this.newCourse.offering = find(this.offerings.available.array,
-              { appearance_code: this.preselectedAppearanceCode }
+              { offering_id: this.preselectedOfferingId }
           );
```
nathanstitt, line 97: @camliu89 sorry I should have tested locally before approving but this has a bug :(   offerings do not have `offering_id`, they have just `id`, so this should be: 
```
this.newCourse.offering = find(this.offerings.available.array,
               { id: this.preselectedOfferingId }
           );
```
However! we've ran into problems with using lodash find before, it uses triple equal and will fail if id is a number and `preselectedOfferingId` is a string.

How about:
```
this.newCourse.offering = this.offerings.available.array.find(o => o.id == this.preselectedOfferingId)
```

camliu89: yeah, that is correct. thanks nathan.

PR #3504, opened 2021-03-16
Subject blocks sort

camliu89:
1. Sort subject blocks by number, course availability, and preview course availability.
2. When adding a new subject, add it at the top. Then scroll the window to the top.

To test: delete the `UISettings.displayedOfferingIds` so it does not show the pre-set order of the offeringIds.


tutor/src/screens/my-courses/dashboard/use-displayed-offerings.tsx
```diff
@@ -1,17 +1,17 @@
...
+import { useAvailableOfferings, useDisplayedOfferingsIds } from '../../../store/offering'
 
 type ReturnUseDisplayedOfferings = [ID[], React.Dispatch<React.SetStateAction<ID[]>>, Offering[], (offeringId: ID, flow?: string) => void]
 
 const useDisplayedOfferings = () : ReturnUseDisplayedOfferings => {
     // getting all the data: offerings and courses
     const courses = useAllCourses()
     const offerings = useAvailableOfferings(courses)
-    const [displayedOfferingIds, setDisplayedOfferingIds] = useState<ID[]>(useDisplayedCourseOfferingIds())
+    const [displayedOfferingIds, setDisplayedOfferingIds] = useState<ID[]>(map(useDisplayedOfferingsIds(courses)))
```
nathanstitt, line 14: what's the purpose of the `map` here?

tutor/src/store/offering.ts
```diff
@@ -1,10 +1,17 @@
...
 import { Offering, Course, ID } from './types'
+import UiSettings from 'shared/model/ui-settings'
 
 const offeringAdapter = createEntityAdapter<Offering>({
-    sortComparer: (a, b) => a.default_course_name.localeCompare(b.default_course_name),
+    sortComparer: (a, b) => {
+        //send an offering that is not available to the bottom of the list
+        if(!a.is_available && !a.is_preview_available) return 1
+        //if the number is higher, it is low priority
+        return b.number.localeCompare(a.number)
```
nathanstitt, line 13: Ok, this needs to be extracted to an external function and a unit spec written for it so we can fully test it.
 
I think we'd decided that only non available should appear at the bottom, and preview available should be sorted normally.    But if both offerings are not available we need to compare them to each other, and no need to use localeCompare since we're sorting by a number.
 
It'll need to be something like this, but I'd expect this to have some bug, hence the need for spec.
```
if ( a.is_available == !b.is_available) return Number(a.number)-Number(b.number);
if (a.is_available && !b.is_available) return -1;
return 1
```

tutor/src/screens/my-courses/dashboard/use-displayed-offerings.tsx
camliu89: nothing, I think I forgot to delete it when I was moving logic around.

tutor/src/screens/my-courses/dashboard/add-subject-dropdown.tsx
```diff
@@ -2,17 +2,30 @@ import React from 'react'
...
+    const courses = useAllCourses()
+    const allOfferings = useAvailableOfferings(courses)
     const offeringsBySubject = groupBy(allOfferings, o => o.subject)
+
+    const onSelectOffering = (offeringId: ID) => {
+        // adding the subject block at the top
+        setDisplayedOfferingIds(prevState => [offeringId, ...prevState])
+        // then scroll to the top to see the new offering block
+        scroller.scrollToTop()
+    }
```
nathanstitt, line 27: scroller has a `deferred` option that will wait 10ms then scroll.  We should use that here because react set state happens asynchronously and the offering probably hasn't been added by the time scroll starts.

nathanstitt (approved):

PR #3506, opened 2021-03-16
update delete course modal message

camliu89:
Screen Shot 2021-03-16 at 1 46 35 PM
Screen Shot 2021-03-16 at 1 46 53 PM

### DIFF
--- a/tutor/specs/e2e/course-settings.e2e.ts
+++ b/tutor/specs/e2e/course-settings.e2e.ts
@@ -27,7 +27,8 @@ describe('Course Settings', () => {
             await expect(page).toHaveSelector('.course-detail-settings-form')
             const courseName = page.$eval('input#course-name', (el: any) => el.value)
             await page.click('testEl=delete-course-btn')
-            await expect(page).not.toHaveSelector('testEl=disabled-delete-course-message', { timeout: 100 })
+            await expect(page).toHaveText('testEl=delete-course-message', 'delete', { timeout: 100 })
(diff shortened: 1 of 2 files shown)


nathanstitt (changes requested):
Per https://github.com/openstax/tutor/issues/1559 , the no students message should read "delete this course" and the one with students should read "leave this course".

In https://github.com/openstax/tutor-js/pull/3507 I added a spec for the no-students case.  Lets get that merged, then add a check for the wording

camliu89:
👀 nice catch.

tutor/specs/e2e/course-settings.e2e.ts
```diff
@@ -27,7 +27,8 @@ describe('Course Settings', () => {
             await expect(page).toHaveSelector('.course-detail-settings-form')
             const courseName = page.$eval('input#course-name', (el: any) => el.value)
             await page.click('testEl=delete-course-btn')
-            await expect(page).not.toHaveSelector('testEl=disabled-delete-course-message', { timeout: 100 })
+            await expect(page).toHaveText('testEl=delete-course-message', 'delete', { timeout: 100 })
```
nathanstitt, line 30: we don't need to use use `timeout` option on a `toHaveXX` expectation.  it's only needed on a `not.toHaveXX` one.  If we do not specify it on the `not` then jest will wait until it timeouts just in case it becomes true at some point

nathanstitt (approved):

PR #3507, opened 2021-03-16
reload page after leaving course

nathanstitt:
needed so that all the stores/models get updated without the course


camliu89 (approved):
it works locally on my machine.

PR #3509, opened 2021-03-16
Filter out concept coach courses

nathanstitt:
They should no longer be displayed anywhere


tutor/specs/e2e/mocker.ts
```diff
@@ -83,17 +83,16 @@ class Mock {
 
     constructor({ page, routes, data, options }: MockArgs) {
         this.page = page
-        this.routes = routes
-        this.mockData = data || {}
-        this.options = options || {}
+        this.routes = clone(routes)
+        this.mockData = clone(data || {})
+        this.options = clone(options || {})
```
camliu89, line 88: why is the `clone` needed in here?

nathanstitt: otherwise if you pass in a object and then mutate it with `mock.course` or `mock.bootstrapData` it could mutate the original and the mutation would persist across specs.  I had that happen with cleaning the courses in one spec also resetting it in others.

camliu89: oh, the whole reference data types... 👍

camliu89 (approved):

PR #3512, opened 2021-03-18
WIP: Upgrade MobX 

nathanstitt:
Use the new modeled-mobx lib


nathanstitt:
closing in favor of #3514 which has better structured commits

PR #3516, opened 2021-03-22
Use longest lasting preview course that is not expired

nathanstitt:
Previously it was returning unusable preview courses


Dantemss (approved):

camliu89 (approved):
this looks good @nathanstitt

PR #3518, opened 2021-03-22
delete also all preview courses when offering block is deleted

camliu89:
Bug: Offering block is deleted but it re-appears when the user refreshes the page.
Cause: User is not getting dropped from all the preview courses when an offering is deleted. The subject block will keep re-appearing in the dashboard because there are courses (in this case preview courses) available for the deleted offering.


nathanstitt (changes requested):
Can you rebase main to get #3517 into this, I hope it fixes the specs.

Then can you add some lines to the my-courses spec to remove the offering, click into a course, then back to my-courses and verify the offering isn't present, then do a `page.reload` and re-verify.  The reload test might not work though, I'm not sure what the fake backend does in that case.

camliu89:
@nathanstitt testing refreshing the page works.

tutor/specs/e2e/my-courses.e2e.ts
```diff
@@ -117,6 +118,46 @@ describe('My Courses', () => {
...
+            // delete an offering block
+            await visitPage(page, '/courses')
+            await page.click('testEl=dashboard-settings-btn')
+            await page.click(`.offering-container[data-offering-id="${offeringIdWithEmptyStudents}"] >> testEl=delete-offering`)
+            await expect(page).toHaveSelector('testEl=delete-offering-modal')
+            await page.click('testEl=delete-offering-modal-btn')
+            await expect(page).not.toHaveSelector(`.offering-container[data-offering-id="${offeringIdWithEmptyStudents}"]`)
+
+            // go to a course
+            await page.click('testEl=dashboard-settings-btn', { force: true })
```
nathanstitt, line 147: why is force needed?  can you add a comment with the reason so we'll know if/when it's safe to remove it.

tutor/specs/e2e/my-courses.e2e.ts
```diff
@@ -117,6 +118,46 @@ describe('My Courses', () => {
...
+            ).toMatch(`/course/${courseIdWithStudents}`)
+
+            // go back to dashboard and make sure the offering block is not shown
+            await page.goBack()
+            await expect(page).not.toHaveSelector(`.offering-container[data-offering-id="${offeringIdWithEmptyStudents}"]`, { timeout: 10 })
+
+            // refreshing the page and make sure the offering block is not shown
+            await visitPage(page, '/courses')
+            await expect(page).not.toHaveSelector(`.offering-container[data-offering-id="${offeringIdWithEmptyStudents}"]`, { timeout: 10 })
+        })
```
nathanstitt, line 160: great job remembering the timeouts, but 10ms is really short for the page to have rendered.  I'd expect it to still be showing a blank screen at that point so of course it won't have the selector.  Maybe use 100?

tutor/specs/e2e/my-courses.e2e.ts
```diff
@@ -117,6 +118,46 @@ describe('My Courses', () => {
             await expect(page).toHaveSelector('testEl=existing-teacher-screen')
             await expect(page).not.toHaveText('testEl=course-card', 'ConceptCoach', { timeout: 100 })
         })
+
+        it.only('deletes offering', async () => {
```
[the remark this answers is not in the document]
camliu89: gahh, i should have that library added in. I am still working on it.

tutor/specs/e2e/my-courses.e2e.ts
camliu89: it is not really needed, I was just testing out something and I forgot to take it out.

PR #3521, opened 2021-03-24
add ap-disclaimer in the subject selection for new users

camliu89:
Screen Shot 2021-03-24 at 1 02 34 PM

### DIFF
--- a/tutor/src/screens/my-courses/new-teacher.tsx
+++ b/tutor/src/screens/my-courses/new-teacher.tsx
@@ -166,6 +166,17 @@ const Content = styled.div`
             }
         }
     }
+
+    .ap-disclaimer {
+        width: 592px;
+        margin: 4rem 0 2rem;
+        
+        p {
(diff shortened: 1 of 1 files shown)


tutor/src/screens/my-courses/new-teacher.tsx
```diff
@@ -166,6 +166,17 @@ const Content = styled.div`
             }
         }
     }
+
+    .ap-disclaimer {
+        width: 592px;
```
nathanstitt, line 171: why do we need to set the width to such a strange value?  Are we attempting to cause the para to break or something? If so, it'd be better to add 2 paras IMO

nathanstitt (approved):

tutor/src/screens/my-courses/new-teacher.tsx
camliu89: that is the width that is set in the offering containers.

PR #3523, opened 2021-03-25
Mobx upgrade find/replace

Dantemss:
Some substitutions to help with the upgrade

### DIFF
--- a/tutor/src/models/course.ts
+++ b/tutor/src/models/course.ts
@@ -1,4 +1,4 @@
-import { BaseModel, field, identifier, hasMany, modelize } from 'shared/model';
+import { BaseModel, field, identifier, model, modelize } from 'shared/model';
 import {
     sumBy, first, sortBy, find, get, endsWith, capitalize, filter, pick, isEmpty,
 } from 'lodash';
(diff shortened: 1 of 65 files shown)


tutor/src/models/course.ts
```diff
@@ -95,19 +93,19 @@ export default class Course extends BaseModel {
...
-    @hasMany({ model: Role, inverseOf: 'course', extend: getters({
+    @model(Role) roles; /* extend: getters({
         student() { return find(this, { isStudent: true }); },
         teacher() { return find(this, { isTeacher: true }); },
         teacherStudent() { return find(this, { isTeacherStudent: true }); },
-    }) }) roles;
-    @hasMany({ model: Student, inverseOf: 'course' }) students;
-    @hasMany({ model: TeacherProfiles, inverseOf: 'course' }) teacher_profiles;
+    }) */
+    @model(Student) students;
```
nathanstitt, line 107: hm, hasMany needs to be set to an empty array, that's how the model code knows how it should be treated.  So this should become `@model(Student) students = []`

You can also omit `;` if it's easier, we no longer require them

PR #3524, opened 2021-03-25
Exercise ap tags

camliu89:
Screen Shot 2021-03-25 at 11 38 29 AM

### DIFF
--- a/tutor/specs/e2e/question-library.e2e.ts
+++ b/tutor/specs/e2e/question-library.e2e.ts
@@ -1,0 +1,25 @@
+import { Mocker, visitPage, setTimeouts } from './helpers'
+
+describe('Question Library', () => {
+
+    Mocker.mock({
+        page,
+        routes: {},
+    })
+
+    beforeEach(async () => {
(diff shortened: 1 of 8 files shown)


tutor/specs/e2e/question-library.e2e.ts
```diff
@@ -0,0 +1,25 @@
...
+    Mocker.mock({
+        page,
+        routes: {},
+    })
+
+    beforeEach(async () => {
+        await setTimeouts()
+        await visitPage(page, '/course/1/questions')
+        await page.evaluate(() => {
+            window._MODELS.feature_flags.set('tours', false);
```
nathanstitt, line 14: is feature_flags needed?  mocker should do that?

nathanstitt (approved):

tutor/specs/e2e/question-library.e2e.ts
camliu89: oh I didn't know Mocker takes care of that. Thanks.

PR #3525, opened 2021-03-25
reset formats and answers when switching question type

camliu89:
1. Reset the formats and answers when the question type is changed.

### DIFF
--- a/shared/specs/model/exercise/question.spec.js
+++ b/shared/specs/model/exercise/question.spec.js
@@ -11,9 +11,7 @@ describe('Exercise Question', () => {
         question.setExclusiveFormat('true-false');
         expect(map(question.formats, 'value')).toEqual(['true-false']);
         question.setExclusiveFormat('multiple-choice');
-        expect(map(question.formats, 'value')).toEqual(['free-response', 'multiple-choice']);
-        question.setExclusiveFormat('multiple-choice');
-        expect(map(question.formats, 'value')).toEqual(['free-response', 'multiple-choice']);
+        expect(map(question.formats, 'value')).toEqual(['multiple-choice', 'free-response']);
         question.setExclusiveFormat('true-false');
         expect(map(question.formats, 'value')).toEqual(['true-false']);
         question.setExclusiveFormat('open-ended');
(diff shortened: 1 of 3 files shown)


shared/specs/model/exercise/question.spec.js
```diff
@@ -78,9 +76,9 @@ describe('Exercise Question', () => {
         question.collaborator_solution_html = '';
         expect(question.collaborator_solutions).toHaveLength(0);
     });
-    it('uses answers for detecting multi-choice', () => {
+    it('uses formats for detecting multi-choice', () => {
         expect(question.isMultipleChoice).toBe(true);
-        question.answers = [];
+        question.formats = ['free-response'];
         expect(question.isMultipleChoice).toBe(false);
```
nathanstitt, line 82: we do not want to switch to using formats for determining multiple-choice.  that should continue to rely on if the exercise has any answers, that way it matches what tutor-server does.  cc @Dantemss to confirm the BE still does that though.

This task was about clearing the answers array when the exercise is saved with it's format not set to multiple-choice

shared/src/model/exercise/question.js
```diff
@@ -114,22 +114,32 @@ class ExerciseQuestion extends BaseModel {
...
+          case 'true-false':
+              if (this.answers.length == 0) {
+                  this.answers.push({ content_html: 'True' });
+                  this.answers.push({ content_html: 'False' });
+              }  
+              break;
       }
-      this.formats = uniq(formats).sort();
+
+      uniq(this.formats).sort();
```
nathanstitt, line 142: we do not want modify the answers until the exercise is saved.  Image the case of an editor who types up very long answers, then inadvertently hits the 'free-response' radio input, quickly switches back to multiple-choice and has lost all their work.

nathanstitt (changes requested):
Hey Alex, I think you mis-understood the acceptance criteria on the card:

> Expected behavior:
When you switching to the Open/Ended or True/False option, answers should be deleted.

We only remove answer choices upon saving the question
When user clicks open-ended, we hide the answer choices, but save them in browser memory
When user saves or publishes, we remove answer choices
If user clicks back to multiple-choice option, answer choices will reappear.
Steps to Reproduce the issue:
1. Create an exercise and choose the Multiple Choice option.
2. And some answers.
3. Switch to the Open-Ended or True/False option.
4. Answers are still persisted in the exercise card.

camliu89:
thanks nathan. for some reason, I didn't see the bullet points in the card that says to keep the answers.

camliu89:
Closing this. This change needs a design session to talk about some stuff.

PR #3533, opened 2021-04-02
fix chapter section spec

jivey:
Minimum to get `yarn test tutor tutor/specs/models/chapter-section.spec.ts` passing but not sure if this `instanceof` check is the right way to think about this in Typescript, feedback appreciated! :)

### DIFF
--- a/tutor/src/models/chapter-section.ts
+++ b/tutor/src/models/chapter-section.ts
@@ -7,11 +7,13 @@ export default class ChapterSection extends BaseModel {
 
     @readonly static blank = new ChapterSection()
 
-    constructor(arg: [number, number] | string = '') {
+    constructor(arg: ChapterSection | [number, number] | string = '') {
         super();
         modelize(this);
         if (isArray(arg)) {
             [this.chapter, this.section] = arg;
+        } else if (arg instanceof ChapterSection) {
+            [this.chapter, this.section] = arg.split();
         } else {
             [this.chapter, this.section] = arg.split('.').map(Number);
         }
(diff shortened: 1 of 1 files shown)


tutor/src/models/chapter-section.ts
```diff
@@ -68,7 +70,6 @@ export default class ChapterSection extends BaseModel {
         return reduceRight(
             this.asArray,
             (memo:number, num: number, index: number) => memo + (num * Math.pow(100, index + 1)),
-            0,
```
jivey, line 71: The original file didn't have this and it caused the spec to fail by adding trailing zeroes to the output.

nathanstitt: nice! ts caught what was probably a bug there

tutor/src/models/chapter-section.ts
```diff
@@ -7,11 +7,13 @@ export default class ChapterSection extends BaseModel {
...
 
-    constructor(arg: [number, number] | string = '') {
+    constructor(arg: ChapterSection | [number, number] | string = '') {
         super();
         modelize(this);
         if (isArray(arg)) {
             [this.chapter, this.section] = arg;
+        } else if (arg instanceof ChapterSection) {
+            [this.chapter, this.section] = arg.split();
         } else {
```
nathanstitt, line 17: yep, `instanceof` is pretty standard, TS will notice you used it and then treat everything in the block as that type.   There's also the concept of [type guards](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#typeof-type-guards)  if found ourselves doing this a lot we could make a `isChapterSection` or something like that

nathanstitt (approved):

PR #3544, opened 2021-04-06
question model - use clear and replace to work around creating an observable typed array

jivey:


camliu89 (approved):

nathanstitt (approved):

jivey:
Hmm looks like these changes might already be in the integration branch, that's my bad.

PR #3551, opened 2021-04-09
Add an autosave to the question library editor

jivey:
- Store editor field state in localStorage. MobX's `autorun` handles firing the autosave when the observables change. A delay might be desirable in the future in case events need to debounce, but since it only occurs on blur or checking boxes, it seems to be performant.
- Set pageIds as a query param so that refreshing the page will put you on the same screen
- Fix a bug where if you started filling out MCQ options, but then switched the tab to Written-response and published, it would create the question as an MCQ, due to it still sending the `options` in the form.

AUTOSAVE_TTL is set to 15 minutes and used to help expired `signed_id`s from being used on uploaded images. AUTOSAVE_VERSION is used in case the form fields ever change significantly. In both cases, the validity (and parsability with a `try...catch`) will be checked before loading, and will clear the autosave if it doesn't match those conditions.

### DIFF
--- a/tutor/src/screens/question-library/dashboard.jsx
+++ b/tutor/src/screens/question-library/dashboard.jsx
@@ -20,7 +20,7 @@ class QuestionsDashboard extends React.Component {
           open: PropTypes.func,
       }),
       history: PropTypes.shape({
-          push: PropTypes.func,
+          replace: PropTypes.func,
       }).isRequired,
   };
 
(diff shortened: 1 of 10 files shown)


tutor/src/screens/question-library/dashboard.jsx
```diff
@@ -37,11 +69,18 @@ class QuestionsDashboard extends React.Component {
...
 
   @action.bound onSelectSections() {
       this.showingDetails = false;
       this.isShowingExercises = false;
+      this.setPageIdsQuery()
+  }
+
+  @action.bound setPageIdsQuery(pageIds = []) {
+      const query = extend(Router.currentQuery(this.windowImpl), { pageIds: pageIds.join(',') })
+      this.props.history.push(this.windowImpl.location.pathname + '?' + qs.stringify(query))
```
nathanstitt, line 83: we probably want to use `history.replace` here, otherwise we'll create multiple entries in history and the user will have to hit back twice or more to get back to the previous page.

nathanstitt (approved):
looks good !

PR #3556, opened 2021-04-13
Added different LO patterns for World History

Dantemss:
The error now shows up immediately when a new LO is added since the value doesn't match the pattern, but what do you think?


jivey (approved):
Works for my locally under normal conditions, but I did notice that if you created the exercise  as WH and then switch to a different book, it checks the old validation:

image

Dunno if that's worth worrying about since it can be fixed by deleting the old LO tag. Related, I we still allow publishing if the validation doesn't pass, also not sure if that's a concern.

Dantemss:
Hm I thought I had fixed this by making the errorMsg computed but I'll re-test on Monday.

nathanstitt (approved):

Dantemss:
Ah I had tested changing the book on the LO component but never the global book. I'll see if I can fix.

nathanstitt (approved):

PR #3560, opened 2021-04-17
WIP Mobx component specs exercises

jivey:


tutor/specs/components/exercises/sectionizer.spec.js
```diff
@@ -1,27 +1,22 @@
 import { FakeWindow } from '../../helpers';
 import Sectionizer from '../../../src/components/exercises/sectionizer';
-import ChapterSection from '../../../src/models/chapter-section';
+import { ChapterSection } from '../../../src/models/chapter-section';
```
nathanstitt, line 3: everything should import from `models`, not from the individual files.  By doing so, we can ensure that things import in the correct order.  This needs to be:
`import { ChapterSection } from '../../../src/models`

nathanstitt (approved):
Looks great Josiah!

tutor/specs/components/exercises/sectionizer.spec.js
jivey: Thanks!

PR #3563, opened 2021-04-19
Removed the admin stats page

Dantemss:
Now handled in metabase.
Please check that I removed ONLY the admin stats page :)


nathanstitt:
thanks @Dantemss  but I think we've already removed this in the mobs-upgrade branch. Lets leave this open though so we can compare after that's merged to make sure we get it all

Dantemss:
Removed in mobx branch

PR #3569, opened 2021-04-21
Mobx upgrade components

jivey:
Some of these aren't fully passing but have some of the other changes we need, maybe it'll be easier to discuss and fix after.

### DIFF
--- a/tutor/src/components/lms-info-card.js
+++ b/tutor/src/components/lms-info-card.js
@@ -3,7 +3,7 @@ import React from 'react';
 import { observer } from 'mobx-react';
 import { action, computed, modelize } from 'shared/model'
 import { Button } from 'react-bootstrap';
-import moment from 'moment';
+import { Time } from '../models';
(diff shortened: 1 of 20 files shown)


tutor/src/components/lms-info-card.js
```diff
@@ -52,21 +52,24 @@ export default class LmsInfoCard extends React.Component {
...
     renderDueDates() {
         const { plan } = this.props;
+
         if (plan.areTaskingDatesSame) {
+            const value = new Time(plan.dateRanges.due.start).toFormat(DUE_FORMAT)
             return (
                 <CopyOnFocusInput
                     label="Due date"
-                    value={moment(plan.dateRanges.due.start).format(DUE_FORMAT)}
+                    value={value.toLowerCase()}
```
jivey, line 61: toLowerCase here because Luxon doesn't seem to have a lowercase AM/PM token.

nathanstitt (approved):

PR #3571, opened 2021-04-22
minor updates to get teacher UI rendering

nathanstitt:


tutor/src/routes.js
```diff
@@ -47,11 +47,9 @@ const getRoutes = (router) => {
                             ],
                         },
                     ],
-                }, {
-                    path: 'metrics/:id', name: 'reviewTask',
-                    renderer: r(() => import('./screens/teacher-review-metrics')) },
```
jivey, line 52: Oh yeah we removed this screen right?

tutor/src/screens/assignment-review/ux.js
```diff
@@ -312,9 +315,11 @@ export default class AssignmentReviewUX {
 
     @action.bound async onSavePlan() {
         await this.editUX.savePlan();
-        Object.assign(this.planScores, pick(this.editUX.plan, ['title', 'description']));
-        this.planScores.grading_template = this.editUX.plan.gradingTemplate;
-        this.isDisplayingEditAssignment = false;
+        runInAction(() => {
```
jivey, line 318: I didn't realize we'd need to use `runInAction` inside something decorated with `@action`? Or is this to process everything a single transaction?

jivey (approved):

tutor/src/screens/assignment-review/ux.js
nathanstitt: yah this is a tricky one, the function is async, which means babel will actually splits it into multiple functions, and the code after the `await` will not be enclosed inside the main action.   https://mobx.js.org/actions.html#asynchronous-actions explains it a bit more

nathanstitt: there's a new way to do this using `flow` but I haven't explored it yet

PR #3572, opened 2021-04-23
Fix answer alignment

jivey:
It's harder to see on the single line example, but the alignment is now on the baseline, so the answer letter baseline should match the answer text. For the checkmark alignment I just went with what felt right, easy to tweak.

This also fixes a FireFox-only bug in Submission Overview with the letter getting squashed when the adjacent text was long.

Before:
image
image

After:
image
image

Before:
image

After:
image


nathanstitt (approved):
looks so much nicer, thanks @jivey

PR #3574, opened 2021-04-23
allow deleting feature and module tags in exercises

jivey:
The card 1624 was for only CNX Feature, but it seems like we should do this for the CNX Module tag as well.

image


nathanstitt:
perfect, yes we should allow deleting everything really

nathanstitt (approved):

PR #3578, opened 2021-04-28
use tutor-server Docker

Dantemss:
Example of how to load the tutor-server docker as a GH action
You'll have to connect the e2e tests to the server that spins up on localhost:3001


nathanstitt (approved):

tutor/specs/e2e/add-edit-question.e2e.js
```diff
@@ -59,7 +56,7 @@ describe('Add/Edit Questions', () => {
         await page.click('testEl=publish-btn')
     })
 
-    it('autosaves', async() => {
+    xit('autosaves', async() => {
```
nathanstitt, line 59: @jivey this spec is failing because when it reloads it goes back to the section picker.

It looks like we've somehow lost the code you'd written to set the page ids in the url when you go through the QL.  Can you check that?

PR #3581, opened 2021-04-29
Removed num_questions_dropped and points_dropped

Dantemss:
These fields were incorrectly calculated and they are no longer sent by the backend.
We can fix the code and add them back if needed.


nathanstitt (approved):
thanks for catching that!

jivey (approved):

PR #3582, opened 2021-04-29
Fix more mobx issues in QL

jivey:


tutor/src/components/add-edit-question/ux.js
```diff
@@ -530,10 +530,12 @@ export default class AddEditQuestionUX {
...
-            this.disableAutosave();
-            this.changed = false;
-            this.enableAutosave();
+            runInAction(() => {
+                this.clearAutosave();
+                this.disableAutosave();
+                this.changed = false;
+                this.enableAutosave();
+            })
         });
```
nathanstitt, line 539: this works fine, but you can also wrap functions in an action:
```js
await this.exercises.createExercise(this.toFormData).then(action(() => {
   this.clearAutosave();
                 this.disableAutosave();
                 this.changed = false;
                 this.enableAutosave();
})
```

nathanstitt (approved):

tutor/src/components/add-edit-question/ux.js
jivey: Oh wow I like that pattern a lot better 👍

PR #3583, opened 2021-04-30
Fix plan update

jivey:
- Fix updating an existing task plan. `settings` was being passed as null (BE doesn't like that,) so I filtered out undefined/null values with `identity`.
- Fix the plan details popup, use Luxon token for 3-letter day of week ('Wed', 'Thu')


nathanstitt (approved):
Thanks Jivey!

PR #3584, opened 2021-05-03
refetch taskPlan so the scores table updates with new extension data

jivey:
I noticed that the Extension icon tooltip doesn't get updated when you grant an extension. Not sure if refetching the task plan here is too much, but it does fix the bug.

### DIFF
--- a/tutor/src/screens/assignment-review/ux.js
+++ b/tutor/src/screens/assignment-review/ux.js
@@ -209,6 +209,7 @@ export default class AssignmentReviewUX {
         if (extensions.length > 0) {
             await this.taskPlan.grantExtensions(extensions);
         }
+        await this.taskPlan.fetch();
         await this.planScores.fetch();
         this.cancelDisplayingGrantExtension();
     }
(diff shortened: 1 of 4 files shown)


tutor/src/screens/assignment-review/ux.js
```diff
@@ -209,6 +209,7 @@ export default class AssignmentReviewUX {
         if (extensions.length > 0) {
             await this.taskPlan.grantExtensions(extensions);
         }
+        await this.taskPlan.fetch();
```
nathanstitt, line 212: instead of refetching, have you investigated what `grantExtension` does with the data that's returned?  Maybe we need to update the taskPlan with it, or set a date somewhere after the request completes.

jivey: Yeah that's why I put this into draft, refetching feels like too much when we can probably do as you say and update the data. I'll investigate more with DropAny work and update this PR with a better approach.

jivey:
Closing this and added a card in New Issues to work out a better approach.

PR #3585, opened 2021-05-04
fix course update method and sending lms setting when unchanged

jivey:
- Fixes the method and path for updating course
- I had to filter out the LMS setting even though it was sent as false — the backend throws an error about the setting change not being allowed.

### DIFF
--- a/tutor/src/models/course.ts
+++ b/tutor/src/models/course.ts
@@ -311,8 +311,9 @@ export class Course extends BaseModel {
         return find(this.roster.teachers, t => t.role_id === teacherRole.id);
     }
 
-    @computed get saveAttrs() {
-        return this.is_access_switchable ? SAVEABLE_ATTRS : without(SAVEABLE_ATTRS, 'is_lms_enabled')
+    @computed get saveData() {
+        const attrs = this.is_access_switchable ? SAVEABLE_ATTRS : without(SAVEABLE_ATTRS, 'is_lms_enabled')
+        return pick(this, attrs)
     }
 
     // called by API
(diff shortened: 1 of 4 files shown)


tutor/src/models/course.ts
```diff
@@ -320,7 +324,7 @@ export class Course extends BaseModel {
     async save() {
         const data = await this.api.request<CourseData>(
             this.isNew ? urlFor('createCourse') : urlFor('updateCourse', { courseId: this.id }),
-            { data: pick(this, SAVEABLE_ATTRS) },
+            { data: pick(this, this.saveAttrs) },
```
nathanstitt, line 327: Since the method's only called from one spot, how about making the method `saveData` and have it do the `pick`

tutor/src/screens/course-settings/course-details.js
```diff
@@ -145,9 +145,12 @@ const CourseDetails = observer(({ course, history }) => {
...
     const onSubmit = ({ courseName, courseCode, timezone }) => {
         setIsSaving(true);
-        course.name = courseName;
-        course.code = courseCode;
-        course.timezone = timezone;
+        runInAction(() => {
+            course.name = courseName;
+            course.code = courseCode;
+            course.timezone = timezone;
+        })
```
nathanstitt, line 152: I think you can make `onSubmit` be an action and then you wouldn't need the runInAction.  The only time you can't do that is if the method is async and the changes occur after the 1st await statement.  In that case babel compiles the remainder of the method to be called in a separate function which isn't decorated as an action so you have to do it yourself

nathanstitt (approved):

jivey:
Good suggestions! Pushed an update with them 👍

PR #3590, opened 2021-05-06
fix create-course and preview-courses e2e specs

jivey:


tutor/specs/e2e/helpers.ts
```diff
@@ -92,6 +92,7 @@ export const findOrCreateTeacherAccount = async ({
         `puts User::FindOrCreateUser[
 username:'${userName}',
 role:'${role}',
+is_test:true,
```
jivey, line 95: Without this, creating a preview course crashes because of a track event related routine.

nathanstitt (approved):

PR #3596, opened 2021-05-07
[Drop Any] Replaced score heading question_id and exercise_id with arrays

Dantemss:
Handles blue triangles and orange icons in homework review screens.
I believe some additional work will be required for the reading review screens.


tutor/src/screens/assignment-review/homework-scores.js
```diff
@@ -170,14 +174,18 @@ const AssignmentHeading = observer(({ ux, heading }) => (
...
+                    data-question-id={`Q${heading.question_id}`}
+                    tooltip={heading.everyQuestionZeroed ? 'Points changed to 0' :
+                             (heading.everyQuestionFullCredit ? 'Full credit given to all students':
+                             'Question dropped for one or more students')}
+                />}
+                {heading.someQuestionsDropped &&
+                     !heading.everyQuestionZeroed &&
+                     !heading.everyQuestionFullCredit ?
+                <OrangeInfoIcon/> : ScoresHelper.formatPoints(heading.points)}
             </HeadingBottom>
```
nathanstitt, line 189: I'm not a big fan of using conditional logic in the middle of a bunch of JSX.  To me it seems to make it difficult to reason about what's actually rendering.

How about moving most of the logic into methods on the header class, and then making a `IndicatorTriangle` or something that does this:
```
const IndicatorTriangle = observer((({ heading }) => {
  if (heading.hasNoDroppedConditions) {
      return  <span>{heading.formattedPoints}</span>
  }

  if (heading.hasMixedDroppedQuestions) {
    return <OrangeInfoIcon/>
  }
  return (
     <CornerTriangle color="blue"
                     data-test-id="dropped-question-indicator"
                     data-question-id={`Q${heading.question_id}`}
                     tooltip={heading.everyQuestionZeroed ? 'Points changed to 0' :
                              (heading.everyQuestionFullCredit ? 'Full credit given to all students':
                              'Question dropped for one or more students')}
                 />
  )
})
```

Dantemss:
Apparently it's not showing up in submission overview, only in the scores page...

PR #3597, opened 2021-05-08
Fix notes/isNew

jivey:
Fixes updating existing notes that were returning true for `isNew` even though they had integer IDs, also fixes some MobX-related warnings.

### DIFF
--- a/shared/src/model.ts
+++ b/shared/src/model.ts
@@ -33,7 +33,7 @@ export class BaseModel {
 
     get isNew() {
         const id = this[(this.constructor as any).idField]
-        return isNil(id) || id === NEW_ID
+        return isNil(id) || id === NEW_ID || id === ''
     }
 
     @action async ensureLoaded(): Promise<void> {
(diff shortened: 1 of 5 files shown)


shared/src/model.ts
```diff
@@ -33,7 +33,7 @@ export class BaseModel {
 
     get isNew() {
         const id = this[(this.constructor as any).idField]
-        return isEmpty(id) || id === NEW_ID
+        return isNil(id) || id === NEW_ID
```
jivey, line 36: Based on the shared spec failures I don't think I can just use the prior `isNil` version, I'll look at fixing it some other way.

nathanstitt: hm, it seems like we do want isEmpty, right?  I'd switched to that since ID is `string | number` and we're sometimes initializing it to `''`

image

nathanstitt: Ug I didn't realize that 😭 

Maybe we shouldn't use type checks and just do: `id != NEW_ID && id != ''`

nathanstitt: hm, actually maybe we should still check isNil tho, so:

`!isNil(id) && id != NEW_ID && id != ''`

jivey: Agree testing the inverse makes sense, I'll add some tests to confirm the behavior and have you take a peek.

nathanstitt (approved):

PR #3600, opened 2021-05-11
Start of drop question modal

nathanstitt:
needs filled in and connected to API, but details view link is working

I think we should get this asap; there's a lot of changes to UX due to the TS conversions

image


tutor/index.js
```diff
@@ -2,14 +2,8 @@ import renderRoot from 'shared/helpers/render-root';
...
+    await renderRoot({
         Component: TutorApp.rootComponent,
         props: { app },
     });
-    if (module.hot) {
-        module.hot.accept('./src/models/app', async () => {
-            const NewApp = await import('./src/models/app');
-            rootRenderer(NewApp.default.rootComponent);
-        });
-    }
```
jivey, line 14: Ah I was wondering what was causing that

jivey (approved):

PR #3601, opened 2021-05-12
use boxible lib to layout controls

nathanstitt:


tutor/src/screens/assignment-review/drop-question-modal.tsx
```diff
@@ -19,52 +21,101 @@ interface DropQuestionModelProps {
     ux: UX
 }
 
+const Light = styled.div`
+font-size: 1.2rem;
+color: #6F6F6F;
```
nathanstitt, line 26: we might have something like this elsewhere but if not we should think about making a file of common things like this

jivey: Yeah, it'd be nice to build out some common styled components that handle common typography

nathanstitt:
closing in favor of #3603

PR #3608, opened 2021-05-13
update question cards to show tooltip and color when dropped

jivey:
Added gray background and blue triangle when the question is dropped. @nathanstitt Great idea to take size to factor the rem scale! Worked out really nicely 👍 

image

### DIFF
--- a/tutor/src/components/homework-questions.js
+++ b/tutor/src/components/homework-questions.js
@@ -1,6 +1,7 @@
 import { React, PropTypes, cn, styled, observer, css } from 'vendor';
 import { colors } from 'theme';
 import Question from 'shared/components/question';
+import { DroppedQuestionIndicator } from './dropped-question';
 
 const HomeworkQuestionsWrapper = styled.div`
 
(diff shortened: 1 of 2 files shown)


tutor/src/components/homework-questions.js
```diff
@@ -186,9 +188,11 @@ const ReviewExerciseCard = observer(({
     questionType = 'teacher-preview',
     styleVariant = 'points',
 }) => {
+    const isDropped = !!info.droppedQuestion;
     return (
         <QuestionPreview className="openstax-exercise-preview">
-            <QuestionHeader variant={styleVariant} className="question-header">
+            <QuestionHeader variant={styleVariant} isDropped={isDropped} className="question-header">
+                {isDropped && <DroppedQuestionIndicator model={info.droppedQuestion} size={1.6} />}
```
nathanstitt, line 195: DroppedQuestionIndicator will only render if `info.drop_method` is set, you shouldn't need the conditional `isDropped`

nathanstitt (approved):

tutor/src/components/homework-questions.js
jivey: I only need it here because it could be undefined from trying to `.find` it, but I do like that approach, maybe later I can change it to use an empty object.

PR #3610, opened 2021-05-14
question headings and other drop-q fixes

nathanstitt:
image

image

image


jivey (approved):
Looks good! I noticed the design has "ID" in the title like here: 
image

PR #3613, opened 2021-05-17
allow two digits before final period in AP LO tag

jivey:
Allows tags for AP Biology and Physics to contain 2 digits before the final period instead of the current 1.

**Examples:**

AP Biology
image

AP Physics
image


jivey:
The only thing I'm not sure about this is if Biology should get this change as well, I asked in the original card and will yank that out if it isn't needed/wanted.

Dantemss (approved):
Looks like they said yes

PR #3618, opened 2021-05-20
A few more drop-question fixes

nathanstitt:
* Hide drop question icon when course is not active
* Choose latest preview course

### DIFF
--- a/tutor/src/screens/assignment-review/drop-question-icon.tsx
+++ b/tutor/src/screens/assignment-review/drop-question-icon.tsx
@@ -1,0 +1,20 @@
+import { React, Theme } from 'vendor';
+import UX from './ux'
+import { Icon } from 'shared'
+
+const DropQuestionIcon:React.FC<{ ux: UX, info: any }> = ({ ux, info }) => {
+    if (!ux.course.isActive) { return null }
+    return (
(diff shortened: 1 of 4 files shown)


tutor/src/screens/assignment-review/drop-question-icon.tsx
```diff
@@ -0,0 +1,20 @@
...
+    if (!ux.course.isActive) { return null }
+    return (
+        <Icon
+            type="minus-circle"
+            data-tour-anchor-id="drop-question-new"
+            tooltip="Drop question"
+            color={Theme.colors.neutral.gray}
+            hoverColor={Theme.colors.neutral.darker}
+            tooltipProps={{ placement: 'bottom' }}
+            onClick={() => ux.displayDropQuestion(info.question)}
```
jivey, line 15: This couples the button to that exact UX class but it's not used anywhere else so probably not a worry.

jivey (approved):

tutor/src/screens/assignment-review/drop-question-icon.tsx
nathanstitt: yeah good point, we could pass the course to it directly.

PR #3620, opened 2021-05-20
only attempt to parse body if it's present

nathanstitt:
Some api methods like onboarding say they're content type json but have
empty body

### DIFF
--- a/shared/src/api/request.ts
+++ b/shared/src/api/request.ts
@@ -87,13 +87,12 @@ async function request<RetT>(methodUrl: MethodUrl, options?: any): Promise<RetT|
         }
         const origin = options?.origin || baseUrl
         const resp = await fetch(`${origin}/${url}`, req)
-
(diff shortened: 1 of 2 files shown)


shared/src/api/request.ts
```diff
@@ -87,13 +87,12 @@ async function request<RetT>(methodUrl: MethodUrl, options?: any): Promise<RetT|
...
         if (resp.ok) {
-            if (resp.status == 204) { // no content
-                return {} as RetT
+            const body = await resp.text()
+            if (body) {
+                return JSON.parse(body) as RetT
             }
-            const respJson = await resp.json()
-            return await respJson as RetT
+            return {} as RetT
```
nathanstitt, line 95: @Dantemss checkout line 107 below for where the 418 response code comes from 😁

Dantemss (approved):

PR #3628, opened 2021-06-01
adjust code to hide modal earlier

nathanstitt:
Get the drop question modal offscreen as soon as visible items are updated

### DIFF
--- a/tutor/src/screens/assignment-review/ux.ts
+++ b/tutor/src/screens/assignment-review/ux.ts
@@ -281,13 +281,16 @@ export default class AssignmentReviewUX {
         } else {
             taskPlan.dropped_questions.push(droppedQuestion);
         }
-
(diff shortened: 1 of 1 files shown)


Dantemss (commented):
Hmm not sure if I'm a fan because this means the modal disappears before saving is complete so what if saving silently fails?

nathanstitt:
no it disappears as soon as *saving* is complete.  It then refreshes the scores in the background.  If that fetch does fail then the error modal will still appear

Dantemss:
Isn't it hiding before the exclusions are saved though?

nathanstitt:
oh right, so we are.  THat's also fine though because if exclusions fail we show the error modal

nathanstitt:
yah the only way to not show the modal is to pass in the `ignoreErrors` flag to `api.request`. The only spot we do that currently is for response validator

PR #3630, opened 2021-06-01
Fix display of dropped questions tooltip on hw score

nathanstitt:
image

image

### DIFF
--- a/tutor/src/components/dropped-question.tsx
+++ b/tutor/src/components/dropped-question.tsx
@@ -76,12 +76,14 @@ DroppedIndicator.displayName = 'DroppedIndicator'
 interface DroppedQuestionHeadingIndicatorProps {
     preventOverflow?: boolean
     size?: number
+    responseCount?: number
     heading: TaskPlanScoreHeading
 }
 
 const DroppedQuestionHeadingIndicator: React.FC<DroppedQuestionHeadingIndicatorProps> = observer(({
     heading,
     preventOverflow = true,
+    responseCount = heading.studentResponses.length,
     size = 1,
 }) => {
     if (!heading.someQuestionsDropped) return null
@@ -96,7 +98,7 @@ const DroppedQuestionHeadingIndicator: React.FC<DroppedQuestionHeadingIndicatorP
             tooltip = 'Points changed to 0 for all students'
         }
     } else {
-        tooltip = `Question dropped for ${pluralize('student', heading.studentResponses.length, true)}`
+        tooltip = `Question dropped for ${pluralize('student', responseCount, true)}`
     }
 
     return (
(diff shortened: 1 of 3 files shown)


tutor/src/components/dropped-question.tsx
```diff
@@ -84,16 +85,15 @@ const DroppedQuestionHeadingIndicator: React.FC<DroppedQuestionHeadingIndicatorP
 }) => {
     if (!heading.someQuestionsDropped) return null
     let tooltip
+    const students = pluralize('student', heading.droppedQuestions.length, true)
     if (heading.isCore) {
```
Dantemss, line 89: I think this is also changing it for homework core which we do not want.

tutor/src/components/dropped-question.tsx
```diff
@@ -108,9 +108,39 @@ const DroppedQuestionHeadingIndicator: React.FC<DroppedQuestionHeadingIndicatorP
...
+    }
+}
+export const DroppedReviewExerciseIndicator: React.FC<DroppedReviewExerciseIndicatorProps> = observer(({
+    info,
+}) => {
+    const model = info.droppedQuestion
+
+    if (!model || !model.drop_method) return null
+
+    let students = pluralize('student', info.responses.length, true)
```
Dantemss, line 127: Same thing here, the change from "all" to a number will also affect homework core which we do not want.

tutor/src/screens/assignment-review/homework-scores.js
```diff
@@ -235,7 +235,7 @@ const TaskResult = observer(({ heading, result, striped }) => {
             isTrouble={result && !result.isUnattemptedAutoZero && result.isTrouble}
             isUnattemptedAutoZero={result && result.isUnattemptedAutoZero}
         >
-            {!heading.isCore && <DroppedQuestionHeadingIndicator heading={heading} />}
```
Dantemss, line 238: I believe this triangle should only appear if the one on https://github.com/openstax/tutor-js/blob/657effafe37bf87e32857010565f70f5356e1f81/tutor/src/screens/assignment-review/homework-scores.js#L83 does not so we need the !isCore check.

Dantemss (changes requested):
I think rather than removing the isCore checks we need to add some checks for reading vs homework.

nathanstitt:
yes it is change it for all HW core, but we do want that as I understand it.  I think we want to show the count at all times.  lets PL

Dantemss (approved):
Looks good to me and Devika said OK to the change from `Question dropped for one or more students` to `Question dropped for # students`

Dantemss (changes requested):
Sorry one more thing... the student count in whatever replaces DroppedReviewExerciseIndicator has to come from info, not the heading I believe. The heading aggregates in ways it shouldn't for DroppedReviewExerciseIndicator so the count will not be right

Dantemss (approved):

PR #3649, opened 2021-06-15
Draft ribbon

nathanstitt:
This way we can identify why exercises may appear twice in the search results

image


jivey (approved):
Looks good!

PR #3659, opened 2021-06-21
only show createCourse block to authorized users

nathanstitt:
No idea why we didn't have this check there already.  maybe it's always been broken?


jivey (approved):
Hmm yeah maybe we should add an unauthorized course creation user to our e2e tests

PR #3679, opened 2021-07-12
When completing or cancelling assignment creation, send instructor back to selected month

Dantemss:
When clicking the calendar to create an assignment, you should go back to that month.
Creating an assignment for the sidebar, however, seems to always default to the current month, so that won't work (but if we want that to change, we can pick a date for sidebar assignments based on the selected month).


jivey (approved):
This is gonna be nice to have!

Dantemss:
Just noticed it doesn't quite work when editing a draft (instead of creating a new one). Let me see if I can fix it real quick...

PR #3680, opened 2021-07-12
update book colors for sociology and us history

nathanstitt:


lwtchu:
:eyes: sorry to barge in. but to better match rex/the cover should we  do white text on navy bg?

jivey (approved):

nathanstitt:
Heya @lwtchu - I'm going off of https://github.com/openstax/tutor/issues/1715#issuecomment-880020368 for the colors.   Should I update it to have white text?  Sorry for not including a screenshot on this, but it currently looks like:

lwtchu:
@nathanstitt i think for text it's been historically either white or black just to reduce the amount of colors to worry about for contrast with highlighting. I'm looking thru the title cards on tutor for the other books and it seems to be the case :eyes:

nathanstitt:
yep, you're correct, I was mistaken and thought that we'd styled the text using the book colors but looking through other books I see that the text is either white or black. I'll update this to be white.  Thanks @lwtchu !

this is what it'll look like with that change:
image


image

lwtchu:
thank you ~ ~ ~

PR #3682, opened 2021-07-15
Make Upgrade to Sociology 3e message a link to creating a new Sociology 3e course

Dantemss:
Not sure if the message also shows up when cloning courses?
But we can't clone 2e into 3e so redirecting to a new course kind of makes sense in that case.

### DIFF
--- a/tutor/src/screens/new-course/sociology-3e-nudge.tsx
+++ b/tutor/src/screens/new-course/sociology-3e-nudge.tsx
@@ -1,5 +1,6 @@
 import { React, styled } from 'vendor'
 import { colors } from '../../theme'
+import TutorLink from '../../components/link';
 
 const Text = styled.div`
     margin-top: 8px;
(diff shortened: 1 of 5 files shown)


tutor/src/screens/new-course/sociology-3e-nudge.tsx
```diff
@@ -28,7 +28,7 @@ const Sociology3eNudge = ({ ux }: { ux: any }) => {
     if (ux.displaySoc3eAvailableNudge) {
         return (
             <Text>
-                Introduction to Sociology 2e is available until Summer 2022. Upgrade to <cite>Introduction to Sociology 3e</cite>.
+                Introduction to Sociology 2e is available until Summer 2022. <a href={`/new-course/offering/${ux.offerings.sociology3e.array[0].id}`}>Upgrade to <cite>Introduction to Sociology 3e</cite>.</a>
             </Text>
```
nathanstitt, line 32: Hehe, so while you can use `a` tags and it'll work that's not the optimal way to change pages.  The browser will do a whole page reload when it follows them. It's better to use the `Link` component like is used here https://github.com/openstax/tutor-js/blob/main/tutor/src/components/my-courses/create-a-course.jsx#L11 That will just push a new entry into the browser's `history` and react-router will update only the needed components.

tutor/src/screens/new-course/ux.js
```diff
@@ -188,12 +188,12 @@ export default class CourseBuilderUX extends BaseModel {
...
         if (!this.offering) return false;
-        return this.offering.isSociology2e && currentOfferings.soc3eExists && !currentOfferings.soc3eAvailable;
+        return this.offering.isSociology2e && this.offerings.soc3eExists && !this.offerings.soc3eAvailable;
     }
 
     @computed get displaySoc3eAvailableNudge() {
         if (!this.offering) return false;
-        return this.offering.isSociology2e && currentOfferings.soc3eAvailable
+        return this.offering.isSociology2e && this.offerings.soc3eAvailable
     }
```
nathanstitt, line 197: nice catch!

nathanstitt (approved):

PR #3684, opened 2021-07-16
Redeem bookstore payments

jivey:
- Add code redemption screens
- Show payment codes alongside purchases

Screenshot 2021-07-15 at 14-07-00 OpenStax Tutor
Screenshot 2021-07-15 at 14-06-49 OpenStax Tutor
Screenshot 2021-07-15 at 14-06-38 OpenStax Tutor
Screenshot 2021-07-15 at 14-16-11 OpenStax Tutor
image


tutor/src/components/onboarding/payment-options.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import type { StudentCourseOnboarding } from './ux/student-course'
+
+interface PaymentOptionsProps {
+    ux: StudentCourseOnboarding
+}
+
+@observer
+export default class PaymentOptions extends React.Component<PaymentOptionsProps> {
+    static className = 'payment-options';
```
nathanstitt, line 13: is className used? It's hard to search for but I didn't see it anywhere

tutor/src/components/onboarding/redeem-code.tsx
```diff
@@ -0,0 +1,146 @@
...
+                        </Tooltip>
+                    }
+                >
+                    <a href="#" aria-label={RedeemCode.tooltipText}>Where to find the access code?</a>
+                </OverlayTrigger>
+                <StyledBox>
+                    <div className={ux.paymentCodeError ? 'has-errors' : ''}>
+                        <input
+                            type="input"
+                            placeholder="OST-100234ABCD"
```
nathanstitt, line 96: do we have any guidance on what the placeholder should be?  using `OST-` seems like it might be misleading if the codes end up being stuff like `GEOAB-` or whatever and the student is expecting a OST prefix.  Might be better to use `XXX-XXXXXXX` as the placeholder to prevent pre-conceptions

tutor/src/components/onboarding/ux/student-course.ts
```diff
@@ -75,7 +104,7 @@ export class StudentCourseOnboarding extends BaseOnboarding {
     @action.bound
     onPaymentComplete() {
         if (this.paymentIsPastDue) {
-        // in this case we have to reload since network requests have been failing silently
+            // in this case we have to reload since network requests have been failing silently
             setTimeout(forceReload);
         } else {
             this.displayPayment = false;
```
nathanstitt, line 110: hm, I'm curious if this is still needed?

nathanstitt (approved):
looks great @jivey !

tutor/src/components/onboarding/payment-options.tsx
jivey: Oh good catch, removing that.

tutor/src/components/onboarding/redeem-code.tsx
jivey: This code comes from the mockups so I assume it's been considered? But good point, maybe we should 🅿️ this on Monday and try to get an example prefix.

tutor/src/components/onboarding/ux/student-course.ts
jivey: Tested removing it locally and it seems to no longer be needed? I'll commit it and we can test more on staging.

PR #3686, opened 2021-07-16
Restore all jest mocks after use

Dantemss:
Not sure if it matters, but I saw it as a recommendation to avoid memory leaks...


nathanstitt (approved):
nice idea!

PR #3694, opened 2021-07-22
fix moment object being used in comparison

jivey:
Fixes the grant extension due date being broken

### DIFF
--- a/tutor/src/screens/assignment-review/grant-extension.js
+++ b/tutor/src/screens/assignment-review/grant-extension.js
@@ -181,7 +181,7 @@ const ExtendModal = observer(({ ux, form: { isValid, values } }) => {
                             timezone={ux.course.timezone}
                             validate={(d) => { // eslint-disable-line consistent-return
                                 d = new Time(d)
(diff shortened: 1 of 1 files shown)


tutor/src/screens/assignment-review/grant-extension.js
```diff
@@ -180,6 +180,7 @@ const ExtendModal = observer(({ ux, form: { isValid, values } }) => {
                             disabledDate={ux.course.isInvalidAssignmentDate}
                             timezone={ux.course.timezone}
                             validate={(d) => { // eslint-disable-line consistent-return
+                                d = new Time(d)
                                 if (!d) { return 'must be valid date' }
```
nathanstitt, line 184: since `d` is now a Time class you can use `d.isValid` here

jivey: Good idea, swapped to that. Hard to test though because the picker automatically overrides it? 😆

nathanstitt (approved):

PR #3698, opened 2021-07-23
update task plan extensions with api response data

jivey:
Should make it so the "E" extension icon shows up immediately after saving the extensions, as the API gives us an update task plan in the response.

### DIFF
--- a/tutor/src/models/task-plans/teacher/plan.ts
+++ b/tutor/src/models/task-plans/teacher/plan.ts
@@ -2,7 +2,7 @@ import {
     BaseModel, field, action, computed, observable, model, lazyGetter,
     array, modelize, NEW_ID, ID, override, hydrateInstance, hydrateModel,
 } from 'shared/model';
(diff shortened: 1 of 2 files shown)


tutor/src/models/task-plans/teacher/plan.ts
```diff
@@ -466,11 +466,12 @@ export class TeacherTaskPlan extends BaseModel {
     async grantExtensions(extensions: TaskPlanExtensionData[]) {
         //if new extensions dates are selected for a student who has already an extension, this will update the student previous extended dates
         const grantedExtensions = unionBy(extensions, this.extensions, 'role_id');
-        const updatedExtensions = this.api.request<TaskPlanExtensionData[]>(
+        const updatedTaskPlan = await this.api.request<TeacherTaskPlanData>(
```
jivey, line 469: Maybe the api returned an array of extensions before, but it's returning the full task plan now.

nathanstitt (approved):

PR #3704, opened 2021-07-27
fix preview course spec

jivey:


tutor/specs/e2e/preview-courses.e2e.ts
```diff
@@ -16,38 +11,35 @@ test.describe('Preview Courses', () => {
...
         })
-        await page.goForward()
-
-        await expect(page).not.toHaveSelector('testId=side-panel >> testId=preview-panel-create-course', { timeout: 100 })
+        await expect(page).not.toHaveSelector('testId=side-panel >> testId=preview-panel-create-course')
     })
 
-    test('hides preview panel for non-available offerings', async ({ page }) => {
+    test('hides preview panel for real courses and the associated preview course', async ({ page }) => {
+        await findOrCreateTeacherAccount({ page, userName: 'teacher01' })
```
nathanstitt, line 36: no need for `findOrCreateTeacherAccount`, we already have `teacher01`,  `teacher02` and review student 01 & 02.  The setup.ts pre-authorizes them and stores their cookies and then switches to using them when `withUser` is called

tutor/specs/e2e/preview-courses.e2e.ts
```diff
@@ -16,38 +11,35 @@ test.describe('Preview Courses', () => {
...
         await page.click('testId=preview-course-card')
         await page.click('testId=preview-panel-create-course')
         await expect(page).toHaveSelector('testId=new-course-wizard')
         await expect(page).toHaveSelector('.select-dates')
+
     })
 
-    test('hides preview panel for non-preview courses', async ({ page }) => {
+    test('hides preview panel if the offering is not available for course creation', async ({ page }) => {
+        await findOrCreateTeacherAccount({ page, userName: 'previewteacher' })
```
nathanstitt, line 24: I think there might already be a `previewteacher` account setup on the backend.  If so we can call `loginAs` which is much faster

nathanstitt (approved):
a few comments on how to make it run faster but seems good otherwise

tutor/specs/e2e/preview-courses.e2e.ts
jivey: Ah I forgot about loginAs :D thanks!

tutor/specs/e2e/preview-courses.e2e.ts
jivey: It doesn't yet but that would be really nice to add, that would make this whole test very speedy

PR #3712, opened 2021-07-30
Fix duplicate sections appearing on the roster when renaming

Dantemss:
Also added a unit test for CoursePeriod save()


jivey (approved):
Nice bug find!

PR #3713, opened 2021-08-02
Multiple Attempts Integration

jivey:


tutor/src/screens/grading-templates/editors.js
```diff
@@ -300,6 +313,79 @@ class TemplateForm extends React.Component {
...
+                                name="allow_auto_graded_multiple_attempts"
+                                label="No"
+                                value={false}
+                                defaultChecked={template.allow_auto_graded_multiple_attempts == false}
+                            />
+                        </Setting>
+                    </div>
+                </FieldsetRow>
+                <Row>
+                    {form.values.allow_auto_graded_multiple_attempts == 'true' &&
```
Dantemss, line 369: Is it weird for this to be the string "true"?

Dantemss (approved):

tutor/src/screens/grading-templates/editors.js
jivey: Yeah it's Formik looking at the HTML attribute for the value, which is fine, but I'm glad you pointed it out because it made me discover a bug with the message not appearing when editing an existing template!

Also getting a boolean makes it read better: https://github.com/openstax/tutor-js/pull/3713/commits/2c38432c60f86a6263353347cb8cbb613bc5eef5

PR #3716, opened 2021-08-09
Use localeCompare in numeric mode for sorting course sections

Dantemss:
https://caniuse.com/localecompare

`localeCompare()` is only weird for the decimal parts of numbers, when the decimal parts have different lengths.
I don't think we care very much about that case though.

### DIFF
--- a/tutor/src/helpers/period.ts
+++ b/tutor/src/helpers/period.ts
@@ -1,5 +1,4 @@
 import type { CoursePeriod } from '../models'
-import { sortBy } from 'lodash';
 import S from './string';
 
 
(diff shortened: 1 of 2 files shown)


tutor/src/helpers/period.ts
```diff
@@ -9,21 +8,11 @@ const PeriodHelper = {
...
-        });
-        return sortBy(periods, (period ) => { // eslint-disable-line consistent-return
-            const number = period.name.match(/[0-9.-]+/g);
-            if (number) {
-                return parseFloat(number[0]);
-            }
-            return undefined
-        });
+        return periods.slice(0).sort(
+            (firstPeriod, secondPeriod) => firstPeriod.name.toLowerCase().localeCompare(
```
jivey, line 12: I forgot about localeCompare, nice to not need lodash all the time.

jivey (approved):

PR #3723, opened 2021-08-12
Echo back attempt number when grading questions

Dantemss:
To handle error cases such as having 2 tabs open and grading at the same time as submitting a new answer.


jivey:
Sorry I should have mentioned I have this covered in the student branch: https://github.com/openstax/tutor-js/pull/3715

jivey:
Oh but you also need it on the grading, I forgot about that :D

Dantemss:
I'll remove the student side one then and leave only the grading one to avoid conflicts

jivey (approved):

PR #3724, opened 2021-08-13
add exercise id and type to task header

jivey:
Moves the question ID and type icons into the header where the Points used to be displayed (points will be displayed in the footer in another PR)

image


tutor/src/screens/task/step/card.js
```diff
@@ -57,13 +57,16 @@ const StepCardHeader = styled.div`
...
-      span {
-        display: none;
-      }
-    }
+  div.question-info {
+    font-weight: bold;
+  }
+
+  .num-questions, .points {
+      display: none;
```
TomWoodward, line 65: why do we hide/show stuff in css instead of the react components?

TomWoodward (approved):

tutor/src/screens/task/step/card.js
jivey: We usually only do it in cases like this one to target responsive breakpoints (this is for desktop, the tablet/mobile view then resets it lower down... I probably should have flipped it the other way around though.)

TomWoodward: oh i see, yea it would have been less confusing for me if it was like, override to hide in certain cases, rather than default to hide and override to show in certain cases, but there are legitimate reasons to organize your breakpoints in different ways so maybe its better like

PR #3730, opened 2021-08-18
Fix for errors when teachers review incomplete assignments between due and close dates

Dantemss:
- Allow teachers to review all steps of an assignment even if it is not closed
- Don't attempt to mark incomplete steps as completed when a teacher is viewing them
- Increased default node max_old_space_size since some tests were running out of memory


tutor/src/screens/task/ux.js
```diff
@@ -405,7 +412,12 @@ export default class TaskUX {
 
     @action.bound markIncorrectAttempt() {
         const step = this.currentStep;
-        if (step.canAnswer && step.answer_id && step.answer_id != step.correct_answer_id) {
+        if (
+            !this.isTeacher && // Teachers reviewing assignments don't get the UI to re-answer
```
jivey, line 416: Good catch

jivey (approved):

PR #3733, opened 2021-08-24
Multiple attempts footer

jivey:
image


tutor/src/screens/task/step/exercise-question.js
```diff
@@ -181,6 +139,25 @@ export default class ExerciseQuestion extends React.Component {
         );
     }
 
+    renderPoints(step) {
+        const points = [step.available_points];
+
+        if (!isNil(step.published_points_without_lateness)) {
+            points.unshift(step.published_points);
```
Dantemss, line 146: It is a little odd to me that published_points_without_lateness is guarded here when it is published_points that is added, but I suppose they are equivalent in this regard.

tutor/src/components/buttons/save-practice.js
```diff
@@ -81,7 +82,7 @@ const SavePracticeButton = observer(({
             practiceQuestion.destroy();
         }
         else {
-            practiceQuestions.create(taskStep.tasked_id);
+            runInAction(() => practiceQuestions.create(taskStep.tasked_id));
```
Dantemss, line 85: Wouldn't destroy above also need this runInAction()?

Dantemss (approved):

tutor/src/screens/task/step/exercise-question.js
jivey: I thought the same thing, it was also there in the old code, so I assumed it had a reason. Meant to ask you about it 😄

tutor/src/components/buttons/save-practice.js
jivey: I almost added it too but then saw that `destroy` here is a wrapper that calls the actual `map.delete` (with `runInAction`) after doing the network request.

PR #3738, opened 2021-08-26
add timeouts to ensure location.search isn't matched too early

jivey:
Should help some of the e2e flakiness.


Dantemss (approved):
Maybe we could implement something like `page.waitForEvaluation(...)` that would behave like `page.waitForSelector(...)` in the future.

jivey:
Yeah maybe I should have looked into the methods here for a better approach https://playwright.dev/docs/api/class-page#page-wait-for-response. I think I want to redo this spec, I'm not sure it should care about the query string, just what is rendered.

PR #3746, opened 2021-09-02
add a basic multiple attemtps e2e spec

jivey:


tutor/specs/e2e/multiple-attempts.e2e.ts
```diff
@@ -0,0 +1,98 @@
...
+} from './test'
+
+const COURSE_ID = 2
+
+let assignmentName = ''
+
+withUser('reviewstudent2')
+
+test.beforeAll(async ({ browser }) => {
+    const context = await browser.newContext({ storageState: 'temp/teacher02-state.json' })
```
TomWoodward, line 16: could all this setup be pre-done in a data file rather than having to do it all through the ui in the `beforeAll`? wouldn't hurt to have a separate test with all this for creating the multiple-attempt questions too if that doesn't exist already

TomWoodward (approved):

tutor/specs/e2e/multiple-attempts.e2e.ts
jivey: Yeah I'd like the demo import to generate different data scenarios including this one, and make manual data creation like this a rare need. In the meantime, I'd like to move this into a helper, as it's also used in the student-task E2E spec.

TomWoodward: can you make a card in zenhub and list the scenarios that youd like added to the demo import? even if we don't do it soon it'd be nice to have a list of the stuff we want to add and where it would be used so we don't forget

PR #3747, opened 2021-09-03
Re-enabled e2e assignment tests and check relative assignment dates

Dantemss:
Need to check that the selected date actually matches the date you can read afterwards, then do the same thing after switching the course to a different timezone.


tutor/specs/e2e/assignment-review.e2e.ts
```diff
@@ -51,30 +51,204 @@ test.describe('Assignment Review', () => {
...
+            )
+
+            await page.click('text="View assignment"')
+            await page.click('testId=edit-assignment')
+
+            await page.click('.due-at')
+            await page.click(
+                '.oxdt-dropdown:not(.oxdt-dropdown-hidden) .oxdt-cell-selected + .oxdt-cell'
+            )
+            await page.click('.oxdt-dropdown:not(.oxdt-dropdown-hidden) button')
```
TomWoodward, line 205: some of these selectors you're using a lot, you could put the selectors in variables for readability and maintainability

tutor/src/screens/assignment-review/details.js
```diff
@@ -281,22 +281,22 @@ const TaskingDates = observer(({ tasking, title }) => {
...
+                    <dd class="due-date">
                         <div>{tasking.dueAt.format(dateFormat)}</div>
                         <div>{tasking.dueAt.format(timeFormat)}</div>
                     </dd>
                 </div>
                 {!tasking.plan.isEvent && (
                     <div>
                         <dt>Close date</dt>
-                        <dd>
+                        <dd class="close-date">
```
TomWoodward, line 299: if these classes are purely used as selectors in the tests, it'd be better to use a `[data-testid="close-date"]` so that nobody confuses it with styling

tutor/specs/e2e/assignment-review.e2e.ts
```diff
@@ -1,12 +1,24 @@
...
 const GRADABLE=3
 
-test.describe('Assignment Review', () => {
+const openCalendarButtonCSS = '.oxdt-dropdown:not(.oxdt-dropdown-hidden) button'
+const openCalendarCellTodayCSS = '.oxdt-dropdown:not(.oxdt-dropdown-hidden) .oxdt-cell-today'
+const openCalendarCellAfterSelectedCellCSS = '.oxdt-dropdown:not(.oxdt-dropdown-hidden) ' +
+                                             '.oxdt-cell-selected + .oxdt-cell'
+
+const detailsDueAtCSS = '.tasking-date-time.row + .tasking-date-time.row .due-at'
+const detailsClosesAtCSS = '.tasking-date-time.row + .tasking-date-time.row .closes-at'
```
nathanstitt, line 14: typically in tests it's best to keep the selectors as generic as possible so they don't break if the DOM's rearranged.  that's the rational behind adding the data-test-id attributes.  Might be better to use those if possible here

nathanstitt (approved):
a few suggestions but seems great and will restore coverage to an area that really needs it 🎉

jivey (changes requested):
No real change needed here, but needs https://github.com/openstax/tutor-js/pull/3771 for the spec to pass.

Dantemss:
Handled by https://github.com/openstax/tutor-js/pull/3771

PR #3749, opened 2021-09-07
fix response validation flexbox issue

jivey:
Fixes an alignment bug with the response validator text by always using `space-between`.

Old:
image

New:
image


TomWoodward (commented):
are you intentionally limiting the width of the left section? you could leave the justification constant and put `flex: 1` on the left text area if you were ok with it filling the space

jivey:
> are you intentionally limiting the width of the left section? you could leave the justification constant and put `flex: 1` on the left text area if you were ok with it filling the space

The Validator text has a max-width there, but didn't really look too closely at it though, maybe the rules could be simplified as you say. I can take a look later today.

Dantemss (approved):

PR #3750, opened 2021-09-08
validate that detailed solution is present when creating exercise

jivey:
The Detailed Solution will now show a validation error if it has not been filled.
image
image

Also re-enables the Add-Edit E2E spec by converting to Typescript.


tutor/specs/e2e/helpers.ts
```diff
@@ -108,9 +108,9 @@ export const selectCalendarSidebarOption = async (page: Page, option: string) =>
     await page.click(`.add-assignment-sidebar >> text="${option}"`)
 }
 
-export const signTerm = async (page: Page, termId: string) => {
-    await page.evaluate(() => {
-        const term = (window as any)._MODELS?.user.terms.get(termId)
+export const signTerm = async (page: Page, _termId: string) => {
+    await page.evaluate((_termId) => {
+        const term = (window as any)._MODELS?.user.terms.get(_termId)
```
TomWoodward, line 113: don't you have to pass termId to the second arg of `evaluate`?

TomWoodward (approved):

tutor/specs/e2e/helpers.ts
jivey: Oh that's right the new runner does need that. Good catch! Updated: https://github.com/openstax/tutor-js/pull/3750/commits/1ad8abc9320b4e095612696417a91f5a30b46e1c

TomWoodward: yea those evaluates are hard because both typescript and my brain can't remember that those lines of code will be yoinked into a different runtime context

PR #3771, opened 2021-09-14
fix course timezone not being used when loading and saving assignment

jivey:
This fixes a regression with the course timezone not being used with moment-timezone when sending the field value to the DateTimeInput component. With this fix, the value loaded should always match the course's timezone. This also changes dataForSave to use the Luxon methods, and fixes a bug with it not converting to the course's timezone before formatting with tokens. This is necessary because the backend doesn't take TZ info from the front end, so the formatted string needs to match the course TZ.

Misc:

- Convert plan details to use Time/Luxon methods to make datetime TZ match the other screens
- moment-timezone is updated to get type info for Typescript
- use data-test-ids for spec, remove TZ replace now that plan details displays them

| Old | New |
|--------------|-----------|
|image |image |


tutor/specs/e2e/assignment-review.e2e.ts
```diff
@@ -138,11 +138,11 @@ test.describe('Assignment Review', () => {
         const dueAtText = await page.$eval(
             '[data-test-id="due-date"]', node => (node as HTMLElement).innerText
         )
-        const newDueAt = DateTime.fromFormat(dueAtText + ' ' + timezone, reviewDateTimeFormat)
+        const newDueAt = DateTime.fromFormat(dueAtText, reviewDateTimeFormat)
```
Dantemss, line 141: Is the timezone in dueAtText now? I had to add it manually before because it was elsewhere in the page.

Dantemss (commented):
Looks OK to me but I know @TomWoodward was trying to remove Moment so I'd like his opinion too.

tutor/specs/e2e/assignment-review.e2e.ts
jivey: Yeah I replaced moment with Luxon in the Details render, so it displays a usable timezone (e.g. EST -> America/New_York):

tutor/src/components/date-time-input.js
```diff
@@ -78,7 +78,12 @@ const DateTimeInput = observer((assignedProps) => {
     const id = props.id || uniqueId(props.name);
     const LabelWrapper = props.labelWrapper || React.Fragment;
 
-    const timeValue = field.value ? new Time(field.value).asMoment : null
+    const momentValue = (value, timezone) => {
+        const time = new Time(field.value)
+        return timezone ? time.inZone(timezone).asMomentTz : time.asMoment
```
TomWoodward, line 83: the time model uses the course timezone as its default, so i'm not sure why you'd have to do this `inZone` thing. what is the value of `field.value`? a javascript date? maybe it is explicitly coming with a timezone overriding the default?

i feel like there is a cleaner way to organize this than passing the timezone around

i assume the time picker component expects a moment object, which is why you're converting to that here, which is fine

TomWoodward (changes requested):

tutor/src/components/date-time-input.js
jivey: `field.value` here is an ISOstring. And yeah you're right the `rc-picker` expects a moment object, would be nice to replace or fork that at some point 😄 

This component was intended to be configurable to aid reusability, so it doesn't assume course TZ. But, could very well remove timezone prop and make it assume course timezone. Though it feels confusing that you have to know about the existence of `syncCourseTimezone`... I personally like the explicitness but would be happy to explore something that feels nicer!

TomWoodward: treating this as a generic component with an input timezone sounds fine, but wouldn't you need to mess with it in onUpdateDate instead? where it triggers onChange? 

maybe for edit it wouldn't matter because the moment already has the tz set, but if there isn't an initial value then timeValue will be set to null, and the resulting moment will probably be in the default timezone

the docs say that onUpdateDate would receive a moment, not an ISO string, but maybe i'm missing something

jivey: Right, as you mentioned `onUpdateDate` does get a moment object from that event. But because we already need to set the timezone for display reasons on render, we still get it back in that event with the same TZ, so nothing more needs to be messed with, as far as I'm aware. It gets converted to an ISOstring at the end of this event chain when getting set to the actual model.

> if there isn't an initial value then timeValue will be set to null, and the resulting moment will probably be in the default timezone

That does seem right to me, if there's no `timezone` prop and the value is null then it defaults to the default. Hmm maybe we should require a timezone to be passed.

TomWoodward: looking at this, if there _is_ a timezone passed, but no initial value, the result will be messaged out using browser default, not the passed timezone, which is a problem. is that not happening?

jivey: The picker does seem to do a bare `moment` call if the value is missing, defined here: https://github.com/react-component/picker/blob/8bc4cce34d2d53e8017e2d8b7a59c4f7db613d0c/src/generate/moment.ts#L8 which will cause what you mention above. In practice it works because of a re-render after setting the value, which then picks up the timezone call now that `value` is present. I can see how that isn't ideal. What do you think about providing our own default Time.now with the Timezone: https://github.com/openstax/tutor-js/commit/55840c62b3b45be815b0472544845e66598eb21d#diff-bd732d805a17eef68d70ecf4680d6532af760bed2daac6b4c09e1a21c1164b60R82

The only downside is we'll lose the ability to show a blank/empty input state, but so far we haven't needed that in a design.

TomWoodward: the alternative is to manipulate the moment to override the timezone to the passed timezone in the `onUpdateDate` function, is there an issue with doing that?

TomWoodward: so what, rc-picker converts to local time when displaying in the picker? so really we need to strip the timezone on the way in and add it on the way out, right?

jivey: `rc-picker` will use the moment object (with or without the `_z` timezone property) you pass in `value`, so if you have it, there's no issue, we just put it into the course timezone before sending it. My last comment was referring to the question about what happens if `DateTimeInput` is given a `timezone`, but no `value`, so we can't create the moment object (in the right TZ) beforehand. From what I can tell, `rc-picker` does a few of these `newDate = value || generateConfig.getNow()` defaults, so if you don't pass `value`, it will basically call `moment()` which will be in the browser timezone.

TomWoodward: ok so now i'm back to 
> manipulate the moment to override the timezone to the passed timezone in the onUpdateDate function, is there an issue with doing that?

the application of the timezone on the input is almost auxiliary, like its good to ensure it actually matches the given timezone for this component, but for our purposes the given input will almost certainly have the correct timezone already, the only situation in which its actually mattering is when the value is being re-applied after being generated by the picker

TomWoodward (approved):

PR #3775, opened 2021-09-15
Fix order date display

TomWoodward:


tutor/specs/components/payments/manage.spec.jsx
```diff
@@ -19,8 +20,15 @@ describe('Student Payments Management', () => {
         Router.makePathname.mockImplementation(() => '/foo');
     });
 
-    it('renders and matches snapshot', () => {
-        expect.snapshot(<C noRef><Payments /></C>).toMatchSnapshot();
+    it('renders and matches snapshot', async () => {
+        const component = renderer.create(<C noRef><Payments /></C>)
+
+        // clearing react lifecycle
```
TomWoodward, line 26: @jivey let me know if there is a cleaner way to do this in the test framework that we have set up here

jivey: Not sure if cleaner, but it looks like we've handled this in other places with a `waitFor` helper that kinda does the same thing:

```
 await waitFor(() => !currentPurchases.api.isPending)
```

TomWoodward: after the api request completes (which happens in one tick because its faked) you have to wait another tick for the react stuff to update, which is why i have to have two awaits, to clear the promise queue twice. it can actually be made a little better than what i had before by using `setImmediate`. the waitFor helper works, but only because it has a minimum wait of 100ms, which is pretty aggressive, i'm not sure i really want to go around sprinkling that everywhere...

jivey (approved):
Added the flaky spec to my list 😄

TomWoodward:
> Added the flaky spec to my list 😄

i'm looking into it here https://github.com/openstax/tutor-js/pull/3776, i thought it was just missing the await, but now i think its an issue with multiple test courses being created with the default name

PR #3776, opened 2021-09-15
fix a test

TomWoodward:


tutor/specs/e2e/course-settings.e2e.ts
```diff
@@ -24,7 +28,6 @@ test.describe('without any students', () => {
 
         await page.click('testId=course-details-tab')
         await expect(page).toHaveSelector('.course-detail-settings-form')
-        const courseName = page.$eval('input#course-name', (el: any) => el.value)
```
TomWoodward, line 27: the initial issue was that this line needed an `await`, passing a promise object into the string concatenation on line 39 was somehow doing something but i'm not sure what.

after fixing that i continued to have issues, and am pretty sure its because there were multiple test courses with the same name created, so the unique course name isolates this test

jivey (approved):

PR #3783, opened 2021-09-21
fix creating time model from moment-tz

TomWoodward:


jivey (approved):
Merging this branch into my date-time-picker branch (with some tweaks) seems to be exactly what we needed, everything is testing well so far 👍

PR #3784, opened 2021-09-21
Add Pardot

jivey:


tutor/src/helpers/pardot.ts
```diff
@@ -0,0 +1,66 @@
...
+export default class Pardot {
+
+    static getConfig(hostname: string) {
+        if (this.isSandbox(hostname)) { return pardotConfig.sand }
+
+        return this.isProduction ? pardotConfig.prod : pardotConfig.sand
+    }
+
+    static get isProduction() {
+        return isProd
```
TomWoodward, line 37: does isProd return true in qa/staging?

TomWoodward (changes requested):

tutor/src/helpers/pardot.ts
jivey: Yeah :/ that's why I'm checking hostname instead. Long term, would be nice to have a sandbox/qa/staging env configs.

TomWoodward: does this logic use the prod config in staging? do we want that?

TomWoodward (changes requested):

tutor/src/helpers/pardot.ts
jivey: I'm glad you pointed that out, I'm guessing we probably don't want that, just like we didn't want it for sandbox. I really should make this do the opposite and only use production if it matches production host, I'll update this PR with that 👍

jivey: Updated to check for production hostname instead, which will let us default to the demo config on sandbox, qa, etc: https://github.com/openstax/tutor-js/pull/3784/commits/3a977478f2b5bae1b1a485b8243437ae7eaa8ad4

TomWoodward (approved):

PR #3817, opened 2021-10-27
fix save to practice by using exercise_id with tasked_id as fallback

jivey:
Working on specs but this should fix a few things:

- taskStep can be undefined when rendering as a fake button when you have no practice questions saved, and the label also wasn't being displayed in that case
- Fix a bug that was introduced in https://github.com/openstax/tutor-js/pull/3789 by going back to the original method of finding by `exercise_id` first, and then falling back to the `tasked_id` if that isn't found.


jivey:
Closing in favor of using UUIDs in https://github.com/openstax/tutor-js/pull/3825 and https://github.com/openstax/tutor-server/pull/2404

PR #3823, opened 2021-11-01
Use the new course role API

Dantemss:
Requires openstax/tutor-server#2402

For courses only (not roster, performance, scores):
- Use student_record instead of students
- Use teacher_record instead of teachers
- Use the new teacher_student_records when using "view as student"


Dantemss:
There's some issue with assuming a non-default role (view as student)

Dantemss:
Requires https://github.com/openstax/tutor-server/pull/2402

Dantemss:
I think the add-edit-questions E2E might need the new Docker image from the backend before it'll pass.

jivey (approved):
Looks good and tested well 👍

PR #3827, opened 2021-11-09
Shuffle Answer Choices

jivey:
- Allow the teacher to toggle Shuffle on and off in the grading template (homework and reading)
image

- Shuffle student answers if the setting is turned on, but only until the first submission, after that "lock" the order by using the `answer_id_order` array

- Allow the teacher to lock answer order importance for teacher-created questions if there are more than 2 options (it's forced on otherwise)
image

- Conditionally show that the answer order note on Submission Overview for homework and reading:
 
| On | Off |
---|---
image | image
---|---
image | image

Note: The E2E spec will fail until https://github.com/openstax/tutor-server/pull/2406 is merged.


shared/src/model/exercise/answer.ts
```diff
@@ -1,12 +1,13 @@
...
 import { isEmpty } from 'lodash';
 import {
-    BaseModel, modelize, field, computed, getParentOf,
+    BaseModel, modelize, field, computed, getParentOf, ID, NEW_ID,
 } from '../../model';
 
 
 export default
 class ExerciseAnswer extends BaseModel {
+    @field id: ID = NEW_ID
```
Dantemss, line 10: I'm surprised this wasn't here before. Didn't we need this to save the selected answer too?

Dantemss (approved):

shared/src/model/exercise/answer.ts
jivey: That was surprising to me as well, it seems it was still getting set on the object though so it worked anyway.

PR #3829, opened 2021-11-11
Fix exercises tag

jivey:
'all' was being created as a tag on exercises because of serialization of the `@model` tags object. This is a quick fix to grab the array instead.

### DIFF
--- a/exercises/src/models/exercises.ts
+++ b/exercises/src/models/exercises.ts
@@ -1,6 +1,6 @@
 import Map from 'shared/model/map';
 import { action, ID, hydrateModel, hydrateInstance } from 'shared/model';
-import { sortBy, last, get } from 'lodash';
+import { sortBy, last, get, merge } from 'lodash';
 import Exercise from './exercises/exercise';
 import urlFor  from '../api'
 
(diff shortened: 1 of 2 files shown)


Dantemss (approved):

Dantemss:
It would probably be a good idea to fix this in modeled-mobx so `toJSON()` doesn't return the `all` method @nathanstitt

nathanstitt:
@Dantemss @jivey ok, at first I thought we did have a bug on modelized-mobx, but after digging in a bit, it looks to me like this is what's going on:

`toJSON` isn't actually a modelized method, it comes from [shared/src/model.ts](https://github.com/openstax/tutor-js/blob/main/shared/src/model.ts#L28-L30) which internally calls `serialize`.   [Serialize](https://github.com/nathanstitt/modeled-mobx/blob/main/src/serialize-hydrate.ts#L63-L86)  looks at the schema that's been registered for the object and only serialized properties that have been registered.

[exercise#tags](https://github.com/openstax/tutor-js/blob/main/shared/src/model/exercise.ts#L33) is a class instance of [TagAssociation](https://github.com/openstax/tutor-js/blob/main/shared/src/model/exercise/tags-association.ts)  which has a `all` property that's registered as a [@model](https://github.com/openstax/tutor-js/blob/main/shared/src/model/exercise/tags-association.ts#L7).  This is the cause of the incorrect `all` property being generated.

I think the best way around this would be to add a `serialize` method onto `TagAssociation` which would return serialize(all),   You can see that [serialize on modeled](https://github.com/nathanstitt/modeled-mobx/blob/main/src/serialize-hydrate.ts#L64-L66) will call that if it's present.  Hopefully something like this will work:
```ts
 serialize() {
    return serialize(this.all)
 }
```

I'm pretty sure that'd do what you want here 🤞 if not, I'm happy to jump in and play with it further.

jivey:
@nathanstitt Ahh that is a really good idea, I think I'll update this to do that.

Dantemss (approved):

PR #3830, opened 2021-11-16
Update OpenStax's pattern-library

Dantemss:
To remove dependency on an old version of node-sass.

Hopefully this will also help reduce the JS file size by removing a ton of old libraries.


jivey (approved):
I'm not really sure how this is used in the app but hopefully any breaking changes will be easy to spot

Dantemss:
I looked at a bunch of Tutor pages and nothing seemed immediately broken to me.

PR #3834, opened 2021-12-02
update mathjax to version 3.2

jivey:
Some changes from the last version:

- The biggest change from the upgrade is needing to load MathJax inside the app JS, as the config has to be set before it is loaded.
- `explorer.js` is included now by default, so the config doesn't need to specifically load it.  
- `Hub.Queue` is replaced with a Promise. 
- Only use debounce when MathJax is actually loaded to prevent some timing issues. 

Otherwise this code mostly works the way it did before: when `typsetMath` is called, look for `data-math` and `<math>` elements, and once they have been typeset, mark them with a `math-rendered` class.


Dantemss (approved):
Need to ask Brooke to do a regression test on some exercises with math.

PR #3859, opened 2022-03-02
More search fixes

Dantemss:
Search now supports limiting results to an approved book or an approved book section, by slug tags

Order books in search box alphabetically by title


jivey (approved):
Tested with the backend PR, seems to work well locally 👍

PR #3868, opened 2022-07-12
Assignment Component Library

jivey:


shared/src/components/question/answer.jsx
```diff
@@ -194,73 +136,26 @@ class Answer extends React.Component {
...
+                iter={iter}
+                answer={answer}
+                chosenAnswer={chosenAnswer}
+                onChangeAnswer={this.onChange}
+                disabled={disabled}
+                onKeyPress={this.onKeyPress}
+                qid={qid}
+                correctIncorrectIcon={<Icon type="check" color="green" />}
+                selectedCount={selectedCount}
+                feedback={feedback}
```
bethshook, line 153: this prop is also not used anymore - maybe this pr just doesn't have the latest changes?

jivey: Yeah I forgot to update this with the newest changes, I'll pull those in and see where these props are used.

PR #3884, opened 2023-01-25
Added more missing help links

Dantemss:
Should be all force links


mwvolo:
@Dantemss I wouldn't worry about these specs, you can close this, I've already redirected these three articles

jivey (approved):
Not sure if we just want to use the redirect but approving in case

Dantemss:
We probably have to release tutor-server too so I fixed the snapshots.

PR #3906, opened 2024-01-23
Add public solution subset

TylerZeroMaster:
# Summary
- New tag with dropdown that supports values `even` and `odd`
- Dropdown is only visible when question is multi-part and solutions are public
- Tag is optional: when it is not set, that means all solutions are public
- When conditions for the dropdown appearing are not met, the value for the tag should be cleared


philschatz (approved):
lgtm!

